Issue #241, opened 2015-06-09
bad resolver recommends "nightly-2015-01-01", which confused me


radix:
If I put `resolver: nightly` in my stack.yaml file (which I just tried), I got this message

```
AesonException "Invalid resolver value: nightly. Possible valid values include lts-2.12, nightly-2015-01-01, and ghc-7.10."
```
This made me think that `nightly-2015-01-01` was the only nightly I could use for some reason. Of course I read "include" as "are", which was my mistake, but an easy one to make.

Showing the *entire* list of possible values would be nice, but perhaps infeasible? It would presumably require an HTTP request to Stackage, and may also be overwhelming given that there are presumably hundreds of nightlies.

bitemyapp:
@radix that or demonstrated the date format, explaining nightly as a special case.

drwebb:
It would be nice to just have 'nightly' as an option, can we use
http://www.stackage.org/download/snapshots.json

snoyberg:
I'm definitely open to improvements to the wording to be less confusing. In fact, if you have a better approach, feel free to just edit it!

@drwebb Taking that approach would remove the possibility of having reproducible builds, as suddenly a new Stackage Nightly would change what your code is built against.

snoyberg:
Thanks for the PR. Should we close this issue?

radix:
Yeah, I'll close it. I mentioned what I think is a nice longer-term solution in https://github.com/commercialhaskell/stack/issues/211#issuecomment-110510693.

Thank you for the review/merge!

PR #242, opened 2015-06-09
make the resolver error show a hopefully more clear description for nightlies

radix:
This is a quick fix for #241. More could be done -- perhaps show a URL to a definitive list of possible values?

### DIFF
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -335,7 +335,8 @@ instance Show ConfigException where
     show (ParseResolverException t) = concat
         [ "Invalid resolver value: "
         , T.unpack t
-        , ". Possible valid values include lts-2.12, nightly-YYYY-MM-DD, and ghc-7.10."
+        , ". Possible valid values include lts-2.12, nightly-YYYY-MM-DD, and ghc-7.10. "
+        , "See https://www.stackage.org/snapshots for a complete list."
         ]
     show (NoProjectConfigFound dir) = concat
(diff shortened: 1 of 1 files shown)


bitemyapp:
@radix https://www.stackage.org/snapshots is the list

https://www.stackage.org/nighty always redirects to latest

radix:
@bitemyapp thanks. I'll add that link. @snoyberg Do you think maybe that /snapshots page could be updated to show the slugs in addition to the pretty names, so it's easy to copy to the resolver field from the main listing?

radix:
Okay, I added the link, but the message is getting really long. I tried to put in a line break, but I realized that the exceptions are just being shown, so the \n is being escaped in the output. Here's what the output looks like now:

```
AesonException "Invalid resolver value: lts-2.9a. Possible valid values include lts-2.12, nightly-YYYY-MM-DD, and ghc-7.10. See https://www.stackage.org/snapshots for a complete list."
```

Pretty long :)

snoyberg:
I don't see a problem, should be an easy change to make.

radix:
Build passed - I guess this is ready for review (I assume I should add the 'ready' label, so I'll do that, but if that's wrong let me know)

chreekat:
I am also guessing, but I think 'ready' means 'ready to be worked on (no questions outstanding)' while the tag you are looking for is 'needs-confirmation' :)

snoyberg:
"ready" does indeed mean "can be worked on." It's provided by waffle.io.

Issue #240, opened 2015-06-09
Network.HTTP.Download.download chokes on Hackage tarballs


snoyberg:
In working on #199, I discovered that calling `download` on the URL http://hackage.haskell.org/package/acme-missiles-0.3/acme-missiles-0.3.tar.gz produces:

> WrongDigest "MD5" "\211\222\&4\245\247\253\231Nx\227\141\218\245\167\\o\215\154}\205\180\239\157w" "09409ff95054443a9adcb9eafc207513"

Not sure if this is a problem with our assumptions or with how Hackage is serving things. At a surface level, it looks like a lack of base16 encoding before comparison.

DanBurton:
The problem was that Hackage doesn't base64-encode the MD5 header. I've changed the `String` to a `CheckHexDigest` type that has slightly different techniques for checking

* a raw String
* a raw ByteString
* a ByteString known to be a "header"

Incidentally, the "raw ByteString" option provides a micro-optimization, since that comparison doesn't go through String anymore.

This touched code in `Data.Fetch`. @snoyberg can you verify that the effected fetch code still works as expected? (Changes are on the download-enhancements branch).

snoyberg:
This is now working.

PR #252, opened 2015-06-10
Download enhacements

DanBurton:


snoyberg:
Steps to be taken still:

* [ ] Fix up against current master (likely uncommenting code I commented out to avoid issue #240)
* [ ] Use MonadBaseControl so that it uses $logInfo instead of putStrLn
* [ ] chunksOverTime feels too complicated to me, I'd like to see that simplified. We can discuss if you're not sure of how to approach that

Issue #199, opened 2015-06-06
Support git repos / http folders in packages stanza


mboes:
Currently, if a package depends on a patched up upstream, the recommendation found in the FAQ is to check out the package as a git submodule, or unpack the source code for that package in some local directory.

However, often it's not so much that a user wants to depend on a patched up package, as the fact that the user is depending on an *unreleased* version of upstream. In this case it would be nice to be able to directly point to a remote repository, or to a commit hash in some remote repository, directly in the stack.yml file without needing to checkout the remote code just for the purpose of building against it.

snoyberg:
If we make some simplifying assumptions, this should be trivial. In particular: assume that the URL or Git commit represents something immutable. Then we unpack/clone into a directory inside `.stack-work` and just treat it as an unwanted local package. Does this sound like a reasonable simplification?

chrisdone:
Where "unwanted" means "not a wanted target" in the Make-like `stack build` step of demanding `stack build TARGET1 TARGET2` or when `stack build` the targets are the implicitly wanted packages listed in the stack.yaml file.

gregwebs:
cabal-meta does much of this and works with darcs also

snoyberg:
I already have an issue open about merging in mega-sdist functionality. Are you opposed to absorbing a few other of the cabal-wrapping tools we've created over the years?

gregwebs:
This is another tool that could possibly stay as a separate executable, except that the tool could benefit from `stack`'s knowledge of `stack.yaml`. The downloading tool could look for a `stack.yaml` in the repo (like how cabal-meta looks for a `sources.txt`) to find where the package is in the repo instead of requiring the user to supply that information (default of course would be `./`).

dbp:
This is great. The only thing I'd add is that some (reasonably prominent) haskell packages put more than one package in the same repository (as subdirectories). For example, all of the `digestive-functors-*` family of packages exist within one git repository. In the tool that I wrote (that I'd love to replace with `stack`), the grammar I defined for these kind of dependencies is:

`repo-path:git-ref+subdir1,subdir2`, where the `git-ref` is optional (if missing, defaults to master), and the `subdir`s are also optional, but when present, are the path to the actual package(s). When missing, it's assumed that the package is at the root. (Note that the comma separated shorthand is unecessary, as you could just have multiple entries, which may be better for `stack`).

snoyberg:
[...]

PR #254, opened 2015-06-10
Use YAML syntax for remote git location identification.

mboes:
As suggested by @greggwebs. The syntax proposed yesterday has the
advantage of concision, but using YAML syntax for the different
components adds marginal verbosity. As it is, the previous syntax made
extensibility problematic, whereas that's exactly where YAML
shines (what if we need an extra field to identify a particular git
commit?). Further, using YAML means people are free to include their own
metadata about package locations, without disturbing Stack.

Here's an example:

```
- location:
    git: https://github.com/kolmodin/binary
    commit: 8debedd3fcb6525ac0d7de2dd49217dce2abc0d9
```

Note that this patch *removes* the old syntax. The reasoning being -
custom user tooling would need to conservatively support both syntaxes
if there was a brief and a long syntax, because it can't assume that
other users utilized the easier to parse syntax.

### DIFF
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -234,17 +234,16 @@ instance ToJSON PackageLocation where
     toJSON (PLHttpTarball t) = toJSON t
     toJSON (PLGit x y) = toJSON $ T.unwords ["git", x, y]
 instance FromJSON PackageLocation where
-    parseJSON = withText "PackageLocation" $ \t ->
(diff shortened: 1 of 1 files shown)


mboes:
cc @gregwebs @3noch. Followup to #199.

snoyberg:
Properly pinging @gregwebs instead :)

chrisdone:
Looks reasonable to me.

snoyberg:
Me too, I agree on the reasoning as well.

gregwebs:
Great!

Issue #244, opened 2015-06-09
Stack interacts badly with GHC-related symlinks


aisamanra:
I have an unusual GHC setup in which all the ghc-related tools are symlinks to a single wrapper script that dispatches to based on how it's invoked (i.e. based on `$0`), so that depending on the symlink name used to invoke it, it can execute a different program.

    ~/bin/ghci     -> ~/bin/ghc-wrapper
    ~/bin/ghc-pkg  -> ~/bin/ghc-wrapper
    ~/bin/runghc   -> ~/bin/ghc-wrapper

Because of this, when I invoke `stack build` with `resolver` set to `ghc-7.10`, I get this error message:

    Process exited with ExitFailure 1: /home/gdritter/bin/ghc-wrapper --no-user-package-db list --global

I believe this is because stack's `System.Process (findExecutable)` calls `System.Directory (canonicalizePath)` on executable names, which means that calls to `ghc-pkg` on my system end up being calls directly to `ghc-wrapper`, which no longer realizes it's supposed to be being invoked as `ghc-pkg`. The correct solution ought to invoke `System.Directory (makeAbsolute)` (or some equivalent, because `makeAbsolute` is only a recent addition to `System.Directory`) rather than `canonicalizePath`, as there is no reason for `stack` to resolve—or even be aware of—symbolic links.

snoyberg:
Seems reasonable. Since it's going to be hard to reproduce the issue on a different system, could you take a stab at making this change?

chrisdone:
No objections to using `makeAbsolute` here, if it does the same job. `canonicalizePath` is partly used for its cleaning up of paths, too. As far as I'm aware we don't need to resolve symbolic links. @manny-fp @snoyberg See any downsides?

snoyberg:
Only the downside that `makeAbsolute` won't be available in the versions of directory shipping with GHC right now (I believe), so we'll need to implement the same logic internally. But otherwise, no, this sounds like a great solution to the problem.

PR #264, opened 2015-06-10
Updated System.Process.Read so it won't try to canonicalize symlinks

aisamanra:
This addresses the problem addresses issue #244, and indeed does solve the problem in tests on my local machine.

The `makeAbsolute` function is available in newer versions of `directory`, but it's a very recent addition and doesn't ship with GHC 7.8, so instead we have to replicate it inline. It's a [pretty trivial function](http://hackage.haskell.org/package/directory-1.2.2.0/docs/src/System-Directory.html#makeAbsolute), so that's not a big deal.


chrisdone:
Thanks!

Issue #277, opened 2015-06-11
stack doesn't respect Buildable:False in tests


qnikst:
If package has a conditionally enabled test, then stack writes a error message.

How to reproduce:

1. take a package with conditional test, that is disabled (for example `network-transport-tcp`)
2. run stack test

Excepted results:
1. output that doesn't mention disable test
2. exit code is `ExitSuccess` (0)

Actual results:
```
Test suite failure for package network-transport-tcp-0.4.1
    TestQC:  executable not found
```
1. Exit code is non zero

snoyberg:
Thanks for the report, this is actually a known issue in the code base:

https://github.com/commercialhaskell/stack/blob/64ff652ba3d7e4a6532276a165a8661de86c4f3b/src/Stack/Package.hs#L226

It's probably not too difficult to address this, just using `pkg` instead of `gpkg`. Want to take a crack at it?

qnikst:
Yes, will try to fix that, thanks for the pointers.

snoyberg:
Cool!

qnikst:
Fixed in https://github.com/commercialhaskell/stack/pull/278 , do I need to do the same for benchmarks? (The only thing that I don't have a good examples under the hands, but can write some).

snoyberg:
Benchmarks aren't necessary (we don't run them ourselves yet), but I wouldn't object to them being added now if you're up for it. It would be very useful to have this same approach used for the executables though, would you be up for adding that to the PR?

qnikst:
Yes, of cause, I'll add this to benchmarks.

qnikst:
Done.

snoyberg:
Merged, thank you!

PR #278, opened 2015-06-11
Do not include unbuildable test suites in build plan.

qnikst:
Fixes Issue: #277.


snoyberg:
See discussion at #277.

Issue #283, opened 2015-06-11
Doing test suites the right way


snoyberg:
This is another issue to try and focus discussion from a bunch of smaller issues. @gregwebs @chrisdone and @qnikst will likely all be interested.

Here's our ideal work flow for tests:

* Go through all packages and build/install their libraries and executables
* Go through all packages that we want to test and build and run their test suites

There's a problem though: Cabal requires reconfiguring a package with test suites enabled to build test suites. This reconfigure step can cause the library to require recompilation, which means that we sometimes have to rebuild dependencies (at least in theory).

Currently, this is getting in the way of at least the following issues: #280 #281 #208 #212 #220. @gregwebs brought up some possible approaches to this in #208 which I was hesitant to move ahead with at the time, but based on what I've seen happen since them, I'm beginning to think may be necessary.

So without drilling into all of the details, I think the workflow we're looking for is:

* Separate out the "build library" and "run test suite" steps completely
* Make the "run test suite" step depend on the "build library" step
* Never unregister or install the library with running the test suite

I think this kind of makes sense, any input?

qnikst:
Suggested workflow looks good for me, and I think it will cover all my use cases.
I'm not sure if proposed solution will cover  #281 because if `run test suite` will depend on `build library` then packages in question (that doesn't have test-suites and are note dependencies of packages with test-suite) will be build and AFAIU @chrisdone wanted them to be skipped entirely.
In this setting #280 becomes orthogonal to the question discussed, but implementing it with new workflow will be straightforward task.

gregwebs:
makes sense to me. From my usage of cabal, it seems like you always want --enable-tests always on but you don't want to require the test dependencies to be installed until the tests are run.

chrisdone:
> I'm not sure if proposed solution will cover #281 because if run test suite will depend on build library then packages in question (that doesn't have test-suites and are note dependencies of packages with test-suite) will be build and AFAIU @chrisdone wanted them to be skipped entirely.

Well, dependencies of packages that need to be tested should be built. It was rather that non-dependency targets that don't have test suites are currently being configured, build and tested and installed when they're not needed.

snoyberg:
Alright, I've implemented this on the 283-test-suite-fix branch. I'd appreciate if others could test before I merge to master since this is a fairly significant overhaul to how execution works. In my testing, it solves the problems I was hoping it would (unnecessary recompiles, cyclical test suites, etc).

PR #287, opened 2015-06-12
Cache package ids

chrisdone:
If the install step doesn't run for a package in `singleBuild` it just uses the package id it already has from the map. Not a necessary optimization but shaves off some time:

```
chris@retina:~/Packages$ time stack build

real	0m1.246s
user	0m4.484s
sys	0m1.760s
```
→
```
chris@retina:~/Packages$ time stack build

real	0m0.911s
user	0m2.532s
sys	0m1.376s
```

If this isn't semantically correct don't bother merging it.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -63,7 +63,7 @@ build bopts = do
 
     if boptsDryrun bopts
         then printPlan plan
-        else executePlan menv bopts baseConfigOpts locals plan
+        else executePlan menv bopts baseConfigOpts locals plan installedMap
   where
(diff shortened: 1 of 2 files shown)


snoyberg:
I think this is correct, but I'm not sure it's a good idea. Reason being: #283 will change how things work such that packages won't be marked as "not installed" if they just require their test suite to be run, and therefore this code path won't even be traversed. I'd rather not have this optimization in place to get confused by that upcoming refactoring.

chrisdone:
I figured the test stuff might be complicated by this. Let's skip this then.

PR #299, opened 2015-06-14
Test suite overhaul

### LINKED ISSUE #283: Doing test suites the right way

snoyberg:
Fixes #283 and a bunch of others, see that issue for details.

### DIFF
--- a/src/Control/Concurrent/Execute.hs
+++ b/src/Control/Concurrent/Execute.hs
@@ -11,7 +11,7 @@ module Control.Concurrent.Execute
     ) where
 
 import           Control.Applicative
-import           Control.Concurrent.Async (Concurrently (..))
+import           Control.Concurrent.Async (Concurrently (..), async)
 import           Control.Concurrent.STM
 import           Control.Exception
 import           Control.Monad            (join)
@@ -37,10 +37,6 @@ data Action = Action
 data ActionContext = ActionContext
     { acRemaining :: !(Set ActionId)
     -- ^ Does not include the current action
-    , acCompleted :: !Int
-    -- ^ Number of actions completed so far
-    , acTotalActions :: !Int
-    -- ^ Total number of actions to be performed
     }
     deriving Show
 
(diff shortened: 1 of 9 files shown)


chrisdone:
Tested it out on my ~/Packages. Works well:

```
chris@retina:~/Packages$ ls -1 | wc -l
28
chris@retina:~/Packages$ stack test
path-0.5.0: configure (test)
path-0.5.0: build (test)
ace-0.6.1: configure (test)
ace-0.6.1: build (test)
descriptive-0.9.3: configure (test)
descriptive-0.9.3: build (test)
lucid-2.9.2: configure (test)
lucid-2.9.2: build (test)
sourcemap-0.1.4: configure (test)
path-0.5.0: test test
sourcemap-0.1.4: build (test)
haskell-docs-4.2.5: configure (test)
haskell-docs-4.2.5: build (test)
descriptive-0.9.3: test test
ace-0.6.1: test test
lucid-2.9.2: test test
sourcemap-0.1.4: test nodejs
haskell-docs-4.2.5: test test
Completed all 6 actions.
chris@retina:~/Packages$
```

Amusingly I was like "hey, haskell-docs doesn't have a test suite. Oh wait, so it does. Go me."

PR #314, opened 2015-06-15
haskell-stack Dockerfile: don't install sudo

gregwebs:


gregwebs:
It also fixes a missing `-y`. I have been using this locally. It shouldn't be used in the CI build.

Issue #130, opened 2015-05-30
Analogue of cabal repl


feuerbach:
IIUC `stack ghci` simply calls ghci with the stack's package db.

`cabal repl` is more than that: if run in a directory with a cabal file, it'll enable the extensions listed in the cabal file (super-useful) and also load all package's modules (moderately useful).

hesselink:
I've never had great luck with `cabal repl`, so if `stack ghci` can improve on that, I'd love it. My use case is usually something like this: I'm developing a set of packages together, and would like to load a file from one of them in the repl. The files it depends on from its package, and preferably also the other packages in the set I'm working on, should come from source, not the package database. For anything else, the package database should be used like now.

I can see how sometimes it can be beneficial to use the compiled version of a package instead, since that speeds things up, so a flag either way looks good.

chrisdone:
@hesselink Indeed, it seems a lot of the time one could load up GHCi with `-i./package1/src -i./package2/src` and optimistically assume the extensions will be non-conflicting. I think that `cabal_macros.h` may also cascade reasonably. 

Brainstorming, perhaps:

    $ cabal repl

* In the root directory: load up all packages specified in stack.yaml.
* In a project-specific directory: load up just that package.

    $  cabal repl foo bar

Load up the `foo` and `bar` packages together, with order being of significance for the order of `-i` and `-X`/`-XNo` specifications.

acfoltzer:
Big :+1: to this in general, and to the improvements laid out in this thread. The biggest current pain point for me is the time it takes to `C-c C-l` after changing anything when running `cabal repl` in `haskell-interactive-mode`.

hesselink:
@chrisdone We use a script that basically uses that approach: it adds all package source directories with `-i`, as well as all cabal macros and autogen directories. We've used it with about 30 packages at the same time, and never had trouble with conflicts between them. We don't specify extensions in cabal files (we put them in the source files instead) so I can't say anything about that.

chrisdone:
Cool, sounds like we're all onboard with this going forward.

@hesselink We've been taking a similar approach of specifying extensions per-file too in anticipation of loading everything on at once for debugging purposes.

ndmitchell:
Note that `ghcid` integration requires `stack ghci` to work as outlined above and load everything up. Requested at https://github.com/ndmitchell/ghcid/issues/38

snoyberg:
This has been merged in, let's start tracking improvements in new issues.

PR #328, opened 2015-06-16
Implement clever stack ghci (aka cabal repl)

chrisdone:
This is for #130.

Here is demonstration output: http://lpaste.net/raw/277892153647038464

This configures with 

* All source directories, cabal macros and autogen directories.
* Extensions.

From every target in the cabal project, and every package in your stack.yaml.

You can call 

    $ stack ghci  

or 

    $ stack ghci projectname

Like in `stack build`.

There're a number of other options in [`BuildInfo`](http://hackage.haskell.org/package/Cabal-1.22.3.0/docs/Distribution-PackageDescription.html#t:BuildInfo):  `ldOptions`, `ccOptions`, `includes` etc.  I'm not sure which of them are actually necessary. Needs to be figured out.

Meanwhile, I'd like to float this by the interested parties: @feuerbach @hesselink @acfoltzer @ndmitchell @snoyberg 

Other points:

* I think some might want a `--load-all` flag which will run `:load X Y Z` on all modules in all packages on start-up. Is that needed for your ghcid @ndmitchell?
* I probably want a `--with-ghc` arg so that I can run `ghci-ng`.

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,4 +1,0 @@
-:set -package optparse-applicative -package optparse-simple -package binary
-:set -isrc/main -isrc
-:set -i.stack-work/dist/x86_64-linux/Cabal-1.18.1.5/build/autogen/
-:set -optP-include -optP.stack-work/dist/x86_64-linux/Cabal-1.18.1.5/build/autogen/cabal_macros.h
--- a/src/Stack/Exec.hs
+++ b/src/Stack/Exec.hs
@@ -1,31 +1,39 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE OverloadedStrings #-}
+
 -- | Execute commands within the properly configured Stack
 -- environment.
 
 module Stack.Exec where
 
 import           Control.Monad.Reader
+import           Control.Monad.Logger
+import           Control.Monad.Catch
+
+
 import           Path
 import           Stack.Types
 import           System.Exit
 import qualified System.Process as P
 import           System.Process.Read
 
+
 -- | Execute a process within the Stack configured environment.
-exec :: (HasConfig r, MonadReader r m, MonadIO m)
+exec :: (HasConfig r, MonadReader r m, MonadIO m, MonadLogger m, MonadThrow m)
         => String -> [String] -> m b
 exec cmd args = do
     config <- asks getConfig
-    liftIO $ do
-        menv <- configEnvOverride config
-                        EnvSettings
-                            { esIncludeLocals = True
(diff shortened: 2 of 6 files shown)


ndmitchell:
I typically have a `.ghci` file which loads the Main module, but I can certainly see that loading all exposed modules seems like a reasonable default.

bitemyapp:
@ndmitchell >loading all exposed modules seems like a reasonable default.

If the code is broken, it won't just exit like Cabal did before 1.22 right?

ndmitchell:
@bitemyapp you'd hope not... I thought that was a Cabal bug, but it might be a feature of ghci. All I require is a prompt which tries to load something and responds to `:reload`.

chrisdone:
> I typically have a .ghci file which loads the Main module, but I can certainly see that loading all exposed modules seems like a reasonable default.

Right, I use .ghci a lot. I'm the .ghci master. Of course, this is aimed at less ghci-savvy users.

> If the code is broken, it won't just exit like Cabal did before 1.22 right? Poking at broken things is half of what I use a REPL for :P

Correct, that's one of the peeves I brought up in #130. cabal repl's failing to launch on compile error is mind-boggling; removing the one thing that can aid me in debugging my code!

> @bitemyapp you'd hope not... I thought that was a Cabal bug, but it might be a feature of ghci. All I require is a prompt which tries to load something and responds to :reload.

IIRC it was fixed by @Philonous and is either on master or on a newer cabal than I've used for a while.

> All I require is a prompt which tries to load something and responds to :reload.

Is this sufficient?

```
chris@retina:~/Work/stack$ stack ghci --ghc-options X.hs
Configuring GHCi with the following projects: stack
GHCi, version 7.8.4: http://www.haskell.org/ghc/  :? for help
Warning: -debug, -threaded and -ticky are ignored by GHCi
Loading package ghc-prim ... linking ... done.
Loading package integer-gmp ... linking ... done.
Loading package base ... linking ... done.
Ok, modules loaded: Main.
Prelude Main> 
```

Or do you need it to figure out that it should load all exposed-modules and main-is modules?

ndmitchell:
If `:reload` and `:show modules` work then that's enough for me. I leave it to you what you put in the environment etc.

snoyberg:
I'd recommend we merge this and then fix things up on master, where more people are likely to test this. It's not like this can really introduce a major regression, since the current ghci command doesn't do anything special (and you can get the current behavior anyway with `stack exec -- ghci`).

chrisdone:
Merging.

Issue #312, opened 2015-06-15
`stack install` could be more clear about how to run installed commands


radix:
Right now when you run `stack install`, you get this rather verbose message at the end of a ton of output:

```
Copying from /Users/radix/.stack/snapshots/x86_64-osx/lts-2.14/7.8.4/bin/happy to /Users/radix/.local/bin/happy
```

How about something like this?

```
Installed the following executables to $HOME/.local/bin
- happy
```

Finally, it'd be nice if stack could detect when the install path is not in $PATH, and print a message (ideally with some colorizing to call attention to it) indicating that the the user should add it to their $PATH.

snoyberg:
The second one is done (though not colorized). I like your version of output better than what we have now. Are you interested in trying to pretty up the output on this? It's a pretty self-contained bit of code.

radix:
@snoyberg Oh, I guess I was confused about the `$PATH` because the time I ran `install` without my PATH set up was with an older stack.

Yeah, I'd be happy to take a look at cleaning up that output.

snoyberg:
PR merged, thanks!

PR #334, opened 2015-06-17
Add an easy-to-read message showing where exes were installed

radix:
This is meant to address #312.

I decided that instead of *replacing* the "Copying ... to ..." output, I'd just add a new message at the end, since the "Copying" lines describe things as they're happening, and my new code just summarizes what was actually successfully installed.

The output looks like this:

```
Installed executables to /Users/radix/.local/bin/:
- stack-integration-tests
- stack
```

I'm sad to say that this is probably the most complex monadic code I've worked with, having never worked with transformer stacks at all, and being a very new Haskeller. So please give it an extra close look, though it's a very small change.

In particular I'm curious if there's an easy way for me to get a function of `[(a, b)] -> Map a [b]`. I ended up accumulating `[(a, [b])]` even though the [b] is always a singleton, and then using `fromListWith (++)` on that.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -232,19 +232,21 @@ executePlan menv bopts baseConfigOpts locals plan = do
 
         currExe <- liftIO getExecutablePath -- needed for windows, see below
 
-        forM_ (Map.toList $ planInstallExes plan) $ \(name, loc) -> do
+        installed <- forM (Map.toList $ planInstallExes plan) $ \(name, loc) -> do
             let bindir =
                     case loc of
                         Snap -> snapBin
                         Local -> localBin
             mfp <- resolveFileMaybe bindir $ T.unpack name ++ ext
             case mfp of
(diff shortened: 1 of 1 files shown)


chreekat:
@radix 

    > :t M.fromListWith (++) . map (fmap (:[]))
    M.fromListWith (++) . map (fmap (:[]))
      :: Ord k => [(k, a)] -> M.Map k [a]

:)

chreekat:
I guess that's the same thing you said you did...

radix:
@chreekat Oh, that makes sense. In some sense it's better than what I did since it lets me accumulate the list as the simpler [(a, b)] and then later turn it into a Map a [b], which is what I was looking for. OTOH the code ends up looking simpler if I just leave it the way I did it... well I'll let you or someone else decide :)

snoyberg:
Looks good to me! I can't think of a better way to do the Map construction. Different? Sure. But not obviously better.

PR #350, opened 2015-06-18
Faster index populate

snoyberg:
Haven't found any slowdown with this


snoyberg:
Pinging @chrisdone

chrisdone:
Ah, I reviewed this but it doesn't merge, forgot to mention it looked ok. :+1:

PR #360, opened 2015-06-19
Line buffering for stdin/stdout, no buffering for stderr.

DanBurton:


chrisdone:
Why? No explanation given here.

Relatedly, [withSticky](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Types/StackT.hs#L253-L257) sets stdout to `NoBuffering` and this is called pretty much for the duration of any stack command, so setting it to `LineBuffering` probably has no effect in practice.

DanBurton:
There was talk on fpco slack about it. It's to ensure consistency for when `stack` isn't being run in a terminal. You're right that for most users, this will probably have no effect.

snoyberg:
@DanBurton Can you add a comment to the source to explain why this is here? Then I think we can merge.

conklech:
On Windows, in an MSYS2/mingw64 shell, 0.0.2 doesn't produce any output until termination. If MSYS2 becomes supported, this would be a user-facing fix.

DanBurton:
@snoyberg done

snoyberg:
@manny-fp I doubt it, but would this have any impact on your fix for the interleaved Haddock stderr output (a0c4f3c2c60470ba3f5543aed624b42ae3f6eeee)?

Issue #143, opened 2015-06-01
Build haddocks for dependencies


feuerbach:
From what I see, haddocks aren't built when installing dependencies.

It would be nice to support these use cases:

1. I only want to install an app and don't care about haddocks, don't waste my time
2. I have plenty of time, build haddocks while installing packages
3. I want to get started ASAP, so for now, only *build* the dependencies, but later I may want to run `stack  haddock` and have it build the haddocks for the dependencies (without rebuilding them, of course). In other words, be able to build haddocks on demand.

taladar:
It might also make sense to add an option to just build the Haddocks for local packages mentioned in stack.yaml? The use-case I have in mind would be to build documentation for locally developed code on a CI server and upload it somewhere.

manny-fp:
@taldar That's currently how `stack haddock` works and that option will definitely stay.

Question for everyone: should the default be to build haddocks for only local packages, or should the default be to build haddocks for all dependencies?

chrisdone:
Good question, this is also relevant for providing an eventual `stack hoogle`.

rvion:
I think the default should be to build the doc for all dependecies

manny-fp:
Another question: for dependencies installed in `~/.stack`, we probably should put the haddocks there as well.  When linking from docs for local packages to dependencies' docs, should we:

1. Use absolute file URLs (e.g. `file:///home/user/.stack/doc/package/`)
2. Use relative file URLs (e.g. `../../../.stack/doc/package/`).  This means links would break if you move your project directory.
3. Always have the links to "../package/", and create filesystem symlinks (or fallback to copies if symlinks not available) pointing to the actual location.

The same question also applies to linking to docs for the standard libraries included with GHC.

I lean toward (3), as this makes for consistent links and makes the documentation easy to copy to a web server or otherwise move around.

manny-fp:
And another question: should `stack haddock` also create a doc index for all packages by default?  What should be included in the index?

1. Only local packages?
2. Combine local packages and dependencies?
3. Separate indices for local packages and dependencies?
4. Both (2) and (3)?

Also, would a "master index" containing all packages installed in `~/.stack` be useful?

rvion:
I vote for 3 ( links to ../package/) too because : 
 - ability to launch warp server to serve docs as you said,
 - ability to mount the .stack server in various docker container with different user name

chrisdone:
[...]

PR #362, opened 2015-06-19
Build haddocks for dependencies (#143)

manny-fp:
This doesn't yet generate the index or do relative links, and it requires rebuilding the package in order to generate haddocks, but does work well in my testing and think it's worth merging with the current feature set (especially as this touches enough code that conflicts are frequent).  

As for building haddocks without rebuilding the packages: this should be doable by adding some extra logic to singleBuild that lets it skip `runhaskell Setup.hs build` when building only haddocks (along with some adjustments to ConfigCache and maybe getInstalled), but interested in opinion about whether that's the right approach or if this would be better done at the Plan level (i.e. split the plan into _build_, _haddock_, and _install_ steps).  Note that haddocks need to be built before installing the package (that way, install takes care of putting them where they belong).

For #143.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -36,6 +36,7 @@ import           Data.Map.Strict                (Map)
 import qualified Data.Map.Strict                as M
 import qualified Data.Map.Strict                as Map
 import           Data.Maybe
+import           Data.Set                       (Set)
 import qualified Data.Set                       as Set
 import           Data.Streaming.Process         hiding (callProcess, env)
 import qualified Data.Streaming.Process         as Process
@@ -115,7 +116,6 @@ printPlan finalAction plan = do
                 DoNothing -> Nothing
                 DoBenchmarks -> Just "benchmark"
                 DoTests -> Just "test"
-                DoHaddock -> Just "haddock"
     case mfinalLabel of
         Nothing -> return ()
         Just finalLabel -> do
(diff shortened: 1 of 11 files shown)


snoyberg:
Will review tomorrow.

snoyberg:
I merged with master and ran all of the tests locally, which passed. Pushing that merge back onto the 143-haddock-deps branch.

src/Stack/Build/Execute.hs
```diff
@@ -593,6 +596,16 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} =
             TTLocal lp -> "build" : map T.unpack (Set.toList $ lpComponents lp)
             TTUpstream _ _ -> ["build"]
 
+    when (shouldBuildHaddock eeBuildOpts eeWanted (packageName package) &&
+          packageHasLibrary package &&
```
snoyberg: Is this redundant with `packageHasExposedModules`? If not, under what circumstances may it be necessary? If it is redundant, looks like a good place to use `assert`.

src/Stack/Build/Execute.hs
```diff
@@ -593,6 +596,16 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} =
...
+          -- Works around haddock failing on bytestring-builder since it has no modules when
+          -- bytestring is new enough.
+          packageHasExposedModules package) $ do
+        announce "haddock"
+        hscolourExists <- doesExecutableExist eeEnvOverride "hscolour"
+        cabal False (concat [["haddock", "--html"]
```
snoyberg: Worth throwing in the `--hoogle` flag?

snoyberg:
LGTM, nice work!

src/Stack/Build/Execute.hs
manny-fp: Pretty sure it's redundant (I think I wrote this before I added the packageHasExposedModules check to shouldBuildHaddock).

PR #368, opened 2015-06-21
Add docs, comments, and type signatures

feuerbach:
While studying the source, I've written a few clarifying comments and added some signatures. Please merge if you find these useful.


snoyberg:
LGTM, better docs are always a good thing

Issue #378, opened 2015-06-22
--stack-yaml command line flag


DanBurton:
Similar to how the `STACK_YAML` environment variable can override the location of `stack.yaml`, we'd like a command line flag that can do the same thing.

snoyberg:
This is now implemented. @DanBurton confirmed?

DanBurton:
Confirmed.

PR #379, opened 2015-06-22
Set the STACK_YAML env var to the --stack-yaml option. #378

DanBurton:
There might be a better way to implement this, but this way was quick and easy.


snoyberg:
:-1: on using environment variables for this. But I'll keep the PR open for now, as the option parsing is fine, and is about 50% of the way there to the necessary solution of passing in an optional `FilePath` to the config parsing function.

PR #404, opened 2015-06-25
Try to avoid any leaks in Stack.PackageIndex

snoyberg:
@mboes Maybe you could review? This should get rid of the possibility of any thunks, and avoid an extra pointer redirection. See any downside? The goal is to address Dawn's performance problem report, though with my own testing I'm doubtful it will actually have a meaningful impact.

### DIFF
--- a/src/Stack/PackageIndex.hs
+++ b/src/Stack/PackageIndex.hs
@@ -18,7 +18,10 @@
 module Stack.PackageIndex
     ( updateAllIndices
     , PackageDownload (..)
-    , PackageCache (..)
+    , PackageCache
+    , pcOffset
+    , pcSize
+    , pcDownload
     , getPackageCaches
     ) where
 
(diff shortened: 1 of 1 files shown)


mboes:
Is the pointer dereferencing happening millions of times in a tight loop? I very highly doubt this fix speeds anything up, though benchmark numbers could tell. If it doesn't then I'd vote for keeping the status quo, whose code is clearer.

The thunk issue can be fixed by changing

```
         addJSON ident lbs =
             case decode lbs of
                 Nothing -> m
                 Just pd -> Map.insertWith
                     (\_ pc -> pc { pcDownload = Just pd })
```

to 

```
         addJSON ident lbs =
             case decode lbs of
                 Nothing -> m
                 Just !pd -> Map.insertWith
                     (\_ pc -> pc { pcDownload = Just pd })
```

In general use of the [strict](https://hackage.haskell.org/package/strict) package can be useful for avoiding thunks with `Maybe`.

snoyberg:
It's not millions of times, it's 50,000 times and growing (once for each .cabal file). I agree that the benefits here may not outweigh the decreased readability. Let's just do the thunk elimination as you commented on master, and if problems still exist for Dawn after that, we'll revisit.

mboes:
Close?

snoyberg:
Yes. I'll leave the branch around for a bit longer though.

PR #406, opened 2015-06-25
Bypass decodeUtf8 in printBuildOutput

fumieval:
`printBuildOutput` (https://github.com/commercialhaskell/stack/blob/706736a6a46810533281eb7cc38714f31a6a5412/src/Stack/Build/Execute.hs) re-prints the output from `ghc`.  However, in my environment (GHC 7.10.1, Windows 8.1), it suffers from indigestion due to `decodeUtf8`:

```
stack: Cannot decode byte '\x81': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```

This patch fixes this so that stack correctly prints compile errors.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -47,6 +47,7 @@ import qualified Data.Text.Encoding             as T
 import           Data.Text.Encoding             (encodeUtf8)
 import           Distribution.System            (OS (Windows),
                                                  Platform (Platform))
+import           Language.Haskell.TH            (Loc(..))
 import           Network.HTTP.Client.Conduit    (HasHttpManager)
 import           Path
 import           Path.IO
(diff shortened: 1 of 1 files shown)


snoyberg:
LGTM. @chrisdone Can you review as well and if you're happy with it merge?

chrisdone:
Damn, I had this as a shower thought that Windows would not like the decodeUtf8. I forgot about writing it up. Reviewing.

chrisdone:
I think there're other decodeUtf8's in the codebase that need reviewing for Windows support.

snoyberg:
Generally speaking it would be better to use `decodeUtf8With lenientDecode` if we have to decode, but in cases where we're just going to spit out to the terminal, sticking with `ByteString` like this PR does is better.

PR #412, opened 2015-06-25
Add include-dirs from cabal file to GHC call.

### LINKED ISSUE #402: 'stack ghci' doesn't add include dirs
If you have a package with an `include-dirs` line in the cabal file, it doesn't load correctly in `stack ghci`. As a workaround, you can pass `--ghc-options -I<path-to-include-dir>`, but I think ideally stack should do this.

hesselink:
Fixes #402.


chrisdone:
Looks good, thanks!

PR #424, opened 2015-06-26
Reduce duplication in file resolver functions

markus1189:
DRY up file resolver functions that only differ whether it's a file or directory by introducing internal helper functions.

### DIFF
--- a/src/Path/IO.hs
+++ b/src/Path/IO.hs
@@ -72,33 +72,35 @@ resolveFile x y =
          where fp = toFilePath x FP.</> y
        Just fp -> return fp
 
+-- Internal helper to define resolveDirMaybe and resolveFileMaybe in one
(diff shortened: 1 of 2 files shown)


snoyberg:
Thanks! I'll leave to @chrisdone to review and comment on.

markus1189:
Hope I fixed the build error due to missing `<$>` import, I just replaced it with `fmap` instead of adding CPP.

chrisdone:
Looks good!

markus1189:
There where still some missing Functor instances for ghc < 7.10 ...  I added them, let's see what travis says this time

chrisdone:
I already fixed it on master, no worries.

markus1189:
Awesome, funny how just using `liftM` in this case is better than `fmap` =)

PR #432, opened 2015-06-27
Add --force flag to init command

markus1189:
New '--force' argument to 'stack init' overwrites the existing stack.yaml file instead of refusing the operation.


chrisdone:
Looks good!

PR #437, opened 2015-06-27
Add --external flag to `stack dot`

### LINKED ISSUE #431: stack dot seems broken
##### stack dot seems broken

I'm on 
```
stack --version
Version 0.1.1.0, Git revision 2cac22d2a26dab6faa00c6f03edf5798affb7840
```

In stack code directory, I did:

```
$ stack dot > test.dot
$ dot -Tpng < test.dot > graph.png
Fontconfig warning: ignoring UTF-8: not a valid region tag
```

here is the result
graph

but I don't see dependencies.

Here is the exact output given by stack dot
```
$ stack dot
digraph deps {
splines=polyline;
{rank=max; "stack"}
}
```

here is the help for stack dot:
```
stack dot --help
Usage: stack dot
  Visualize your project's dependency graph using Graphviz dot
```

markus1189:
When given `--external`, `stack dot` will include external dependencies.

By default it keeps the old behavior and only includes local packages and their interdependencies.
If `--external` is activated local packages will have the "dashed" line attribute so that one can distinguish local packages from external dependencies.

As an example, given a tree like this with two projects `one` and `two`:

```
├── [4.0K]  one
│   ├── [ 686]  one.cabal
│   └── [  46]  Setup.hs
├── [  78]  stack.yaml
└── [4.0K]  two
    ├── [  46]  Setup.hs
    └── [ 710]  two.cabal
```

`stack dot` will produce:
foo

while `stack dot --external` will produce:
foo

This would also fix #431.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -14,7 +14,8 @@
 
 module Stack.Build
   (build
-  ,clean)
+  ,clean
+  ,withLoadPackage)
   where
 
 import           Control.Monad
@@ -101,7 +102,13 @@ mkBaseConfigOpts bopts = do
         }
 
 -- | Provide a function for loading package information from the package index
-withLoadPackage :: M env m
+withLoadPackage :: ( MonadIO m
+                   , HasHttpManager env
+                   , MonadReader env m
+                   , MonadBaseControl IO m
+                   , MonadCatch m
+                   , MonadLogger m
+                   , HasEnvConfig env)
                 => EnvOverride
                 -> ((PackageName -> Version -> Map FlagName Bool -> IO Package) -> m a)
                 -> m a
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -1,53 +1,185 @@
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
-module Stack.Dot where
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Stack.Dot (dot
+                 ,DotOpts(..)
+                 ,dotOptsParser
+                 ,resolveDependencies
+                 ,printGraph
+                 ) where
 
-
-import           Control.Monad (when)
+import           Control.Monad (void)
 import           Control.Monad.Catch (MonadCatch)
-import           Control.Monad.IO.Class (MonadIO)
+import           Control.Monad.IO.Class
 import           Control.Monad.Logger (MonadLogger, logInfo)
 import           Control.Monad.Reader (MonadReader)
+import           Control.Monad.Trans.Control (MonadBaseControl)
 import qualified Data.Foldable as F
-import           Data.Monoid ((<>))
+import qualified Data.HashSet as HashSet
+import           Data.Map (Map)
 import qualified Data.Map as Map
+import           Data.Monoid ((<>))
+import           Data.Set (Set)
 import qualified Data.Set as Set
-import qualified Data.Text as T
+import           Data.Text (Text)
+import qualified Data.Text as Text
+import qualified Data.Traversable as T
+import           Network.HTTP.Client.Conduit (HasHttpManager)
+import           Options.Applicative
+import           Options.Applicative.Builder.Extra (boolFlags)
+import           Stack.Build (withLoadPackage)
 import           Stack.Build.Source
 import           Stack.Build.Types
+import           Stack.Constants
 import           Stack.Package
 import           Stack.Types
 
+-- | Options record for `stack dot`
+data DotOpts = DotOpts
+    { dotIncludeExternal :: Bool
+    -- ^ Include external dependencies
+    , dotIncludeBase :: Bool
+    -- ^ Include dependencies on base
+    , dotDependencyDepth :: Maybe Int
+    -- ^ Limit the depth of dependency resolution to (Just n) or continue until fixpoint
+    }
+
+-- | Parser for arguments to `stack dot`
+dotOptsParser :: Parser DotOpts
+dotOptsParser = DotOpts <$> includeExternal <*> includeBase <*> depthLimit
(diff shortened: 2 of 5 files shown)


rvion:
as discussed in #431, should we make --external the default (without command option) and come up with an other option name (like `--local-only`) for current behaviour ?

snoyberg:
I'd rather have the current behavior as default, with flags to force it either direction. Then we can change the default in the future and people can keep backwards compatibility easily. Let me review this patch. I can make the addition of a `--no-external` or equivalent myself after review unless @markus1189 beats me to it :).

snoyberg:
I don't think this is working as intended: it currently only goes one level deep. For example, when run on the stack repo, it tells you all of the packages that stack depends on, but not the packages that they depend on as well. Am I correct in understanding that that was the desired behavior?

markus1189:
I did not find an easy way to resolve the PackageName into something we can use to extract further dependencies without changing even more of the original version. So currently the intended behavior is that only one level is shown. 

But I agree that it would be very cool to get a full dependency graph generated. Maybe that could be done as an extension though I don't know how much work it will be.

snoyberg:
I can probably either add that myself tomorrow or, if you'd prefer, add information to this issue explaining where to get the missing information.

markus1189:
If you want to explain it that's great because then I can give it a shot
myself :)

snoyberg:
OK, I'll give more details tomorrow when I'm at my computer instead of on my phone. Probably the Stack.Build.Sources module is the right place to start.

snoyberg:
Alright, a bit more information:

* `loadSourceMap` will give you a `Map` from `PackageName` to some information on package in question
* For local packages, you can just grab the `Package` value with `lpPackage` like the code does now
* For non-local packages, you'll want to use `withLoadPackage` in `Stack.Build`.

I hope that helps, let me know if you have any questions

markus1189:
Yes thanks that definitely helps.  I think I will have something working soon, I'll let you know.

markus1189:
Okay so I got something working.
I added two more flags, `--[no-]include-base` and `--depth DEPTH`
```
Usage: stack dot ([--external] | [--no-external]) ([--include-base] |
                 [--no-include-base]) [--depth DEPTH]
  Visualize your project's dependency graph using Graphviz dot

Available options:
  --external               Enable inclusion of external dependencies
  --no-external            Disable inclusion of external dependencies
  --include-base           Enable inclusion of dependencies on base
  --no-include-base        Disable inclusion of dependencies on base
                           limit)
```

The reason for `--include-base` is that most (all?) dependencies will have a edge to base which makes the graph harder to read.

In addition `--depth N` limits the depth of the shown dependencies, as it will become very big for larger projects.

As an example take two cabal projects `one` and `two`

Dependencies for `one`:
```
build-depends:       base >=4.8 && <4.9, free
```
and for `two`
```
build-depends:       base >=4.8 && <4.9, free, mtl, transformers, one
```

`stack dot` gives
default

`stack dot --external --depth 0 --no-include-base` gives
ext0

`stack dot --external --depth 1 --no-include-base` gives
ext1

`stack dot --external --depth 2 --no-include-base` gives
onetwo

As an example why `--no-include-base` is handy,  compare
`stack dot --external --include-base`
with_base
with `stack dot --external --no-include-base`
without_base

However I noticed that for packages that come with `ghc` it seems like I have to do something different to get the dependencies, as they are not in the `SourceMap`, is that correct?
That would be something I would like to change before this can be merged in addition to some tests so this is only a preview :)

snoyberg:
This is looking awesome. On my phone so I can't review properly, I'll provide feedback later.

snoyberg:
It looks like this doesn't build with GHC 7.8 due to lack of Applicative => Monad.

markus1189:
Yeah that trips me up every time because I currently only have ghc 7.10 installed :)

snoyberg:
You could just use `stack setup` ;)

markus1189:
Did I mention that my internet is SUPER slow? About 30 kB/s is the normal rate :P

src/Stack/Build.hs
```diff
@@ -87,7 +88,13 @@ mkBaseConfigOpts bopts = do
         }
 
 -- | Provide a function for loading package information from the package index
-withLoadPackage :: M env m
+withLoadPackage :: ( MonadIO m
```
snoyberg: :+1:

snoyberg:
Ouch, I think I'd go crazy at that rate :(

The patch looks good, barring the 7.8 issues I'd be happy to merge it. I don't mind fixing those up myself, would you like me to?

And thanks again for the contributions, not sure if you noticed, but:

__Achievement Unlocked: Commit Bit__

markus1189:
The 7.8 issue should be fixed.  There is still a problem with packages included with `ghc`, like `template-haskell`, `ghc-prim` and others.  I guess I have to use something else to get those dependencies, any ideas?

And I would like to add some tests for the graph building before this is merged.

snoyberg:
I'm not sure how useful it is to include dependencies of wired-in packages (those that are tied into the compiler), since their dependencies are somewhat irrelevant (they can never be reinstalled). I'd say showing the `ghc` package as a leaf is a good call.

I'll hold off on the merge then.

markus1189:
Is there an easy way to check whether it is wired-in? Or is checking for an empty dependency set  enough?

snoyberg:
It's not exactly the same thing, but I'd just trust that empty dependency set means it should be displayed as a leaf. This will apply to some non-wired-in packages that are in the global database as well if I'm not mistaken.

But to answer your question much more directly: look at `Stack.Constraints.wiredInPackages`.

markus1189:
Okay I think this can be merged.  Packages from `Stack.Constants.wiredInPackages` are drawn as rectangle nodes, though there are more leaf nodes if they are in the global package db I think.
I chose not to add a "ghc" node because the graph already is complicated enough for large projects and adding this pseudo node just hinders readability.

I also added some tests although not as many as I would like to have because for that I would need to also create `LocalPackage`s which is a little unwieldy manually :)

I split the pr into 3 commits but if you want I can also just squash them into one.

Best,
Markus

mboes:
Thanks!

PR #441, opened 2015-06-28
Skip unreachable directories during init

uvNikita:
Catch permission denied exception in findFiles function and ignore such
directories.

In my project I have directory with content of linux container. After running `stack init` command I got an error "Permission denied" on this directory.

Fixed such behavior and added test for this usecase.

### DIFF
--- a/test/integration/tests/skip-unreachable-dirs/Main.hs
+++ b/test/integration/tests/skip-unreachable-dirs/Main.hs
@@ -1,0 +1,8 @@
+import StackTest
+import System.Directory (setPermissions, emptyPermissions, createDirectory)
+
+main :: IO ()
+main = do
+    createDirectory "unreachabledir"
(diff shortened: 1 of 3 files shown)


test/integration/tests/skip-unreachable-dirs/Main.hs
```diff
@@ -0,0 +1,7 @@
+import StackTest
+import System.Directory (setPermissions, emptyPermissions)
+
+main :: IO ()
+main = do
+    setPermissions "unreachabledir" $ emptyPermissions
```
snoyberg: I think you need to first create the directory.

snoyberg:
@chrisdone Can you look this one over? I think this should do the right thing for all our use cases.

chrisdone:
Works for me.

uvNikita:
@snoyberg Updated. Create "unreachable" directory now in test.

Issue #391, opened 2015-06-24
Directory symlinks for latest install


mgsloan:
The idea is to have `stack build` (and other commands which involve a build) add directory symlinks to `.stack-work` which point to the latest install and logs folders.  In other words, after a build which puts its files in `.stack-work/install/x86_64-linux/lts-2.14/7.8.4/`, `.stack-work/install/latest` will also point to the directory.

This is useful to the user, as it provides a short path to the latest binaries / docs.  It's also useful for other tools / build scripts which need to access the results of `stack build`.  Instead of hardcoding paths that depend on OS version and need to be updated when we change ghc version / docker image tag, we can use paths that don't change.

snoyberg:
What's wrong with `stack exec` for running executables? It also seems like `stack path` will go a long way towards answering these questions.

snoyberg:
I'm overall against this idea since it introduces state that doesn't need to be there. With this in place, we suddenly have a piece of metadata floating around that could contradict what's in `stack.yaml` because we haven't yet run `stack build`.

alanz:
What about a specific command in stack to symlink the current context? This will help for people using traditional tooling.

e.g. I would be interested in being able to have a symlink from a given .stack-work context to .dist

snoyberg:
If `stack path` gave you the directory you were looking for, how different would that be from `ln -s $(stack path --dist-dir) dist`?

(Pinging @chrisdone, perhaps the `stack path` output should be modified to make the above command actually work.)

alanz:
That would do it, but I suspect it may be a fairly common operation
(initially at least), and it may be useful to give it a synonym as a stack
command.

chrisdone:
@snoyberg I agree with this. I'd considered it while implementing; specifically requesting one item should omit the key. I'll do it now.

chrisdone:
Done.

    $ stack path --dist-dir
    .stack-work/dist/x86_64-linux/Cabal-1.18.1.5/

mgsloan:
Having such a command is sufficient, cool!

I've edited the issue to make it clear that this isn't about logs.  That was a mistaken afterthought.

snoyberg:
After speaking with @mgsloan about this yesterday, I think I understand the motivation a bit better, and it makes sense. Let's move ahead with it.

snoyberg:
[...]

PR #442, opened 2015-06-29
Add a symlink to the last install directory #391

mgsloan:
This implements the feature discussed in #391.

It creates a symlink from `.stack-work/install/last` at the beginning of executing the build plan.  I found `last` to be a more sensible name for this directory than the name choice of `latest` from the issue.

Since `.stack-work/dist/<platform>/<cabal>/` also exists, I added a `.stack-work/dist/last` directory as well, for consistency, though it seems less useful than the install directory symlink.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -18,6 +18,7 @@ import           Control.Concurrent.Lifted (fork)
 import           Control.Concurrent.Execute
 import           Control.Concurrent.MVar.Lifted
 import           Control.Concurrent.STM
+import           Control.Exception.Enclosed     (catchIO)
 import           Control.Exception.Lifted
 import           Control.Monad
 import           Control.Monad.Catch            (MonadCatch, MonadMask)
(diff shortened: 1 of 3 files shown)


src/Stack/Build/Execute.hs
```diff
@@ -896,3 +899,26 @@ getSetupHs dir = do
...
+  bc <- asks getBuildConfig
+  target <- distDirFromDir (bcRoot bc)
+  source <- lastInstallSymlink
+  updateSymlink target source
+
+updateSymlink :: M env m => Path Abs Dir -> Path Abs File -> m ()
+updateSymlink target source = do
+  createTree target
+  removeFileIfExists source
+  liftIO $ createSymbolicLink (toFilePath target) (toFilePath source)
```
snoyberg: This needs some `catchIO` wrapping to work on Windows.

mgsloan: Indeed, I haven't tested this on windows - I wish travis CI supported windows!

Hmm, would the `catchIO` wrapping be for `createSymbolicLink`?  Unless there are some unusual circumstances, that ought to not throw any exceptions, because the target exists and the source doesn't.  Might still be good to wrap it with a more descriptive error message, though?

`createTree` shouldn't ever fail, because it's just a no-op if the tree doesn't exist, and otherwise creates the whole path.  `removeFileIfExists` catches the does-not-exist exception [here](https://github.com/fpco/stack/blob/e9860699de3111dd1d7509f21de8d08d35e86a11/src/Path/IO.hs#L157).

snoyberg: As I mentioned in the issue, the implementation of this function on windows just throws an exception.

mgsloan: Ah, I see, good point!  I've added some thoughts on windows support here: https://github.com/commercialhaskell/stack/issues/470

mgsloan:
@snoyberg How's this updated version look?

mgsloan:
As mentioned in #470, we're not going to add this.

PR #443, opened 2015-06-29
Specifypath

### LINKED ISSUE #342: Stack install should take a --path option
I tend to keep my user binaries in ~/bin instead of ~/.local/bin. The directory path seems to be set in.

```
src/Stack/Config.hs
186:        localDir <- liftIO (getAppUserDataDirectory "local") >>= parseAbsDir
```
getAppUserDataDirectory is defined in `System.Directory`. Is it should return $HOME/.local always, so it doesn't appear to be able to take a flag or anything. The install function takes both `BuildOpts` and `GlobalOpts` parameters, so perhaps such a parameter could live in one of there as a Maybe and be passed as a --path option from the command line.

https://github.com/commercialhaskell/stack/blob/118cb3ed42ff6aba02b5acdea19e6c2cbbc52302/src/main/Main.hs#L346

drwebb:
Stack install will take a --path argument to control to which directories the bins are written to. Should close #342.


src/Stack/Build/ConstructPlan.hs
```diff
@@ -149,9 +149,13 @@ constructPlan mbp0 baseConfigOpts0 locals extraToBuild0 locallyRegistered loadPa
                 , planFinals = M.fromList finals
                 , planUnregisterLocal = mkUnregisterLocal tasks dirtyReason locallyRegistered
                 , planInstallExes =
-                    if boptsInstallExes $ bcoBuildOpts baseConfigOpts0
-                        then installExes
-                        else Map.empty
+                    let (doInstallFlag,maybeUserDir) = 
+                            boptsInstallExes $ bcoBuildOpts baseConfigOpts0
+                        in if (doInstallFlag)
+                           then case maybeUserDir of
```
snoyberg: Why not:

    InstallExesPlan True maybeUserDir installExes

src/Stack/Build/Types.hs
```diff
@@ -280,7 +281,7 @@ data BuildOpts =
             ,boptsDryrun :: !Bool
             ,boptsGhcOptions :: ![Text]
             ,boptsFlags :: !(Map (Maybe PackageName) (Map FlagName Bool))
-            ,boptsInstallExes :: !Bool
+            ,boptsInstallExes :: !(Bool, Maybe (Path Abs Dir))
```
snoyberg: I think a sum type `data InstallDest = NoInstall | DefaultInstall | InstallDir (Path Abs Dir)` would make more sense, to avoid such confusing values as `(True, Just _)`.

drwebb:
Thanks for the suggestions, I though that tuple was ugly and confusing as well

src/Stack/Build/ConstructPlan.hs
```diff
@@ -149,9 +149,13 @@ constructPlan mbp0 baseConfigOpts0 locals extraToBuild0 locallyRegistered loadPa
                 , planFinals = M.fromList finals
                 , planUnregisterLocal = mkUnregisterLocal tasks dirtyReason locallyRegistered
                 , planInstallExes =
-                    if boptsInstallExes $ bcoBuildOpts baseConfigOpts0
-                        then installExes
-                        else Map.empty
+                    let installExesPlan = 
+                            boptsInstallExes $ bcoBuildOpts baseConfigOpts0
+                        in case installExesPlan of
```
snoyberg: This code still makes me uncomfortable, as you're encoding the same information as both a `Bool` and by making an empty `Map`. Is there any reason to keep the `Bool` around?

src/Stack/Build/Execute.hs
```diff
@@ -229,10 +229,13 @@ executePlan menv bopts baseConfigOpts locals sourceMap plan = do
             , eeGlobalDB = globalDB
             }
 
-    unless (Map.null $ planInstallExes plan) $ do
+    when (installExesPlanFlag $ planInstallExes plan) $ do
+        -- Also install executables
         snapBin <- (</> bindirSuffix) `liftM` installationRootDeps
         localBin <- (</> bindirSuffix) `liftM` installationRootLocal
-        destDir <- asks $ configLocalBin . getConfig
```
snoyberg: Why not just have the flag on the command line cause `configLocalBin` to get overridden with the provided value? That way you don't end up with an unused install path floating around that may accidentally get used.

drwebb: Okay, I've been staring at this tonight for far too long and am finally making sense of it. So, I can keep the hardcode "$HOME/.local/bin" in the config file and then call `loadConfigWitithOpts` overriding the `destDir`. Then I can clean up the code in the middle to remove the floating install path? I've pushed the WIP, but it's my first time now making sense of the local config Monoid.

drwebb:
Copying the comment above in the line note so it stay's visible: Okay, I've been staring at this tonight for far too long and am finally making sense of it. So, I can keep the hardcode "$HOME/.local/bin" in the config file and then call loadConfigWitithOpts overriding the destDir. Then I can clean up the code in the middle to remove the floating install path? I've pushed the WIP, but it's my first time now making sense of the local config Monoid.

drwebb:
Closing in favor of "global -p option" #542

PR #455, opened 2015-06-30
Remove usage of `ByteString.breakByte`

markus1189:
Warning and solution printed during compilation:
```
Deprecated: "It is an internal function and should never have been
exported. Use 'break (== x)' instead. (There are rewrite rules that
handle this special case of 'break'.)"
```


snoyberg:
Thanks!

PR #487, opened 2015-07-02
Add --prune to `stack dot`

markus1189:
TL;DR: Add `--prune p1,p2,p3` to `stack dot` to remove nodes from the dependency graph

### The Motivation

Playing around with some haskell projects and their `stack dot` graphs, it becomes apparent that it it a little hard to look at them because they get quite huge.  As an example, take the graph for `wreq` generated by `stack dot --external --no-include-base`:
wreq

okay that is a LOT.  Looking a bit more we see that `wreq-examples` is also there which depends on a bunch of the dependencies.

The new feature is that `--prune` takes a comma separated list of package names that will be pruned.

So here is `wreq` without dependencies of `wreq-examples` and `lens`:
wreq_pruned_lens
Still a lot but more manageable.

Of course we can have a dependency graph that is a little more focused by pruning more (randomly chosen) dependencies from the graph:

`stack dot --external --no-include-base --prune lens,wreq-examples,http-client,aeson,tls,http-client-tls`
wreq_pruned

### The implementation
- new flag `--prune PACKAGE1,PACKAGE2,...`
- add pruning functionality to `Stack.Dot`
- properties in `DotSpec` for pruning
- new dependency for stack library: `split` *without* version bound, **do we need one**?
- new dependency in stack tests: QuickCheck, should not be questionable

### Help text for `stack dot`
```
Usage: stack dot ([--external] | [--no-external]) ([--include-base] |
                 [--no-include-base]) [--depth DEPTH] [--prune PACKAGES]
  Visualize your project's dependency graph using Graphviz dot

Available options:
  --external               Enable inclusion of external dependencies
  --no-external            Disable inclusion of external dependencies
  --include-base           Enable inclusion of dependencies on base
  --no-include-base        Disable inclusion of dependencies on base
                           limit)
  --prune PACKAGES         Prune each package name from the comma separated list
                           of package names PACKAGES

```

PS: I also took the liberty to add the changes to CHANGELOG, or is this not wanted / do you prefer to do it yourself in the future?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,8 @@
 ## Unreleased
 
-* Add --ignore-subdirs flag to init command [#435](https://github.com/commercialhaskell/stack/pull/435)
+* Add `--prune` flag to `stack dot` [#487](https://github.com/commercialhaskell/stack/issues/487)
+* Add `--[no-]external`,`--[no-]include-base` flags to `stack dot` [#437](https://github.com/commercialhaskell/stack/issues/437)
+* Add `--ignore-subdirs` flag to init command [#435](https://github.com/commercialhaskell/stack/pull/435)
 * Handle attempt to use non-existing resolver [#436](https://github.com/commercialhaskell/stack/pull/436)
 * Add `--force` flag to `init` command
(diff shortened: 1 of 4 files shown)


zudov:
@markus1189 Wow, I was just going to ask for that feature. Thanks a lot.

Issue #399, opened 2015-06-24
LGPL licensing restrictions on Windows because of integer-gmp


quyse:
Default version of GHC on Windows produces binaries with integer-gmp linked statically. As integer-gmp's license is LGPL, it means that resulting binaries [should be provided with source code or object files](http://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic). I guess that's not a thing everyone would want.

On UNIXes integer-gmp is linked dynamically, so these restrictions don't apply.

So, in order to fix that we can either:
* Provide a version of GHC which builds Windows executables with integer-gmp linked dynamically (haven't tried, but seems like [it's possible](http://haskell.forkio.com/gmpwindows)).
* Provide a version of GHC [with integer-simple wired in](https://code.google.com/p/hmpfr/wiki/GHCWithoutGMP) instead of integer-gmp (some packages have problems, but overall works good for me).
* Or at least write a big red warning on stack's homepage that your Haskell software on Windows has to be essentially a free software.

Additional Links:
[ReplacingGMPNotes](https://ghc.haskell.org/trac/ghc/wiki/ReplacingGMPNotes)

UPDATE:
* More up-to-date link: [Design/IntegerGmp2](https://ghc.haskell.org/trac/ghc/wiki/Design/IntegerGmp2)
* Clarification, just to avoid confusion: license of integer-gmp (the Haskell library) is BSD3, but it uses C library [The GNU Multiple Precision Arithmetic Library](https://gmplib.org/) which is LGPL.

vyp:
Might help: http://copyleft.org/guide/comprehensive-gpl-guidech12.html#x15-11100011.5

N.B. Have no idea if that website/document can be considered legal advice.

andrewthad:
I would love it if there was a way to use integer-simple with stackage. As a warning, `blaze-textual`, which is supported by stackage, has a [very long standing issue](https://github.com/bos/blaze-textual/issues/8) that prevents it from working with `integer-simple` (and this library is used by `postgres-simple`, which is used by a lot of people). If stackage were to start supporting integer-simple GHC builds, some packages would have to get patched first.

snoyberg:
I don't think a big red warning is the right answer, but having this well documented in the FAQ seems reasonable. I would love to get rid of the LGPL code entirely, but as @andrewthad points out, there's no real chance of being able to do that right now.

@quyse Would you be able to add a FAQ entry for this?

kantp:
[...]

PR #495, opened 2015-07-03
For windowsintegersimple resolve flags as if Windows OS

3noch:
I have very little confidence I did this right, but here's my good faith effort. :)

See #399

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -47,7 +47,7 @@ import qualified Data.Set as S
 import qualified Data.Text as T
 import           Data.Text.Encoding (encodeUtf8, decodeUtf8)
 import qualified Data.Yaml as Yaml
-import           Distribution.System (OS (Windows), Platform (..), buildPlatform)
+import           Distribution.System (OS (..), Platform (..), buildPlatform)
 import qualified Distribution.Text
 import           Distribution.Version (simplifyVersionRange)
 import           Network.HTTP.Client.Conduit (HasHttpManager, getHttpManager, Manager, parseUrl)
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -63,7 +63,7 @@ import           Distribution.Package hiding (Package,PackageName,packageName,pa
 import           Distribution.PackageDescription hiding (FlagName)
 import           Distribution.PackageDescription.Parse
 import           Distribution.Simple.Utils
-import           Distribution.System (OS, Arch, Platform (..))
+import           Distribution.System (OS (..), Arch, Platform (..))
 import           Distribution.Text (display)
 import           Distribution.Version (intersectVersionRanges)
 import           Path as FL
@@ -615,7 +615,7 @@ mkResolveConditions :: Version -- ^ GHC version
 mkResolveConditions ghcVersion (Platform arch os) flags = ResolveConditions
     { rcFlags = flags
     , rcGhcVersion = ghcVersion
-    , rcOS = os
+    , rcOS = if isWindows os then Windows else os
     , rcArch = arch
     }
 
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -234,11 +234,6 @@ stripTrailingSlashT t = fromMaybe t $ T.stripSuffix
         (T.singleton FP.pathSeparator)
         t
 
-isWindows :: OS -> Bool
-isWindows Windows = True
-isWindows (OtherOS "windowsintegersimple") = True
-isWindows _ = False
-
 -- | Ensure GHC is installed and provide the PATHs to add if necessary
 ensureGHC :: (MonadIO m, MonadMask m, MonadLogger m, MonadReader env m, HasConfig env, HasHttpManager env, MonadBaseControl IO m)
           => SetupOpts
--- a/src/Stack/Types/BuildPlan.hs
+++ b/src/Stack/Types/BuildPlan.hs
@@ -22,6 +22,7 @@ module Stack.Types.BuildPlan
     , MiniPackageInfo (..)
     , renderSnapName
     , parseSnapName
+    , isWindows
     ) where
 
 import           Control.Applicative
@@ -51,7 +52,7 @@ import           Data.Time                       (Day)
 import qualified Data.Traversable                as T
 import           Data.Typeable                   (TypeRep, Typeable, typeOf)
 import           Data.Vector                     (Vector)
-import           Distribution.System             (Arch, OS)
+import           Distribution.System             (Arch, OS (..))
 import qualified Distribution.Text               as DT
 import qualified Distribution.Version            as C
 import           GHC.Generics                    (Generic)
@@ -386,3 +387,9 @@ data MiniPackageInfo = MiniPackageInfo
     }
     deriving (Generic, Show, Eq)
 instance Binary.Binary MiniPackageInfo
+
+
+isWindows :: OS -> Bool
+isWindows Windows = True
+isWindows (OtherOS "windowsintegersimple") = True
+isWindows _ = False


3noch:
GHC 7.10.1 builds but with xz compression:

  * [32-bit](https://drive.google.com/file/d/0B1Q2wj7eQ5WbOUJCRHhDanpqU0E/view?usp=sharing)
  * [64-bit](https://drive.google.com/file/d/0B1Q2wj7eQ5WbbWRERHltSVBxZEk/view?usp=sharing)

3noch:
I'm out for the holiday weekend, so do whatever you wish to this in the meantime. I'll be back Monday.

3noch:
I think there's still something afoot. I wonder if there's a Cabal version mismatch, but my builds are failing. More details next week if you don't beat me to it.

3noch:
FWIW:

```
--  While building package unordered-containers-0.2.5.1 using:
      C:\\Users\\Elliot Cameron\\AppData\\Roaming\\stack\\programs\\i386-windowsintegersimple\\ghc-7.10.1\\bin\\runhaskell.exe -package=Cabal-1.22.2.0 -clear-package-db -global-package-db -package-db=C:\Users\Elliot Cameron\AppData\Roaming\stack\snapshots\i386-windowsintegersimple\nightly-2015-07-02\7.10.1\pkgdb\ C:\Users\ELLIOT~1\AppData\Local\Temp\stack14932\Setup.hs --builddir=.stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\ build
    Logs have been written to: "C:\\Code\\windows-client\\.stack-work\\logs\\unordered-containers-0.2.5.1.log"

    Configuring unordered-containers-0.2.5.1...
    Building unordered-containers-0.2.5.1...
    Preprocessing library unordered-containers-0.2.5.1...
    [1 of 8] Compiling Data.HashMap.UnsafeShift ( Data\HashMap\UnsafeShift.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\UnsafeShift.o )
    [2 of 8] Compiling Data.HashMap.PopCount ( Data\HashMap\PopCount.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\PopCount.o )
    [3 of 8] Compiling Data.HashMap.Unsafe ( Data\HashMap\Unsafe.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\Unsafe.o )
    [4 of 8] Compiling Data.HashMap.Array ( Data\HashMap\Array.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\Array.o )
    [5 of 8] Compiling Data.HashMap.Base ( Data\HashMap\Base.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\Base.o )

    Data\HashMap\Base.hs:85:1: Warning:
        The import of `Data.Functor' is redundant
          except perhaps to import instances from `Data.Functor'
        To import instances alone, use: import Data.Functor()
    [6 of 8] Compiling Data.HashMap.Strict ( Data\HashMap\Strict.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\Strict.o )
    [7 of 8] Compiling Data.HashMap.Lazy ( Data\HashMap\Lazy.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashMap\Lazy.o )
    [8 of 8] Compiling Data.HashSet     ( Data\HashSet.hs, .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\Data\HashSet.o )
    C:\Users\Elliot Cameron\AppData\Roaming\stack\programs\i386-windowsintegersimple\ghc-7.10.1\mingw\bin\ar.exe: .stack-work\dist\i386-windowsintegersimple\Cabal-1.22.2.0\build\libHSunordered-containers-0.2.5.1-AP0g8NmrBD35Ls9w1ZDHH1.a-13844\libHSunordered-containers-0.2.5.1-AP0g8NmrBD35Ls9w1ZDHH1.a: No such file or directory
```

3noch:
Also, perhaps something needs to be uploaded still?

```
C:\Code\windows-client\build>stack --arch=x86_64 setup
Unable to find installation URLs for OS key: windowsintegersimple64
```

src/Stack/Config.hs
```diff
@@ -47,7 +47,7 @@ import qualified Data.Set as S
 import qualified Data.Text as T
 import           Data.Text.Encoding (encodeUtf8, decodeUtf8)
 import qualified Data.Yaml as Yaml
-import           Distribution.System (OS (Windows), Platform (..), buildPlatform)
```
snoyberg: This change appears unnecessary.

src/Stack/Package.hs
```diff
@@ -63,7 +63,7 @@ import           Distribution.Package hiding (Package,PackageName,packageName,pa
 import           Distribution.PackageDescription hiding (FlagName)
 import           Distribution.PackageDescription.Parse
 import           Distribution.Simple.Utils
-import           Distribution.System (OS, Arch, Platform (..))
+import           Distribution.System (OS (..), Arch, Platform (..))
```
snoyberg: As does this one

snoyberg:
Actually, looks verygood. There are two seemingly unnecessary import changes, but that's not a problem. Merging, thanks!

src/Stack/Package.hs
3noch: We need `Windows`.

src/Stack/Config.hs
3noch: Oops... true.

3noch:
@snoyberg Any idea what's going on with `unordered-containers`? Is it a bad GHC build or perhaps the build plan is concluding that the dependency hash is one thing but the actual build (which is not depending on the same build of `hashable` as normal) produces a different hash?

snoyberg:
No I'm not sure. I can try to investigate next week, but I don't have much
experience with custom builds like this.

3noch:
I think there must still be something amiss with the build plan because it works when I hijack the GHC installation. In other words, the following process produces a working build:

  1. Delete every conceivable stack cache, both globally and in the repo
  2. Remove `os: windowsintegersimple` constraint (we'll use normal GHC bindist)
  3. Run `stack --arch=i386 setup` and wait for it to download the normal 32-bit GHC 7.10.1 bindist
  4. In `AppData\Roaming\stack\i386-windows` rename `ghc-7.10.1` to something else and copy in the custom integer-simple build
  4. Run `stack --arch=i386 build`
  5. Voila! It works.

I've also noticed that the `windowsintegersimple` download ends up in `AppData\Roaming\stack\programs` instead of `AppData\Local\stack` like the normal bindist.

3noch:
Also note that when hijacking the GHC folder, I **cannot** build my project without specifying custom flags to turn off integer-gmp. It fails with an error that I need `integer-gmp`.

3noch:
Perhaps treating `integer-simple` GHC as a new OS is too heavy handed? It seems that there are numerous places where `stack` is simply confusing the real OS. Would adding a different configuration setting to select some sort of custom build for a given OS make the job easier?

snoyberg:
I think you're right, we will need some kind of extra tagging for GHC builds. I think @manny-fp is going to need something like that for adding variants of GHC for Linux based on GMP versions. Perhaps you two could come up with some kind of a plan for what this would look like? I think basically we just want to extend `Platform` to include an extend field containing some arbitrary text.

manny-fp:
So would we want to use the extended Platform basically wherever Distribution.System.Platform is used now (in the configuration, etc)?  Having only glanced at the code, that looks like it would probably work.  I could move the libgmp.so.3 detecting code (https://github.com/commercialhaskell/stack/commit/15ce80ca9dad9674621f1e6cdd2747395c8695ff) to where the config is loaded, and that would allow it to be overridden on the command-line.

snoyberg:
Yes, that's my idea, and make sure that all directories (such as inside
dist) use the tag as well.

3noch:
That's the direction I was thinking too when I asked the question. Being largely unfamiliar with the code, I can't tell if `Platform` is primarily targeted at answering "What build of GHC do I use?" or if it is integrated in more complex ways. Of course, which GHC build is used ties into a lot of other things. In other words, I have a hard time peering deeply enough into the code's intent to know where to place the tagging.

3noch:
It almost seems as if GHC could merely be part of the dependency tree somehow. Of course, it would not fall under the treatment of typical cabal dependencies. I suppose this connection is already present with the `resolver` deciding which GHC to use. Perhaps we need the custom tagging to exist on the resolver itself.

3noch:
Shall I open a new issue to add custom build tagging on the `resolver`?

snoyberg:
Discussing in a new issue is a good idea

Issue #496, opened 2015-07-03
Allow passing options to benchmarks


Toxaris:
I want to pass `--output something.html` to a criterion benchmark. `stack bench` seems to not support passing options to benchmarks, or maybe I'm missing something.

With cabal, I would do `cabal bench --benchmark-option="--output=$benchmark.html"`. If there were two benchmarks `bench1` and `bench2` in my project, that would build them both and then run them as `bench1 --output bench1.html` and `bench2 --output bench2.html`.

snoyberg:
We have code in place for this to work for tests, it should be almost identical for benchmarks. Would you like to take a stab at implementing this?

markus1189:
Implemented in #512.  The stack equivalent is:
```
$ stack bench --benchmark-arguments '--output=$benchmark.html'
```
@snoyberg I'd say close :)

snoyberg:
Good call, thanks again @markus1189!

Toxaris:
Ah excellent, I upgraded stack to get that feature and it seems to work fine. Sorry for the noise.

PR #512, opened 2015-07-03
Implement `stack bench --benchmark-arguments`

markus1189:
Implements #496.

I renamed the `boptsTestArgs` to `boptsAdditionalArgs` and use that field for both testing and benchmarking.  I also chose to use `--benchmark-arguments` instead of `--benchmark-options` because for tests it is `--test-arguments`.

With this `stack bench --benchmark-arguments='-o $benchmark.html --csv $benchmark.csv'` works as expected.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -123,7 +123,7 @@ printPlan finalAction plan = do
     let mfinalLabel =
             case finalAction of
                 DoNothing -> Nothing
-                DoBenchmarks -> Just "benchmark"
+                DoBenchmarks _ -> Just "benchmark"
                 DoTests _ -> Just "test"
     case mfinalLabel of
         Nothing -> return ()
@@ -405,8 +405,8 @@ toActions runInBase ee (mbuild, mfinal) =
     mfunc =
         case boptsFinalAction $ eeBuildOpts ee of
             DoNothing -> Nothing
-            DoTests rerunTests -> Just (singleTest rerunTests, checkTest)
-            DoBenchmarks -> Just (singleBench, checkBench)
+            DoTests topts -> Just (singleTest (toRerunSuccessful topts), checkTest)
+            DoBenchmarks _ -> Just (singleBench, checkBench)
 
     checkTest task =
         case taskType task of
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -701,7 +701,8 @@ singleTest topts ac ee task =
                     TTLocal lp -> lpDirtyFiles lp
                     _ -> assert False True) ||
                 not testBuilt
-            needHpc = boptsCoverage (eeBuildOpts ee)
+
+            needHpc = toCoverage topts
 
             componentsRaw =
                 case taskType task of
@@ -721,19 +722,19 @@ singleTest topts ac ee task =
             setTestBuilt pkgDir
 
         toRun <-
-            if (boptsNoTests (eeBuildOpts ee))
-                then do
-                    announce "Test running disabled by --no-run-tests flag."
-                    return False
-                else if toRerunTests topts
-                    then return True
-                    else do
-                        success <- checkTestSuccess pkgDir
-                        if success
-                            then do
-                                unless (null testsToRun) $ announce "skipping already passed test"
-                                return False
-                            else return True
+            if toDisableRun topts
+              then do
+                  announce "Test running disabled by --no-run-tests flag."
+                  return False
+              else if toRerunTests topts
+                  then return True
+                  else do
+                      success <- checkTestSuccess pkgDir
+                      if success
+                          then do
+                              unless (null testsToRun) $ announce "skipping already passed test"
+                              return False
+                          else return True
 
         when toRun $ do
             bconfig <- asks getBuildConfig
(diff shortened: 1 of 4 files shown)


snoyberg:
Looks good at first blush. I'm on my phone right now and probably won't be able to properly review until Sunday. If someone else wants to review and merge I have no objection.

src/Stack/Build/Execute.hs
```diff
@@ -883,8 +883,9 @@ singleBench ac ee task =
             config <- asks getConfig
             cabal (console && configHideTHLoading config) ["build"]
 
+        let args = boptsAdditionalArgs (eeBuildOpts ee)
         announce "benchmarks"
-        cabal False ["bench"]
+        cabal False ["bench","--benchmark-options=" <> (intercalate " " args) <> ""]
```
snoyberg: This won't properly handle arguments that have spaces in them, we'd need some kind of quoting mechanism to be used instead. I'm also unsure what the `<> ""` at the end is trying to do.

markus1189: The `<> ""` was a check for your watchful eyes, obviously :)

markus1189:
You're right about the need for a quoting mechanism, any suggestions?

snoyberg:
I can't think of a good prebuilt solution. @chrisdone are you aware of anything?

manny-fp:
What about using multiple `--benchmark-option` (singular) arguments instead?

manny-fp:
Alternatively, could just take `--benchmark-arguments` as a string rather than parsing it into a list, and pass it verbatim to `--benchmark-options`?  But I think I like the former suggestion better, that way the quoting behaviour is consistent with others in stack and not platform-dependent.

markus1189:
Actually things like
```
$ stack bench --benchmark-arguments='--csv $benchmark.csv -o $benchmark.html -v 2'
```
work, we get `args = ["--csv","$benchmark.csv","-o","$benchmark.html","-v","2"]` which (curiously to me) works with `cabal False ["bench","--benchmark-options=" <> unwords args]`

The problems occur if one of the arguments has spaces, as noted correctly above by @snoyberg:

```
$ stack bench --benchmark-arguments='--csv "my $benchmark".csv -o "my $benchmark.html" -v 2'
```
The `args` list looks like this: `["--csv","my $benchmark.csv","-o","my $benchmark.html","-v","2"]`.

I don't think we can use `--benchmark-option`, because we still have the problem about quoting the `my $benchmark.html` filename containing spaces...

markus1189:
Okay I just switched out the `argsOption` parser for optparse's `strOption`, as `argsOption` unescapes the string.  Now it works:

```
$ stack bench --benchmark-arguments='--csv "my $benchmark.csv" -o "my $benchmark.html" -v 2'
```
results in `args = ["--csv \"my $benchmark.csv\" -o \"my $benchmark.html\" -v 2"]`.

I chose to still rename the field in `BuildOpts` to `boptsAdditionalArgs` and use it for benchmarks, in theory we could also keep the old `boptsTestArgs` and add a special field `boptsBenchArgs`.

What speaks for the first is that we keep the number of fields in `BuildOpts` minimal, what speaks for the second is that `boptsTestArgs` is `[String]` which does not make sense for benchmarks because we use only one string and pass it on, though we can just use a singleton list as I am doing right now.

TL;DR: Rename and reuse `boptsTestArgs` or add special field `boptsBenchArgs` to `BuildOpts`?

snoyberg:
I'll throw a new option into the mix: add fields to the `FinalAction` type so that test and benchmark specific options do not appear in `BuildOpts`. I've already done that for rerunning tests, I think it would make sense to do that more generally. This could look like:

```haskell
data FinalAction
    = DoTests TestOpts
    | DoBenchmarks BenchmarkOpts
    | DoNothing
```

Anyone see a downside to that?

chrisdone:
I thought about adding stuff to `FinalAction` a couple times, and extending the args parser so that each action can have its own parser. Not opposed.

snoyberg:
@markus1189 Not intending to hoist extra work onto you that you didn't intend to sign up for, but are you interested in taking the stab at making this change? It should be mostly mechanical and pretty well isolated from other things.

markus1189:
No problem at all, but thanks for asking anyway.

Summary:
- add `TestOpts` to `DoTests` and `BenchmarkOpts` to `DoBenchmarks`
- remove the previously introduced `boptsAdditionalArgs` in favor of the fields in `FinalAction`

and I also took the liberty to move the flags for `--no-run-tests` and `coverage` to `TestOpts` as well as fix #537 because I was touching it anyway.

Do you want me to squash the commits into one with a summary in the body?

src/Stack/Build/Types.hs
```diff
@@ -320,19 +315,28 @@ defaultBuildOpts = BuildOpts
...
     , boptsOnlySnapshot = False
-    , boptsCoverage = False
     , boptsFileWatch = False
     , boptsKeepGoing = Nothing
-    , boptsNoTests = False
     }
 
+-- | Options for the 'FinalAction' 'DoTests'
+data TestOpts =
+  TestOpts {toRerunTests :: !Bool -- | Whether successful tests will be run gain
```
snoyberg: I think you want to use `-- ^` instead

snoyberg:
I just had a minor comment about haddock syntax (`-- ^` instead of `-- |`). Other than that, this looks great, thank you!

markus1189:
Good catch. Changed the haddock for `TestOpts` and `BenchmarkOpts`.

Issue #430, opened 2015-06-26
Request: Some way of logging that tests build but don't pass


JulianBirch:
I'd like to be able to "auto commit" anything where the tests build, but log a different message depending on whether the tests pass. I don't think this is currently possible. `stack test --no-test` or similar would solve the problem, as would a more general system where you could insert commands into the process.

Apologies if this is a solved problem that I can't figure out.

snoyberg:
It's not solved yet, and should be pretty trivial to implement. Want to give a shot at a PR? The relevant code is the `singleTest` function in `Stack.Build.Execute`, and you'll likely want to add a field to `BuildOpts` (or perhaps add a field to the `DoTests` constructor of `FinalAction`).

JulianBirch:
Thanks I'll give it a go. (I'm a rank beginner, so this may take me a while.)

snoyberg:
Hit us up with questions, happy to help

JulianBirch:
Okay, I really can't figure this bit out: after adding something to BuildOpts, where do I register the command line flag and command line documentation? I'm finding it pretty hard to find anything that actually creates one except from Upgrade.hs.

snoyberg:
@JulianBirch Just make the change and try to compile, GHC should help you here. The relevant location is src/main/Main.hs, the `buildOpts` parser.

JulianBirch:
Thanks, I should have thought of that. I've also learned about strict fields. :)

snoyberg:
Pull request sent and merged, thanks!

Blaisorblade:
Sorry if I ask, but why is building tests only reachable through `test` (creating the weird `test --no-tests` or `test --no-run-tests`) rather than `build --test`?

snoyberg:
Not sure if I parsed your comment correctly, but I think that's exactly how stack behaves today. Perhaps toy can clarify with a command line that's not doing what you expected?

Blaisorblade:
I think `cabal test --no-run-tests` shouldn't exist (but it does exist on 0.1.3.1), but apparently the interface changed further, and there's (also) a more sensible interface.

snoyberg:
I actually use that command line occasionally.

PR #513, opened 2015-07-03
Add --no-tests parameter to stack test, which disables the running of tests.

JulianBirch:
The addresses #430. I've also split out the actual "run tests" part of `singleTest`. I can undo that if you'd prefer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.swp
 *.tag
 *~
+*_flymake.hs
 .hsenv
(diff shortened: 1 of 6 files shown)


snoyberg:
I haven't reviewed this yet, but just noticed that this has a merge conflict. Could you rebase against master?

JulianBirch:
Okay, the `singleTest` has been changed sufficiently, I've gone with a different approach.

Issue #506, opened 2015-07-03
Stack path should strip trailing path separators from output


DanBurton:
Context: http://www.reddit.com/r/haskell/comments/3byf25/stack_hdevtools_together_at_last/csqzlf4

@snoyberg verify that I've correctly summarized the issue?

snoyberg:
This is correct. Should be easy to implement for someone looking to dive into the code base.

PR #518, opened 2015-07-04
Drop trailing slashes in `stack path` output

markus1189:
Implements #506. Is it that easy or did I miss anything special? :)


DanBurton:
Should be that easy. I'm away so I'll leave it to others to merge.

snoyberg:
Awesome, thanks!

PR #524, opened 2015-07-06
Report `dist` path as `HASKELL_DIST_DIR` env variable

jdnavarro:
This is my naive attempt to make the `dist` path available for cases like running `doctest`. See https://groups.google.com/forum/#!topic/haskell-stack/OoP88_VckhY.

I tried to imitate how the other env variables are set without fully understanding what the consequences may be. Testing it manually works as I would expect.


snoyberg:
Thanks!

Blaisorblade:
For reference: haskell/cabal#3483 is about adding the same interface to cabal.

Issue #519, opened 2015-07-05
Allow global options both before and after command


snoyberg:
I believe this is a regression, since it seemed to work properly before. See #505. Both of the following should work, but currently only the latter does. This may be an issue in optparse-simple:

```
stack build --resolver lts-2.16
stack --resolver lts-2.16 build
```

Assigning to @manny-fp, but pinging @chrisdone who may have a quick answer.

DanBurton:
I thought it had been this way for a long time / always. (Global opts must
come before the command.)

On Sunday, July 5, 2015, Michael Snoyman <notifications@github.com> wrote:

> I believe this is a regression, since it seemed to work properly before.
> See #505 <https://github.com/commercialhaskell/stack/issues/505>. Both of
> the following should work, but currently only the latter does. This may be
> an issue in optparse-simple:
>
> stack build --resolver lts-2.16
> stack --resolver lts-2.16 build
>
> Assigning to @manny-fp <https://github.com/manny-fp>, but pinging
> @chrisdone <https://github.com/chrisdone> who may have a quick answer.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/commercialhaskell/stack/issues/519>.
>


-- 
-- Dan Burton

3noch:
I recall getting tripped up several times and needing to move `--arch` before the command, ever since that option was introduced. I think it's been this way for a while.

manny-fp:
I'm pretty sure it's been this way since the start, and it is normal behaviour for optparse-applicative.  That said, it's really annoying and worth fixing.

chrisdone:
Haven't rebuilt stack for a week, seems that `--resolver` is handled in both cases on my commit:

```
chris@retina:~/Packages/ace$ stack build --resolver lts-2.12
Downloading lts-2.12 build plan ...^Cuser interrupt
chris@retina:~/Packages/ace$ stack  --resolver lts-2.12 build
Downloading lts-2.12 build plan ...^Cuser interrupt
chris@retina:~/Packages/ace$ stack --version
Version 0.1.1.0, Git revision 66ce026d5cf61d0d5e34dd2379f4096578ac8e33
chris@retina:~/Packages/ace$ 
```

So if this isn't working on master, it's a regression.

chrisdone:
> I'm pretty sure it's been this way since the start, and it is normal behaviour for optparse-applicative. That said, it's really annoying and worth fixing.

If that were true, `stack build -v` would never have worked.

[...]

PR #542, opened 2015-07-08
global -p option

drwebb:
@snoyberg Any suggestions for naming these options: `--local-bin-path` seems to make sense, but I'm not a fan of `-p`. Also note that the invocation will be a lot more natural once #519 is implemented.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -29,6 +29,7 @@ import qualified Codec.Archive.Tar as Tar
 import qualified Codec.Compression.GZip as GZip
 import           Control.Applicative
 import           Control.Concurrent (getNumCapabilities)
+import           Control.Exception  (IOException)
 import           Control.Monad
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
@@ -153,9 +154,9 @@ configFromConfigMonoid configStackRoot mproject configMonoid@ConfigMonoid{..} = 
        Just userPath -> do
            tryPath <- try (liftIO $ canonicalizePath userPath >>= parseAbsDir)
            case tryPath of
-               Left (_ :: SomeException) ->
-                   error $ "Could not locate user specified directory \"" ++
-                      userPath ++ "\""
+               Left (e :: IOException) ->
+                   error ("Could not locate user specified directory \"" ++ userPath ++ "\"\n" ++
+                          "IOException: " ++ (show e))
                Right absPath -> return absPath
 
      configJobs <-
@@ -227,7 +228,6 @@ configOptsParser docker =
             idm
     <*> optional (strOption
              ( long "local-bin-path"
-             <> short 'p'
              <> metavar "DIR"
              <> help "Install binaries to DIR"
               )) 
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -146,9 +146,17 @@ configFromConfigMonoid configStackRoot mproject configMonoid@ConfigMonoid{..} = 
                 return $ progsDir </> $(mkRelDir stackProgName) </> platform
             _ -> return $ configStackRoot </> $(mkRelDir "programs") </> platform
 
-     configLocalBin <- do
-        localDir <- liftIO (getAppUserDataDirectory "local") >>= parseAbsDir
-        return $ localDir </> $(mkRelDir "bin")
+     configLocalBin <- case configMonoidLocalBin of
+       Nothing -> do
+           localDir <- liftIO (getAppUserDataDirectory "local") >>= parseAbsDir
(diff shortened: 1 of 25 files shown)


snoyberg:
I'm -1 on the `-p` shorthand. I'd say `--install-path` would be better, but perhaps @manny-fp would have input. Frankly, even `--path` seemed OK, but now that it's a global option I can see making it a bit more explicit would be worthwhile.

src/Stack/Config.hs
```diff
@@ -146,9 +146,17 @@ configFromConfigMonoid configStackRoot mproject configMonoid@ConfigMonoid{..} =
...
 
-     configLocalBin <- do
-        localDir <- liftIO (getAppUserDataDirectory "local") >>= parseAbsDir
-        return $ localDir </> $(mkRelDir "bin")
+     configLocalBin <- case configMonoidLocalBin of
+       Nothing -> do
+           localDir <- liftIO (getAppUserDataDirectory "local") >>= parseAbsDir
+           return $ localDir </> $(mkRelDir "bin")
+       Just userPath -> do
+           tryPath <- try (liftIO $ canonicalizePath userPath >>= parseAbsDir)
```
snoyberg: This won't work if the directory doesn't already exist. Possible solution: call `createDirectoryIfMissing True userPath` first. Not sure if it's worth it.

src/Stack/Config.hs
```diff
@@ -146,9 +146,17 @@ configFromConfigMonoid configStackRoot mproject configMonoid@ConfigMonoid{..} =
...
+           case tryPath of
+               Left (_ :: SomeException) ->
```
snoyberg: This shouldn't be a `SomeException`, but an `IOException`. Also, displaying that information to the user would be a good thing.

drwebb:
Summary of commit be8dbb1:

Dropped `-p` short option. I like `--path` as well but `--local-bin-path` is a little more explicit that we are just copying the binaries to that location. I'm still up to changing to `--install-path` though might imply other things than just the bin. As it it `--local-bin-path` is clear at the expense of looking a little unconventional.  

I am opposed to employing `createDirectoryIfMissing`. We want the user to see something like
```
λ church stack → λ git userlocalbin* → ~/.local/bin/stack --local-bin-path ./binz install
Could not locate user specified directory "./binz"
IOException: ./binz: canonicalizePath: does not exist (No such file or directory)
```

Basically fail fast and hard on a typo, not silently doing what the user doesn't want.

The other change was to match only on IOException, which is the only thing that I think canonicalizePath can throw. Looking at the code for Path, it can throw a PathParseException which I will go ahead and cover.

drwebb:
Summary of 55a5476:
Also catching PathParseExceptions right there, never been able to trigger it (canonicalizePath seems to catch any nastyness) though.

drwebb:
@snoyberg: I talked with @manny-fp about the naming and he also liked `--local-bin-path` as one can already use `stack path --local-bin-path`. I should also make sure the new field in the `stack.yaml` will follow the same convention, so if this looks good to you I'll handle the merge.

snoyberg:
@manny-fp Please merge when you're ready, I'm fine with this PR.

src/Stack/Config.hs
```diff
@@ -146,9 +147,23 @@ configFromConfigMonoid configStackRoot mproject configMonoid@ConfigMonoid{..} =
...
-        return $ localDir </> $(mkRelDir "bin")
+     configLocalBin <-
+         case configMonoidLocalBin of
+             Nothing -> do
+                 localDir <- liftIO (getAppUserDataDirectory "local") >>= parseAbsDir
+                 return $ localDir </> $(mkRelDir "bin")
+             Just userPath ->
+                 (liftIO $ canonicalizePath userPath >>= parseAbsDir)
+                 `catches`
+                 [Handler (\(ex :: IOException) -> error
```
manny-fp: Should not be using `error`, instead throw a new [ConfigException](https://github.com/commercialhaskell/stack/blob/fc963d1f0faded90fb8ca7e5375613a0b11b67c4/src/Stack/Types/Config.hs#L515) constructor.

src/Stack/Types/Config.hs
```diff
@@ -500,6 +504,7 @@ instance FromJSON ConfigMonoid where
          configMonoidExtraIncludeDirs <- obj .:? "extra-include-dirs" .!= Set.empty
          configMonoidExtraLibDirs <- obj .:? "extra-lib-dirs" .!= Set.empty
          configMonoidConcurrentTests <- obj .:? "concurrent-tests"
+         configMonoidLocalBin <- obj .:? "local-bin"
```
manny-fp: Should this be `"local-bin-path"` for consistency with the command-line argument name?

Issue #538, opened 2015-07-07
Support options like --package for the ghci command.


seanparsons:
It would be really handy to be able to launch a REPL with some random dependencies. Since the runghc command supports this I would assume this to be trivial?

chrisdone:
Currently you can run `stack ghci --ghc-options "-package x"`.

snoyberg:
I had read @seanparsons's comment as requesting a way to tell `stack` to install a package before running `ghci`, like `stack runghc -package foo bar.hs` would force the foo package to be available before running bar.hs. Am I correct?

@seanparsons The code used for `runghc` should transfer easily to ghci, would you be interested in trying out a PR to add this?

seanparsons:
@snoyberg That is exactly what I was thinking. :+1:

seanparsons:
I might give adding that feature a bash.

snoyberg:
Cool, just ping if you need some advice.

markus1189:
@seanparsons did you decide to work on this? Otherwise I will give this a shot myself :)

seanparsons:
@markus1189 I've quite literally just forked the repo to give it a bash myself! :)

markus1189:
have fun =)

seanparsons:
@snoyberg Since this is in master, will this make it into the 0.2.0.0 milestone?

snoyberg:
It will be in the next release.

PR #550, opened 2015-07-09
Added in support for using the --package option with ghci.

seanparsons:
I believe this should fix https://github.com/commercialhaskell/stack/issues/538


chrisdone:
Looks good in the commit. Going to test it. :-)

chrisdone:
Looks good! Thanks!

PR #551, opened 2015-07-09
Use `value` instead of fromMaybe for arguments

markus1189:
Instead of using the pattern:

```haskell
fmap (fromMaybe "default value")
     (optional (_parser (_meta)))
```

we can use `optparse-applicative`'s [value](http://hackage.haskell.org/package/optparse-applicative-0.11.0.2/docs/Options-Applicative-Builder.html#v:value) function to specify a default value:

```haskell
_parser (_meta <> value "default value")
```

This is the benefit of being able to show defaults automatically by enabling it with [showDefault](http://hackage.haskell.org/package/optparse-applicative-0.11.0.2/docs/Options-Applicative-Builder.html#v:showDefault) in the help text, e.g. for `stack ghci --help` we now get (note the help for `--with-ghc`):

```
Usage: stack ghci [TARGET] [--ghc-options OPTION] [--with-ghc GHC] [--no-load]
                  [--package ARG]
  Run ghci in the context of project(s)

Available options:
  TARGET                   If none specified, use all packages defined in
                           current directory
  --ghc-options OPTION     Additional options passed to GHCi
  --with-ghc GHC           Use this command for the GHC to run (default: "ghc")
  --no-load                Don't load modules on start-up
  --package ARG            Additional packages that must be installed
```

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -183,14 +183,15 @@ main = withInterpreterArgs stackProgName $ \args isInterpreter ->
                               (many (strArgument
                                        (metavar "TARGET" <>
                                         help "If none specified, use all packages defined in current directory"))) <*>
-                         fmap (fromMaybe [])
-                              (optional (argsOption (long "ghc-options" <>
-                                                     metavar "OPTION" <>
-                                                     help "Additional options passed to GHCi"))) <*>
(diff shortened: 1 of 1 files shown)


snoyberg:
Awesome, thanks!

chrisdone:
:+1:

PR #557, opened 2015-07-10
Propose fuzzy-matched versions

### LINKED ISSUE #504: Consider versions "2.1" and "2.1.0" as same
I just had a minor but confusing problem that I've put `protocol-buffers-2.1` into `extra-deps`, but got a message `Didn't see protocol-buffers-2.1 in your package indices, updating and trying again` from `stack build`. The reason, as you might guess, is that it couldn't match versions `2.1` and `2.1.0`.

Please confirm that improving version-matcher to handle this would be a way to go. If so – I could take a look at this issue later.

k-bx:
Fixes #504.

When you get a "Didn't see pkg-<ver> in your package indices" message,
also see list of candidates with same major version. Might be that you
forgot some minor thing like ".0" in the end of a version string.

### DIFF
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -22,57 +22,59 @@ module Stack.Fetch
     , withCabalLoader
     ) where
 
-import qualified Codec.Archive.Tar as Tar
-import qualified Codec.Archive.Tar.Check as Tar
-import           Codec.Compression.GZip (decompress)
+import qualified Codec.Archive.Tar              as Tar
+import qualified Codec.Archive.Tar.Check        as Tar
+import           Codec.Compression.GZip         (decompress)
 import           Control.Applicative
-import           Control.Concurrent.Async (Concurrently (..))
-import           Control.Concurrent.MVar.Lifted (newMVar, modifyMVar)
-import           Control.Concurrent.STM          (TVar, atomically, modifyTVar,
-                                                  newTVarIO, readTVar,
-                                                  readTVarIO, writeTVar)
-import           Control.Exception (assert)
-import           Control.Monad (liftM, when, join, unless, void)
+import           Control.Concurrent.Async       (Concurrently (..))
+import           Control.Concurrent.MVar.Lifted (modifyMVar, newMVar)
+import           Control.Concurrent.STM         (TVar, atomically, modifyTVar,
+                                                 newTVarIO, readTVar,
+                                                 readTVarIO, writeTVar)
+import           Control.Exception              (assert)
+import           Control.Monad                  (join, liftM, unless, void,
+                                                 when)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
-import           Control.Monad.Reader (asks)
-import           Control.Monad.Reader (runReaderT)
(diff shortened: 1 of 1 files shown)


src/Stack/Fetch.hs
```diff
@@ -22,57 +22,59 @@ module Stack.Fetch
...
+import           Data.Set                       (Set)
+import qualified Data.Set                       as Set
+import qualified Data.Text                      as T
+import           Data.Text.Encoding             (decodeUtf8)
+import           Data.Typeable                  (Typeable)
+import           Data.Word                      (Word64)
 import           Network.HTTP.Download
 import           Path
-import           Prelude -- Fix AMP warning
+import           Prelude
```
snoyberg: This comment shouldn't get lost, as someone may then accidentally remove the import.

k-bx: @snoyberg sorry, it was lost because of stylish-haskell, will bring it back

src/Stack/Fetch.hs
```diff
@@ -274,38 +276,66 @@ withCabalLoader menv inner = do
...
+    case Map.lookup ident caches of
+        Nothing -> return Nothing
+        Just (index, cache) -> do
+            [bs] <- flip runReaderT env
+                  $ withCabalFiles (indexName index) [(ident, cache, ())]
+                  $ \_ _ bs -> return bs
+            return $ Just bs
+
+-- TODO: use 'Maybe (NonEmpty PackageIdentifier)' return-type
+fuzzyLookupCandidates :: PackageIdentifier -> PackageCaches
```
snoyberg: It would probably be more efficient to convert `PackageCaches` into a `Map PackageName (Set Version)` to avoid traversing the entire collection for every mismatched name.

k-bx:
I should note that I tested it on original repo with a problem:

Before:

```
➜  riak-haskell-client git:(protocol-buffers-nonfork) stack build
Didn't see protocol-buffers-2.1 in your package indices, updating and trying again
```

After:

```
➜  riak-haskell-client git:(protocol-buffers-nonfork) ✗ stack build
Didn't see protocol-buffers-2.1 in your package indices. Possible candidates: protocol-buffers-2.1.0. Updating and trying again.
```

src/Stack/Fetch.hs
k-bx: @snoyberg yeah, I decided to not go for efficiency, as this seems like a rare piece of functionality. But since you mentioned it – I have no problem making it efficient then.

snoyberg:
I'm going to merge in now, sans the minor efficiency concern. Feel free to send another PR if interested :)

src/Stack/Fetch.hs
k-bx: @snoyberg I updated function to be better in terms of alg complexity. It now uses `Map.splitLookup` to do a logN-split, filtering smaller elements. It then filters bigger elements by doing `takeWhile` (takes only satisfying head).

Thus, I think I don't understand your comment in full. Since we're only running this algorithm for one package (I tested, if two packages didn't match we still fail and error only for the first one), the map-reconstruction for better version-matching wouldn't make a lot of sense. Might be that it would make sense in future, when we show errors for many packages at once.

PR #560, opened 2015-07-10
Use `retry` package for download retry logic

markus1189:
I hope @DanBurton doesn't mind me stealing his todos ;)

Background: I saw the todo in `Network.HTTP.Download.Verified`:
```
-- TODO(DanBurton): use Control.Retry instead.
```
and I didn't know that package but I also had some personal code where I went with the custom `retry` function approach so I was curious about `Control.Retry`.  Turns out the `retry` package is pretty nice!

Changes:
1. add dependency on `retry >= 0.6` because the retry policy was improved starting with 0.6
2. replace the custom retry function with [Control.Retry.recovering](http://hackage.haskell.org/package/retry-0.6/docs/Control-Retry.html#v:recovering), using `limitRetries 3` as policy, so no delay (**do we want delay?**)
3. **remove the `ZeroRetries` constructor from `VerifiedDownloadException`**

The semantics of the old `retry` function should be kept, e.g. the last exception is rethrown by `recovering` if all retries fail.

My understanding of 3.:
The `ZeroRetries` case was meant to be used in case the number of retries is <= 0.  But from my point of view that is not a problem even if it is negative: that just means we won't *retry* the action at all.  Due to this reasoning I just removed it completely. If I got anything wrong please correct me here.

### DIFF
--- a/src/Network/HTTP/Download.hs
+++ b/src/Network/HTTP/Download.hs
@@ -4,7 +4,7 @@
 module Network.HTTP.Download
     ( verifiedDownload
     , DownloadRequest(..)
-    , drRetriesDefault
+    , drRetryPolicyDefault
     , HashCheck(..)
     , CheckHexDigest(..)
     , LengthCheck
@@ -70,7 +70,7 @@ download req destpath = do
             { drRequest = req
             , drHashChecks = []
             , drLengthCheck = Nothing
-            , drRetries = drRetriesDefault
+            , drRetryPolicy = drRetryPolicyDefault
             }
     let progressHook = return ()
     verifiedDownload downloadReq destpath progressHook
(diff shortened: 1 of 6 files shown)


DanBurton:
No complaints here!

My original code semantically mixed up the concept of "tries" with "retries", and this code fixes that problem, so now "retries" actually means "retries". :+1:

As a next step (not necessary for merging this PR) I think the `drRetries` field should be replaced with `drRetryPolicy :: RetryPolicy`, and `drRetryPolicyDefault` can add a short delay, e.g.

`limitRetry 3 <> constantDelay 100`.

DanBurton:
n.b. that's why `ZeroRetries` was there... it actually meant zero _tries_. Eliminating that constructor was the correct thing to do here.

markus1189:
Okay I just went ahead and replaced `drRetries` with `drRetryPolicy` as requested.  I added a default constant delay of 100ms (100000 μs) as suggested.  The only gotcha was that `DownloadRequest` can no longer derive `Show` but seems like it wasn't used anyway.  If we want `DownloadRequest`s to be showable we could add one that just ignores the retry policy.

markus1189:
Okay I don't know what happened here, but I just noticed that it's not yet in `master`.  Really strange I can't remember why I would close this pr, anything I forgot?

PR #564, opened 2015-07-12
Refactoring Options Parsers

drwebb:
All Options Parsers have been moved out from Main.hs
and other Modules and in to one single location at Stack.Options.

Other types that existed solely for configuration, have been
moved to Stack.Types.Config.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -21,8 +21,8 @@
 -- probably default to behaving like cabal, possibly with spitting out
 -- a warning that "you should run `stk init` to make things better".
 module Stack.Config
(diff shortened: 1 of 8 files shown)


drwebb:
@chrisdone: Could you take a look and see if you like where everything is ending up? This change is relatively minor, but goes a long way in cleaning out Main.hs and makes automatic documentation a possibility.

chrisdone:
FWIW it's usually preferable to have every commit compile; I was confused as to where all the code went. Anyhoo, looks good.

drwebb:
yeah, that was me forgetting to `git add`

PR #572, opened 2015-07-13
Better fuzzy-matching algorithm (n -> logn)

k-bx:
I noticed that you merged [my previous PR](https://github.com/commercialhaskell/stack/pull/557) (thanks!), this PR is an improvement from what was asked there.

Copy of my comment from there:

@snoyberg I updated function to be better in terms of alg complexity. It now uses Map.splitLookup to do a logN-split, filtering smaller elements. It then filters bigger elements by doing takeWhile (takes only satisfying head).

Thus, I think I don't understand your comment in full. Since we're only running this algorithm for one package (I tested, if two packages didn't match we still fail and error only for the first one), the map-reconstruction for better version-matching wouldn't make a lot of sense. Might be that it would make sense in future, when we show errors for many packages at once.

### DIFF
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -331,11 +331,16 @@ fuzzyLookupCandidates :: PackageIdentifier -> PackageCaches
 fuzzyLookupCandidates (PackageIdentifier name ver) caches =
     if null sameMajor then Nothing else Just (map fst sameMajor)
   where
-    sameIdentCaches = filter (\(PackageIdentifier n _, _) -> name == n)
-                             (Map.toList caches)
     sameMajor = filter (\(PackageIdentifier _ v, _) ->
                              getMajorVersion ver == getMajorVersion v)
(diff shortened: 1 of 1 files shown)


snoyberg:
I may not be reading everything carefully enough. I thought it was possible to be generating more than one fuzzy-error-message at a time, in which case generating a package version lookup `Map` in advance and sharing it for multiple `fuzzyLookupCandidate` calls would help with the overhead. But I could be mistaken.

Your call on whether this should be merged: I see it as an improvement, but you understand the code better than I do.

And TIL about `splitLookup`, which is good :)

k-bx:
@snoyberg yeah, as soon as we want multi-package-error-reporting we might need a bigger refactoring as you noted, but it would be an overkill to do right now for single-error case.

I'm +1 for merging this, as it seems as quite simple improvement.

p.s.: I also didn't know about splitLookup before this :)

PR #576, opened 2015-07-13
Allow "--force" with `stack new` (#575)

### LINKED ISSUE #575: "stack new" not functional if a license is already present.
When a license was already present in my project directory, the "stack new" command did nothing but cause a message to display. I think it would be better if 'stack' would do everything it normally does, other than create a license, when a license is already present. 

By the way, I am very impressed with 'stack', and very grateful to you for providing it. 'threadscope' is not available from Stackage, so I installed it using the Github repository. 'stack' ascertained that a dependency was part of GHC-7.10.1 and installed it.  

Back to the issue. The message I received was:
```
NOTE: Currently stack new functionality is very rudimentary
There are plans to make this feature more useful in the future
For more information, see: https://github.com/commercialhaskell/stack/issues/137
For now, we'll just be generating a basic project structure in your current directory
The following files already exist, refusing to overwrite:
- LICENSE
```

markus1189:
Resolves #575.

`new` was already using the `initOptsParser` which has the `forceOverwrite` option, so I just forwarded that to `Stack.New.newProject` and check for that options value before refusing the overwrite.

### DIFF
--- a/src/Stack/Init.hs
+++ b/src/Stack/Init.hs
@@ -199,7 +199,7 @@ data InitOpts = InitOpts
     { ioMethod :: !Method
     -- ^ Preferred snapshots
     , forceOverwrite :: Bool
-    -- ^ Force overwrite of existing stack.yaml
+    -- ^ Overwrite existing files
     , includeSubDirs :: Bool
     -- ^ If True, include all .cabal files found in any sub directories
     }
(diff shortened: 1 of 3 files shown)


snoyberg:
Beautiful, that additional output to the user is the perfect level of guidance.

Issue #401, opened 2015-06-24
Provide a simple way to produce opencontainers.org image(s) for executables


dysinger:
Instead of having Dockerfiles all over our Haskell projects, let's find a simple way to package a stack project's executable(s) as container(s).  

The simplest case would be that the user provides a base container (specified in stack.yml).  Stack would then take the base image & layer on top an ENTRYPOINT set to run one executable.  There may be more than one executable in a Haskell project.  There can be only one ENTRYPOINT in a docker container. We may or may not need to produce multiple images from a Haskell project. You would receive 1 container image for each Haskell executable by default. 

If the user wanted to have a VOLUME, EXPOSE ports or add ONBUILD, they would need to include that in the base image.  The base image could generic or be specific to 1 Haskell project. It depends on the user's needs.

We could handle more cases later, but this simple approach of requiring a base container would work for a lot of projects immediately.

manny-fp:
Feel free to assign to me, as this interests me.  I think providing the option to set Dockerfile metadata as well (rather than requiring the base image to have it) would be a big benefit, and wouldn't be difficult to implement (I do something like this in the stack image build tool already).

ketzacoatl:
OMG +1.

manny-fp:
More thoughts: the stack.yaml could specify multiple images to be generated, each with its own base image, ENTRYPOINT and extra metadata.

mgsloan:
+1 this would be quite awesome

manny-fp:
I think this'd be easier to reason about if we know a bit more about what you and @ketzacoatl want to see the Docker image look like, since you have been doing a lot more deploying of images on servers than I have.  Some more unfocused thoughts from me on this include:

- By default, create images based FROM the `fpco/stack-run:lts-x.y` image corresponding to the resolver (and stack-build/stack-full image) in the stack.yaml, since that guarantees any runtime libraries required to run the executable will be available.  However, it's a fairly large base to be starting from, so if small images are your goal this will not be the way. 

- Since the stack-run image is based FROM phusion/baseimage, you might want to use phusion's my_init rather than directly setting the ENTRYPOINT to the executable.  With stack-run, if you set CMD to the executable it'll run under my_init automatically (since the default ENTRYPOINT uses it).

- You may need to augment the stack-run image with some additional data files from the repo, so there should be an easy to to COPY or ADD them.  One way could be that each target can specify a Dockerfile that's used to augment the base image.

[...]

PR #578, opened 2015-07-14
Stack can create docker image(s) from your executable(s)

dysinger:
There is still more to add to this but it works (outside of docker-in-docker)

### DIFF
--- a/src/Path/IO.hs
+++ b/src/Path/IO.hs
@@ -27,7 +27,8 @@ module Path.IO
   ,copyFile
   ,copyFileIfExists
   ,copyDirectoryRecursive
-  ,createTree)
+  ,createTree
+  ,dropRoot)
   where
 
 import           Control.Exception hiding (catch)
@@ -38,6 +39,7 @@ import           Data.Either
 import           Data.Maybe
 import           Data.Typeable
 import           Path
+import           Path.Internal (Path(..))
 import qualified System.Directory as D
 import qualified System.FilePath as FP
 import           System.IO.Error
@@ -223,6 +225,12 @@ copyDirectoryRecursive srcDir destDir =
                 case stripDir srcDir srcSubDir of
                   Nothing -> return ()
                   Just relSubDir -> copyDirectoryRecursive srcSubDir (destDir </> relSubDir))
+
+
+-- | Drop the root (either @\/@ on POSIX or @C:\\@, @D:\\@, etc. on
+-- Windows).
+dropRoot :: Path Abs t -> Path Rel t
+dropRoot (Path l) = Path (FP.dropDrive l)
 
 -- Utility function for a common pattern of ignoring does-not-exist errors.
 ignoreDoesNotExist :: MonadIO m => IO () -> m ()
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ dist/
 tags
 /_release/
 .vagrant/
+*.imports
+/.idea/
+/*.iml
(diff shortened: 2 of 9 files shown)


dysinger:
docker in docker works too as soon as stack-image-builder is updated

cocreature:
I was just trying this out and run into a small bug (at least I consider it to be one):
You need to run `stack install` before `stack image container`, otherwise it fails with `copyFile: does not exist (No such file or directory)`.

I don't really see why it should be required to install the executables globally. If there is a good reason for that this should happen automatically or at least provide a better error message.

cocreature:
When I specify `snoyberg/haskell-scratch` as my base image then my executables don't seem to be contained in the image. Changing this to `fpco/stack-build` and using the same path for the executables works just fine. No idea what could be causing this.

Also a feature request (I can open another issue for that if you want):

I'd like to have different containers for my executables. The rational behind this is that I put packages in one project that somehow logically fit together (e.g. a web scraper and a webservice presenting the scraped data) but imho it really would make more sense to have separate images for this. Obviously I could have one image and then use different containers which start separate executables but since they don't really share anything I feel like having separate images here is better.

src/Path/IO.hs
```diff
@@ -224,6 +226,12 @@ copyDirectoryRecursive srcDir destDir =
                   Nothing -> return ()
                   Just relSubDir -> copyDirectoryRecursive srcSubDir (destDir </> relSubDir))
 
+
+-- | Drop the root (either @\/@ on POSIX or @C:\\@, @D:\\@, etc. on
+-- Windows).
+dropRoot :: Path Abs t -> Path Rel t
```
manny-fp: This seems like it might better belong in Path itself (since it does no I/O).

manny-fp:
LGTM, so as far as I'm concerned go ahead and merge it whenever you're ready.

manny-fp:
> I'd like to have different containers for my executables [...]
@cocreature please file an "enhancement" issue.

dysinger:
@cocreature please do criticize :) I have thick skin & could use helpful critique.  Also this docker stuff is tricky.  I'm not completely satisfied with this effort as of today.

src/Path/IO.hs
dysinger: I know.  Chris suggested that I put it there for now. (slack chat) He said it was a staging area for possible Path additions.

mgsloan:
> I was just trying this out and run into a small bug (at least I consider it to be one):
You need to run stack install before stack image container, otherwise it fails with copyFile: does not exist (No such file or directory).

@cocreature , I ran into this as well - fixed here https://github.com/commercialhaskell/stack/commit/c028b81f478ecabef753c64d54b11952533b2021

dysinger:
thanks @mgsloan

fizruk:
Is there any documentation for this anywhere?

Also, I believe this has solved #401 (so I am adding this reference, because there was none and it took me some time to find this PR).

fizruk:
Found documentation [here](https://github.com/commercialhaskell/stack/wiki/stack.yaml#image).

nicemaker:
I got yesterday a docker image running, build with stack. But I have to call it like 
`docker run -p 5000:5000 {imagename} {app-exe}`.

What I really want is that the app-exe is started automatically, I thought `endpoint` is suppose to do that. I tried

```
endpoint:
    - app-exe
```

That doesn't work for me unfortunately. Am I missunderstanding entrypoint?

Stack 1.3.2
Docker  for Mac

Issue #446, opened 2015-06-29
Proper handling of warning/error output


snoyberg:
The output of commands like `stack dot` can sometimes be intruded upon by warning and error output. We should ensure that warning/error output is sent to stderr instead of stdout. In addition, as discussed elsewhere, we may want to silence some specific warnings that are not dangerous (specifically: the missing files referenced from a .cabal file).

chrisdone:
Note to self or usurper: Warnings/error output can be tweaked in `Stack.Types.StackT`.

chrisdone:
This was implemented in 4460c36814e8312afd8429461497faafe5955175 which is merged in.

chrisdone:
@snoyberg Can you check on your side that this does the job you intended?

snoyberg:
Confirmed

PR #585, opened 2015-07-14
Log LevelError and LevelWarn to stderr

markus1189:
Usurpation of #446.

Assumption: Only `LevelError` and `LevelWarn` are logged to stderr, the rest (`LevelInfo`, `LevelOther "sticky[-done]"` and **`LevelDebug`**) go to stdout. Or should `LevelDebug` go to stderr as well?


chrisdone:
I agree with this as-implemented. Debug is requested information, not an error, seems okay to put it to stdout.

chrisdone:
Thanks!

PR #603, opened 2015-07-15
111 custom snapshot

### LINKED ISSUE #111: Custom snapshots
Custom snapshots are essential for me and generally important for user freedom.

By 'custom' I mean the ability to specify both custom constraints and a custom download repository.

I am happy to try to work on this myself, but wanted to find out your thoughts first.

snoyberg:
Resolves #111

### DIFF
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -36,7 +36,7 @@ import           Control.Monad.State.Strict      (State, execState, get, modify,
                                                   put)
 import           Control.Monad.Trans.Control (MonadBaseControl)
 import qualified Crypto.Hash.SHA256 as SHA256
-import           Data.Aeson.Extended (FromJSON (..), withObject, withText, (.:))
+import           Data.Aeson.Extended (FromJSON (..), withObject, withText, (.:), (.:?), (.!=))
 import           Data.Binary.VersionTagged (taggedDecodeOrLoad)
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString.Base16 as B16
@@ -206,8 +206,11 @@ data ResolveState = ResolveState
     }
 
 toMiniBuildPlan :: (MonadIO m, MonadLogger m, MonadReader env m, HasHttpManager env, MonadThrow m, HasConfig env, MonadBaseControl IO m, MonadCatch m)
-                => BuildPlan -> m MiniBuildPlan
-toMiniBuildPlan bp = do
+                => Version -- ^ GHC version
+                -> Map PackageName Version -- ^ cores
+                -> Map PackageName (Version, Map FlagName Bool) -- ^ non-core packages
+                -> m MiniBuildPlan
+toMiniBuildPlan ghcVersion corePackages packages = do
     $logInfo "Caching build plan"
 
     -- Determine the dependencies of all of the packages in the build plan. We
@@ -216,9 +219,9 @@ toMiniBuildPlan bp = do
     -- remove those from the list of dependencies, since there's no way we'll
     -- ever reinstall them anyway.
     (cores, missingCores) <- addDeps True ghcVersion
-        $ fmap (, Map.empty) (siCorePackages $ bpSystemInfo bp)
+        $ fmap (, Map.empty) corePackages
 
-    (extras, missing) <- addDeps False ghcVersion $ fmap goPP $ bpPackages bp
+    (extras, missing) <- addDeps False ghcVersion packages
 
     assert (Set.null missing) $ return MiniBuildPlan
         { mbpGhcVersion = ghcVersion
@@ -229,7 +232,6 @@ toMiniBuildPlan bp = do
             ]
         }
   where
-    ghcVersion = siGhcVersion $ bpSystemInfo bp
     goCore (PackageIdentifier name version) = (name, MiniPackageInfo
                 { mpiVersion = version
                 , mpiFlags = Map.empty
@@ -238,11 +240,6 @@ toMiniBuildPlan bp = do
                 , mpiExes = Set.empty
                 , mpiHasLibrary = True
                 })
-
-    goPP pp =
-        ( ppVersion pp
-        , pcFlagOverrides $ ppConstraints pp
-        )
 
     removeMissingDeps cores mpi = mpi
         { mpiPackageDeps = Set.intersection cores (mpiPackageDeps mpi)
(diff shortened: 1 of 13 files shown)


3noch:
I'm getting this error with `resolver: "custom: integer-simple-snapshot.yaml"` where `integer-simple-snapshot.yaml` is in the same directory.

```
> stack build
CreateDirectory "C:\\Users\\Elliot Cameron\\AppData\\Roaming\\stack\\snapshots\\i386-windows\\custom: integer-simple-snapshot.yaml\\7.10.1": invalid argument (The filename, directory name, or volume label syntax is incorrect.)
```

src/Stack/Types/Config.hs
```diff
@@ -377,6 +381,7 @@ instance FromJSON Resolver where
 renderResolver :: Resolver -> Text
 renderResolver (ResolverSnapshot name) = renderSnapName name
 renderResolver (ResolverGhc (MajorVersion x y)) = T.pack $ concat ["ghc-", show x, ".", show y]
+renderResolver (ResolverCustom url) = "custom: " <> url
```
3noch: Is this not using built-in YAML syntax?

src/Stack/Types/Config.hs
```diff
@@ -365,6 +365,10 @@ data Resolver
   -- ^ Require a specific GHC major version, but otherwise provide no build
   -- plan. Intended for use cases where end user wishes to specify all upstream
   -- dependencies manually, such as using a dependency solver.
+
+  | ResolverCustom !Text
+  -- ^ A custom resolver based on the given URL. This file is assumed to be
```
3noch: I suppose this means that if you need to change your snapshot, you need to copy/paste and rename? Is there any way to detect if this assumption has been broken?

src/Stack/BuildPlan.hs
```diff
@@ -667,3 +678,43 @@ shadowMiniBuildPlan (MiniBuildPlan ghc pkgs0) shadowed =
...
+            (Map.fromList $ map addFlags $ Set.toList $ csPackages cs)
+
+data CustomSnapshot = CustomSnapshot
+    { csGhcVersion :: !Version
+    , csPackages :: !(Set PackageIdentifier)
+    , csFlags :: !(Map PackageName (Map FlagName Bool))
+    }
+instance FromJSON CustomSnapshot where
+    parseJSON = withObject "CustomSnapshot" $ \o -> CustomSnapshot
+        <$> o .: "ghc"
```
3noch: Perhaps, to be thinking of the future, this could be `compiler` and "ghc-X.Y" would refer to GHC itself. I can imagine GHCJS, Haste, or others appearing here, possibly?

src/Stack/BuildPlan.hs
```diff
@@ -667,3 +678,43 @@ shadowMiniBuildPlan (MiniBuildPlan ghc pkgs0) shadowed =
...
+        <*> o .: "packages"
+        <*> o .:? "flags" .!= Map.empty
```
3noch: Certainly not necessary at this point, but it could be quite nice to have a "base" on which this snapshot builds, similar to how `extra-deps` overrides the snapshot. You could imagine a hierarchy of snapshots which cascade into the final product. I would use it to base a snapshot on one of stackage's snapshots, and override a few things.

This seems like it would be a lot more work for not a lot more gain, but I figured I'd mention it.

src/Stack/BuildPlan.hs
snoyberg: I like that. It also makes some quoting in the yaml file unnecessary. Want to modify the branch?

src/Stack/Types/Config.hs
snoyberg: For local files, easily. It's for remote files that we have a problem since it would be inefficient to download every time. This actually seems necessary to handle relative files properly...

src/Stack/Types/Config.hs
snoyberg: It's going to be `resolver: "custom: ..."`. I originally did the same `custom-` kind of prefix as everything else, but it looks too weird to say "custom-https://..."

snoyberg:
Alright, there's a flaw in this around the naming. I think we need to modify this so that there are two parts to the resolver definition:

* The location of the file
* A unique name for the snapshot database

Perhaps this would look like:

```yaml
resolver:
    location: http://...
    name: foobarbaz
```

Then the package database would be in ~/.stack/.../custom-foobarbaz. We would then identify contents of the .yaml file by their hash to resolve problems with a user modifying that file and it not getting tracked.

Hope that wasn't horribly vague :)

snoyberg:
OK, I pushed my vague idea to this branch, if you could give this a shot I'd be obliged :)

3noch:
Great! I might not be able to test this until next week. Just out of curiosity, would it make sense to move the name *into* the snapshot definition? I'm guessing you thought of that and decided against it.

snoyberg:
I did think about it a bit, and I'm not sure. The advantage I saw with this setup is that many snapshots could be published by different people without having to worry about name collisions, and it just because the responsibility of the users of those snapshots to disambiguate. I'm not convinced this is a very strong argument, however. There may have also been a technical reason about knowing the name earlier in the process that pushed me this direction, but I don't remember for certain.

3noch:
After thinking about it, I think your `name` field is in the best place for it. However, I think there might a "best" way. If it's too much work, then using a manually defined `name` is 2nd best I think.

Here's the "best" way I've been able to think of :wink::

  * For any custom snapshot, download (or just read, if it's local) the definition and create a hash of the meaningful contents.
  * Cache this hash in `.stack-work` for the project.
  * Install the snapshot dependencies into user-wide snapshot folder for that hash (say `.../snapshots/custom/b366e93o`, for example).
  * For future builds, refer to the cached hash instead of the real file. We could cache the snapshot file itself too, if that is needed.
  * If the user deletes his project cache or clones the repo again, this process begins again.

Why this is nice:

  * We only need to download the snapshot definition once.
  * It guarantees that we have an immutable reference to the snapshot. (If `.stack-work` gets deleted, then we don't care what it was before anyway.) So if/when mutation *does* happen, we are guarded against it.
  * However, if the user *really wants* to mutate the snapshot, they can **without breaking other people**. They have to clear their cache, but that is not surprising behavior. Folks with the older hash continue using it. Folks with the new hash will generate it and use it.

snoyberg:
Iiuc, this is closer to what I'd originally envisioned, but I've specifically gone against it due to Roman's preference. The idea is that of a user makes a small tweak to a snapshot, it won't force a recompilation of everything

3noch:
This can actually be simulated with the `name` field pretty well, just by writing a script that generates the hash. I'm cool with using a `name`. Also, if we wanted to improve it in the future, it can be done pretty easily I think without breaking back-wards compatibility.

3noch:
Just tried it out. Works great! I haven't done anything heavy-duty. I used a local snapshot file and `stack` built everything in a user-wide folder.

When you say "makes a small tweak to a snapshot", do you mean, changes/additions that the user *knows* won't break his project, even if other folks are still using an older version of the snapshot?

3noch:
Would it be possible to generate these snapshots with `stack solver` or some other option? If something can automatically write `extra-deps` it seems building custom snapshots would not be difficult. I think it could be also helpful if you could select a pre-existing snapshot as a base. For example, something like this: `stack solver --resolver=nightly > my-custom-snapshot.yaml`.

snoyberg:
> When you say "makes a small tweak to a snapshot", do you mean, changes/additions that the user *knows* won't break his project, even if other folks are still using an older version of the snapshot?

Presumably. Personally, I'm in favor of fully-immutable stuff on this front, but there are certainly arguments to be made for a custom snapshot to get such minor tweaks.

> Would it be possible to generate these snapshots with `stack solver` or some other option? If something can automatically write `extra-deps` it seems building custom snapshots would not be difficult. I think it could be also helpful if you could select a pre-existing snapshot as a base. For example, something like this: `stack solver --resolver=nightly > my-custom-snapshot.yaml`.

Yes, I think this shouldn't even be that difficult. Perhaps we should make this the default behavior of `stack init --solver`. I'm not going to be able to get around to this right now, but the modifications shouldn't be too difficult to make if you want to take a crack at it.

Issue #604, opened 2015-07-16
Make output of `stack path` independent of logging.


ntc2:
The output of `stack path` is produced using `Control.Monad.Logger.logInfo`. This makes `stack path --<some specific path>` unusable, since the output can't be separated from other log output.

## Expected

I'm writing script that copies Stack-produced binaries. I would like to do e.g.
```
cp "$(stack path --local-install-root)"/bin/saw tmp/release/bin
```

## Actual

However, I'm exporting `STACK_YAML` -- we build against both GHC 7.8 and 7.10 -- and so the output of `stack path --local-install-root` includes a warning on the first line which causes the `cp` to fail:
```
Getting project config file from STACK_YAML environment
/home/conathan/v/saw-script.git/.stack-work/install/x86_64-linux/lts-2.15/7.8.4
```

Note that using `--verbosity <verbosity level>` doesn't help, since the warning and the path are both printed using `logInfo`, so you get both or neither.

## Ideas for fixes

In any case, the regular output of the `stack path` command should not be produced using `logInfo`. So, change `$logInfo` to `T.putStrLn` at https://github.com/commercialhaskell/stack/blob/8b023a8cd2ec5023be808b63897538a00bb50e59/src/main/Main.hs#L310.

I would also suggest that log messages go to stderr by default. If you have a reason for sending log messages to stdout, then `--verbosity silent` could still be used to suppress them, so this is not essential.

## Workaround

In the mean time I could reimplement `stack path --local-install-root` in terms of `stack path`:
```
stack --verbosity info path | sed -ne 's/local-install-root: //p'
```

Another workaround is to do `stack --stack_yaml=$STACK_YAML path --local-install-root`, but that will break if I stop exporting `STACK_YAML` later.

mgsloan:
> In any case, the regular output of the stack path command should not be produced using logInfo. So, change $logInfo to T.putStrLn at https://github.com/commercialhaskell/stack/blob/8b023a8cd2ec5023be808b63897538a00bb50e59/src/main/Main.hs#L310.

Sounds like a good change to me!  Feel free to make a PR :D

chrisdone:
Errors and warnings were changed in the last couple days to go to stderr by default. Did you try recent master?

manny-fp:
[...]

PR #615, opened 2015-07-16
Make `stack path` output independent of logging. (Issue #604)

ntc2:
E.g.

    $ (export STACK_YAML=stack.yaml && stack path --local-install-root)
    Getting project config file from STACK_YAML environment
    /home/conathan/Desktop/mirrored/stack-fix/stack.git/.stack-work/install/x86_64-linux/lts-2.17/7.8.4

    $ (export STACK_YAML=stack.yaml && stack --verbosity silent path --local-install-root)
    /home/conathan/Desktop/mirrored/stack-fix/stack.git/.stack-work/install/x86_64-linux/lts-2.17/7.8.4

but I still think it's a good idea to also send log messages to stderr.

### DIFF
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -13,7 +13,7 @@ import           Control.Applicative
 import           Control.Monad (void)
 import           Control.Monad.Catch (MonadCatch)
 import           Control.Monad.IO.Class
-import           Control.Monad.Logger (MonadLogger, logInfo)
+import           Control.Monad.Logger (MonadLogger)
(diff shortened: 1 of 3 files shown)


mgsloan:
Looks good so far!  Only issue is that this will break `stack dot`.  Could you please update it with some `putStrLn`s?  https://github.com/commercialhaskell/stack/blob/master/src/Stack/Dot.hs

ntc2:
@mgsloan: I added the `putStrLn`s to `stack dot`.

mgsloan:
@ntc2: Thanks, looks good to me!

PR #622, opened 2015-07-17
Allow setting the docker image name, fixes #621

### LINKED ISSUE #621: Set image name for containers
I'd like to be able to set the name that stacks gives to the docker images when using `stack image container`. At the moment it just guesses them from the directory.

cocreature:
Because just complaining is a bit lame, I took a shot at this feature. Feel free to yell at me and tell me why this is a bad idea or what needs to be changed.

### DIFF
--- a/src/Stack/Image.hs
+++ b/src/Stack/Image.hs
@@ -131,7 +131,8 @@ createDockerImage :: M e m => FilePath -> m ()
 createDockerImage dir = do
     config <- asks getConfig
     bconfig <- asks getBuildConfig
-    case maybe Nothing imgDockerBase (imgDocker (configImage config)) of
+    let dockerConfig = imgDocker (configImage config)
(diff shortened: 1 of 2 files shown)


snoyberg:
@dysinger Can you review this?

@cocreature Looks like this PR no longer merges with master, would you be able to update it?

cocreature:
@snoyberg Sorry I didn't see that it broke. Rebased and force pushed.

dysinger:
We probably want to factor in the image name here too https://github.com/cocreature/stack/blob/6784b04b2501f49ac209a4468bae85efb547c7ad/src/Stack/Image.hs#L184 where we extend the image to include specific executable entrypoints.

cocreature:
Good catch, I changed that as well. I'll be away the next few days so sorry if I don't reply.

PR #626, opened 2015-07-18
Use absolute paths in GHC error messages

snoyberg:


drwebb:
I confirmed the fix in emacs Haskell mode, all looks good to me.

Issue #313, opened 2015-06-15
Upload to hackage without cabal-install


andrewthad:
If I'm using stack to build a library that I want to upload to hackage, how can I build something that hackage will accept? I can't find any mention of this in the wiki, and [hackage's upload page](https://hackage.haskell.org/upload) doesn't give any instructions on how to make the tarball on your own.

snoyberg:
I actually just added `stack upload` support, which if given a directory will generate the tarball and then upload it. Under the surface right now, it actually does call out to cabal-install, but in the future we can change this. I originally wrote the code to do `runghc Setup.hs sdist`, but that generate tarballs in the wrong format.

I'd like to leave this open to track doing sdist without `cabal-install`. It shouldn't actually be that difficult, we'd want to:

* Call `runghc Setup.hs sdist --list-sources=tmpfile`
* Process the tmpfile for a list of all filenames
* Generate the tarball

If you or someone else reading is interested in taking a crack at this, let me know, it shouldn't be too hard.

andrewthad:
If I remember this when I'm less busy a week from now (and if no one has already done it), I may give it a try. If I switch to using the master branch of stack (which I assume has the upload command) instead of the beta that I downloaded a week ago, should that give me the ability to upload? Also, in your opinion, does that seem like a reasonably safe idea?

andrewthad:
I'm renaming this to be more consistent with its purpose on the milestone.

gregwebs:
@andrewthad given that it uses cabal-install it shouldn't have any bugs (other than those in cabal-install). I would recommend using `stack` or `stackage-upload` as soon as you can because they at least use https.

andrewthad:
I'd like to use stack right now, especially I don't even have cabal-install installed anymore. Given that `stack upload` doesn't seem to be available in the version of stack in the ubuntu repository, is it safe to use the master branch of this github repo?

snoyberg:
Yes, it's safe to use master, but unfortunately (because of what I mentioned earlier) it still depends on cabal-install being available. It shouldn't be hard to overcome that though.

andrewthad:
Thanks. That shouldn't be a problem.

andrewthad:
@snoyberg I'm using the master branch of stack, and it doesn't recognize `stack upload`. Is this still on another branch, or am I supposed to be running it differently?

snoyberg:
It's the master branch. Are you certain that the executable you're running is the one you built? What's the output of `stack --version`?

[...]

PR #627, opened 2015-07-18
Use Setup.hs to get sdist list + add tar.gz code #313

mgsloan:
Instead of using the `cabal` binary to create sdists, this uses `runhaskell Setup.hs sdist --list-sources` to get a list of paths to include in the tarball.  It then uploads this tarball without writing it to a file.

### DIFF
--- a/src/Path/IO.hs
+++ b/src/Path/IO.hs
@@ -4,6 +4,8 @@
 
 module Path.IO
   (getWorkingDir
+  ,parseRelAsAbsDir
+  ,parseRelAsAbsFile
   ,listDirectory
   ,resolveDir
   ,resolveFile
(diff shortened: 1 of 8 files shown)


snoyberg:
@mgsloan Would you be able to rebase against/merge in master?

mgsloan:
@snoyberg Sure, I've pushed a rebased commit

mgsloan:
A big part of this change is moving `getCabalSetupRunner` out of `withSIngleContext`.  Why?  To avoid needing an `ExecuteEnv`.  Computing the `ExecuteEnv` requires `BuildOpts` / `BaseConfigOpts` / etc, which aren't available to `uploadCmd`.

It's possible that there's a simpler way to do this, possibly using a default `BuildOpts` or having upload take build opts arguments.  @snoyberg Thoughts?

snoyberg:
I haven't looked at this yet, but I'm OK with the idea of using a default `BuildOpts` value in the upload/sdist code (we do stuff like that elsewhere). Would you want to refactor this PR based on that feedback, before I review?

mgsloan:
@snoyberg I gave that a try, and I think it does work out nicer this way.

Now my only qualms with this change is the addition of new exports to `Stack.Build` and `Stack.Build.Execute`.  Thoughts on those?

snoyberg:
Good work @mgsloan, merged!

PR #632, opened 2015-07-20
Set lower dependency restrictions on project-template.

qnikst:
project-template version is not precise so dependency
solver could chose older version that stack is incompatible
with.


snoyberg:
Thanks!

PR #653, opened 2015-07-21
Restore install command

afcowie:
Revert addition of a warning not to use install command. Explain that the install prefix defaults to _~/.local/bin_ and is overridable with `--local-bin-path`.

The next step is to add `--prefix` and some notion of `DESTDIR` to change that default across both binaries and other installation artifacts.

AfC

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -94,7 +94,7 @@ main = withInterpreterArgs stackProgName $ \args isInterpreter ->
                         (buildCmd DoNothing)
                         (buildOptsParser Build False)
              addCommand "install"
-                        "Identical to 'build --copy-bins', not actually a managed installation tool!"
+                        "Copy binaries to the target prefix location, presently hardcoded to ~/.local/bin. This is the only command that will mutate your system outside of this project directory"
                         installCmd
                         (buildOptsParser Build True)
              addCommand "uninstall"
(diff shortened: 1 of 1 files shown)


snoyberg:
~/.local/bin isn't hard-coded, there's already a --local-bin-path flag.

snoyberg:
This also ignores the discussion that led up to this, where users were very much confused by the fact that install didn't manage installations, leading to this wording. We can discuss whether people _should_ be confused by that naming, but the confusion certainly existed.

mgsloan:
I don't interpret the message as telling you not to use `stack install` - it isn't deprecated.  Instead it's just clarifying its meaning.

afcowie:
@snoyberg The presence of a multi-line warning message "clarifying" to not use it is morally equivalent to deprecated :)

snoyberg:
@afcowie Honestly, I'm really tired of having constant naming debates, this is not helping anyone. You saw the previous discussion that I linked to. This message will help prevent some confusion from others, and I don't see how it in any way harms people who know better. The best I can possibly imagine is moving this from Warn to Info, but I'm far beyond caring about this until someone demonstrates we're not just bikeshedding.

afcowie:
I read the bugs and threads you asked me to. The premise was wrong. There are conventions for this. `--prefix` is where stuff gets installed to. That is a one-way process; no one should expect otherwise; there hasn't been a viable non-package-managed uninstall anywhere in computing; the presence of the occasional `make uninstall` (which no one ever trusts and never work right anyway) is the outlier. The existence of GNU stow(8) is testament to that.

People _do_ expect that `make install` or the language specific equivalent will take stuff out of your project and plonk it, destructively, out in the correct places on the filesystem.

A package manager will build with `--prefix=/usr`; we've spent years getting the XDG default directories for `$HOME` et al into place, which is why I was thrilled to see you use _~/.local_ as the prefix for your binary installs.

So the whole thing can simplify down: you can get rid of `stack build --copy-bins` and `--no-copy-bins` (I mean, huh?); that **is** the purpose of `$tool install`, by convention, in every language out there, and what I'm hoping you will consider returning to for Haskell.

And looking ahead, making `stack install` the first class citizen that _does_ take responsibility for destructive updates outside of your project tree / build tool cache means that it can be the home for installing (see? look what you made me do) other important pieces to _$PREFIX/share_, _$PREFIX/lib_, and so on -- which is something we have needed in the past; cabal didn't help us there, but it would be marvelous if stack would consider it.

AfC

afcowie:
> ~/.local/bin isn't hard-coded, there's already a --local-bin-path flag.

Fixed comment in help to reflect that.

PR #654, opened 2015-07-22
Add 'stack sdist' command

mgsloan:


src/main/Main.hs
```diff
@@ -574,6 +579,18 @@ uploadCmd args go = do
                 (tarName, tarBytes) <- getSDistTarball pkgDir
                 liftIO $ Upload.uploadBytes uploader tarName tarBytes
 
+sdistCmd :: [String] -> GlobalOpts -> IO ()
+sdistCmd [] _ = error "To create an sdist tarball for the current package, please run 'stack sdist .'"
```
snoyberg: In the `sdist` case, I think defaulting to current directory is fine. This *is* in contradiction to `stack upload`, since the latter has major side-effects.

snoyberg:
Just had the one comment, once that's resolved feel free to merge.

src/main/Main.hs
mgsloan: Good point!

mgsloan:
As discussed elsewhere, now `stack sdist` with no arguments builds tarballs for all packages.

PR #660, opened 2015-07-22
Create bin install dir if missing

chreekat:


drwebb:
Is there an issue for this? It was something I thought was a "bad idea". In my opinion a copy should fail if the directory is missing, prevents typos. When you want to be explicit, you just do a `mkdir` or whatever.

chreekat:
Consider this 'raising the issue'. I think the 'typo' argument is weak,
since I could just as well make a typo with mkdir, and I never appreciate
being required to take the extra step of creating the directory I'm asking
a tool to put things into.

I don't feel strongly about it, however.

drwebb:
I know a lot of build tools are out there that happily create a directory for you. I'm assuming you have a work flow where you noticed the behavior. I'd rather maximize this features usability in the end though, so this PR is fine with me if you think it actually cuts down on some hassle.

chreekat:
@drwebb I'm actually doing this in the process of making "yesod keter" use stack. I definitely think "yesod keter" should build the (local) bin directory as needed, but I could just as easily make yesod-bin handle that instead of stack. I'll do it there.

PR #666, opened 2015-07-23
Preserve file permissions when unpacking sources

Fuuzetsu:
The `tar` package, in name of portability, doesn't preserve file
permissions when creating and unpacking archives. A sample package which
relies on permissions to be kept as they were packed with is
`ghc-parser` which executes a bash script as part of its build process.

Before this patch we'd get

```
    /run/user/1000/stack26234/ghc-parser-0.1.7.0/Setup.hs:2:1: Warning:
        Module ‘System.Cmd’ is deprecated: Use "System.Process" instead
    /bin/sh: ./build-parser.sh: Permission denied
    Configuring ghc-parser-0.1.7.0...
    Warning: 'hs-source-dirs: src-7.8.3' directory does not exist.
```

but now the package builds fine.

cc @mboes

### DIFF
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -24,6 +24,7 @@ module Stack.Fetch
 
 import qualified Codec.Archive.Tar              as Tar
 import qualified Codec.Archive.Tar.Check        as Tar
+import qualified Codec.Archive.Tar.Entry        as Tar
 import           Codec.Compression.GZip         (decompress)
 import           Control.Applicative
 import           Control.Concurrent.Async       (Concurrently (..))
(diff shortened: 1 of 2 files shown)


snoyberg:
Testing on Windows before merging, but this looks good.

snoyberg:
Hmm, I'm getting errors like this when running the integration test on Windows:

    setPermissions: does not exist (No such file or directory)

Do you have access to a Windows system to test this on?

Fuuzetsu:
@snoyberg No Windows machine access. I'm not sure why that problem would occur: `FP.</>` uses native separator, and `fromTarPath` (https://hackage.haskell.org/package/tar-0.4.2.0/docs/Codec-Archive-Tar-Entry.html#v:fromTarPath) is documented to convert to native separator. Simple way to check would be to print the path we get but as I said, I don't have access to a Windows machine. I'd be leaning towards `fromTarPath` not doing its job properly but can't check.

snoyberg:
It seems like Windows doesn't like setting permissions on a directory when a trailing backslash is present. I think I can fix this easily.

snoyberg:
Fix added for Windows and merged, thanks!

PR #667, opened 2015-07-24
Never throw an exception when stack.yaml is missing

snoyberg:
I'm not sure this is a good idea, which is why I'm putting it on a
separate branch for discussion. Pinging @manny-fp and @chrisdone.

My reasoning for this change: we currently have weird behavior where
some commands (e.g. stack install) will work outside of a project, but
others (e.g., stack build) will not. This makes a lot of sense for some
commands (like stack test), since it's almost certainly user error to
run `stack test` outside a project. However:

* This logic is difficult to explain to people
* My guesses about usage were wrong, e.g. people really do want
  `stack build somelibrary` to work outside a project for things like
  `stack runghc` to work nicely
* Consistent behavior > smart behavior

So this new approach is: we always fall back to the implicit global
project, but notify the user about it.


snoyberg:
OK, it builds again

PR #670, opened 2015-07-24
Filelock -- ensure mutual exclusion of multiple stack processes

rrnewton:
This initial version of the functionality simply enforces complete mutual exclusion.

It uses an standard empty lock file, e.g., `~/.stack/lockfile`.

Here is a list of all current stack commands and what this PR does with their locking behavior.  If you think one of them is wrong, please raise it in the discussion here.

##### Unlock immediately:

 * sdist
 * ide
 * list-dependencies

##### Unlock when control transfers to the subprocess:

 * exec
 * ghci
 * runghc - this reduced to the `exec` case.

##### Keep lock for entire run:

 * build
 * install
 * uninstall
 * test
 * bench
 * haddock
 * new
 * init
 * solver
 * setup
 * path
 * unpack
 * update
 * upgrade
 * upload
 * dot
 * ghc
 * clean
 * docker
 * image
 
The policy here is to default to the safe option of holding the lock unless there is a compelling reason to carve out an exception.

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -299,7 +299,7 @@ pluginShouldHaveRun _plugin _globalOpts = do
 -- support others later).
 pathCmd :: [Text] -> GlobalOpts -> IO ()
 pathCmd keys go =
-    withBuildConfig
+    withBuildConfigAndLock
         go
         ExecStrategy
         (do env <- ask
@@ -481,10 +481,10 @@ withUserFileLock cfg act = do
                  (\_lk -> act)
 
 
-withConfig :: GlobalOpts
+withConfigAndLock :: GlobalOpts
            -> StackT Config IO ()
            -> IO ()
-withConfig go@GlobalOpts{..} inner = do
+withConfigAndLock go@GlobalOpts{..} inner = do
     (manager, lc) <- loadConfigWithOpts go
     withUserFileLock (lcConfig lc) $
      runStackT manager globalLogLevel (lcConfig lc) globalTerminal $
@@ -492,11 +492,11 @@ withConfig go@GlobalOpts{..} inner = do
             runStackT manager globalLogLevel (lcConfig lc) globalTerminal
                 inner
 
-withBuildConfig :: GlobalOpts
+withBuildConfigAndLock :: GlobalOpts
                 -> NoBuildConfigStrategy
                 -> StackT EnvConfig IO ()
                 -> IO ()
-withBuildConfig go@GlobalOpts{..} strat inner = do
+withBuildConfigAndLock go@GlobalOpts{..} strat inner = do
     (manager, lc) <- loadConfigWithOpts go
     runStackT manager globalLogLevel (lcConfig lc) globalTerminal $
         Docker.reexecWithOptionalContainer (lcProjectRoot lc) $ do
@@ -515,7 +515,7 @@ withBuildConfig go@GlobalOpts{..} strat inner = do
                  inner
 
 cleanCmd :: () -> GlobalOpts -> IO ()
-cleanCmd () go = withBuildConfig go ThrowException clean
+cleanCmd () go = withBuildConfigAndLock go ThrowException clean
 
 -- | Helper for build and install commands
 buildCmdHelper :: StackT EnvConfig IO () -- ^ do before build
@@ -524,7 +524,7 @@ buildCmdHelper beforeBuild strat finalAction opts go
     | boptsFileWatch opts = fileWatch inner
     | otherwise = inner $ const $ return ()
   where
-    inner setLocalFiles = withBuildConfig go strat $ do
+    inner setLocalFiles = withBuildConfigAndLock go strat $ do
         beforeBuild
         Stack.Build.build setLocalFiles opts { boptsFinalAction = finalAction }
 
@@ -546,24 +546,24 @@ copyCmd :: BuildOpts -> GlobalOpts -> IO ()
 copyCmd opts = buildCmdHelper (return ()) ExecStrategy DoNothing opts { boptsInstallExes = True }
 
 uninstallCmd :: [String] -> GlobalOpts -> IO ()
-uninstallCmd _ go = withConfig go $ do
+uninstallCmd _ go = withConfigAndLock go $ do
     $logError "stack does not manage installations in global locations"
     $logError "The only global mutation stack performs is executable copying"
     $logError "For the default executable destination, please run 'stack path --local-bin-path'"
 
 -- | Unpack packages to the filesystem
 unpackCmd :: [String] -> GlobalOpts -> IO ()
-unpackCmd names go = withConfig go $ do
+unpackCmd names go = withConfigAndLock go $ do
     menv <- getMinimalEnvOverride
     Stack.Fetch.unpackPackages menv "." names
 
 -- | Update the package index
 updateCmd :: () -> GlobalOpts -> IO ()
-updateCmd () go = withConfig go $
+updateCmd () go = withConfigAndLock go $
     getMinimalEnvOverride >>= Stack.PackageIndex.updateAllIndices
 
 upgradeCmd :: Bool -> GlobalOpts -> IO ()
-upgradeCmd fromGit go = withConfig go $
+upgradeCmd fromGit go = withConfigAndLock go $
     upgrade fromGit (globalResolver go)
 
 -- | Upload to Hackage
@@ -589,10 +589,10 @@ uploadCmd args go = do
                     Upload.defaultUploadSettings
             liftIO $ Upload.mkUploader config uploadSettings
     if null dirs
-        then withConfig go $ do
+        then withConfigAndLock go $ do
             uploader <- getUploader
             liftIO $ forM_ files (canonicalizePath >=> Upload.upload uploader)
-        else withBuildConfig go ExecStrategy $ do
+        else withBuildConfigAndLock go ExecStrategy $ do
             uploader <- getUploader
             liftIO $ forM_ files (canonicalizePath >=> Upload.upload uploader)
             forM_ dirs $ \dir -> do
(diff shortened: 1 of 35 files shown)


rrnewton:
Ok, in the most recent commit I improved it a little bit, issuing a warning message if the lock is not acquired on the first try and the process is waiting on another stack instance:

```
Failed to grab lock ("/Users/rrnewton/.stack/lockfile"); other stack instance running.  Waiting...
Lock acquired, proceeding.
```

However, these are currently printed with `putStrLn`.  I need a pointer on how to route output through the logger from the IO monad.  That is' we're not inside a `MonadLogger` like `StackT`, but we are able to access the `Config`.  Is it possible?

Furthermore, it seems like this is an existing issue in `Main.hs`.  There are some uses of `$logInfo`, some uses of `putStrLn` (stdout), and some uses of `T.hPutStr stderr`.

3noch:
I'll do some ad-hoc testing on Windows.

3noch:
The locking is working on Windows. One build holds off another. Killing the build causes an unlock.

Does this serialize calls to `stack exec`?

3noch:
If this serializes `stack exec` I expect there will serious issues with that. :cry: If there is no easier way to do locking, then, to my chagrin, we might actually need something like an `--allow-concurrency-problems` switch.

rrnewton:
So my proposal for that is that we unlock early with stack exec. 

But when, exactly?  Right before the subprocess launch?

Should it work the same for "stack ghci"?  Are there any other exceptions to carve out after that?  That's where I'm looking for some feedback.

rrnewton:
Hah, this has already proved useful.  I had *no idea* that the structured-haskell-mode setup I installed months ago is actually stack-aware and using stack, and I was inadvertently doing concurrent stack actions.  I didn't realize until I saw the output from "Failed to grab lock" message.

rrnewton:
One thing to test here is CPU usage.  I caught the previous version exhibiting a weird behavior where it would use 2% CPU when blocked for a while, but then it would spike to 100% CPU.  I think this is some kind of bug in the `retrying` package, because it should have woken up only a couple times a second (quick exponential backoff to 500ms delay).

Anyway, the problem is fixed now by doing a blocking call after the first failure.  On Mac, I can confirm that even though the stack process has 13 threads on a 4 core machine,

rrnewton:
Note that it's important that the other HECs quiet down and go to sleep as they should (related to #680).  In my current testing after the most recent commit this is the case.

rrnewton:
Ok, so that should be it for an initial version.  I carved out the exception for `stack ghci` and `stack exec` so that they release the lock before "tail calling" to ghci or the executed command.

The changes to `Main.hs` are not deep, but they do touch many places.  So merging/rebasing on master can be a pain for this PR.  It would be great to get it merged soon if it is acceptable.

rrnewton:
Any feedback on this one?

3noch:
@rrnewton This new version is working much better with `stack exec`.

src/main/Main.hs
```diff
@@ -470,21 +473,41 @@ setupCmd SetupCmdOpts{..} go@GlobalOpts{..} = do
...
-withConfig :: GlobalOpts
+-- | Enforce mutual exclusion of every action running via this function
+-- on this users account.  Currently, stack uses this to completely
+-- exclude concurrent execution of stack commands.  In the future,
+-- stack may refine this to a finer-grained locking approach.
+withUserFileLock :: Config
+                 -> (FileLock -> IO a)
+                 -> IO a
+withUserFileLock cfg act = do
+    lockfile <- parseRelFile "lockfile"
```
snoyberg: Typically we use `$(mkRelFile "lockfile")` for these kinds of things to avoid a runtime check and catch typos at compile time as well.

src/main/Main.hs
```diff
@@ -470,21 +473,41 @@ setupCmd SetupCmdOpts{..} go@GlobalOpts{..} = do
...
+    let pth = configStackRoot cfg </> lockfile
+    fstTry <- tryLockFile (toFilePath pth) Exclusive
```
snoyberg: Technically this code isn't async-exception safe. I'm not sure how important that safety is to the algorithm.

src/main/Main.hs
```diff
@@ -608,7 +632,7 @@ uploadCmd args go = do
 
 sdistCmd :: [String] -> GlobalOpts -> IO ()
 sdistCmd dirs go =
-    withBuildConfig go $ do
+    withBuildConfigAndLock go $ \_ -> do
```
snoyberg: Does this actually need a lock?

src/main/Main.hs
```diff
@@ -626,69 +650,80 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+ideCmd (targets,args) go@GlobalOpts{..} = withBuildConfigAndLock go $ \_ -> do
       ide targets args
 
 -- | Pull the current Docker image.
 dockerPullCmd :: () -> GlobalOpts -> IO ()
 dockerPullCmd _ go@GlobalOpts{..} = do
     (manager,lc) <- liftIO $ loadConfigWithOpts go
-    runStackTGlobal manager (lcConfig lc) go $
-        Docker.preventInContainer Docker.pull
+    -- TODO: can we eliminate this lock if it doesn't touch ~/.stack/?
```
snoyberg: @manny-fp is the expert here, but I believe none of the `docker` commands themselves require locking.

stack.cabal
```diff
@@ -1,5 +1,5 @@
 name:                stack
-version:             0.1.2.1
+version:             0.1.2.9999
```
snoyberg: Probably accidental

snoyberg:
Added some comments inline, but LGTM.

src/main/Main.hs
```diff
@@ -626,69 +650,80 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
                     Nothing
         ExecOptsEmbellished {..} ->
-           withBuildConfig go $ do
+           withBuildConfigAndLock go $ \lk -> do
                let targets = concatMap words eoPackages
                unless (null targets) $
                    Stack.Build.build (const $ return ()) defaultBuildOpts
                        { boptsTargets = map T.pack targets
                        }
+               liftIO $ unlockFile lk -- Unlock before transferring control away.
```
3noch: Not sure how idiomatic it would be or if it would be much value-add, but `lk` could be an unlocking function instead. E.g. `withBuildConfigAndLock go $ \unlock -> do ..... unlock; exec`.

src/main/Main.hs
rrnewton: Ah, ok, these uses of TH are neat but new to me.  I'll try this.

Btw, it seems like many of them could probably use the new typed TH interface, no?   (The one confusing thing about reading the APIs that use TH is the loss of type info in all the `Q Exp`s.)

3noch:
I'm not familiar enough with the code to have confidence about which commands are being locked all the way, which are being unlocked early, and which have no locking. I'm particularly interested in things like `ghc`, `runghc`, `ide`. Perhaps we could get a rundown of which commands have what locking behavior? It might even be useful to have that in docs or `--help` somewher.

src/main/Main.hs
rrnewton: Ah, good catch, thanks.

Ok, let's see.  So if anything goes really wrong and the process exits, then the `filelock` lib is designed so that file descriptors get closed and the locks get released.

But the issue here is that if the thread running `withUserFileLock` gets killed by an async exception from another Haskell IO thread -- while holding the lock from `tryLockFile` -- then the lock will not get released until the process exits.

I think this is ok for now because the lock is global, and per-process.  But it's better to fix this and use bracket at the outset.  Ok, just did a commit with that fix.

src/main/Main.hs
rrnewton: I dunno!  I was being conservative and locking on all commands, except for the carved out exception for `stack exec`/`stack ghci`.  I wasn't totally sure which commands might imply other build actions recursively.

But... `sdistCmd` *is* a clean one..  It's effects seem to be visible in its body without reading over too much other code.   Its effects seem obviously local to the current directory.  Ok, I've made sdist another exception.  For now it instantly unlocks rather than never locking, but that could be improved.

Any more exceptions?

stack.cabal
rrnewton: Well, I wanted to mark this dev version and make sure we could rely on the locking functionality on our continuous integration infrastructure.  Is there a naming convention for dev versions a la GHC's "7.11.DATE".

3noch:
How do `runghc` and `ide` deal with locking? I imagine they could both be long running?

stack.cabal
rrnewton: Actually, it's unclear to me by the PVP whether this PR should technically bump the major version number.  It changes an executable rather than a library, and any usage that worked before 
*should* still work.  (Not so if `stack exec` kept the lock though, in which case you could easily construct usages that *required* concurrent execution.)

src/main/Main.hs
rrnewton: Sure, could do.  My gut feeling on this one is just to defer generalization until its needed because reading the type `(FileLock -> IO a)` should give someone a much stronger idea of what's going on than reading the type `(IO () -> IO a)`.

src/main/Main.hs
rrnewton: So would the idea there be to release the `stack` process lock just before calling `Docker.whatever`, and let it contend for the (per docker container?) lock at that point?

rrnewton: So would the idea there be to release the `stack` process lock just before calling `Docker.whatever`, and let it contend for the (per docker container?) lock at that point?

rrnewton:
@3noch, good point regarding adding a list.  I created a checklist at the top of this PR.

`runghc` is already covered, and I just added `ide` as another exception.

rrnewton:
In the process of resolving merge conflicts, I just made `list-dependencies` another exception.

At this point I've addressed @snoyberg's points, I believe.  And given his "LGTM", I believe it should be ready for a merge.

We can easily tweak the policy moving forward, and in particular can add the docker commands as another exception.  But I believe this is a safe starting point.

I've already found it quite useful -- not just for catching races between my emacs-created stack instances and other stack builds -- but also the explicit "can't grab lock" warning message has been able to remind me that I have some backgrounded or forgotten stack process lying around that I need to kill.

Longer term, after everyone confirms that this starting point works fine, it would be good to start refining the locks to per-snapshot, but that will be a separate PR.

snoyberg:
@manny-fp Deferring to you on merge, I'm good with this to go.

Issue #638, opened 2015-07-20
stack list-dependencies


bitemyapp:
Would help with some custom Dockerization stuff, ideally we'd be able to run it and collect a set of all packages we need across multiple projects and their respective versions so we can install the deps before our code gets into the Docker image.

creichert:
This would be quite nice. It would also help for generating the tags for
a stack megarepo and all it's dependencies (similar to how `codex` works
and mentioned in the `codex issue tracker).

manny-fp:
Would the output of `stack dot --external` be usable for this?  I realize it's not in the ideal format, but might work as something to tide you over.

bitemyapp:
@manny-fp doesn't have versions and would require munging but might save us some work. Thank you!

manny-fp:
A PR implementing this would be most welcome, and happy to answer any questions or give direction.  Hopefully the `stack dot` can be used as a starting point.

markus1189:
I think I will give this a go, what's the desired output? Just lines of <package name>-<version>?

bitemyapp:
@markus1189 tab separated for easier shell processing would be even better, if it's not too much trouble.

`$packagename<TAB>$version`

creichert:
Something like `cabal list --simple-output` would work perfectly for my use.

markus1189:
A first version is here: https://github.com/markus1189/stack/tree/638-list-dependencies. Example

```
$ stack list-dependencies --separator '-'
Cabal-1.18.1.5
MonadRandom-0.3.0.2
StateVar-1.1.0.0
adjunctions-4.2.1
aeson-0.8.0.2
ansi-terminal-0.6.2.1
ansi-wl-pprint-0.6.7.2
array-0.5.0.0
asn1-encoding-0.9.0
asn1-parse-0.9.1
asn1-types-0.3.0
async-2.0.2
attoparsec-0.12.1.6
auto-update-0.1.2.2
base-<unknown>
base16-bytestring-0.1.1.6
base64-bytestring-1.0.0.1
bifunctors-4.2.1
binary-0.7.1.0
blaze-builder-0.4.0.1
blaze-html-0.8.0.2
blaze-markup-0.7.0.2
byteable-0.1.1
bytestring-0.10.4.0
bytestring-builder-0.10.6.0.0
case-insensitive-1.2.0.4
cereal-0.4.1.1
chunked-data-0.1.0.1
cipher-aes-0.2.11
cipher-des-0.0.6
[...]

PR #682, opened 2015-07-27
New command `list-dependencies`

markus1189:
Implements #638 .

Adds a new command `stack list-dependencies` that will list all transitive dependencies for the local packages.

It's not able to show dependencies for the test and benchmark packages, because I could not figure out how to do it, nevertheless I though I should probably pr this as is for the added functionality.

The problem I had with test/benchmark in a nutshell: 

> It seems like `loadSourceMap` and `loadLocals` by default assume that they are not enabled (package config sets it to false). Once I changed that it seems like it somehow does not pickup the dependencies for test packages like QuickCheck when using `withLoadPackage`'s function to load packages...

If you see what I am doing wrong I'd be happy to change this :)

The nice part is that it reuses the structure generated by `Stack.Dot` so in theory we could also display versions for the `stack dot` output.

### DIFF
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -125,11 +125,12 @@ listDependencies :: (HasEnvConfig env
                  -> m ()
 listDependencies sep = do
   (locals,_,_) <- loadLocals defaultBuildOpts Map.empty
-  let localNames = Set.fromList (map (packageNameString . packageName . lpPackageFinal) locals)
-      dotOpts = DotOpts True True Nothing localNames
+  let localNames = Set.fromList (map (packageName . lpPackageFinal) locals)
+      dotOpts = DotOpts True True Nothing Set.empty
 
   resultGraph <- createDependencyGraph dotOpts locals
-  void (Map.traverseWithKey go (snd <$> resultGraph))
+  let graphWithoutLocals = F.foldl' (flip Map.delete) resultGraph localNames
+  void (Map.traverseWithKey go (snd <$> graphWithoutLocals))
     where go name v = liftIO (Text.putStrLn $
                                 Text.pack (packageNameString name) <>
                                 sep <>
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -38,7 +38,7 @@ import           Stack.Package
 import           Stack.Types
 import           Stack.Types.Internal (HasLogLevel)
 
--- | Options record for `stack dot`
+-- | Options record for @stack dot@
 data DotOpts = DotOpts
     { dotIncludeExternal :: Bool
     -- ^ Include external dependencies
@@ -73,6 +73,10 @@ dot dotOpts = do
         prunedGraph = pruneGraph localNames pkgsToPrune resultGraph
     printGraph dotOpts locals prunedGraph
 
+-- | Create the dependency graph, the result is a map from a package
+-- name to a tuple of dependencies and a version if available.  This
+-- function mainly gathers the required arguments for
+-- @resolveDependencies@.
 createDependencyGraph :: (HasEnvConfig env
                          ,HasHttpManager env
                          ,HasLogLevel env
@@ -82,8 +86,8 @@ createDependencyGraph :: (HasEnvConfig env
                          ,MonadIO m
                          ,MonadMask m
                          ,MonadReader env m)
-                      => DotOpts
-                      -> m (Map PackageName (Set PackageName, Maybe Version))
+                       => DotOpts
+                       -> m (Map PackageName (Set PackageName, Maybe Version))
 createDependencyGraph dotOpts = do
   (_,locals,_,sourceMap) <- loadSourceMap defaultBuildOpts
   let graph = Map.fromList (localDependencies dotOpts locals)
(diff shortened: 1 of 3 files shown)


creichert:
Awesome, +1 from me

manny-fp:
I'd prefer the default separator to be `-`, to match `ghc-pkg list` and many other things (but very nice to see the `--separator` option, that will make scripting much easier).

I'm also currently seeing the version shown as `<unknown>` for any packages included with GHC.  Is that expected?
```
base <unknown>
ghc-prim <unknown>
integer-gmp <unknown>
template-haskell <unknown>
```

markus1189:
Well I think two things speak for ` ` as the default.
1. `cabal list --installed` uses it as well
2. you can more easily extract only the version/only the name when using a space because it is not part of a valid package name (right?), whereas `-` is, e.g. `x509-system 1.5.0` is easier to take apart than `x509-system-1.5.0`.

Having said that I have to strong feelings for any default so I am also fine with `-` as a separator.

> I'm also currently seeing the version shown as <unknown> for any packages included with GHC. Is that expected?

Uhm no, for me it's not the case, I have 
```
base 4.7.0.2
ghc-prim 0.3.1.0
integer-gmp 0.5.1.0
template-haskell 2.9.0.0
```

Strange, any idea on what I could do to reproduce? I am using the standard `stack.yaml` checked in the repo.

manny-fp:
Yeah I don't have a strong preference for the separator either (I've just gotten used to `-`), so I'm fine with leaving at as `space`.

Here's my --verbose output.  Can you compare to yours?
```
$ stack --verbose list-dependencies
Version 0.1.2.1, Git revision 67220bfed687e80d69cc64df1dd48026ecc2b55f
2015-07-29 16:47:11.30497: [debug] Checking for project config at: /Users/manny/fpco/tmp/stack/stack.yaml @(stack-0.1.2.1:Stack.Config src/Stack/Config.hs:460:9)
2015-07-29 16:47:11.307415: [debug] Loading project config file stack.yaml @(stack-0.1.2.1:Stack.Config src/Stack/Config.hs:483:13)
2015-07-29 16:47:11.318005: [debug] Run process: ghc --numeric-version @(stack-0.1.2.1:System.Process.Read src/System/Process/Read.hs:255:3)
2015-07-29 16:47:11.377504: [debug] Run process: ghc-pkg --no-user-package-db field --simple-output Cabal id @(stack-0.1.2.1:System.Process.Read src/System/Process/Read.hs:255:3)
2015-07-29 16:47:11.402206: [debug] Run process: ghc-pkg --no-user-package-db list --global @(stack-0.1.2.1:System.Process.Read src/System/Process/Read.hs:255:3)
2015-07-29 16:47:11.426672: [debug] Checking resolver: lts-2.17 @(stack-0.1.2.1:Stack.Build.Source src/Stack/Build/Source.hs:87:13)
2015-07-29 16:47:11.795129: [debug] Run process: ghc-pkg --global --no-user-package-db dump --expand-pkgroot @(stack-0.1.2.1:System.Process.Read src/System/Process/Read.hs:255:3)
2015-07-29 16:47:11.83475: [debug] Run process: ghc-pkg --user --no-user-package-db --package-db /Users/manny/.stack/snapshots/x86_64-osx/lts-2.17/7.8.4/pkgdb/ dump --expand-pkgroot @(stack-0.1.2.1:System.Process.Read src/System/Process/Read.hs:255:3)
2015-07-29 16:47:11.915845: [debug] Run process: ghc-pkg --user --no-user-package-db --package-db /Users/manny/fpco/tmp/stack/.stack-work/install/x86_64-osx/lts-2.17/7.8.4/pkgdb/ dump --expand-pkgroot @(stack-0.1.2.1:System.Process.Read src/System/Process/Read.hs:255:3)
[...]
```

markus1189:
What git revision is that?  `origin/638-list-dependencies` is 10b24285c2b8989685adf6f24126dfdbf3efe200 for me, otherwise the output looks the same (ignoring the paths and time): https://www.diffchecker.com/ur99obyp

manny-fp:
Ah, that was after I merged in upstream master.  I reset to 10b24285c2b8989685adf6f24126dfdbf3efe200 and still see the `<unknown>`s.

markus1189:
Hm even after clean and removing `.stack-work` it's still not the case here, that's really mysterious

src/Stack/Dot.hs
```diff
@@ -50,71 +53,128 @@ data DotOpts = DotOpts
...
-pruneGraph :: (F.Foldable f, F.Foldable g)
+        thrd :: (a,b,c) -> c
+        thrd (_,_,x) = x
+
+-- Given an 'Installed' try to get the 'Version'
+libVersionFromInstalled :: Installed -> Maybe Version
+libVersionFromInstalled (Library ghcPkgId) =
+    case ghcPkgIdPackageIdentifier ghcPkgId of
+       PackageIdentifier _ v -> Just v
+libVersionFromInstalled (Executable _) = Nothing
```
markus1189: This is where the nothing comes from in the end

markus1189:
Seems like it also works on travis: https://travis-ci.org/markus1189/stack/jobs/73232119

manny-fp:
After blowing away my old `~/.stack`, it's reporting version numbers for all packages.  I spent some time looking into the difference, but so far not finding anything and I don't know that it's worth putting more effort into.

mgsloan:
I just ran into the fact that it uses space instead of `-`.  Are we sure that's a reasonable default? :/

I wanted to do `stack list-dependencies | xargs stack unpack`.

mgsloan:
Opened this issue: https://github.com/commercialhaskell/stack/issues/1614

PR #689, opened 2015-07-29
allow building containers regardless of docker builds being enabled

### LINKED ISSUE #659: `stack image container` should work even if docker: enable: true is on in stack.yaml
Currently stack responds to this situation with "This command must be run on host OS (not in a Docker container)."

We disabled the ability to try to build an image inside of another docker instance for a couple reasons:
1) there's no easy way to map the docker server socket inside the container & give the container user access.
2) even if we could provide a docker server socket inside the container, it's fragile. there are breaking changes in between releases of docker (eg, 1.6.2 server & 1.7.0 client combo breaks)
3) stack lays down it's executables it produces in a GHC-version-specific directory under .stack-work.  This is very likely different between the docker image & your desktop/server host.

So... we have to think of another way.  What's been discussed so far is allowing in-docker container building but placing (possibly using an optional flag) the binaries into a neutral place under .stack-work (say .stack-work/image) and providing the ability to run the docker image process on the host outside of docker after building.  This way we can find the executables built by docker & we aren't trying to build docker images inside of docker.

@manny-fp said, in discussion, that

"We can add a variant of `Docker.reexecWithOptionalContainer` that takes a `postInner` (or something) argument which is always run after `inner`, outside the container. (edited)"

dysinger:
… or disabled. closes #659 /cc @manny-fp @cocreature @snoyberg

### DIFF
--- a/src/Stack/Image.hs
+++ b/src/Stack/Image.hs
@@ -9,12 +9,13 @@
 
 -- | This module builds Docker (OpenContainer) images.
 module Stack.Image
-       (imageDocker, imgCmdName, imgDockerCmdName, imgOptsFromMonoid,
+       (stageContainerImageArtifacts, createContainerImageFromStage,
+        imgCmdName, imgDockerCmdName, imgOptsFromMonoid,
         imgDockerOptsFromMonoid, imgOptsParser, imgDockerOptsParser)
        where
 
 import           Control.Applicative
-import           Control.Exception.Lifted
+import           Control.Exception.Lifted hiding (finally)
 import           Control.Monad
 import           Control.Monad.Catch hiding (bracket)
 import           Control.Monad.IO.Class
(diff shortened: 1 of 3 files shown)


src/Stack/Image.hs
```diff
@@ -25,154 +26,120 @@ import           Data.Char (toLower)
...
--- specified in the project's stack.yaml.  The new image will contain
--- all the executables from the packages in the project as well as any
--- other specified files to `add`.  Then new image will be extended
--- with an ENTRYPOINT specified for each `entrypoint` listed in the
--- config file.
-imageDocker :: M e m => m ()
-imageDocker = do
-    tempDirFP <- liftIO SD.getTemporaryDirectory
-    bracket
-        (liftIO (createTempDirectory tempDirFP "stack-image-docker"))
```
manny-fp: Possibly use [System.IO.Temp.withSystemTempDirectory](http://haddock.stackage.org/nightly-2015-07-28/temporary-1.2.0.3/System-IO-Temp.html#v:withSystemTempDirectory)?

manny-fp:
LGTM

PR #707, opened 2015-08-02
Filelock3 -- release snapshot lock early for local-only builds

rrnewton:
This is the third and final (for now) refinement of the locking mechanism.

"Filelock2" (#700) refines the locking to be per-snapshot, rather than global/per-user.  This patch goes further and allows the snapshot lock to be released early by any build command whose actions are *all local*.

The next reasonable step would be to allow mixed snapshot/local builds to release early once the snapshot portion of the build is done.  (This could be accomplished simply by making the unlock itself an `Action`, and scheduled as part of the dependence graph of actions.)

That last refinement would be nice to have, but "filelock3" already handles the most important common case -- a server running CI jobs again and again builds the snapshot bits only on the first execution (and occasionally thereafter).  The vast majority of the time it runs local-only builds.

Thus, with this patch a CI server can run many local-only builds at the same time, safely, while still protecting the shared snapshot state.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -26,8 +26,8 @@ import           Control.Monad.Logger
 import           Control.Monad.Reader (MonadReader, asks)
 import           Control.Monad.Trans.Resource
 import           Data.Function
-import           Data.Map.Strict (Map)
 import qualified Data.Map as Map
+import           Data.Map.Strict (Map)
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Network.HTTP.Client.Conduit (HasHttpManager)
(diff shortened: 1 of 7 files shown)


3noch:
Phenominal, can't wait to get this.

chrisdone:
I think this has a bad interaction with docker... 

```
chris@retina:~/Work/foo/bar/mu/proj$ rm /home/chris/.stack/lockfile
rm: cannot remove ‘/home/chris/.stack/lockfile’: No such file or directory
chris@retina:~/Work/foo/bar/mu/proj$ stack test .
Warning: /home/chris/Work/foo/stack.yaml: Unrecognized field in ImageOptsMonoid: docker
Warning: /home/chris/Work/foo/stack.yaml: Unrecognized field in ImageOptsMonoid: docker
Failed to grab lock ("/home/chris/.stack/lockfile"); other stack instance running.  Waiting...
  c c^C*** Init system aborted.
```

I have to figure out how to update my stack in docker to see whether this still occurs with a new stack inside the docker image…

chrisdone:
Seems impossible to run the docker stack due to this locking mechanism.

chrisdone:
Ok, so I bissected as far back as d7e90fd and it works there. @rrnewton Are you able to test the docker integration to see what's up?

chrisdone:
Okay, 52f88b5 works. 42cdfac and afterwards exhibits this behaviour.

Meanwhile I'm going to try to revert the part that applies to Docker, because this is preventing regular use of Docker.

chrisdone:
Hmm, generally I think `exec` shouldn't use a lock… it makes it impossible to run ghci in one place or a server program in one place and run anything to do with stack elsewhere.

3noch:
`exec` *should* immediately unlock

3noch:
So it won't block other things, but I suppose it can itself be blocke.d

rrnewton:
Ah, thanks for the bisect.  I was not previously set up to do docker tests but I can set something up now to help figure this out.

Would it help as a quick fix to not grab *any* lock on docker commands?

That's what Michael hinted should be possible -- just let docker do its own locking.

rrnewton:
Ah, thanks for the bisect.  I was not previously set up to do docker tests but I can set something up now to help figure this out.

Would it help as a quick fix to not grab *any* lock on docker commands?

That's what Michael hinted should be possible -- just let docker do its own locking.

chrisdone:
That would be very helpful as a temporary fix. Currently I have to use that old commit as we use the Docker support for our development at FP Complete.

rrnewton:
I didn't follow how exactly that error is happening in `stack test .`.  Is stack test invoking one of the stack docker commands and then failing?  Is the lock it fails to get the one inside the docker image?

Or is it just that it somehow recursively invokes stack before releasing the lock via some other (non-docker?) command?

rrnewton:
Oh, another small comment here -- in its current state this works really well with `--only-snapshot`.  Basically a build or test with `--only-snapshot` guarantees that the subsequent build or test after it will be local-only, which means that it will release the lock early.

Issue #651, opened 2015-07-21
Add --test and --bench options to build command


snoyberg:
Motivation: you could write something like the following:

    stack build --test --bench --haddock --copy-bins

This just makes the commands far more composable than they are right now. We should *not* remove or deprecate `stack test` and `stack bench` IMO: they are useful shortcuts for common cases.

manny-fp:
What do you think about an `--exec` option to execute a command after the build (particularly useful in conjunction with `--file-watch`)?  It's a slightly different case than the others because `stack exec` isn't one of the build commands (unlike `stack test` etc), and currently uses the `execve` system call, so the `stack build --exec` option would probably have to follow a different code path than `stack exec`.

snoyberg:
I'm in favor of it, though we will probably need to bikeshed the name to
death.

snoyberg:
I'd like to write up clearly my plans for overhauling how targets, "wanted"s,
and components will be handled. The goals of this overhaul are to make
tests/benchmarks more modular, make error reporting of bad arguments more
reliable, and make it possible to do things like running a single test suite. This procedure should address the following issues:

* #387
* #651
* #687

1. `stack test` is a synonym for `stack build --test`. Similarly, `stack bench` is a synonym for `stack build --bench`. At this point: `--copy-bins`, `--haddock`, `--test`, and `--bench` are composable with each other.
2. When running `stack build`, you provide a list of **targets**. Targets can be in one of the following formats:
    * *package:comptype[:compname]*, e.g. `bar:exe:bar-exe1`. This will specify as a target the given *component* of a local package (fails for any non-local package). If the package name is not available locally, this will be an error. `comptype` can be `lib`, `exe`, `test`, and `bench`. When using `lib`, `:compname` must not be used.
    * *package:compname*, e.g. `bar:bar-exe1`. Same as above, but determines the `comptype` from the `compname`. This will fail if the same `compname` is used by two different `comptype`s in a package (e.g., both a executable and test suite named `foobar`).
    * *:compname*, e.g. `:bar-exe1`. Same as above, but also determines the package name. Will fail for any ambiguity.
    * *package*, e.g., `foo`. Fails for non-local package. For a local package, adds all of the library and executables components available in the given package. If `--test` is passed in, all test suites are also added. If `--bench` is passed in, all benchmarks are also added.
[...]

PR #726, opened 2015-08-06
Stack ghci build

chrisdone:
Pinging @snoyberg to review. I don't know what the sans-tests/with-tests distinction is for with `lpPackage` vs `lpPackageFinal`. In this case the semantics of the bools changes somewhat in `packageConfigEnableTests` and `packageConfigEnableBenchmarks` because it doesn't necessarily mean an action will be performed, but it does mean stanzas of test suites and benchmarks are enabled.

With this change, it would make `lpPackageFinal` redundant, but I don't know what that's for. I struggled generally to read the codebase.

This passes the test suite at least and behaves as expected: http://lpaste.net/4831131355167326208

### DIFF
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -200,20 +200,15 @@ loadLocals bopts latestVersion = do
         name <- parsePackageNameFromFilePath cabalfp
         let wanted = validWanted && isWanted' dir name
             config = PackageConfig
-                { packageConfigEnableTests = False
-                , packageConfigEnableBenchmarks = False
+                { packageConfigEnableTests = wanted && boptsEnableTests bopts
+                , packageConfigEnableBenchmarks = wanted && boptsEnableBenchmarks bopts
                 , packageConfigFlags = localFlags (boptsFlags bopts) bconfig name
                 , packageConfigGhcVersion = envConfigGhcVersion econfig
                 , packageConfigPlatform = configPlatform $ getConfig bconfig
                 }
             configFinal = config
-                { packageConfigEnableTests =
-                    case boptsFinalAction bopts of
-                        DoTests _ -> wanted
-                        _ -> False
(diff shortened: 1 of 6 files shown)


snoyberg:
Unless I'm misreading, this will break the test-cycle case: when package B depends on package A, but package A's test suite depends on package B. That's the reason for the two different sets of dependencies: so we can build and install the library and only after that deal with the test suite.

snoyberg:
It seems to me like some of what you're aiming for here may be achieved if we implemented #651. More generally, we may want to start thinking of individual components within a package as being much more autonomous than we think of them today.

chrisdone:
> Unless I'm misreading, this will break the test-cycle case: when package B depends on package A, but package A's test suite depends on package B. That's the reason for the two different sets of dependencies: so we can build and install the library and only after that deal with the test suite.

I see, then this approach can't work.

> It seems to me like some of what you're aiming for here may be achieved if we implemented #651. More generally, we may want to start thinking of individual components within a package as being much more autonomous than we think of them today.

Right, if #651 were implemented then this would be doable more directly (enabling both test and benchmark for build, but skipping the final action step). All I'm after here is to make sure that the test suite and benchmark's dependencies are available before starting GHCi. 

In parallel I have an issue to enable specific stanzas when launching `stack ghci` (#727), which would allow to just load up the library and executable.

If #651 is a big upheaval to do right now, a possible alternative implementation for this task could be that, seeing as I already know all the packages that I _want_ to specify as `-package x -package y -package z` to GHCi, then I could run a separate build for those dependencies as if I'd run `stack install x y z`. So in this case it'd be `stack install temporary hspec` etc. The final alternative is to just flag up to the user, "these packages aren't installed, you may want to run `stack install x y z`", but doing this automatically seems more in line with stack's goals of doing the right thing by default.

(All these points share the same code with `stack ide`, so it applies there too.)

Issue #105, opened 2015-05-29
Respect TemplateHaskell addDependentFile dependency changes


DanBurton:
For example, when building `stack` itself.

* Make a local change
* `stack build`
* gitrev writes the version as "dirty"
* commit
* `stack build`

What should happen:
* stack recognizes that the dependency changed and rebuilds accordingly

What currently happens:
* stack does not recognize that the dependency changed

DanBurton:
Here's my intended approach to working on this.

When formulating a build plan (a specific arch, ghc, lts are selected)

1. locate the place(s) where "get source files for this local package'" is calculated
2. locate .hi files (if present) for the .hs/.lhs files already discovered
3. calculate additional file dependencies
  - 3.1 for TH, it's just file names, relative to that package's root (I don't believe there's such a thing as extra-source-dirs?)
  - 3.2 for module dependencies not already accounted for, look for these in any of the relevant hs-source-dirs
4. print warnings about these additional dependencies if they are not already accounted for in the cabal file, suggest adding them to the cabal file with copy/pasteable output. (extra-source-files for 3.1, other-modules for 3.2)
5. return the augmented list of source files

As I understand it, the build cache should smoothly take it from there, as long as we tell it about these extra files.

Some nuances and thoughts regarding:

* locating .hi files:  
  `getInstallationRoot` gets us to `.stack-work/install/$arch/$lts/$ghc/`.  
  From there we need to drill down into `lib/$arch-ghc-$ghc/$package-$packageVer/`,
  and the .hi files for that package are in module hierarchy locations from there.
* dependencies of dependencies  
  Steps 2 thru 5 may need to be repeated until no new file dependencies are discovered.  
  Limit N repeats for sanity?
* .hi files absent prior to first build  
  The build process should work correctly anyways since it is a clean build.  
  I'm guessing subtle bugs may creep in here.  
  The warnings, as outlined above, won't trigger unless we add a post-build check.
* first time build cache  
  The first build cache won't include these extra files, which may lead to extra rebuilding.  
  This is a second reason to add a post-build check: it can update the build cache at that time.
* build cache details  
  We might need some extra way of marking a module as "needs rebuild due to dirty file dependency". I'm not quite sure what the implementation details are here.

I might stub out the "module dependencies" part (3.2) if it is tricky.

DanBurton:
One shortcoming of this approach is that it only works for a library target. I can't find the `.hi` file for the `main-is` in an executable target.

[...]

Issue #32, opened 2015-05-19
Detect when a module is compiled but not listed in the cabal file


snoyberg:
Would have caught this bug @chrisdone just introduced ;)

```
/home/vagrant/haskell/stackage-common/dist/build/libHSstack-0.0.0.a(Config.o):(.text+0x1cb7f): undefined reference to `stackzm0zi0zi0_StackziTypesziDocker_zdwzdcshowsPrec_info'
/home/vagrant/haskell/stackage-common/dist/build/libHSstack-0.0.0.a(Config.o):(.data+0x3250): undefined reference to `stackzm0zi0zi0_StackziTypesziDocker_zdfFromJSONZLzmzgZRzuzdcparseJSON_closure'
/home/vagrant/haskell/stackage-common/dist/build/libHSstack-0.0.0.a(Config.o):(.data+0x3490): undefined reference to `stackzm0zi0zi0_StackziTypesziDocker_zdfFromJSONZLzmzgZRzuzdcparseJSON_closure'
/home/vagrant/haskell/stackage-common/dist/build/libHSstack-0.0.0.a(Config.o):(.data+0x3658): undefined reference to `stackzm0zi0zi0_StackziTypesziDocker_zdwzdcshowsPrec_closure'
/home/vagrant/haskell/stackage-common/dist/build/libHSstack-0.0.0.a(Config.o):(.text+0x2797): undefined reference to `stackzm0zi0zi0_StackziTypesziDocker_zdwzdcshowsPrec_info'
/home/vagrant/haskell/stackage-common/dist/build/libHSstack-0.0.0.a(Config.o):(.data+0x9c0): undefined reference to `stackzm0zi0zi0_StackziTypesziDocker_zdwzdcshowsPrec_closure'
```

FYI: I'm fixing that problem in the codebase right now

snoyberg:
Just spoke with Dan about this, it will fit in with other work he's doing.

borsboom:
Discussion in PR #734.

PR #734, opened 2015-08-07
Detect unlisted modules and TH dependent files (#32,#105)

borsboom:
This seems to basically work.  The approach it takes is to have GHC dump the interfaces (using `-ddump-hi -ddump-to-file`) while it builds, and then it parses the dumped interfaces when checking for changes (before the next build) to detect any unlisted modules and TH dependencies.

I'd like input on some of these before moving further.

- [ ] If there are unlisted/TH dependencies, the second time you `build` after a clean (without changing any files in the interim) it thinks those files are new and so it runs the build anyway.  That's because the dumped HI files didn't exist for the first `build` so it didn't record the state of the unlisted files.  Is this a corner case worth fixing?  To do so, it'd have to do a pass looking over the `.dump-hi` files immediately after the build to find the unlisted files and record their state.

- [ ] It reads all the `.dump-hi` files before every build.  That seems to be pretty fast, but caching would be faster.  Worth doing?  (this and the previous item probably make sense to do together)

- [ ] Currently there are a number list->Set->list conversions.  To fix that, it'd mean changing a bunch of stuff in the module to use Sets instead of lists.  Any objection to doing so?

- [ ] I noticed `nub` is used a bunch in the existing code.  I think converting the code to use Sets would eliminate many of those `nub`s, but any remaining should probably become `nubOrd`s (from extra) instead.  Agreed?

- [ ] Stack.Ghci will now get the unlisted modules in `ghciPkgModules` (previously it only got the listed ones).  Is that the correct behaviour?

- [ ] The `Stack.Iface` module Dan added isn't used at all (this code draws inspiration from it, though).  Should I just remove that?

ping @chrisdone @snoyberg


borsboom:
An idea mentioned in https://github.com/commercialhaskell/stack/issues/105#issuecomment-122026872 is to display warnings for any files not listed from the .cabal file.  For modules, this should be straightforward, but for TH dependencies I don't think there's a good way to know whether the file should be listed or not.  For example, using gitrev makes a module _optionally_ depend on `.git/packed-refs` and `.git/index`, which obviously should _not_ be be listed in extra-source-files.

stack.cabal
```diff
@@ -178,6 +178,7 @@ executable stack
   ghc-options:    -Wall -threaded -rtsopts
   other-modules:     Plugins
                      Plugins.Commands
+                     Paths_stack
```
snoyberg: Was this discovered by this code itself?

snoyberg:
> If there are unlisted/TH dependencies, the second time you build after a clean (without changing any files in the interim) it thinks those files are new and so it runs the build anyway

Actually, I think your current solution is the right one, it avoids an ugly race condition where a TH-loaded file is edited after the build is started.

> That seems to be pretty fast, but caching would be faster. Worth doing?

Not yet, let's wait until we see it cause trouble in practice

> Currently there are a number list->Set->list conversions. To fix that, it'd mean changing a bunch of stuff in the module to use Sets instead of lists. Any objection to doing so?

None at all, but let's wait till this is merged to master and then do the cleanup on master itself.

> I noticed nub is used a bunch in the existing code. I think converting the code to use Sets would eliminate many of those nubs, but any remaining should probably become nubOrds (from extra) instead. Agreed?

Absolutely

> Stack.Ghci will now get the unlisted modules in ghciPkgModules (previously it only got the listed ones). Is that the correct behaviour?

I'll need to defer to @chrisdone on that

>  The Stack.Iface module Dan added isn't used at all (this code draws inspiration from it, though). Should I just remove that?

Yes, and may as well include it in this branch (though doing on master after merge is also fine).

* * *

Overall: I like the approach, and the code is shorter than I was hoping for, which is great. Having a few integration tests around this would be nice, but I'm OK merging this to master now. Please do so when you're ready. I'm doing some testing on Windows and will shout if there's a problem, but I doubt that will be the case.

snoyberg:
In testing (at least on Windows): rebuilding did *not* occur after a change to a TH-included file. I used `embedFile` to include `foo.txt`, built, and then modified that file with no rebuild triggered.

borsboom:
There seem to be worse issues on Windows.  Just saw this, seems to be an error from GHC itself while writing the .dump-hi file:

```
$ stack build  --copy-bins
stack-0.1.2.7-192640abed725b4dd084095d9f476aa3: unregistering (local file changes)
stack-0.1.2.7: build
Building stack-0.1.2.7...
Preprocessing library stack-0.1.2.7...
[40 of 64] Compiling Stack.Package    ( src\Stack\Package.hs, .stack-work\dist\i386-windows\Cabal-1.18.1.5\build\Stack\Package.o )
[41 of 64] Compiling Stack.Build.Types ( src\Stack\Build\Types.hs, .stack-work\dist\i386-windows\Cabal-1.18.1.5\build\Stack\Build\Types.o ) [Stack.Package changed]
[42 of 64] Compiling Stack.GhcPkg     ( src\Stack\GhcPkg.hs, .stack-work\dist\i386-windows\Cabal-1.18.1.5\build\Stack\GhcPkg.o ) [Stack.Build.Types changed]
.stack-work\dist\i386-windows\Cabal-1.18.1.5\build\src/Stack\GhcPkg.dump-hi: commitBuffer: invalid argument (invalid character)

--  While building package stack-0.1.2.7 using:
      C:\\Users\\emanu\\AppData\\Local\\Programs\\stack\\i386-windows\\ghc-7.8.4\\bin\\runhaskell.exe -package=Cabal-1.18.1.5 -clear-package-db -global-package-db -package-db=C:\Users\emanu\AppData\Roaming\stack\snapshots\i386-windows\lts-2.17\7.8.4\pkgdb\ C:\Users\emanu\AppData\Local\Temp\stack1620\Setup.hs --builddir=.stack-work\dist\i386-windows\Cabal-1.18.1.5\ build --ghc-options -hpcdir .stack-work\dist\i386-windows\Cabal-1.18.1.5\hpc\.hpc\ -ddump-hi -ddump-to-file
```

snoyberg:
That's interesting, I did _not_ trigger that bug in my testing.

borsboom:
@snoyberg TH dependency changes should now work on Windows (was a newline issue).

For the "invalid character" problem, have you tried building stack itself on Windows using this version of stack?  I've tried to reproduce on Linux using `LC_ALL=C`, but wasn't able to.  I tried the `chcp 65001` trick that works for Hakyll on Windows, but that didn't help.

borsboom:
Probably triggered by this bit of output it's trying to write to the .dump-hi (note the `α` in `(GHC.Types.IO α -> m α)`):

```
  $wunregisterGhcPkgId :: Control.Monad.IO.Class.MonadIO m
                          -> Control.Monad.Logger.MonadLogger m
                          -> Control.Monad.Catch.MonadCatch m
                          -> Control.Applicative.Applicative GHC.Types.IO
                          -> Control.Applicative.Applicative m
                          -> GHC.Base.Monad GHC.Types.IO
                          -> GHC.Base.Monad m
                          -> (GHC.Types.IO α -> m α)
                          -> System.Process.Read.EnvOverride
                          -> Path.Internal.Path Path.Abs Path.Dir
                          -> Stack.Types.GhcPkgId.GhcPkgId
                          -> m ()
```

snoyberg:
I can confirm that the original bug I reported is resolved. I'll play a bit more and see if I can reproduce the other problem.

snoyberg:
If you set your `LANG` environment variable on Windows, does that solve the problem?

borsboom:
No, doesn't seem to make any difference.

snoyberg:
Do you have any interesting environment variables set by any chance?

stack.cabal
borsboom: Yup!

borsboom:
I merged the Windows code page fix to master and then rebased this PR onto it.  Solves the "invalid character" for me.

chrisdone:
I'm still gonna review this when I get round to it, btw.

PR #750, opened 2015-08-11
`stack path` shows documentation roots.

duplode:
New options: `--snapshot-doc-root` and `--local-doc-root`

Motivation:
* Troubleshooting info; and
* Quick access to the documentation (e.g. `elinks $(stack path --snapshot-doc-root)/index.html`).

Coding style note: I was unsure about whether you'd rather have me adding reader monad actions to `Stack.Types.Config`, just like to those for the other paths. In the end, I went for the least invasive option, simply concatenating the paths *in situ*, imitating `Stack.Build.Haddock.snapDocDir`.


snoyberg:
Thanks!

PR #753, opened 2015-08-11
651 target parse overhaul

### LINKED ISSUE #651: Add --test and --bench options to build command

snoyberg:
Not quite ready for merge to master. I want to review the diff and do some cleanup, and then it's up to @borsboom on timing. I originally said this should be merged after the upcoming release, but this is already looking pretty stable, so I'm not necessarily opposed to including it now.

Pinging @chrisdone for review

Merging this to master resolves #651

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,7 @@
 
 * Detect unlisted modules and TemplateHaskell dependent files (#32, #105)
 * Overhauled target parsing, added `--test` and `--bench` options [#651](https://github.com/commercialhaskell/stack/issues/651)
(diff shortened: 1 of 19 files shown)


snoyberg:
Reassigning to @borsboom. I'm happy with the current state of this, so it's up to you when this should be merged to master.

snoyberg:
Discussed with @borsboom, we're moving ahead with this.

PR #760, opened 2015-08-11
Add text versions of strArgument and strOption

markus1189:
A common pattern found repeatedly is:

`fmap T.pack (strArgument _)` or `fmap T.pack (strOption _)`

This pr adds two functions in `Options/Applicative/Builder/Extra`, `textOption` and `textArgument` which are like  `strArgument` and `strOption` just for Text instead of String.  
It might be possible to instead use an upper bound on `optparse-applicative` depending on the outcome of pcapriotti/optparse-applicative#147, though it might be easier to just maintain local copies for now inside `stack`.  In order to prevent conflicts in the near future the local versions are called `text-` instead of `txt-` as in the  `optparse-applicative` pr.

### DIFF
--- a/src/Options/Applicative/Builder/Extra.hs
+++ b/src/Options/Applicative/Builder/Extra.hs
@@ -7,13 +7,18 @@ module Options.Applicative.Builder.Extra
   ,enableDisableFlags
   ,enableDisableFlagsNoDefault
   ,extraHelpOption
-  ,execExtraHelp)
+  ,execExtraHelp
+  ,textOption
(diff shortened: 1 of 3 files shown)


markus1189:
I get repeated build errors, is it possible that due to caching we miss them when building? It fails due to `mono-traversable` [history](https://travis-ci.org/markus1189/stack/builds) and [example](https://travis-ci.org/markus1189/stack/jobs/75126245)

chrisdone:
Merged!

PR #775, opened 2015-08-13
Switch to LTS 3.0

snoyberg:
No rush on getting this in


snoyberg:
Note that the downside of making this move is that it's more likely that people will break the build for GHC 7.8.

borsboom:
I'm all for doing it right away, and travis will catch any 7.8.4 broken builds.

Issue #736, opened 2015-08-08
Specify full version of GHC, not just major version WAS: Stack installs wrong version of GHC


hannes-landeholm:
Using resolver `nightly-2015-06-16` which is based on `ghc-7.10.1` causes stack to install `ghc-7.10.2` when I run `stack setup`.

It's also not possible to specify `ghc-7.10.2` as a resolver. I can only specify `ghc-7.10` which makes no sense as that's not even a real version of ghc. It doesn't mean anything as version numbers are arbitrary incremented.

borsboom:
Stack only cares that the correct GHC major version (e.g. 7.10.*) is used for a snapshot, and `stack setup` will only install the latest for that major version (to be precise, the version specified in [stack-setup.yaml](https://github.com/fpco/stackage-content/blob/master/stack/stack-setup.yaml)).  However, if you already have an older GHC 7.10 (e.g. 7.10.1), it should use that (won't force you to upgrade to 7.10.2).

hannes-landeholm:
Why? The [snapshot nightly-2015-06-16](https://www.stackage.org/nightly-2015-06-16) clearly states that it's based on `ghc-7.10.1`. `ghc-7.10.2` contains back compatibility breaking changes.

Related back compatibility breaking change:

https://downloads.haskell.org/~ghc/7.10.1/docs/html/libraries/ghc/TcRnDriver.html

https://downloads.haskell.org/~ghc/7.10.2/docs/html/libraries/ghc-7.10.2/TcRnDriver.html

Related issue: https://github.com/ucsd-progsys/liquidhaskell/issues/420

borsboom:
I think we'll have to see if @snoyberg can give more detail about the reasoning, but my assumption is because it was thought to be the more desirable behaviour in most cases.  And we must have believed "[Patchlevels are bug-fix releases only, and never change the programmer interface to any system-supplied code.](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/version-numbering.html)"

I could definitely see supporting a more strict interpretation of GHC versions being desirable, at least optionally, so worth discussing here.  In the mean time, you should be able to work around this by installing GHC 7.10.1 yourself and using `stack --system-ghc`.

snoyberg:
This was by design originally, for exactly the reasons @borsboom mentioned, but I was conflicted about it, and in hindsight think it was the wrong decision. I'm in favor of moving to:

* Specifying GHC version explicitly
[...]

PR #784, opened 2015-08-13
Use exact GHC versions rather than major-version #736

mgsloan:
To test before https://github.com/fpco/stackage-content/pull/8 is merged, replace [this line](https://github.com/commercialhaskell/stack/blob/736-exact-ghc-versions/src/Stack/Setup.hs#L497) with

```
    req = "https://raw.githubusercontent.com/fpco/stackage-content/c9165a5845370cf83b2dd373a66cb9a80f9bf74d/stack/stack-setup-2.yaml"
```

Description from the commit:

* Requires the exact GHC version specified by the snapshot.

* Deprecates major-version (ghc-7.10, ghc-7.8) resolvers, but still
  supports them.  They resolve to the newest installed / available ghc
  matching that number.

* Changes the format of stack-setup.yaml, and so changes which URL is
  used to find it (in order to not break old stack versions).

* Refactors ensureTool code, as it already had a lot of special cases,
  which I found confusing.  Main cause is that I needed to pass in a
  'CompilerVersion' instead of 'Version', but just for installing ghc,
  not for git.

* Introduces a 'CompilerVersion' type, and changes some naming to
  specify that compiler versions are being passed around rather than ghc
  versions.  The change could be a simpler without this, but this will
  be helpful for GHCJS support.

### DIFF
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -735,7 +735,8 @@ data CustomSnapshot = CustomSnapshot
 instance FromJSON CustomSnapshot where
     parseJSON = withObject "CustomSnapshot" $ \o -> CustomSnapshot
         <$> ((o .: "compiler") >>= (\t -> maybe (fail $ "Invalid compiler: " ++ T.unpack t) return $ do
-                GhcVersion v <- parseCompilerVersion t
-                return v))
+                cv <- parseCompilerVersion t
+                case cv of
+                    GhcVersion v -> return v))
         <*> o .: "packages"
         <*> o .:? "flags" .!= Map.empty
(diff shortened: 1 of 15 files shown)


mgsloan:
Note: While working on GHCJS atop this, I've just realized that a few things ought to be changed here. (so no merging till I've fixed that!)

src/Stack/BuildPlan.hs
```diff
@@ -735,7 +735,7 @@ data CustomSnapshot = CustomSnapshot
 instance FromJSON CustomSnapshot where
     parseJSON = withObject "CustomSnapshot" $ \o -> CustomSnapshot
         <$> ((o .: "compiler") >>= (\t -> maybe (fail $ "Invalid compiler: " ++ T.unpack t) return $ do
-                t' <- T.stripPrefix "ghc-" t
-                parseVersionFromString $ T.unpack t'))
+                GhcVersion v <- parseCompilerVersion t
```
snoyberg: Probably shouldn't use this kind of match, as the compiler will hide the incomplete pattern match, and you're about to add more constructors (I presume).

snoyberg: I'll make the change myself, it will also give me a chance to compile/test it locally.

PR #789, opened 2015-08-14
Add --git-repo to stack upgrade

### LINKED ISSUE #787: Allow git repo for stack upgrade to be configurable
In the source I already see

`                    , "https://github.com/commercialhaskell/stack" -- TODO allow to be configured`

but no relevant issue

Fuuzetsu:
Default is the same as always and the flag does nothing worthwhile if
--git is not specified too.

Closes #787


snoyberg:
Thanks!

Issue #799, opened 2015-08-16
stack --version should show architecture (x86 / 64)


joozek78:

borsboom:
Moved to #800.

PR #800, opened 2015-08-16
Add arch info to --version #799

joozek78:


snoyberg:
I'd recommend instead using the values from Distribution.System, like we do at:

https://github.com/commercialhaskell/stack/blob/2200266ac823b7fa60757b27b692b8e935ee9df4/src/Stack/Config.hs#L128

It will avoid the need for CPP or hard-coding a list of values.

joozek78:
I was hoping for such simple way of getting that, but couldn't find it.
Will update in a day or two

snoyberg:
Thanks! For future reference, it's a good idea to ping the PR with a comment after making a change, as Github doesn't send any notifications.

Issue #337, opened 2015-06-17
GHCJS?


jkozlowski:
Are there plans to support GHCJS in the same first class way as GHC (i.e. getting the binaries etc.)?

snoyberg:
Not yet, but I like the idea. Let's use this issue as a discussion point for it.

jkozlowski:
Thanks for a quick reply. I already saw some calls for better distribution on the ghcjs github (https://github.com/ghcjs/ghcjs/issues/231). I think if it would be just as painless to get ghcjs through stack as it is ghc (including getting ghcjs packages from stackage) it would greatly help with adoption.

cocreature:
Just to show that there's interest: I'm interested :)

bitemyapp:
This hinges in part on GHC 7.10 compatibility does it not?

mgsloan:
Nah, GHCJS works fine with 7.8.4.  I didn't realize there were compatibility issues with 7.10? (though further discussion of that should go in a different issue, if one doesn't already exist)

Anyway, agreed that this would be a nice thing to have!  It's particularly useful when sharing code between the client and server.  Currently I need to remember to run my ghcjs build script, and otherwise the protocol types can go out of sync.

bitemyapp:
Having an ez-pz way to get people going with GHCJS would be wonderful.

jkozlowski:
> This hinges in part on GHC 7.10 compatibility does it not?

GHCJS compatiblity? Granted that it's on a branch, but I am running on 7.10 fine (https://github.com/ghcjs/ghcjs/wiki/GHCJS-with-GHC-7.10)

bitemyapp:
@jkozlowski I thought Stack was GHC 7.8 only and didn't know GHCJS worked with 7.8

jkozlowski:
So, would the first step be to get binary builds setup somewhere/somehow? I don't know if FP Complete already does that somehow, or it exists somewhere else (no official ones, according to my issue on ghcjs github).

snoyberg:
We do already have Docker images with ghcjs binaries. Another option is just automating the process of building ghcjs

krisajenkins:
+1! Stack has now addressed every problem I have with Haskell adoption, save this one. I'm using Elm for frontend stuff, but I'm using it at its limits and would love to use GHCJS if you could work your supermagic on the install-process. :-)

mgsloan:
[...]

PR #802, opened 2015-08-17
Initial implementation of GHCJS support

mgsloan:


vertexcite:
If this can already be used with a stack built off the git master tip, please provide some instructions on how to use it. Thanks in advance.

mgsloan:
There are some instructions in the comment here: https://github.com/commercialhaskell/stack/issues/337#issuecomment-131677050

GHCJS will need to be on your PATH.  I used these instructions to get it setup with ghc-7.10 and the improved-base version of ghcjs: https://github.com/ghcjs/ghcjs/wiki/GHCJS-with-GHC-7.10

Issue #703, opened 2015-08-01
Exclude files from --file-watch and more debugging info


agrafix:
I'm currently moving a project to stack, and it's been working out great so far! I like the `--file-watch` flag, but it seems to rebuild way to often. I think this is due to my emacs configuration which creates temporary files in the source tree. Would it be possible to read .hgignore or .gitignore files to exclude files from being watched? For debugging purposes, it would also be very useful if `stack build --file-watch` gave some information why it thinks a rebuild is useful, eg:

Rebuilding due to changes in files X.hs, Z.hs, ...

snoyberg:
All of the filewatch code is in its own module at `Stack.FileWatch`, and should be relatively easy to modify without diving into the rest of the code. In particular, indicating which files changed should be a trivial change. Interested in trying your hand at a PR for this?

agrafix:
Sure, I'll look into this!

snoyberg:
I've pushed a commit to master that significantly improves this experience. I'm going to close this issue, please ping if master doesn't do what you're looking for.

agrafix:
Alright, great, thank you! I'm very busy this week, but I'll provide feedback next week.

PR #831, opened 2015-08-22
--file-watch should respect VCS boring/ignore files

agrafix:
( A first attempt of Part 1 for #703 )

This is a very basic implementation and I've moved out most of the logic to a separate package ( http://hackage.haskell.org/package/ignore ) as I would like to reuse it for other projects. Don't know if this is the desired approach, looking for feedback :-)

### DIFF
--- a/src/Stack/FileWatch.hs
+++ b/src/Stack/FileWatch.hs
@@ -19,6 +19,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.String (fromString)
 import Data.Traversable (forM)
+import Ignore
 import Path
 import System.FSNotify
 import System.IO (stderr)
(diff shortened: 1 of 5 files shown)


snoyberg:
Thanks!

luigy:
One of `ignore`'s dependencies requires having `pcre` headers installed
I had to install them through my package manager first on OS X 10.10... I thought they would come with Xcode :/

agrafix:
Yeah, to correctly interpret mercurials hgignore files you need pcre :-(

snoyberg:
Hmm, that's actually a serious problem, since it will greatly complicate
the windows installation story. Is there any subset of functionality we can
get without the library dependency?

agrafix:
The simplest thing would probably be to add a flag to the `ignore` package to build w/o pcre and then use that on windows. .gitignore files will still work fine as the run on globs, but .hgignore files will only work with `syntax: glob` enabled. Would that be a solution?

snoyberg:
It's a little awkward to provide different functionality by OS. And this
may be a problem on Mac as well, I'm not sure. Purely from the ease of
install perspective, my answer would be to fully drop mercurial support,
but that's not a great answer either.

borsboom:
It's not ideal on Linux either (I just had to install libpcre on a system for the latest version of stack).

agrafix:
I'll add a flag to the ignore package to remove pcre support for now as a quick fix, and then we can look for a long term fix.

Issue #810, opened 2015-08-17
`stack ide start` fails to start if the project has compile errors


CRogers:
[This gist has the minimal failing example](https://gist.github.com/CRogers/4c1ea6336404a6f12e9f). The expected behaviour of `stack ide start` is to start up stack-ide even if there are compile errors. However, stack now tries to build the project first which fails with compile errors:

```
$ stack --version
Version 0.1.3.1, Git revision 28a8f6a643d43a1cdbd3d2bc734359e4b74c7255
```

```
$ stack ide start --verbose
Version 0.1.3.1, Git revision 28a8f6a643d43a1cdbd3d2bc734359e4b74c7255
2015-08-17 18:41:19.778771: [debug] Checking for project config at: /Users/callumr/Dropbox/Programming/atom/atom-stack-ide/haskell/test-data/one-source-error/stack.yaml @(stack_5d6BW9mjZpYKU0uYDUjIqC:Stack.Config src/Stack/Config.hs:466:9)
2015-08-17 18:41:19.779075: [debug] Loading project config file stack.yaml @(stack_5d6BW9mjZpYKU0uYDUjIqC:Stack.Config src/Stack/Config.hs:489:13)
2015-08-17 18:41:19.784362: [debug] Run process: ghc --info @(stack_5d6BW9mjZpYKU0uYDUjIqC:System.Process.Read src/System/Process/Read.hs:257:3)
2015-08-17 18:41:19.845723: [debug] Run process: ghc --numeric-version @(stack_5d6BW9mjZpYKU0uYDUjIqC:System.Process.Read src/System/Process/Read.hs:257:3)
2015-08-17 18:41:19.880178: [debug] Run process: ghc-pkg --no-user-package-db field --simple-output Cabal id @(stack_5d6BW9mjZpYKU0uYDUjIqC:System.Process.Read src/System/Process/Read.hs:257:3)
2015-08-17 18:41:19.907664: [debug] Run process: ghc-pkg --no-user-package-db list --global @(stack_5d6BW9mjZpYKU0uYDUjIqC:System.Process.Read src/System/Process/Read.hs:257:3)
2015-08-17 18:41:19.933647: [debug] Checking resolver: lts-3.0 @(stack_5d6BW9mjZpYKU0uYDUjIqC:Stack.Build.Source src/Stack/Build/Source.hs:147:17)
[...]

PR #835, opened 2015-08-23
Still start up stack-ide and stack ghci even if there are project compilation errors.

CRogers:
See issue #810.

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -207,6 +207,7 @@ ghciSetup mainIs stringTargets = do
         , boptsBenchmarkOpts = (boptsBenchmarkOpts base)
           { beoDisableRun = True
           }
+        , boptsBuildSubset = BSOnlyDependencies
         }
       where
         base = defaultBuildOpts


src/Stack/Ghci.hs
```diff
@@ -192,7 +192,7 @@ ghciSetup mainIs stringTargets = do
         forM locals $
         \(name,(cabalfp,components)) ->
              makeGhciPkgInfo sourceMap (map fst locals) name cabalfp components
-    unless (M.null realTargets) (build (const (return ())) Nothing bopts)
+    try (unless (M.null realTargets) (build (const (return ())) Nothing bopts)) >>= (\x -> return (x :: Either SomeException ()))
```
CRogers: This is not great, both in that it needs the extra `>>=` to do the type annotation and it listens to all exception types. For the type annotation, I tried annotated with `try (build ...) :: m (Either SomeException ())` but ghc can't resolve `m` with `m` used in the type of the function. I tried to use `ScopedTypeVariables` but couldn't make it do what I want. Any ideas?

borsboom: Would something like this work?
```
(_ :: Either SomeException ()) <- try (unless (M.null realTargets) (build (const (return ())) Nothing bopts))
```

But catching SomeException isn't great, because that will also hide asynchronous exceptions (like hitting Ctrl-C) that should be allowed to propagate up the call stack.  What about catching `StackBuildException` instead?

@chrisdone, can you review this?  I'm not sure about the implications.  Should the build exception be displayed (maybe as a warning)?

CRogers:
After @rvion's [hint](https://github.com/commercialhaskell/stack/issues/810#issuecomment-134370231) I've instead made stack ghci/ide only build dependencies, as that will make sure that the correct packages are installed and will still start up stack ghci/ide even if there are compile errors in the project. It also has the benefit over the `try` approach of not polluting stdout with stuff that isn't related to stack ide, especially since api consumers will be expecting and trying to deserialise json.

chrisdone:
Looks good to me!

PR #853, opened 2015-08-26
Create tarballs rather than gzipping naked binaries during release.

mboes:
This has several advantages:
* The executable bit is set by default, so no need to `chmod +x` after
  decompressing.
* No need to rename binary to `stack` after decompressing.
* Can use tarball as the source argument for Docker `ADD` commands,
  for conveniently installing a stack binary into a docker image when
  the package manager not appropriate / not available.

The tarball still only contains just the `stack` binary.

### DIFF
--- a/scripts/release/release.hs
+++ b/scripts/release/release.hs
@@ -1,5 +1,5 @@
 #!/usr/bin/env stack
--- stack --install-ghc runghc --package=shake --package=extra --package=zip-archive --package=mime-types --package=http-types --package=http-conduit --package=text --package=conduit-combinators --package=conduit --package=case-insensitive --package=aeson --package=zlib --package executable-path
+-- stack --install-ghc runghc --package=shake --package=extra --package=zip-archive --package=mime-types --package=http-types --package=http-conduit --package=text --package=conduit-combinators --package=conduit --package=case-insensitive --package=aeson --package=zlib --package executable-path --package tar
 {-# OPTIONS_GHC -Wall -Werror #-}
 {-# LANGUAGE RecordWildCards #-}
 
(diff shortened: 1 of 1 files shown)


borsboom:
@mboes I haven't tried this out yet, but does it put the executable in a subdirectory in the tar that matches the tar's basename, as is conventional?  Even if it only contains one file, I'd still rather that be the case (I tend to be annoyed when archives don't follow this convention, and `tar --strip-components` works if you do actually want to uncompress without the extra directory).  It might be nice to put the README in there as well.  I'm fine with taking care of this myself for the next release.

mboes:
Oh... I didn't know `ADD` is no longer recommended. The Docker documentation is extremely confusing on this matter. As an SO commenter says, at once point `COPY` was deprecated, but looks like they ["deprecated their deprecation warning"](http://stackoverflow.com/questions/24958140/docker-copy-vs-add)...

The official reasons for preferring `COPY` are given [here](https://docs.docker.com/articles/dockerfile_best-practices/#add-or-copy), but make no sense to me. They say don't `ADD this there; RUN tar xf ...`, due to diskspace concerns, but isn't the point of `ADD` precisely that it does the unpacking automatically? Whether the temporary tarball gets automatically deleted by Docker sounds like an implementation detail to me.

That said, I'm happy either way if we include a subdir or not. The nice thing about the convention is that it avoids bombing your current directory with unpredictable new entries. Though it must be said that in this case arguably we aren't doing any worse: we're adding a single entry to the directory, whose name is a prefix of the tarball name - it's just that this entry happens to be an executable, not a directory.

mboes:
Ah! I understand now... `ADD` only decompresses *local* tarballs, not remote ones. Presumably due to some misguided security concern (misguided because the alternative `RUN curl ... && tar xf` is no more secure), or maybe due to buffering.

Okay so that makes `ADD` pretty useless. Not worth catering for that use case.

mboes:
Feature request submitted upstream: https://github.com/docker/docker/issues/15934.

PR #864, opened 2015-08-27
add some wording to soothe users about installing GHC

radix:
To address the concern that some users have had about stack installing GHC and potentially messing up their existing installations, this PR updates the README slightly to be more explicit about that, and also adds some output to the "setup" command so that they hopefully won't panic and try to interrupt the installation.

I thought of doing this while reading the intro to Snoyman's new Stack guide, where it says:

> Finally, stack is isolated: it will not make changes outside of specific stack directories (described below). Do not be worried if you see comments like "Installing GHC": stack will not tamper with your system packages at all. Additionally, stack packages will not interfere with packages installed by other build tools like cabal.

I also wanted to refer to a reddit thread I remember seeing where a user described their experience of running "stack setup" after being instructed to do so, and then panicking when they saw what it was doing, since they were afraid that it would mess up their existing GHC installation. I can't find the thread, though. :P

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ projects. It is aimed at Haskellers both new and experienced.
 
 It features:
 
-* Installing GHC automatically.
+* Installing GHC automatically, in an isolated location.
 * Installing packages needed for your project.
 * Building your project.
 * Testing your project.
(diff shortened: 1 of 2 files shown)


borsboom:
Thanks!

Issue #854, opened 2015-08-26
Resource vanished while downloading new ghc


k-bx:
```
➜  stack git:(master) stack setup
ghc-7.10.2:    6.12 MiB / 140.12 MiB (  4.37%) downloaded...<socket: 18>: hGetBuf: resource vanished (Connection reset by peer)
➜  stack git:(master) stack setup
ghc-7.10.2:   31.77 MiB / 140.12 MiB ( 22.67%) downloaded...<socket: 18>: hGetBuf: resource vanished (Connection reset by peer)
➜  stack git:(master) stack setup
ghc-7.10.2:   45.75 MiB / 140.12 MiB ( 32.65%) downloaded...<socket: 18>: hGetBuf: resource vanished (Connection reset by peer)
```

I don't have the best internet in the world, I guess (but I downloaded bigger files also). Ideal would be to retry with continuing download, something similar to `wget -c`.

markus1189:
Just an idea, but if stack uses `verifiedDownload` for the ghc download, I think we only have to include the exception from `hGetBuf` as another handler [here](https://github.com/commercialhaskell/stack/blob/master/src/Network/HTTP/Download/Verified.hs#L203).  Seems like `ResourceVanished` is an `IOError` currently we recover only from `HttpException`.

I'll do a pr later.

@k-bx is this reproducible, i.e. does this always happen if we disconnect from the internet while downloading?

k-bx:
@markus1189 the problem isn't reproduced by disconnecting from internet. Disconnecting from internet makes download just freeze (without continuation on reconnect). In order to reproduce this bug we'd need to write a small server (instead of the one from where we download), and make it "stop" (exit or kill thread without closing socket) before we finish our download.

markus1189:
Oh after rereading your first message I see it already did retry 3 times above.  That means we do not need to add more handlers,  one more idea would be to adjust the `drRetryPolicyDefault` to try a little harder?

k-bx:
@markus1189 while I don't fully understand current logic, I believe your initial suggestion is correct. It didn't retry 3 times, I manually re-launched it 3 times, so either adding `IOError` handling to stack, or converting it to `HttpException` on http-library level seems to make sense.

markus1189:
Oh okay then I misread the first message.  Actually the current logic is easy:

```haskell
drRetryPolicyDefault :: RetryPolicy
drRetryPolicyDefault = limitRetries 3 <> constantDelay onehundredMilliseconds
[...]

PR #865, opened 2015-08-27
Retry verifiedDownload for ResourceVanished IOException

markus1189:
As mentioned in #854 I think we should retry `verifiedDownload` for **some** `IOException`s.  Of course there are many cases where a retry will not have a chance of success, so I chose to make the handler as specific as possible.

Is this **too** specific, because it seems like it will be very `GHC` bound, due to the fact that we access functions from `import GHC.IO.Exception`?


DanBurton:
:+1:

borsboom:
Thanks!

PR #887, opened 2015-08-31
Full msys2 support

snoyberg:
Instead of using PortableGit, this moves stack over to using the full msys2. The advantage: it comes with pacman. This change also makes it so that the relevant --extra-lib-dirs and --extra-include-dirs are automatically set, meaning the following actually works:

    stack exec -- pacman -S mingw64/mingw-w64-x86_64-icu
    stack build text-icu. It's working on the msys2 branch

Props to @ygale for the thread at http://sourceforge.net/p/icu/mailman/message/34125921/.

I'd like to get some Windows user tests of this besides myself before this is merged.


vrdhn:
Is a binary available for this branch ?

snoyberg:
No, not yet. I'd recommend building from master (`stack upgrade --git`), since we'll likely be fixing things for a few newly merged features for the next few days.

Issue #897, opened 2015-09-02
Setup.hs cannot use local snapshot dependencies WAS stack failing to resolve local dependencies of multiple package project


mitchellwrosen:
Hi, I can't seem to reproduce this with a smaller example repo so I'll just point to the one that's failing: https://github.com/mitchellwrosen/cabal-macosx/tree/b6e5f714fbf451a72c8b7e70577ce7d9e342f8d0

I'm trying to stackify this three year old repo, and the example projects, which depend on the `cabal-macosx`, fail to build:

```
--  While building package gtkHello-0.1.0 using:
[...]

PR #899, opened 2015-09-02
Give Setup.hs all project dependencies #897

mitchellwrosen:


snoyberg:
Two thoughts:

1. I think using `-package-id` with the `GhcPkgId` would be more resilient
2. There's some funny business around the Cabal library itself, namely: stack always builds with the global Cabal package for various reproducibility issues, whereas the snapshot may contain a different version of the Cabal library. I think- though I'm not certain- that the `deps` should exclude any `Cabal` package listed there.

Would you be able to try out these two changes and confirm that it still works for your use case?

Thanks for looking into this!

mitchellwrosen:
Yeah. I also seem to have broken `stack sdist` :)

Am I correct in thinking that for `sdist`, it doesn't really matter if Setup.hs has access to all global, snapshot, and local packages? Or do the same reproducibility objections apply?

snoyberg:
I think that's correct, but perhaps @mgsloan can weigh in on the requirements when running `Setup.hs` for the sdist command.

mgsloan:
I'm not sure if the same reproducibility objections apply.  However, to me it makes sense to run Setup.hs in exactly the same way for both cases.

The current change will certainly break `sdist`, because having `deps = Set.singleton ident` means that Setup.hs depends on the package itself!  It should be possible to get dependencies for sdist as well.

mitchellwrosen:
Yeah, I know why it's failing, @mgsloan. I'm having a second look over the code - it's not entirely clear to me how to get a list of dependencies without actually executing a build and such with all of the infrastructure in Stack.Build.Execute.hs

snoyberg:
One thing worth considering: make a first stab that just change the behavior for `stack build`, but for `sdist` keep the old behavior, with a note that this could be improved in the future.

mitchellwrosen:
@snoyberg Ok, updated

snoyberg:
Thanks!

Issue #889, opened 2015-09-01
Replace Binary with something more reliable for file storage


snoyberg:
`Binary` instances are quite problematic for file storage, since they store no metadata about the schema itself. This has led multiple times to out-of-memory issues because an `Int` was parsed as a list size, for example. We have some hacks in the code base to essentially embed magic numbers to work around this, but there are certainly better ways we can solve this with improved libraries.

A non-solution: use JSON. We tried this already, and the performance hit was significant. We need a fast (de)serialization.

phadej:
https://github.com/well-typed/binary-serialise-cbor https://www.youtube.com/watch?v=60gUaOuZZsE it's not released but might be an option. [CBOR](http://cbor.io/) is essentially binary json, there are tags indicating what data comes next in the binary stream.

snoyberg:
Awesome, thanks for the link. I remembered hearing @dcoutts discuss this at some point, but didn't know what came of it. Duncan: are there plans to release this in the near term? And there doesn't seem to be any `Generic` support in the type class yet. Is that intentional, or would `Generic` support be a good thing for the library?

phadej:
@snoyberg I have made a small package https://github.com/phadej/binary-tagged during "the morning hack". It's missing a bit of essential instances; but could it be used as a possible solution to this? It's not yet on Hackage, but it could be as soon as I verify API in our own dependant project.

snoyberg:
It definitely looks like an improvement. My concerns are:

* It looks like this requires explicit versioning, am I misreading that?
* How big a performance hit would this introduce?

phadej:
First question: it allows explicit versioning, but data is implicitly versioned by its structure: generic-derivable `HasNominalSop`. Explicit versioning might be needed sometimes, but with enough `newtypes` it should be needed (time shows). Actually one could write tests using this, to track when data formats actually change, i.e. hardcode hash into tests.

Second question: If only top-level data is tagged, then the performance hit shouldn't be noticeable. I'll write some trivial benchmarks to verify that.

snoyberg:
Sounds good, I'm liking where this is headed :)

phadej:
With https://github.com/phadej/binary-tagged/blob/wip/bench/Bench.hs (which is quite nice usage example as well) the result is:
![fireshot capture - criterion report - http___localhost_8000_bench html](https://cloud.githubusercontent.com/assets/51087/9719788/aa4ea418-5590-11e5-87eb-8dca842b617d.png)

[...]

PR #902, opened 2015-09-02
Adapt to upcoming Cabal installed package identifier format change

### LINKED ISSUE #851: Stack assumes IPIDs have a specific format
`parseGhcPkgId` assumes that an installed package ID is always a package name, a version, and a hash. IPIDs are opaque identifiers generated by Cabal, and Cabal reserves the right to change them (and will be changing them soon, c.f. https://github.com/haskell/cabal/issues/2745 )

snoyberg:
Fixes  #851

### DIFF
--- a/src/Stack/Types/GhcPkgId.hs
+++ b/src/Stack/Types/GhcPkgId.hs
@@ -8,24 +8,22 @@ module Stack.Types.GhcPkgId
   (GhcPkgId
   ,ghcPkgIdParser
   ,parseGhcPkgId
-  ,ghcPkgIdString
-  ,ghcPkgIdPackageIdentifier)
+  ,ghcPkgIdString)
   where
 
 import           Control.Applicative
 import           Control.Monad.Catch
 import           Data.Aeson.Extended
-import           Data.Attoparsec.ByteString.Char8
+import           Data.Attoparsec.ByteString.Char8 as A8
+import           Data.Binary (getWord8, putWord8)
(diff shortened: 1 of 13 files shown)


src/Stack/Types/GhcPkgId.hs
```diff
@@ -36,13 +34,24 @@ instance Show GhcPkgIdParseFail where
...
 
 instance Hashable GhcPkgId
-instance Binary GhcPkgId
+instance Binary GhcPkgId where
+    put (GhcPkgId x) = do
+        -- magic string
+        putWord8 1
+        putWord8 3
+        putWord8 4
+        putWord8 7
```
chrisdone: What's this shenanigans?

chrisdone:
I'm happy with this patch. Seems pretty mechanical. Laborious to do, but straight-forward updates.

src/Stack/Types/GhcPkgId.hs
snoyberg: Make sure that the binary format being read is the same one that was written. See: https://github.com/commercialhaskell/stack/issues/889

snoyberg:
I'd say: let's merge this in immediately after the next release is cut. I don't want to rock the boat on this release for such a non-urgent change, but would like to maximize testing during the next release cycle. Similar thoughts apply to #887. @borsboom sound good?

chrisdone:
:+1:

PR #903, opened 2015-09-02
Hacky workaround for optparse-applicative issue with stack exec --help

### LINKED ISSUE #806: stack exec --help is invalid
```
Invalid option `--help'

Usage: stack exec CMD [-- ARGS (e.g. stack ghc -- X.hs -o x)] ([--plain] |
                  ([--ghc-package-path] | [--no-ghc-package-path])
                  ([--stack-exe] | [--no-stack-exe]) [--package ARG])
  Execute a command
```

snoyberg:
Fixes #806

Downside: usage information now looks like:

Usage: stack exec [CMD -- ARGS (e.g. stack ghc -- X.hs -o x)]

Notice how "CMD" is inside the square brackets. Still seems like a
worthwhile tradeoff to me.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -31,6 +31,7 @@ Bug fixes:
 * Create missing directories for `stack sdist`
 * Don't ignore .cabal files with extra periods [#895](https://github.com/commercialhaskell/stack/issues/895)
 * Deprecate unused `--optimizations` flag
+* Hacky workaround for optparse-applicative issue with `stack exec --help` [#806](https://github.com/commercialhaskell/stack/issues/806)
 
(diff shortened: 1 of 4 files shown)


chrisdone:
Took me a while to understand what this patch does. I see the problem it solves is `stack exec CMD` requires CMD to be passed regardless of any other options, so `stack exec foo --help` will print the help, but `stack exec --help`  complains, in its stupid way, that CMD is missing (but actually says `--help` is an invalid option).

borsboom:
Merged this, but I think I'll keep the original issue open for now in the hopes of finding a non-hacky way to fix it.

Issue #894, opened 2015-09-01
Proposal: expanded documentation on stack's relationship with other tools


kairuku:
(I apologize if these topics have already been covered somewhere else on the Web and I've overlooked this.)

I'm somewhat confused on how stack relates to some other tools (besides cabal), such as *stackage-cli, cabal-src and cabal-meta,* especially whether or not stack is *replacing* any other tools (or perhaps might someday in the future), including any other tools I'm not aware of and didn't mention, but which are also seemingly similar.

For instance, it seemed in the initial public announcements of both stack and stackage-cli that both were covering some similar workflows, and now I wonder if stack has completely superseded either stackage-cli outright or some of the stackage-cli *plugins*.

So I thought maybe it could be useful to add documentation, possibly even a separate markdown file, on specifically this aspect of stack, including any relevant roadmap.

snoyberg:
This could be a great addition to: https://github.com/commercialhaskell/stack/blob/master/GUIDE.md#comparison-to-other-tools. Would you mind sending a PR with an initial comparison to other tools? A newcomer's impression to this would be very good. I'll be happy to fill in any details that need explaining.

kairuku:
I'm new to forking and pull-requests (and at the sub-novice level with Haskell too) so please bare with me :)  I didn't go to deeply yet into differentiation of tools, and in this first branch have mainly just explored some ideas in creating a new directory structure for documentation (I'm happy to undo any of this too). I'll need to repair some URLs in a subsequent commit if this were to be approved for merging.

kairuku:
Sorry, I did not see [#900](https://github.com/commercialhaskell/stack/issues/900) before making my pull-request

Issue #900, opened 2015-09-02
Documentation cleanup


snoyberg:
We've had a few side discussions floating around, so I wanted to try and congregate a bunch of different documentation ideas into one place:

* In many cases: move Wiki files to the repo itself. (Recommendation: make a `doc/` subdirectory, point old files to that location, and possibly in the future host that `doc/` directory on something like https://readthedocs.org/.)
* Update the CONTRIBUTING.md about:
    * Where documentation is kept, and how to update it
    * Issue tracking behavior
* Possibly add a dedicated document for future improvements (alternatively: keep in a Github milestone)
* Make sure the README.md and Wiki homepage both reflect all of these changes.

I'll likely do some minor parts of this myself if they're non-controversial, but will hold off on any significant changes.

borsboom:
Should we move GUIDE.md into `doc/` (and update any links to it)?

drwebb:
+1 My eyes would look for a doc directory before scanning for the file
manfest for markdown files.

Emanuel Borsboom writes:

> Should we move GUIDE.md into `doc/` (and update any links to it)?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/commercialhaskell/stack/issues/900#issuecomment-137276245

-- 
Tristan Webb PhD
Application Engineer
FP Complete

snoyberg:
Yes, but I'd keep a file in the old location too with a link to the new
location, at least for the next few months.

borsboom:
@kairuku, are you interested in taking a stab at moving additional content from the wiki into main repo?  I did a quick run-through of the pages that exist, and I think:

Should move to repo:
* Architecture
* Build command
* Custom snapshot
* Docker
* stack.yaml

Should stay on wiki:
* Home (but anything it links to not already linked in the README should probably be copied to the README).
* Lowering the bar to open source contributions
* New to Haskell
* Nonstandard project initialization
* Script interpreter
* Signing key (this will go on the new web site once it's ready)
* Stack's origins
* Transition guide

Not sure:
* Checklist
* Downloads
* FAQ
* Shell autocompletion (might be best to copy this content into GUIDE, and leave the old version on the wiki for now with a link to the GUIDE)
* Travis
* Visualizing dependencies with Graphviz (maybe should be copied into GUIDE?)

kairuku:
[...]

PR #923, opened 2015-09-04
initial restructuring of docs and barebones inclusion of other tools re: commercialhaskell/stack#894

kairuku:
-- creates new directory structure for documentation, aiming for smaller topic-based files and
   limiting repetition of content as much as possible (could potentially use table of content
   files to point to these, and potentially further nest groups of related topic files into new
   subdirectories, letting the "toc" files hide this organizational detail)

-- creates a separate doc for motivation behind stack, tools comparison, and future roadmap

-- pulls the "why" section out of the readme and the "comparison" section out of the guide,
   moving these both to the new document mentioned above

-- moves as much as possible among documentation-esque files from the root directory of the repo
   into the new docs directory, leaving only the readme in the top level directory

-- edits readme to point to the new document (URLs will need to be fixed in a subsequent commit)

-- divides docs directory into subdirectories for maintainer docs, and docs for everyone
   (inspired initially by a "release" document found elsewhere in the repo)

-- creates suggested stub directories for haddocks and man pages, but nothing has been moved here
   (one idea is that haddocks placed here with all the other documentation could be rsynced to any
   "dist" directory during any build process if needed)

-- creates suggested example stub file for wiki migration (assuming this is a desirable paradigm),
   specifically a "faq" file


kairuku:
(moving this comment from the "issue" to this "conversation")

I'm new to forking and pull-requests (and at the sub-novice level with Haskell too) so please bare with me :)  I didn't yet go deeply into differentiation of tools, and in this first branch have mainly just explored some ideas in creating a new directory structure for documentation (I'm happy to undo any of this too). I'll need to repair some URLs in a subsequent commit if this were to be approved for merging.

Minor notes: I wasn't sure specific markdown flavor was being targeted (I've just heretofore leaned on the syntax guide in Haroopad, but noted some difference in, e.g., the GUIDE.md file), nor whether it was desirable to break lines in the markdown files around column 80 or if this was just incidental and not necessary.

kairuku:
ref [#894](https://github.com/commercialhaskell/stack/issues/894)

README.md
```diff
@@ -3,8 +3,7 @@
 [![Build Status](https://travis-ci.org/commercialhaskell/stack.svg?branch=master)](https://travis-ci.org/commercialhaskell/stack)
 [![Release](https://img.shields.io/github/release/commercialhaskell/stack.svg)](https://github.com/commercialhaskell/stack/releases)
 
-`stack` is a cross-platform program for developing Haskell
-projects. It is aimed at Haskellers both new and experienced.
+`stack` is a cross-platform program for developing Haskell projects. It is aimed at Haskellers both new and experienced.
```
snoyberg: Why the collapsed line?

README.md
```diff
@@ -16,6 +15,8 @@ It features:
 * Testing your project.
 * Benchmarking your project.
 
+`stack` replaces `stackage-cli`. For more details on the motivation for the creation of `stack`, a comparison of `stack` to other tools, and the roadmap for the future of `stack', [see this document]().
```
snoyberg: I'm not sure stackage-cli was ever popular enough for this message to warrant being front and center in the README, I think it will end up leading to more confusion around "what's stackage-cli?"

docs/primary/topic_based_expanded_reference/GUIDE.md
```diff
@@ -1352,30 +1352,6 @@ reminder for future uses of stack:
 * stack.yaml can override the flag settings for individual packages
 * stack.yaml specifies which packages to include
 
-## Comparison to other tools
```
snoyberg: I'm not certain this should be moved out of the guide. If it is:

1. @wolftune just did a significant overhaul of the text in the guide, so the updated text should be copied over.
2. We should have a link to the new file added here.

snoyberg:
Thanks for getting started on this! I've added some inline comments, two meta comments:

1. I'm not sure if a deeply-nested directory structure is beneficial here, it may simply prevent people from finding documentation.
2. I'd like to keep a GUIDE.md in the original location, containing a link to the new location, for those already sharing the link.

README.md
wolftune: reference to stackage-cli looks like unneeded trivia to me that most people don't need to know at all.

README.md
kairuku: Sorry, I wasn't sure it was intentional, and it's just force of habit for me to remove line breaks in anything but stuff that is only likely to be read inside a small terminal window.

README.md
kairuku: Understood. I was having the opposite problem; I didn't even realize stackage-cli was deprecated until I stumbled upon its actual Hackage page, after opening #894 :)

docs/primary/topic_based_expanded_reference/GUIDE.md
kairuku: I was just trying to break things down where possible into smaller "topic" type files that could potentially be referenced from multiple places, e.g., "Installation and Upgrading" could've been another potential "topic".

kairuku:
@snoyberg I've scaled back the PR some

snoyberg:
I don't know how realistic a future merger of stack and Halcyon is, and I wouldn't want to include that without @mietek's blessing. As for comparison to other tools:

* cabal-meta inspired a lot of the multi-package functionality of stack. If you're still using cabal-install, cabal-meta is relevant. For stack work, the feature set is fully subsumed by stack.
* cabal-src is mostly irrelevant in the presence of both stack and cabal sandboxes, both of which make it easier to add additional package sources easily. The mega-sdist executable that ships with cabal-src is, however, still relevant. Its functionality may some day be folded into stack
* stackage-cli was an initial attempt to make cabal-install work more easily with curated snapshots, but due to a slight impedance mismatch between cabal.config constraints and snapshots, it did not work as well as hoped. It is deprecated in favor of stack.

Does that sound about right?

kairuku:
@snoyberg I think it looks great, and you know more about this than I :) I added your explanations verbatim if that is okay. Hopefully this section could be of some value to someone in the future and won't be just trivia; I cleared up my own initial confusion about stack vs stackage-cli after opening the PR.

I removed the Roadmap section because its only content current was an example link to the Google Groups discussion, but I still kind of think the general concept of a Roadmap section located adjacent to any Comparison content (or place in a separate file, linked to from any relevant section in any other file) could possibly be useful.

I could still make a /doc directory, I just wasn't sure if there was still a waiting period for comments on #900 to settle on some naming scheme, etc, but these things could also be addressed in a separate PR too. I'm not sure what anyone might have thought about (or the implications of) moving things like the man page(s), haddocks, LICENSE , CHANGELOG, etc, under /doc, it just seemed related in my mind.

snoyberg:
Thanks! I think a separate PR for the doc directory would be great.

PR #926, opened 2015-09-05
Fix warning about unicode specifier

emwap:
src/Stack/Docker.hs:573:31:
     warning: \u used with no following hex digits; treating as '\' followed by identifier [-Wunicode]
                       ,maybe [] (\u -> ["--username=" ++ u]) (dockerRegistryUsername docker)

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -570,7 +570,7 @@ pullImage envOverride docker image =
                 "docker"
                 (concat
                    [["login"]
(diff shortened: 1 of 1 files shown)


borsboom:
I'll merge this since it does no harm, but it probably points to a strange setup on your system.  I'm going to guess you're on OS X where these sorts of warnings happened for some people when OS X transitioned to clang (from GCC).

emwap:
Ok, thanks for the tip.

Issue #840, opened 2015-08-24
PATH check should canonicalize paths


DanielG:
I'm getting `Installation path /home/dxld/.local/bin/ not found in PATH environment variable` because `.local` is a symlink on my machine but the target of this symlink is actually on the PATH. I can't be bothered to dig up where this comparison is being made right now but `canonicalizePath` should probably be applied to both sides before checking for equality.

snoyberg:
```
$ grep 'not found in PATH' -rn src/
src/Stack/Build/Execute.hs:336:                , " not found in PATH environment variable"
```

PR welcome, it should be a matter of just `mapM canonicalizePath`, though you should be careful to handle the case where a directory doesn't exist.

bollu:
@snoyberg - This is going to sound silly, but going through the `stack` source code is kind of daunting, since I've never seen "professional" Haskell code before. It would be a great help if you could point to as to where to start from.

snoyberg:
My previous comment gives the relevant filename/line number: `src/Stack/Build/Execute.hs:336`.

bollu:
Yes, I saw that :)

So, well, i wasn't talking about this change - I meant in general, about
how to go through Haskell source when you're unfamiliar with the patterns
and libraries involved.

If it were OO, then I would look for interfaces. Otherwise, in, say,
python, I'd explore types using the interpreter and play with chunks of the
code. I'm not sure what the equivalent is for Haskell.

PR #932, opened 2015-09-06
Check canonicalized `local-bin-path`

bixuanzju:
Silent warning if `local-bin-path` is a symlink to a target that is on
the PATH.

Ref #840


snoyberg:
Thanks! This isn't necessary, but just a comment. This implementation (`fmap any`) will force all of the directories to be canonicalized, since the IO actions are not lazy. An `anyM` implementation could be more efficient, but admittedly would not be as elegant/composable as what you did here.

Issue #811, opened 2015-08-18
"stack upgrade --git" rebuilds even when stack if up-to-date


rvion:
`stack upgrade --git` rebuilds stack even when stack is up-to-date
To repro, one can run `stack upgrade --git` twice and see what happens

I have
```
stack --version
Version 0.1.3.1, Git revision 3da2a5b1faf8ef045fed550840647571cda9ff3e
```

snoyberg:
This is enough of a corner case that I wasn't worried about it. If you'd like to improve the situation, PR welcome, all of the code should be in Stack.Upgrade.

rvion:
works for me, I'll give it a try once I'll have taken the time to finish setting up appveyor

snoyberg:
Resolved by #937

PR #937, opened 2015-09-06
Don't rebuild stack if it is up-to-date

bixuanzju:
Ref #811

I've tested on my local fork, it works. But I am not sure if I am doing it in a robust way. Perhaps I could have made a customised `readProcess` just like what you have as [`callProcess`](https://github.com/commercialhaskell/stack/blob/master/src/System/Process/Run.hs#L63).

### DIFF
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE TemplateHaskell       #-}
 module Stack.Upgrade (upgrade) where
 
+import           Control.Monad               (when)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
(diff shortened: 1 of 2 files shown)


src/Stack/Upgrade.hs
```diff
@@ -38,6 +41,9 @@ upgrade gitRepo mresolver = withSystemTempDirectory "stack-upgrade" $ \tmp' -> d
     tmp <- parseAbsDir tmp'
     mdir <- case gitRepo of
       Just repo -> do
+        remote <- liftIO $ readProcess "git" ["ls-remote", repo, "master"] []
+        let latestCommit = head . words $ remote
+        when (latestCommit == $(gitHash)) $ error "Already up-to-date, no upgrade required"
```
snoyberg: Just for future reference: you can replace `$(gitHash)` with `$gitHash`.

snoyberg: Also: I don't think we should consider this an error condition, as that could break automated tooling that would simply want to know that it's using the newest stack. Can you instead just print a message and return `Nothing` like the non-Git case does?

bixuanzju: Sure!

bixuanzju: @snoyberg Done! Please have a look

snoyberg:
Looks great, thanks! I'll just wait for the Travis green light and then merge.

snoyberg:
Thanks!

PR #939, opened 2015-09-07
complete copy-edit, clean up of GUIDE.md

wolftune:
I decided to read the GUIDE.md by editing it.
I thoroughly cleaned up the file, removing extraneous words,
fixing line lengths, making grammar fixes, clarifying unclear bits,
improving formatting, and generally improving anywhere I noticed room for improvement.

I know not every change may be precisely the preference of others, but I really didn't change any content, just made it more readable, cleaner, etc. I think the vast majority of this is just clear improvements. Unfortunately, the diff is a bit hard to follow because line changes in places make git unable to pick out what were mostly only small changes, just lots of them. The reflowing in cases makes the diff a bit less skimmable though.

One content change though: I removed the bit about the needed extra flag for bug prior to 0.1.4 because that's out now, so the note about "won't be needed once that is out" is no longer relevant.

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -9,7 +9,7 @@ with the Haskell packaging system or other build tools.
 stack handles the management of your toolchain (including GHC — the Glasgow
 Haskell Compiler — and, for Windows users, MSYS), building and registering
 libraries, building build tool dependencies, and more. While it can use existing
-tools on your system, stack's has the capacity to be your one-stop shop for all
+tools on your system, stack has the capacity to be your one-stop shop for all
 Haskell tooling you need. This guide will follow that stack-centric approach.
 
 ### What makes stack special?
@@ -242,7 +242,7 @@ In this subsection, we'll dissect the helloworld example in more detail.
 
 ### Files in helloworld
 
-Before moving studying stack more, let's understand our project a bit better.
+Before studying stack more, let's understand our project a bit better.
 
 ```
 michael@d30748af6d3d:~/helloworld$ find * -type f
@@ -281,13 +281,13 @@ extra-deps: []
 resolver: lts-3.2
 ```
 
-As is obvious if you're familiar with YAML, the `flags` and `extra-deps` keys
-have empty values. We'll see more interesting usages for these fields later.
-Let's focus on the other two fields. `packages` tells stack which local packages
-to build. In our simple example, we have only a single package in our project,
-located in the same directory, so `'.'` suffices. However, stack has powerful
-support for multi-package projects, which we'll elaborate on as this guide
-progresses.
+If you're familiar with YAML, you may recognize that the `flags` and
+`extra-deps` keys have empty values. We'll see more interesting usages for these
+fields later.  Let's focus on the other two fields. `packages` tells stack which
+local packages to build. In our simple example, we have only a single package in
+our project, located in the same directory, so `'.'` suffices. However, stack
+has powerful support for multi-package projects, which we'll elaborate on as
+this guide progresses.
 
 The final field is resolver. This tells stack *how* to build your package:
 which GHC version to use, versions of package dependencies, and so on. Our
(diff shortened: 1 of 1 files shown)


GUIDE.md
```diff
@@ -1,64 +1,57 @@
...
+This guide takes a new stack user through the typical workflows. This guide
+will not teach Haskell or involve much code, and it requires no prior experience
+with the Haskell packaging system or other build tools.
+
+## Stack's functions
+
+stack handles the management of your toolchain (including GHC — the Glasgow
+Haskell Compiler — and, for Windows users, MSYS), building and registering
+libraries, building build tool dependencies, and more. While it can use existing
+tools on your system, stack's has the capacity to be your one-stop shop for all
```
snoyberg: s/stack's/stack

GUIDE.md
```diff
@@ -246,12 +238,11 @@ Test suite not yet implemented
...
 ## Inner Workings of stack
 
-In this subsection, we'll dissect in more detail the helloworld
-example.
+In this subsection, we'll dissect the helloworld example in more detail.
 
 ### Files in helloworld
 
-Before moving on with understanding stack a bit better, let's understand our project just a bit better.
+Before moving studying stack more, let's understand our project a bit better.
```
snoyberg: There are some words missing

GUIDE.md
```diff
@@ -290,10 +281,10 @@ extra-deps: []
 resolver: lts-3.2
 ```
 
-If you're familiar with YAML, you'll see that the flags and extra-deps keys
+As is obvious if you're familiar with YAML, the `flags` and `extra-deps` keys
```
snoyberg: This change in wording comes across as slightly condescending to me

GUIDE.md
```diff
@@ -587,11 +581,10 @@ No GHC found, expected version 7.8.4 (x86_64) (based on resolver setting in /hom
 ```
 
 This fails, because GHC 7.8.4 (which lts-2.22 uses) is not available on our
-system. The first lesson is: when you want to change your GHC version, modify
-the resolver value. Now the question is: how do we get the right GHC version?
-One answer is to use `stack setup` like we did above, this time with the
-`--resolver lts-2` option. However, there's another way worth mentioning: the
-`--install-ghc` flag.
+system. This shows us that we can change GHC version by modifying the resolver
```
snoyberg: "change GHC version" doesn't sound quite right.

GUIDE.md
```diff
@@ -1288,22 +1331,22 @@ I installed the stm package via --package stm
...
     stack exec ghci
 
-However, this doesn't do anything particularly intelligent, such as loading up
-locally written modules. For that reason, the `stack ghci` command is
-available.
-
-NOTE: At the time of writing, `stack ghci` was still an experimental feature,
-so I'm not going to devote a lot more time to it. Future readers: feel free to
-expand this!
+But that won't loading up locally written modules for access. For that, use the
```
snoyberg: s/loading/load ?

GUIDE.md
```diff
@@ -1288,22 +1331,22 @@ I installed the stm package via --package stm
...
+`stack ghci` command. To then load modules from your project, use the `:m` (for
+"module" command followed by the module name.
```
snoyberg: Missing closing paren

GUIDE.md
```diff
@@ -1522,34 +1601,43 @@ built packages, meaning that subsequent builds will be much faster.
 
 Two notes for future improvement:
 
-* One Travis whitelists the stack .deb files, we'll be able to simply include stack in the `addons` section, and automatically use the newest version of stack, avoiding that complicated `before_install` section
-* Starting with stack-0.1.4.0, there are improvements to the test command, so that the entire script section can be `stack --no-terminal --install-ghc test`
+* One Travis whitelists the stack .deb files, we'll be able to simply include
```
snoyberg: Typo you inherited from me actually: s/One/Once

GUIDE.md
wolftune: actually, there's just an extra still. "moving" should be gone. will fix

GUIDE.md
wolftune: Okay, I don't think condescending is the right term, but I'll just adjust it.

snoyberg:
Looks great, thanks! I've looked this over and provided some inline comments. Just ping me when those are addressed and we can merge.

GUIDE.md
wolftune: Yeah, I had *no* clue what "One Travis" was, so I just left it alone :P will fix

snoyberg:
Awesome, thanks again!

PR #948, opened 2015-09-08
Add stack path --global-pkg-db / --ghc-package-path

mgsloan:
It can be handy to know which global db stack is using.  In order to add some hacks for xmonad recompilation, I needed to hardcode a global db path and compute my own GHC_PACKAGE_PATH (https://github.com/mgsloan/compconfig/blob/master/xmonad.hs#L505).  This patch also adds `stack path --ghc-package-path`.

One downside of this is that it adds an additional invocation of `ghc-pkg`.  Benchmarked with `time`, this seems to have a very negligible affect on speed of the `stack path` command.

### DIFF
--- a/src/Stack/GhcPkg.hs
+++ b/src/Stack/GhcPkg.hs
@@ -21,7 +21,8 @@ module Stack.GhcPkg
   ,findGhcPkgDepends
   ,findTransitiveGhcPkgDepends
   ,listGhcPkgDbs
-  ,ghcPkgExeName)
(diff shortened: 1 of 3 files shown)


snoyberg:
LGTM. Can you update the ChangeLog?

mgsloan:
Yup, updated: https://github.com/commercialhaskell/stack/commit/439b2359b3e1e07a9059e1d316201ff2fcf26c14

PR #954, opened 2015-09-08
Use binary-tagged, remove WithTag

### LINKED ISSUE #889: Replace Binary with something more reliable for file storage

phadej:
I tried to use the old stack and one with this patch one after each other, both pruned other's caches.

Fixes #889


snoyberg:
That was a lot faster than I was expecting, thanks! This looks good to merge, just waiting for Travis. Could you also add something to the ChangeLog about this?

phadej:
Indeed, it was quite straightforward change.

phadej:
And for the record: `binary-tagged` will catch changes like https://github.com/commercialhaskell/stack/pull/902/files#diff-8c2944535368646d36cd90e08a658087R74 so one doesn't need to manually bump `SemanticVersion`.

snoyberg:
Thanks!

PR #955, opened 2015-09-08
Resolve #953. Use deepseq own generic derivation

### LINKED ISSUE #953: Drop dependency to deepseq-generics
`deepseq` itself supports generic derivation from `1.3` (lts-2.17 has it).

There seems to be both both used across codebase. That is confusing.

I can make the PR if this change is ok.

phadej:


snoyberg:
Thanks!

Issue #928, opened 2015-09-06
Create a "for new contributors" label?


bollu:
I'm pretty new to Haskell in general. I know a decent amount of the language and how it works, but I've never written anything "real world" with it.

I was hoping to find a few issues on stack to snipe off, so I could get into haskell properly. But, there doesn't seem to be issues labeled as "for new contributors" or some parallel of that. That's somewhat of a bummer.

snoyberg:
We have both `newcomer` and `awaiting-pr` labels. They'd both be somewhat relevant for this I think.

bollu:
Oh, that's pretty cool! I didn't notice that. 

Um, shall I create a PR to highlight that on the `readme.md` page? Create a section for newbies who are interested to contribute?

Also, since we're talking about this, do you have any issues I could start with?

snoyberg:
Sure, README updates sound good, thanks!

Some thoughts on first issues: #840 #593 #508 #620 #792 

Welcome to the team, let me know if you have any questions

bollu:
I don't see a section on where to build `stack` from source. Could you point me to that?

snoyberg:
It's just a `stack build` (dogfooding to the max). Clone the repo, `stack setup` if necessary, and `stack build` should do it.

bollu:
Would a pre-existing install with `cabal` interact with the from-source build weirdly?

snoyberg:
No

PR #962, opened 2015-09-09
canonicalize environment paths before executing

### LINKED ISSUE #840: PATH check should canonicalize paths

bollu:
fixes #840 

As discussed in #928, @snoyberg pointed me to this issue.


bollu:
@snoyberg - is this redundant now, after #932?

snoyberg:
Yes it is, sorry for the confusion. I've closed the relevant issue.

PR #987, opened 2015-09-13
Add filesystem polling via --file-watch-poll flag

tedkornish:
When developing in a Docker container with code injected using a volume, `stack test --file-watch` run from inside the container won't register changes to the code made in the host environment because fsnotify events don't propagate down into the Docker container. This patch gives the option to poll the filesystem instead of relying on events, which is the less performant but more portable option. It's a pretty concise change to make since polling is already built into fsnotify - literally just flipping a bool.

I've sanity checked it against the example above (tests running in Docker container, code being edited outside the container) and it works like a charm. Let me know what you think!

### DIFF
--- a/src/Stack/Types/Build.hs
+++ b/src/Stack/Types/Build.hs
@@ -25,6 +25,7 @@ module Stack.Types.Build
     ,Plan(..)
     ,TestOpts(..)
     ,BenchmarkOpts(..)
+    ,FileWatchOpts(..)
     ,BuildOpts(..)
     ,BuildSubset(..)
     ,defaultBuildOpts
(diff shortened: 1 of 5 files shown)


src/Stack/Types/Build.hs
```diff
@@ -448,6 +449,7 @@ defaultBuildOpts = BuildOpts
     , boptsPreFetch = False
     , boptsBuildSubset = BSAll
     , boptsFileWatch = False
+    , boptsFileWatchPoll = False
```
snoyberg: It's a bit unclear what the semantics are when both `boptsFileWatch` and `boptsFileWatchPoll` are set. I'd probably restructure this a bit as:

```haskell
data FileWatch
  = NoFileWatch
  | FileWatch
  | FileWatchPoll

boptsFileWatch :: FileWatch
```

Then when parsing command line arguments, you can use the `Alternative` instance and get something like:

```haskell
flag' FileWatch (long "file-watch" <> ...) <|> flag' FileWatchPoll (long "file-watch-poll" <> ...)
  <|> pure NoFileWatch
```

tedkornish: Great call! Will implement and push tomorrow.

snoyberg:
Looks good, I just had a minor comment on the parsing of the command line arguments. Let me know if you have any questions on how to implement that.

And nice job updating the ChangeLog, it's often overlooked ;)

tedkornish:
I implemented your suggestion - let me know what you think. Nifty trick with `class Alternative` and `<|>` for flag parsing!

Haha, I actually missed the ChangeLog the first time through, but I was looking through past pull requests before I submitted mine and noticed a trend. ;)

snoyberg:
Thanks!

PR #990, opened 2015-09-14
Extra package database support

vigoo:
This pull-request adds support for creating and using reusable _extra package database_ with stack.

[This test repository](https://github.com/vigoo/stack-sandbox-chain-test) demonstrates how it would work:
- `stack build` in `lib1` builds the library into the local package database and nothing else
- `stack build` in `lib2` builds its dependencies (`ansi-wl-pprint` and everything it needs) **and** `lib2` itself to its local package database.
- `stack build` in `app` builds the executable using both `lib1` and `lib2` from the previously built package databases.

The change adds two new options to the stack yaml:
- `extra-package-dbs` is a list of relative or absolute paths to extra package databases, which will be added on top of the _global_ package database, before snapshot and local
- With `force-install-local` we still use the snapshots to determine the package versions to be installed, but they are _not_ installed to the shared snapshot package database but always to the current local package database. This is required to make the built package databases redistributable.

**NOTE** Currently this does not work. But it worked before I rebased it, so I'm opening the PR to get some help in fixing it before we can merge. To reproduce the issue, build this branch, and run the test script from the above linked repository, giving it the built stack executable as its parameter.

### DIFF
--- a/src/Stack/Build/Installed.hs
+++ b/src/Stack/Build/Installed.hs
@@ -81,15 +81,15 @@ getInstalled menv opts sourceMap = do
 
     let loadDatabase' = loadDatabase menv opts mcache sourceMap
 
-    (installedLibs0, globalInstalled) <- loadDatabase' [] Nothing []
+    (installedLibs0, globalInstalled) <- loadDatabase' Nothing []
     (installedLibs1, _extraInstalled) <-
-      (snd <$> (foldM (\(prevDBs, lhs') pkgdb -> do
-        lhs'' <- loadDatabase' prevDBs (Just (ExtraGlobal, pkgdb)) (fst lhs')
-        return (prevDBs ++ [pkgdb], lhs'')) ([], (installedLibs0, globalInstalled)) extraDBPaths))
+      (foldM (\lhs' pkgdb -> do
+        lhs'' <- loadDatabase' (Just (ExtraGlobal, pkgdb)) (fst lhs')
+        return lhs'') ((installedLibs0, globalInstalled)) extraDBPaths)
     (installedLibs2, _snapInstalled) <-
-        loadDatabase' extraDBPaths (Just (InstalledTo Snap, snapDBPath)) installedLibs1
+        loadDatabase' (Just (InstalledTo Snap, snapDBPath)) installedLibs1
     (installedLibs3, localInstalled) <-
-        loadDatabase' extraDBPaths (Just (InstalledTo Local, localDBPath)) installedLibs2
+        loadDatabase' (Just (InstalledTo Local, localDBPath)) installedLibs2
     let installedLibs = M.fromList $ map lhPair installedLibs3
 
     case mcache of
@@ -133,13 +133,12 @@ loadDatabase :: (M env m, PackageInstallInfo pii)
              -> GetInstalledOpts
              -> Maybe InstalledCache -- ^ if Just, profiling or haddock is required
              -> Map PackageName pii -- ^ to determine which installed things we should include
-             -> [Path Abs Dir] -- ^ extra package databases this database depends on
              -> Maybe (InstalledPackageLocation, Path Abs Dir) -- ^ package database, Nothing for global
              -> [LoadHelper] -- ^ from parent databases
              -> m ([LoadHelper], [DumpPackage () ()])
-loadDatabase menv opts mcache sourceMap extraDBs mdb lhs0 = do
+loadDatabase menv opts mcache sourceMap mdb lhs0 = do
     wc <- getWhichCompiler
-    (lhs1, dps) <- ghcPkgDump menv wc (extraDBs ++ (fmap snd (maybeToList mdb)))
+    (lhs1, dps) <- ghcPkgDump menv wc (fmap snd (maybeToList mdb))
                 $ conduitDumpPackage =$ sink
 
     let lhs = pruneDeps
--- a/src/Stack/Build/Installed.hs
+++ b/src/Stack/Build/Installed.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell       #-}
 -- Determine which packages are already installed
 module Stack.Build.Installed
@@ -69,6 +70,7 @@ getInstalled :: (M env m, PackageInstallInfo pii)
 getInstalled menv opts sourceMap = do
     snapDBPath <- packageDatabaseDeps
     localDBPath <- packageDatabaseLocal
+    extraDBPaths <- packageDatabaseExtra
 
     bconfig <- asks getBuildConfig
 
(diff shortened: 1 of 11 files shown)


borsboom:
Thanks for the PR.  Just to make sure I'm not missing any background, is there an issue where this idea was discussed?  I have some misgivings about the feature, because it breaks one of the fundamental purposes of `stack`: reproducible builds.  Since `stack` doesn't control what's in the extra package DBs, it means the build result depends on what happens to be on the developer's system already.  Perhaps it would help if you can describe the problem that this is solving.  @snoyberg, I'd value your opinion as well.

snoyberg:
@borsboom Check out this email thread:

https://groups.google.com/d/msg/haskell-stack/bCzgxqcbKKA/D5DuMSCTFwAJ

The idea here is that we're not making the situation any worse than we already have it from allowing packages in the global database, but we additionally gain the advantage of allowing easy sharing of binary package databases between systems.

snoyberg:
When running this on my system, I see the following package databases:

```
--package-db=clear
--package-db=global
--package-db=/home/vagrant/Desktop/stack-sandbox-chain-test/lib1/.stack-work/install/x86_64-linux/lts-2.22/7.8.4/pkgdb/
--package-db=/home/vagrant/.stack/snapshots/x86_64-linux/lts-2.22/7.8.4/pkgdb/
--package-db=/home/vagrant/Desktop/stack-sandbox-chain-test/lib2/.stack-work/install/x86_64-linux/lts-2.22/7.8.4/pkgdb/
```

At the very least, the order of these databases is incorrect; the snapshot database ended up in the middle of the lib1 and lib2 databases, and the local database does not appear anywhere in that list. It should be ordered:

```
--package-db=clear
--package-db=global
--package-db=additional-extra-package-databases
--package-db=snapshot
--package-db=local
```

vigoo:
I think in the above case the script is building lib2 which means that
lib2's local is the local, so the order is correct. Isn't it?

vigoo:
So there are 3 stack runs, and I think the correct package-db order for each is:

* **lib1**: `global -> snapshot -> local(lib1)`
* **lib2**: `global -> local(lib1) -> snapshot -> local(lib2)`
* **app**: `global -> local(lib1) -> local(lib2) -> snapshot -> local(app)`

snoyberg:
The problem is that you're reinstalling global packages - which are depended on by the Cabal library - in the local database. This means that (at least for GHC 7.8, which your example is using) the local version shadows the global version, and the dependencies of Cabal are unavailable. Concretely, I'm seeing on my system that Cabal depends on bytestring-0.10.4.0-d6f1d17d717e8652498cab8269a0acd5, but bytestring-0.10.4.0-9a9bbd95e0709e7cafd40d5934840c75 is in the local database.

vigoo:
I saw that too but wasn't sure it's the real problem. I'm a but confused here. Why does the `MiniBuildPlan` coming from the snapshot resolver (here: https://github.com/commercialhaskell/stack/blob/master/src/Stack/Build/Source.hs#L179) contain these global packages at all? Is there an easy way to filter them out?

snoyberg:
Different GHC installations may place different packages in the global
database, so we never hardcode which packages should be considered global.
Instead, the installed list tells us which database packages are installed
in, and we use that to determine which packages are already installed and
therefore don't need to be reinstalled.

It seems like this PR is attempting to do more than we originally
discussed, specifically change the behavior of where packages are
installed. I'm not too thrilled with the implicit nature of this, though it
could be that I'm misunderstanding how it works.

vigoo:
I did not want to make it implicit, that's why added a separate option for it (`force-install-local`).
Without this my solution only works if all the dependent packages are added as extra dependencies and no snapshot is used. This is a limitation I can work with, but while I was adding the extra package database support it seemed relatively easy to keep the advantage of having the snapshot by forcing the installed packages to be installed to the local database. 

If you think it's too complicated and/or against the way stack should work, I can remove the `force-install-local` parts from the PR. The disadvantage is then that I have to maintain in all my `stack.yaml`s a consistent set of dependent libraries with explicitly specified versions. I could generate these from higher level (the gradle plugin in this case), by fetching the snapshot config from stackage to get the versions. What do you think?

snoyberg:
I wasn't aware that `force-install-local` was added and a separate feature. That's probably fine.

To be honest though, I'm still not convinced that this is the best way to achieve your goal of cached binary builds, but I have no objection to having the multiple package database support in stack.

vigoo:
I tried out what I described above, and it works. https://github.com/vigoo/stack-sandbox-chain-test/commit/667cb75936dcd5998830bd80188d7b48a79e7619
I think the simplest and cleanest is to go this way, so I'll clean up this pull request soon and make sure the tests are passing.

vigoo:
I rebased this again; is there anything else I can do to make this merged?

borsboom:
I'm OK with this, I think it'll probably come in handy even though I still think the example use case is a bad idea.  @snoyberg, can you make the final determination and merge if you agree?

src/Stack/Build/Installed.hs
```diff
@@ -126,13 +133,15 @@ loadDatabase :: (M env m, PackageInstallInfo pii)
...
+             -> [Path Abs Dir] -- ^ extra package databases this database depends on
+             -> Maybe (InstalledPackageLocation, Path Abs Dir) -- ^ package database, Nothing for global
              -> [LoadHelper] -- ^ from parent databases
              -> m ([LoadHelper], [DumpPackage () ()])
-loadDatabase menv opts mcache sourceMap mdb lhs0 = do
+loadDatabase menv opts mcache sourceMap extraDBs mdb lhs0 = do
     wc <- getWhichCompiler
-    (lhs1, dps) <- ghcPkgDump menv wc (fmap snd mdb)
-                  $ conduitDumpPackage =$ sink
+    (lhs1, dps) <- ghcPkgDump menv wc (extraDBs ++ (fmap snd (maybeToList mdb)))
```
snoyberg: It shouldn't be necessary to pass in the extraDBs I think, the packages in those databases should be inherited automatically from the passed in `lhs0`. Do things break without them?

src/Stack/Build/Source.hs
```diff
@@ -183,16 +182,18 @@ parseTargetsFromBuildOpts needTargets bopts = do
...
             needTargets
             (bcImplicitGlobal bconfig)
             snapshot
-            (flagExtraDeps <> bcExtraDeps bconfig)
+            extraDeps'
             (fst <$> rawLocals)
             workingDir
             (boptsTargets bopts)
-    return (mbp0, cliExtraDeps <> flagExtraDeps, targets)
+    return (mbp0, extraDeps', targets)
```
snoyberg: I'm not certain why this change would be necessary for this feature, and it seems like it's likely to introduce a regression.

snoyberg:
Mostly looks OK, I added some inline comments.

src/Stack/Build/Source.hs
vigoo: You are right, I'm reverting it.

src/Stack/Build/Installed.hs
vigoo: If you don't pass the extra databases to `ghc-pkg` it will report the packages depending on any of them in the actual database as broken

snoyberg: Stack itself does the dependencies check and ignores ghc-pkg's opinion on the matter. You can see that we currently do the databases one at a time.

vigoo: Ok, I'm checking what happens if I'm not passing them

vigoo: Seems to work without it, I'll test a bit more and clean it up soon

vigoo:
Commented issues fixed

snoyberg:
Thanks!

PR #1000, opened 2015-09-16
docs900b2 branch

kairuku:
This branch begins some of the work in #900


kairuku:
@borsboom I put some additional ideas into this first PR; I copied over everything from the Wiki and have mostly refrained from file editing/combining/deletion for this initial stage, but have some other limited changes beyond the Wiki.

borsboom:
Overall, looks good.  A few thoughts:

* In general, I'm not sure about moving the documents currently in the root do `doc/`, but I don't feel really strongly about it.  In particular:

    * Consider leaving `CONTRIBUTING.md` in the root, because that's where Github looks for it to put a notice for users submitting issues/PRs (see https://github.com/blog/1184-contributing-guidelines).  I actually do think `doc` is the better place, but there should be as few barriers as possible in front of people seeing it.  Alternatively, leave the "pointer" document in root indefinitely and make the text more like "please see doc/contributing-guide" rather than "this document has moved".

    * Maybe leave LICENSE in the root?  Not exactly sure why, just feels right to me.  If not, update `stack.cabal` to point to the new location.

    * If the ChangeLog goes in `doc/`, update `stack.cabal` to point to new location.  But I tend to expect the changelog to be in the root rather than the documentation.

* Maybe "USERS-GUIDE" would be a better name than "new-users-guide".  I'm far from a new user, but I still refer to the guide regularly.  And maybe upper-case the name for emphasis since it's probably the first thing people should look at after the README?

* Could we change the links in the README to relative links (e.g. `doc/architecture.md`).  That way, you'll get the documentation for the commit you're looking at.  If that's not possible, maybe we can point them to the `release` branch, which will always be the current release (could also have a small "(development version)" link next to them that points at master).

kairuku:
@borsboom @snoyberg I wondered if you'd mind checking the current state when you have a moment to see if it's merge-worthy

CONTRIBUTING.md
```diff
@@ -1,61 +1,2 @@
-## Bug Reports
```
snoyberg: As @borsboom mentioned, this file should not move, since it has special meaning to Github.

CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+Please see [doc/docs/project/CHANGELOG.md
```
snoyberg: I'm still not a fan of the deep hierarchy, I don't think we have so many files that it makes sense.

LICENSE
```diff
@@ -1,24 +0,0 @@
-Copyright (c) 2015, stack
```
snoyberg: LICENSE file in the root is a pretty well accepted practice in the Haskell package world, I'd rather not change that.

README.md
```diff
@@ -16,24 +16,7 @@ It features:
 * Testing your project.
 * Benchmarking your project.
 
-#### How to install
```
snoyberg: I don't think we should remove this content from the README, we still want the README to function as a quickstart guide without having to jump to other files.

CHANGELOG.md
kairuku: I was just thinking that if you did eventually host the docs somewhere like readthedocs.org, it would also be nice to have the rest of the documentation there too, but still have permanent pointers to the "hosted" docs in the repo root as well. I thought maybe this could satisfy all potential scenarios.

snoyberg: But why doesn't that work if we just put all of the markdown files directly in `doc/`?

kairuku: Well there used to be a man/ directory in the repo too (it has since disappeared though), and to me things like man pages are also documentation (and I don't know if you'd ever want to include haddocks as well), but I don't think those should necessarily be mixed in with anything you maybe potentially host on readthedocs,org. Man pages and haddocks just feel like a different category of documentation to me, but they are still documentation of course. At least if all categories of documentation are under doc/ (even with subdirectories), rather than scattered about, in my mind it made it easier to find any possible category of docs without having to crawl the directories too much.

snoyberg: Maybe we're talking past each other. What I'm saying is: +1 to have documentation in the doc/ directory, -1 on having multiple directories inside doc/. When we get to the point of having more than 30 or so files in the doc/ directory, we can start to talk about subdirectories, but it's not relevant yet.

snoyberg: And I appreciate the work you're putting into this, thanks!

kairuku:
@borsboom @snoyberg I've scaled back the PR some. It now just basically imports the Wiki repo, renames those files, deletes Home.md, appends a TOC to README, updates links in README, and adds two additional resources to doc/GUIDE.md (the tutorial link from the Wiki Home page, and an fpco blog post about stack). Wiki references in doc/GUIDE.md haven't been updated yet though.

snoyberg:
Thanks!

kairuku:
No problem, sorry I couldn't get back to it sooner

snoyberg:
Just a heads-up on this, I removed a few of the files from the repo that I don't consider authoritative enough, and am now updating the Wiki to have links to the new locations for files.

PR #1003, opened 2015-09-17
Fix #997 (pass --ghc-options -hpcdir) only on --coverage

### LINKED ISSUE #997: Allow "--ghc-options -hpcdir" to be selectively disabled via CLI
Hello folks,

first of all I'm not expecting this to be considered a stack issue, as I'm not sure it is. I'm writing this down as this project is highly maintained and having this kind of reports might help others in the future. I have ported a `snap` project I have at work to use `stack`. So far so good, pretty much smooth sailing. If you are not familiar with `snap` it supports automatic recompilation leveraging [hint](https://hackage.haskell.org/package/hint). I have therefore enabled the "dev" mode calling:

```
stack install --flag chronos:development
```

Which built just fine. Upon startup though, I realised that `hint` was trying to look for packages in my global package index and, unable to find my deps, was failing. This is similar to an issue someone reported here:

http://hub.darcs.net/jcpetruzza/hint/issue/5

Dutifully, I tried something like this:

```
GHC_PACKAGE_PATH=$(stack path --local-pkg-db) chronos
```

Which generated:

```
o port specified, defaulting to port 8000
Listening on http://0.0.0.0:8000/
chronos: Unknown interpreter error:

flags: '-hpcdir .stack-work/dist/x86_64-osx/Cabal-1.18.1.5/hpc/.hpc/' not recognized
```

And this is because - I suspect - stack is passing such flag dowstream and Hint's interpreter code is choking on it. Is Hint right here?

The project seems quite abandoned (but I might be totally wrong!) and unless someone knows better than me Snap's plan for 1.0 release, they seems to still leverage hint for automatic recompilation. This means this problem will hinder Stack's adoption for people using Snap.

Of course if I have missed something completely trivial and/or Snap is ditching hint in 1.0 I guess this issue is moot, but I thought would have been useful to share anyway!

Alfredo

adinapoli:
This patch extends `extraBuildOptions` in `Execute.hs` to accept
an extra argument of type `BuildOpts`, as well as checking if coverage
is enabled. In case it is, `-hpcdir` is passed as `--ghc-option`.

This is needed to avoid certain in-memory REPL (like
[hint](https://hackage.haskell.org/package/hint)) to choke
on that command line option.

This was the result of a compilation with latest `stack`:

screen shot 2015-09-17 at 12 45 00

And this is compiling with my patched stack:

```
☁  chronos [issue-489] ⚡ /Users/adinapoli/programming/haskell/stack/.stack-work/dist/x86_64-osx/Cabal-1.22.4.0/build/Stack/stack clean
☁  chronos [issue-489] ⚡ /Users/adinapoli/programming/haskell/stack/.stack-work/dist/x86_64-osx/Cabal-1.22.4.0/build/Stack/stack install --flag chronos:development
```

screen shot 2015-09-17 at 14 12 19

A couple of things I have noticed (which I think are orthogonal to my patch):

* Every time I tried to compile my project switching stack version (between my binary and the default one) a complete rebuild (with all the deps!) was triggered. Is this by design? Is quite inconvenient with massive stack (no pun intended). At the moment my project has 143 deps, and counting.

* To "toggle" between the behaviour introduced by my patch and the official one I had to do a `stack clean`. How so?

* Enabling `--coverage` **both on official and my binary**, the install step was failing at the final phase with:

```
Linking .stack-work/dist/x86_64-osx/Cabal-1.18.1.5/build/chronos-ctl/chronos-ctl ...
chronos-0.15.0.0: install
Installing executable(s) in
/Users/adinapoli/work/chronos/.stack-work/install/x86_64-osx/lts-2.16/7.8.4/bin
/Users/adinapoli/work/chronos/.stack-work/install/x86_64-osx/lts-2.16/7.8.4/hpc/: getDirectoryContents: does not exist (No such file or directory)
```

Last thing I want is to cause a regression, so does someone have a clue on why is this happening? Can you think of any case this can be introduced by this patch?

Apart from that, should be all good.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -927,7 +927,7 @@ singleBuild runInBase ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} in
 
         () <- announce "build"
         config <- asks getConfig
-        extraOpts <- extraBuildOptions
+        extraOpts <- extraBuildOptions eeBuildOpts
         cabal (console && configHideTHLoading config) $
             (case taskType of
                 TTLocal lp -> concat
@@ -1043,7 +1043,7 @@ singleTest runInBase topts lptb ac ee task installedMap = do
             case taskType task of
                 TTLocal lp -> writeBuildCache pkgDir $ lpNewBuildCache lp
                 TTUpstream _ _ -> assert False $ return ()
-            extraOpts <- extraBuildOptions
+            extraOpts <- extraBuildOptions (eeBuildOpts ee)
             cabal (console && configHideTHLoading config) $
                 "build" : (components ++ extraOpts)
             setTestBuilt pkgDir
@@ -1187,7 +1187,7 @@ singleBench runInBase beopts _lptb ac ee task installedMap = do
                 TTLocal lp -> writeBuildCache pkgDir $ lpNewBuildCache lp
                 TTUpstream _ _ -> assert False $ return ()
             config <- asks getConfig
-            extraOpts <- extraBuildOptions
+            extraOpts <- extraBuildOptions (eeBuildOpts ee)
             cabal (console && configHideTHLoading config) ("build" : extraOpts)
             setBenchBuilt pkgDir
         let args = maybe []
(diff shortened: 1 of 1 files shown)


snoyberg:
> Every time I tried to compile my project switching stack version (between my binary and the default one) a complete rebuild (with all the deps!) was triggered. Is this by design? Is quite inconvenient with massive stack (no pun intended). At the moment my project has 143 deps, and counting.

If they're local packages, I'm not surprised. We have a change on master for how binary files are saved, so in fact stack master and stack 0.1.4.0 store their metadata in different formats.

> To "toggle" between the behaviour introduced by my patch and the official one I had to do a stack clean. How so?

I'm not sure exactly what you mean, but my guess is that `stack clean` forced a reconfigure. You could also try the new `--reconfigure` flag.

I'm not sure about that last point, so I'll defer to @mgsloan on it. @mgsloan: this patch is good to go IMO, assuming you don't see any way it's introduced a regression (I don't see any).

adinapoli:
Thanks for the explanations Michael, and good to hear about `--reconfigure`. Will wait for the :+1: from the other Michael then!

mgsloan:
Looks good to me!

I don't think that this will cause any regressions, so merging.

mgsloan:
I've opened up this issue to track resolving the issue with install https://github.com/commercialhaskell/stack/issues/1006 .

Issue #1017, opened 2015-09-20
InvalidAbsDir "./stack-setupXXXXX" from Path.IO.listDirectory


robstewart57:
When I execute `stack setup` I'm getting this error:

    $ stack setup
    InvalidAbsDir "./stack-setup31658"

This is thrown by Path.parseAbsDir in the path package (see [line 79](https://github.com/chrisdone/path/blob/09996f160573edba61166293c27438ea1d342050/src/Path.hs#L79)). I've modified Path.IO.listDirectory to print the filepath it is calling `parseAbsDir` on, so that the file path is printed each time and also so that I can get the stack trace up to that point.

This is the printed file paths and the stack trace:

```
"/home/rob/.stack/programs/i386-linux/ghc-7.10.2.tar.xz"
"/home/rob/.stack/programs/i386-linux/.."
*** Exception (reporting due to +RTS -xc): (THUNK_1_0), stack trace:
  Path.CAF
  --> evaluated by: Path.IO.listDirectory.\,
  called from Path.IO.listDirectory,
  called from Stack.Setup.listInstalled,
  called from Stack.Setup.ensureGHC,
  called from Stack.Types.StackT.>>=,
  called from Main.setupCmd.\,
  called from Main.withUserFileLock.\,
  called from Main.withUserFileLock,
  called from Main.setupCmd,
  called from Main.main.\,
  called from Data.Attoparsec.Args.withInterpreterArgs,
  called from Main.main
```

I've raised this question with @chrisdone who believes that stack should canonicalize paths before calling  `parseAbsDir`.

mgsloan:
configLocalPrograms seems to be well canonicalized, [here](https://github.com/commercialhaskell/stack/blob/c774f9905b3b112f58fc3789dfb9c26389d33b24/src/Stack/Config.hs#L167), so it must not be the input to `listDirectory` that's the issue.

It's very surprising that there'd be a bug in `listDirectory`, though (you'd think it'd always happen, and be caught quickly).

Lets test out chris's thought - does this fix it?

```diff
diff --git a/src/Path/IO.hs b/src/Path/IO.hs
index ca1bed9..79e6d39 100644
--- a/src/Path/IO.hs
+++ b/src/Path/IO.hs
@@ -170,8 +170,9 @@ listDirectory dir =
   do entriesFP <- liftIO (D.getDirectoryContents dirFP)
      maybeEntries <-
        forM (map (dirFP ++) entriesFP)
-            (\entryFP ->
[...]

PR #1019, opened 2015-09-20
Canonicalises path from withSystemTempDirectory

robstewart57:
… Stack.Setup

The System.IO.Temp withSystemTempDirectory function from
System.IO.Temp used in Stack.Setup returns a non-canonicalised
path. However, the parseAbsDir function in the Path module of the path
library expects canonicalised paths.

The change is applied to stack setup, `sanityCheck` and stack upgrade
in Stack.Setup .

Fixes commercialhaskell/stack#1017

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -287,7 +287,7 @@ withExecuteEnv :: M env m
                -> m a
 withExecuteEnv menv bopts baseConfigOpts locals globals sourceMap inner = do
     withSystemTempDirectory stackProgName $ \tmpdir -> do
-        tmpdir' <- parseAbsDir tmpdir
+        tmpdir' <- parseAbsDir =<< liftIO (D.canonicalizePath tmpdir)
         configLock <- newMVar ()
         installLock <- newMVar ()
         idMap <- liftIO $ newTVarIO Map.empty
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -439,7 +439,7 @@ upgradeCabal menv wc = do
                 , " to replace "
                 , T.pack $ versionString installed
                 ]
-            tmpdir' <- parseAbsDir tmpdir
+            tmpdir' <- parseAbsDir =<< liftIO (D.canonicalizePath tmpdir)
             let ident = PackageIdentifier name newest
             m <- unpackPackageIdents menv tmpdir' Nothing (Set.singleton ident)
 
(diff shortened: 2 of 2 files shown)


robstewart57:
Hang off accepting this PR actually, this same problem is causing a problem elsewhere. More anon...

snoyberg:
Looks good. One thought: perhaps instead of modifying each place in the code, we could add a `withSystemTempDir` function to `Path.IO` that would handle the canonicalizing and `parseAbsDir` call, to make sure that future temp directory usages don't hit this.

By any change, do you have a temporary directory environment variable set to a relative path?

robstewart57:
@snoyberg thanks for the review, and I agree. Rather than moving a version of `withSystemTempDirectory` to the path library, I've submitted a pull request to the temporary library.

https://github.com/batterseapower/temporary/pull/16

If this gets accepted to temporary, I have a corresponding stack commit that uses the `withCanonicalisedSystemTempDirectory` function throughout, in place of `withSystemTempDirectory`. And I will submit another stack PR with this commit if or when the 6db0f9952d66d74363cb127d00b7586aefe3c53e temporary commit gets reflected on hackage.

So I'm closing this PR.

snoyberg:
Actually, Path.IO is in stack itself, so it could just be added here. I'd still recommend going that route, as it will be easier to deal with dependencies that way.

robstewart57:
@snoyberg ah ha! I've got to the bottom of this bug. It's a really interesting corner case.

The `withSystemTempDirectory` is used in a number of places throughout stack. It's [implementation](https://github.com/batterseapower/temporary/blob/master/System/IO/Temp.hs#L35) in the temporary library uses `getTemporaryDirectory` from System.Directory. The [documentation](https://hackage.haskell.org/package/directory-1.2.3.1/docs/System-Directory.html#v:getTemporaryDirectory) says:

> On Unix, getTemporaryDirectory returns the value of the TMPDIR environment variable or "/tmp" if the variable isn't defined.

I've looked into my `~/.bash_profile` and sure enough,

    # for bibtex2html
    export TMPDIR=.

The bib2texhtml [documentation](https://www.lri.fr/~filliatr/bibtex2html/) says to do this:

> Note for users using a version of bibtex2html prior to 1.98: TeXlive 2010 prevents bibtex2html to run bibtex in a temporary directory, resulting in an error message such as .... A workaround consists in telling bibtex2html to use the current directory for temporary files, using the following shell command before running bibtex2html:
>
>  export TMPDIR=.

The path starting "." is eventually passed back to stack in all places where `withSystemTempDirectory` is used in stack. The stack code base always assumes a canonicalised path is provided from this function, though as this `TMPDIR=.` corner case demonstrates that this is not necessarily the case.

@snoyberg What do think is the best resolution?

1. Ignore this corner case, the number of users who've specified the `TMPDIR` environment variable, for bibtex2html or otherwise, is tiny.

2. Canonicalise the path received from `withSystemTempDirectory`, as per my first PR.

3. Copy the temporary library functions to Path.IO in stack, with each modified to ensure that all with* functions return canonicalised paths.

snoyberg:
I'd still say 3 is the right approach. Just to be clear though, I don't
anticipate any code being copied from the temporary package, those library
functions can just be wrapped.

robstewart57:
Done https://github.com/commercialhaskell/stack/pull/1047 .

PR #1023, opened 2015-09-21
Set default FileWatch command to force a rebuild

tedkornish:
Just something which could be kind of nice.

### DIFF
--- a/src/Stack/FileWatch.hs
+++ b/src/Stack/FileWatch.hs
@@ -110,6 +110,7 @@ fileWatchConf cfg getProjectRoot inner = withManagerConf cfg $ \manager -> do
                     "watched" -> do
                         watch <- readTVarIO watchVar
                         mapM_ (putStrLn . toFilePath) (Map.keys watch)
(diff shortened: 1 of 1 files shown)


snoyberg:
I'm OK with this, but I think it should still let the user know about the `help` command for discoverability. (Yes, it mentions help at the beginning, but that's easy to miss/forget.) Maybe just add it to the "waiting for next build" line?

tedkornish:
^ disregard: that's a mistake and not what you're looking for (fc2d25b).

> I think it should still let the user know about the help command for discoverability
> Maybe just add it to the "waiting for next build" line?

Could you clarify what you mean by "it"? As far as I'm aware, every instance of telling the user about the `help` command is still there.

snoyberg:
I mean that pressing enter by itself today gives the user some guidance about what's going on. I can picture someone not understanding how to exit and just hitting enter, for example.

snoyberg:
And you can ignore the AppVeyor build failures, I'm just getting that set up now.

PR #1027, opened 2015-09-22
Use partial SHA1 instead of full platform name and Cabal version on Windows

anton-dessiatov:
I've had problems using Stack with our project on Windows. The issue was that when installing dependencies (in particular, aeson-pretty-0.7.2), Stack failed to compile Paths_aeson_pretty module because full path to intermediate files exceeded Windows 260 characters limit.

This pull request makes build artifacts path shorter by replacing "platform/cabal" part of it with first 8 characters of Base16-encoded SHA1 of it. In my case, this helped to get aeson-pretty succesfully installed.

### DIFF
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE OverloadedStrings #-}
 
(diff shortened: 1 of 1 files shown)


snoyberg:
This has two downsides:

* Paths on Windows will be less user friendly
* Existing built libraries will be affected (likely unusable)

This is still probably a good move, but would you mind emailing the Stack
mailing list for feedback first?

snoyberg:
This is becoming enough of a problem on Windows that I've gone ahead and merged without discussion. The testing we get before the next release will be sufficient.

PR #1047, opened 2015-09-24
Canonicalizes temporary directory paths

robstewart57:
When the $TMPDIR environment variable is set, the directory paths
provided by `withSystemTempDirectory` and `withTempDirectory` from
System.IO.Temp provided by the temporary library are not
canonicalised. This commit wraps these functions into canonicalized
versions.

See an earlier PR for discussion commercialhaskell/stack#1019

Fixes commercialhaskell/stack#1017

### DIFF
--- a/src/Path/IO.hs
+++ b/src/Path/IO.hs
@@ -33,7 +33,9 @@ module Path.IO
   ,createTree
   ,dropRoot
   ,parseCollapsedAbsFile
-  ,parseCollapsedAbsDir)
+  ,parseCollapsedAbsDir
+  ,withCanonicalizedSystemTempDirectory
+  ,withCanonicalizedTempDirectory)
   where
 
 import           Control.Exception hiding (catch)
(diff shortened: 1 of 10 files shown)


src/Path/IO.hs
```diff
@@ -289,3 +292,18 @@ dropRoot (Path l) = Path (FP.dropDrive l)
 ignoreDoesNotExist :: MonadIO m => IO () -> m ()
 ignoreDoesNotExist f =
     liftIO $ catch f $ \e -> unless (isDoesNotExistError e) (throwIO e)
+
+withCanonicalizedSystemTempDirectory :: (MonadMask m, MonadIO m)
+    => String            -- ^ Directory name template.
+    -> (FilePath -> m a) -- ^ Callback that can use the canonicalized directory
```
snoyberg: I'd actually been thinking this would be a `Path Abs Dir`, to make it fit in with the rest of this module and avoid the need for the `parseAbsDir` in all of the call sites.

Issue #115, opened 2015-05-29
Add commands for automating stack.yaml management


snoyberg:
Originally discussed in #27. Exact requirements here are uncertain, some ideas:
- Upgrade to a new snapshot
- Prune the extra-deps
- Add an extra package to the list of packages (and remove one?)
- Use a dependency solver to automatically add necessary extra-deps

chreekat:
+1 for easy way to upgrade snapshot (and probably extra-deps at the same time). Looks like the current way to do it is 
* move stack.yaml
* (re)run stack build/init (if/when init exists) — which should not fare very well
* see if any extra-deps have made it into the new snapshot
* manually check the extra-deps to see if any newer versions are available
* merge the old stack.yaml into the new one

Lots of hairballs there.

rvion:
+1, I find myself doing naively the same as @chreekat for now

snoyberg:
Following #390: running `stack config set resolver nightly` should set to the newest nightly. Other similar kinds of shortcuts should be allowed too (e.g., lts-2 for newest LTS 2.x).

alanz:
Would `stack config set resolver nightly` update the per-project config or the global (on a given machine) config?

Because if the latter, it makes it simple to move a given dev environment as a whole onto a different snapshot/nightly.

snoyberg:
I think it would do both: `stack config` should work the same way as `stack exec`/`stack ghci`, using a local project config if available, otherwise falling back to the default global (aka `ExecStrategy` in the code base). Thanks for pointing out this requirement.

ekmett:
:+1: I just found myself in the situation of needing some form of `stack config set resolver nightly` that doesn't involve a hack that goes and retrieves the current version from the json document and found my way to this ticket.

drwebb:
`stack config set resolver <snapshot>` is currently in master. There is another branch where I am adding `stack config add extra-dep <packagename>-packagever [--suggested]`. It would be great to have more generic add and sets, but it's easier to just add in special cases for the %80 of actual uses.

harendra-kumar:
[...]

PR #1048, opened 2015-09-25
--push flag for stack image container

snoyberg:
@dysinger can you look this one over?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,12 @@ Major changes:
 
 * "stack setup" now supports building and booting GHCJS from source tarball.
 
+Other enhancements:
+
+* `stack image container` now accepts a `--push` option
+
+Bug fixes:
+
 ## 0.1.5.0
 
 Major changes:
--- a/src/Stack/Image.hs
+++ b/src/Stack/Image.hs
@@ -56,10 +56,11 @@ stageContainerImageArtifacts = do
 -- extended with an ENTRYPOINT specified for each `entrypoint` listed
 -- in the config file.
 createContainerImageFromStage :: Assemble e m
-                              => m ()
-createContainerImageFromStage = do
+                              => Bool -- ^ should we follow up with a docker push?
+                              -> m ()
+createContainerImageFromStage push = do
     imageDir <- imageStagingDir <$> getWorkingDir
-    createDockerImage imageDir
+    createDockerImage push imageDir
     extendDockerImageWithEntrypoint imageDir
 
 -- | Stage all the Package executables in the usr/local/bin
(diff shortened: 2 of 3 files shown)


fizruk:
I am not sure it actually makes sense to have Docker commands replicated in Stack.
While `stack image container` is useful to build an image without having to specify custom `Dockerfile` (thus removing reasonable amount of boilerplate), managing Docker images is not Stack's job.
That is, I don't think Stack should `push`/`run`/`exec`/etc.
It won't remove any boilerplate and we can name the image in `stack.yaml` to then use `docker` to do something to it.

snoyberg:
I'm thinking about the case of automated tooling (like Travis CI, for instance) where you don't want to have to hard-code the name of the Docker tag in two places.

fizruk:
@snoyberg well, you obviously know better if it makes sense to have this in Stack :) nonetheless...

It would be somewhat convenient feature for a simple CI scenario.
But in some more complicated scenarios I would like also (from the top of my head) to automatically tag images (e.g. with Git tags), run and test container based on this image. Would these features also fit into `stack image container`? Or should they be left for `docker`?

snoyberg:
I'm not certain. I wasn't even certain of this PR, which is why it's a PR :)

I think the tagging would make sense, but I'd probably draw the line at
running/testing containers. That's just an arbitrary line in the sand that
makes sense to me though.

On Fri, Sep 25, 2015 at 11:54 AM, Nickolay Kudasov <notifications@github.com

dysinger:
@snoyberg I'm neutral if it should be in or out of stack.  I buy the argument that it's a simple command contained in a tool that's already installed. 

The P/R looks fine. You would need to have a very specific remote-friendly 'name' setting in stack.yaml for it to work.  Might want to error if there's no 'name' setting in stack.yaml because docker will just error underneath if you are trying to push "cool-app"

eg, the project dir might be "cool-app" but docker push of an image called "cool-app" would just error.  You would be required to have a name like "https://docker-registry.my.domain/myuser/myimage" for the push to "just work."

snoyberg:
Maybe we just need a `stack image container-name` command or something like that, which will print the image name to stdout. Then scripts could look like:

```
stack image container
$NAME=$(stack image container-name)
docker push $NAME
docker tag $NAME $NAME-$GITREV
```

And so on.

fizruk:
As an alternative `stack image container` might receive `name` through a command line argument like this:

```
$ stack image container --name $NAME
```

Then we just have to set `NAME` variable and use it for both `stack` and `docker`:

```
stack image container --name $NAME
docker push $NAME
docker tag $NAME $NAME-$GITREV
```

I think I would prefer this to special `container-name` command.

fizruk:
I have just thought that maybe a command printing an option from `stack.yaml` might be useful.
But instead of an arbitrary command for specific option (like `container-name`) we might introduce a command to get any (or almost any) option. It then can be used to get container name like this:

```
$ stack option image.container.name
```

I am however not sure what else people might use such `stack option` for.

fizruk:
I believe my `stack option` suggestion connects to `stack config` in #115.

snoyberg:
Taking the option on the command line defeats the idea of a general purpose tool that would be able to build/push any user's repository.

While in theory having `stack option` may be sufficient, it wouldn't work well for cases when Stack is figuring out the image name itself.

dysinger:
@snoyberg I didn't mean to kill the P/R.  I get the need for this.  CI needs to build with stack & then if it finishes, it needs to know the name of the image to push to a remote repo.  The name can be derived or over-ridden in the stack.yaml.  The CI kinda needs to know what's happened in order to do anything with `docker push`.  So we either need to provide output that's grok-able from the `image container` command or else be able to list the image names that it would produce.

snoyberg:
I closed because I think outputting the name would be a better approach

PR #1055, opened 2015-09-26
Fix #959: Inline haskell on the command-line

### LINKED ISSUE #959: Inline Haskell on the command-line with stack
Could `stack ghc -- -e "<haskell code>"` be abbreviated to something like `stack do "<haskell code>"`? I don't care so much about the particular name.

sumitsahrawat:
+ Generalize `execOptsParser` to accept a second optional argument for metavar.

+ Add a new command `eval`, where `stack eval CODE [ARGS]` evaluates to `stack ghc -- -e CODE [ARGS]`.

### DIFF
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -450,8 +450,9 @@ ghciOptsParser = GhciOpts
 
 -- | Parser for exec command
 execOptsParser :: Maybe String -- ^ command
+               -> Maybe String -- ^ metavar for arguments
                -> Parser ExecOpts
-execOptsParser mcmd =
+execOptsParser mcmd mmvr =
     ExecOpts
         <$> pure mcmd
         <*> eoArgsParser
@@ -463,9 +464,10 @@ execOptsParser mcmd =
     eoArgsParser :: Parser [String]
     eoArgsParser = many (strArgument (metavar meta))
       where
-        meta =
-            (maybe ("CMD ") (const "") mcmd) ++
-            "-- ARGS (e.g. stack ghc -- X.hs -o x)"
+        meta = case mmvr of
+                   Nothing -> (maybe ("CMD ") (const "") mcmd) ++
+                              "-- ARGS (e.g. stack ghc -- X.hs -o x)"
+                   Just x -> x
 
     eoEnvSettingsParser :: Parser EnvSettings
     eoEnvSettingsParser = EnvSettings
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -9,6 +9,7 @@ module Stack.Options
     ,dockerCleanupOptsParser
     ,dotOptsParser
     ,execOptsParser
+    ,evalOptsParser
     ,globalOptsParser
     ,initOptsParser
     ,newOptsParser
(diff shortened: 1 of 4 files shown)


src/main/Main.hs
```diff
@@ -848,6 +852,16 @@ execCmd ExecOpts {..} go@GlobalOpts{..} = do
                munlockFile lk -- Unlock before transferring control away.
                exec eoEnvSettings cmd args
 
+-- | Evaluate some haskell code inline.
+evalCmd :: ExecOpts -> GlobalOpts -> IO ()
+evalCmd e@ExecOpts { eoArgs = args } go@GlobalOpts {..} =
+    if length args /= 1
+    then error "Expecting ONE string argument to 'eval'"
```
borsboom: Would it be possible to tweak the options parser so that optparse-applicative reports the invalid command-line, rather than have this check here?

src/Stack/Options.hs
```diff
@@ -463,9 +464,10 @@ execOptsParser mcmd =
     eoArgsParser :: Parser [String]
     eoArgsParser = many (strArgument (metavar meta))
       where
-        meta =
-            (maybe ("CMD ") (const "") mcmd) ++
-            "-- ARGS (e.g. stack ghc -- X.hs -o x)"
+        meta = if isNothing mmvr
```
borsboom: Stylistically, it would be better to parren match is `mmvr`.  E.g.,

```
case mmvr of
    Nothing -> [...]
    Just mvr -> mvr
```

src/main/Main.hs
sumitsahrawat: I was trying to reuse the `execOptsParser`, which should not throw an error when given multiple arguments. 

I can create a separate parser for this, but it would have to duplicate some code from `execOptsParser`. As I expect `stack eval CODE` to behave exactly like `stack exec ghc -- -e CODE`, this is a fragile solution.

borsboom: Yeah, I get what you're saying.  Parsers are pretty composable but probably not worth the trouble in this case.

sumitsahrawat: Yup, I forgot to mention that I'm not very experienced with `optparse-applicative`. Thinking about it, there should be a way to parse and declare the parse failed if it doesn't match this condition.

Do you think such a solution is viable + possible + worth the trouble?

sumitsahrawat: I meant something similar to this, borrowing some intuition from parsec:

```haskell
parser :: Parser A

condition :: A -> Bool

verifyingParser :: Parser A
verifyingParser = do
  a <- parser
  if condition a
    then return a
    else fail "Condition failed"
```

sumitsahrawat:
@borsboom I've squashed the changes into one single commit. Thanks for the review :+1:.

src/main/Main.hs
borsboom: Ok, looked at the code a bit more and I don't think it'd be too hard to do this "right".  The options that would be common to `exec`/`runghc` and `eval` are already split out into a separate `ExecOptsExtra`, so you could reuse that data structure in a new `EvalArgs`.  You can factor the options parser for that out of `execOptsParser`.

sumitsahrawat:
@borsboom It seems like an issue with AppVeyor. Maybe restarting the build would set things right.

sumitsahrawat:
I cannot trigger the appveyor build by closing-and-opening.

borsboom:
Looks great.  Just add an entry to the ChangeLog.md and it'll be ready to merge.

sumitsahrawat:
Thanks :+1:.

PR #1077, opened 2015-09-28
Add documentation about the dataflow of the path command

NightRa:
Add documentation about the dataflow of the path command
Also, the name of the function "getMinimalEnvOverride" is misleading.
Should I change it?

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -336,7 +336,8 @@ ensureCompiler sopts = do
                 arch /= expectedArch
         isWanted = isWantedCompiler (soptsCompilerCheck sopts) (soptsWantedCompiler sopts)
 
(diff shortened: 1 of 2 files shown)


borsboom:
It's minimal in the sense that all none of the optional environment additions are enabled (it uses an `EnvSettings` with all fields `False`).  If you can think of a better name that still communicates that idea, feel free to change it.

NightRa:
The path environment variable is changed though..
I'll take another look into it soon

borsboom:
Yes, all the non-optional environment changes are still made.

borsboom:
Didn't hear back about changing the name of `getMinimalEnvOverride`, but the rest of this is good the way it is so merging.

NightRa:
:+1: I was just busy with other things.
If anything, I'll make another PR :)

PR #1078, opened 2015-09-28
Quickstart guide to stack proposition

yogsototh:
Hi! I saw the proposed new Download page for Haskell here: 

http://haskellorg.snoyman.com/downloads#choosing

My only remark was that the `stack` guide while excellent to learn stack might be a bit overwhelming for a real beginner that just want to go to the next stage. Being, write code and launch a REPL.

So here is a proposition to make an easier and straight to the point QUICKSTART.md document, which of course lack all the gory detail a beginner could find in the GUIDE (which is of course linked in this document).

Tell me your impression,
Yann.


borsboom:
I like it, thanks. I'm thinking it should link to the installation instructions, just in case someone start ends up looking at the QUICKSTART before going through the README.  Another stab at something like this was made on the wiki: https://github.com/commercialhaskell/stack/wiki/New-to-Haskell.  That page is deprecated, but it might be worth having a look at that page to see if there's anything worth bringing over.  I like the simplicity of your quickstart, though, so I don't think it needs to have all that info (which is in the GUIDE as well, and belongs there).

Probably worth adding a link from the README to the QUICKSTART as well.

snoyberg:
How about also adding a link at the end of the quickstart to other resources (including the current guide) for people looking for more information?

yogsototh:
@snoyberg yes of course, which link do you think we should add?

Are you thinking to https://github.com/commercialhaskell/stack#the-haskell-tool-stack ?

snoyberg:
I was talking about linking to [the guide](https://github.com/commercialhaskell/stack/blob/master/doc/GUIDE.md). But the link you just provided raises an important point: how is the quickstart intended to be different from the instructions in the README? Would it make more sense to just improve the README?

yogsototh:
While the README contains most informations, it is more a document explaining what stack is, while the quick start is more an instruction guide which helps you to start without taking the time to understand much, but make you use Haskell as fast as possible.

I've first created this page to link for your download haskell page proposition. When I first followed the link to the guide I found it was a very long read which can be a bad first experience for a beginner that generally don't really want to understand everything and just want to start as soon as possible.

So the choice is yours. We might also add a quick start section in the README if you find this is a better idea.

snoyberg:
If you look at the section you link to in the README, it contains much the same information as the quickstart guide. I don't like the idea of having two very-similar-yet-slightly-different pieces of content: it's extra maintenance, and presents confusion as to where people should be linked to. As I see it, we have three different goals we're trying to address:

* General information about what Stack is: link to the top of the README
* In-depth information on using Stack: link to the guide
* Quick start to using Stack: now we'd have two options, either the linked-to section of the README, or the newly proposed quickstart guide

If there's a different purpose for the quickstart versus that section of the README, let's clarify it. Otherwise, we should either:

1. Not create a new page and focus on improving the README
2. Remove the quickstart content from the README and instead link to the new quickstart guide

snoyberg:
@yogsototh Did you have any thoughts on my comments here?

yogsototh:
I believe adding a quick start section in the README is the best thing to do. Sorry for my late reply, I'm on vacation until tomorrow. If I remember correctly there is already a section in the README very close to the quick start. We might just rename it quick start and slightly modify it if needed to make it short.

yogsototh:
I've added a Quick Start section in the README. It is close to the section "How to Use" but I didn't want to remove anything so for now I just left it untouched.

PR #1083, opened 2015-09-30
Fix Monoid SetupInfo instance

fizruk:
Current `Monoid` instance does not allow to override behaviour as specified in [`stack-info` documentation](https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md#setup-info).


borsboom:
Thanks!

PR #1086, opened 2015-10-01
Adds `stack sig sign sdist` and `stack upload --sign`

dysinger:


src/Stack/Sig/Cabal.hs
```diff
@@ -0,0 +1,45 @@
...
+import qualified Data.Version as V
+import qualified Distribution.Package as D
+import qualified Distribution.PackageDescription as D
+import qualified Distribution.PackageDescription.Parse as D
+import qualified Distribution.Verbosity as D
+import           Stack.Types
+
+-- | Extract the @PackageIdentifier@ given an exploded haskell package
+-- path.
+cabalFilePackageId
```
snoyberg: I think this can be replaced by some functions in Stack.Package, which will (1) avoid code duplication, and (2) ensure that Cabal library workarounds (like BOM stripping) are centralized to one place.

src/Stack/Sig/GPG.hs
```diff
@@ -0,0 +1,109 @@
...
+
+-- | Extract the full long @fingerprint@ given a short (or long)
+-- @fingerprint@
+fullFingerprint
+    :: (Monad m, MonadIO m, MonadThrow m)
+    => Fingerprint -> m Fingerprint
+fullFingerprint (Fingerprint fp) = do
+    (code,out,err) <-
+        liftIO
+            (readProcessWithExitCode
```
snoyberg: This should use the functions from System.Process.Read so that the modified environment is used, which will be useful especially for Windows.

src/Stack/Sig/GPG.hs
```diff
@@ -0,0 +1,109 @@
...
+        else maybe
+                 (throwM
+                      (GPGFingerprintException
+                           ("unable to extract full fingerprint from output:\n " <>
+                            out)))
+                 return
+                 (let hasFingerprint = (==) ["Key", "fingerprint", "="] .
+                          take 3
+                      fingerprint = T.pack .
+                          concat .
```
snoyberg: Should this be `unwords` instead?

src/Stack/Sig/Sign.hs
```diff
@@ -0,0 +1,139 @@
...
+import           Stack.Types
+import           System.Directory (getDirectoryContents)
+import           System.Process (readProcessWithExitCode)
+
+-- | Sign a haskell package with the given url of the signature
+-- service and a path to a tarball.
+sign
+    :: (MonadCatch m, MonadBaseControl IO m, MonadIO m, MonadMask m, MonadLogger m, MonadThrow m)
+    => String -> FilePath -> m ()
+sign url filePath = do
```
snoyberg: It would definitely be better to use `tar` here, which will avoid the unnecessary filesystem activity.

src/Stack/Sig/Sign.hs
```diff
@@ -0,0 +1,139 @@
...
+                          (head cabalFiles)
+                  pkg <-
+                      Cabal.cabalFilePackageId
+                          (toFilePath
+                               (tempDir </> cabalFile))
+                  signPackage url pkg filePath))
+
+-- | Sign a haskell package with the given url to the signature
+-- service, a package tarball path (package tarball name) and a lazy
+-- bytestring of bytes that represent a the tarball bytestream.  The
```
snoyberg: s/a the/the

src/main/Main.hs
```diff
@@ -773,9 +788,9 @@ upgradeCmd (fromGit, repo) go = withConfigAndLock go $ globalFixCodePage go $
     upgrade (if fromGit then Just repo else Nothing) (globalResolver go)
 
 -- | Upload to Hackage
-uploadCmd :: ([String], Maybe PvpBounds) -> GlobalOpts -> IO ()
-uploadCmd ([], _) _ = error "To upload the current package, please run 'stack upload .'"
-uploadCmd (args, mpvpBounds) go = do
+uploadCmd :: ([String], Maybe PvpBounds, Bool) -> GlobalOpts -> IO ()
+uploadCmd ([], _, _) _ = error "To upload the current package, please run 'stack upload .'"
+uploadCmd (args, mpvpBounds, shouldSign) go = do
```
snoyberg: I'd like to make it easier for people to automatically sign, likely with some setting in their stack.yaml file. Can that be added as a default in case `--sign` is not provided on the command line?

borsboom:
@dysinger / @snoyberg: I see there have been some more commits since the last comments.  Is this ready to merge?

snoyberg:
I still see that some of my comments seem to be unaddressed.

dysinger:
@borsboom can you take a look at this when you are able? I think it's ready

src/Stack/Sig/GPG.hs
dysinger: I forgot this one.  but as I am doing it, I realized that STDERR contains the output of a successful verify with GPG.  So I can't just use readProcess in that package (AFAIK).

dysinger:
I think the only one I haven't address is switching to System.Process.Read.  This would be hard for one of the functions (verify) which needs to stream in on stdin and consume stderr after (& see a successful exit code).  @borsboom talked about maybe adding to System.Process.Read in the future.  I think this branch is ready to merge (after CI).

borsboom:
Can you add an entry to the change log, and also add/update any relevant documentation in `doc/`?  If you'd rather I just merge this now and you'll come back to do those later, that's ok too, just let me know.

dysinger:
added a few lines to the ChangeLog.md & also to the GUIDE.md

borsboom:
Cool.  Might be worth moving this to the "major changes" section of the changelog; I think it's worth calling extra attention to it.

Issue #532, opened 2015-07-06
Provide proxy to latest binary release


3noch:
GitHub makes it very hard (if not impossible) to provide a download link for the *latest* release of a file.

Would it makes sense to host a proxying URL that allows script writers to always have access to the latest binary release of stack? Even `stack upgrade` could use it potentially. But I imagine the primary benefit would be for "bootstrapping" scripts or provisioning tools to simply download the latest binary and put it on the `PATH`.

Since `stack` takes you the rest of the way, it would be possible to provide a shell script and a PowerShell script and reach almost every OS on the planet. By copying from those templates, users could make their systems run with potentially zero system dependencies. That would be awesome.

phadej:
For bootstrapping scripts *latest* version is dangerous concept, especially when back-incompatible changes are introduced. 

Which got me thinking, does stack download the latest GHC in major series (i.e. will download 7.10.2 when it comes out for say `nightly-2015-07-06` resolver), or are snapshots tied to exact GHC version?

phadej:
And by the way, starting from `stack-0.1.2.0` you can add `stack upgrade` to bootstrap script: done.

3noch:
@phadej Good idea. I'm pretty sure the "danger" is highly related to what features you need. I sure hope that bootstrapping to the latest in order to run `stack build` is not dangerous! But I see your point.

DanBurton:
> does stack download the latest GHC in major series?

Yes. Snapshots are associated with a GHC major series, not with a precise
GHC version. Pinging @snoyberg to double check that this is still the case.

On Monday, July 6, 2015, Elliot Cameron <notifications@github.com> wrote:

> @phadej <https://github.com/phadej> Good idea. I'm pretty sure the
> "danger" is highly related to what features you need. I sure hope that
> bootstrapping to the latest in order to run stack build is not dangerous!
> But I see your point.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/commercialhaskell/stack/issues/532#issuecomment-118944250>
> .
>


-- 
-- Dan Burton

phadej:
@3noch to clarify myself. I think it's a bad idea that there are time-dependent variables in CI / provision etc scripts. Something that worked yesterday may stop working tomorrow: introduced regression; accidental exploiting of a bug, fixed later. The latter is not the case for `stack build`, but that should work with older version too.

[...]

PR #1092, opened 2015-10-02
Update travis instructions to use stack-0.1.5.0

bergmark:
I looked if there was a way to just link to the latest version, but I didn't find anything useful.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1603,7 +1603,7 @@ before_install:
 # Download and unpack the stack executable
 - mkdir -p ~/.local/bin
 - export PATH=$HOME/.local/bin:$PATH
-- travis_retry curl -L https://github.com/commercialhaskell/stack/releases/download/v0.1.4.0/stack-0.1.4.0-x86_64-linux.tar.gz | tar xz -C ~/.local/bin
+- travis_retry curl -L https://github.com/commercialhaskell/stack/releases/download/v0.1.5.0/stack-0.1.5.0-x86_64-linux.tar.gz | tar xz -C ~/.local/bin
 
 # This line does all of the work: installs GHC if necessary, build the library,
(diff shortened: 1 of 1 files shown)


borsboom:
It's possible to get the latest version's URL using an incantation like this:

    curl -sSL https://api.github.com/repos/commercialhaskell/stack/releases/latest \
        | grep '"browser_download_url": ".*/stack-[0-9\.]\+-x86_64-linux.tar.gz"' \
        | sed 's/.*"\(.*\)"$/\1/'

Not sure if you'll want to include something like this, so it might be better to just replace version numbers with `<VERSION>` and add a note to replace it with the current version, rather than needing to remember to update this document for every new version.

See are planning to include an easier "download the current version" link on the website once it's ready.  See #532.

snoyberg:
@borsboom One concern I have is that we've had a different bindist format for the past three releases. Are we sure we're going to have something stable now? Otherwise, hard-coding a version into the document still seems prudent.

borsboom:
@snoyberg I don't foresee changing it again anytime soon; maybe some files will be added, but there will continue to be a `stack` executable in the top-level directory of the tarball.

phadej:
it should be:
```sh
curl -L https://github.com/commercialhaskell/stack/releases/download/v0.1.5.0/stack-0.1.5.0-x86_64-linux.tar.gz | tar xz --strip-components=1 -C ~/.local/bin
```

i.e. `--strip-components=1` for `0.1.5.0`

borsboom:
Even better is
```
curl -L https://github.com/commercialhaskell/stack/releases/download/v0.1.5.0/stack-0.1.5.0-x86_64-linux.tar.gz | tar xz --strip-components=1 -C ~/fpco/tmp/xx stack-0.1.5.0-x86_64-linux/stack
```
(which only extracts the `stack` executable, and skips the documentation)

Issue #783, opened 2015-08-13
local templates


legrostdg:
It would be cool if "stack new" and "stack templates" could use local template files, instead of getting templates from https://github.com/commercialhaskell/stack-templates.git only.

tfausak:
:+1: I would like this, too! 

In fact, I would like to be able to specify a remote template. I am able to do that with [hi](https://github.com/fujimura/hi), which I would consider `stack new`'s closest competitor.

lukexi:
Agreed! Allowing a list of repo URLs to be entered in the global stack.yaml (defaulting to commercialhaskell's to start with) would be particularly great as devs could then offer template repos to go with their packages.

AaronFriel:
Also agreed, I'd prefer if Stack cached templates that were available a la `apt-get update` versus `apt-get install`.

I write this not coincidentally because my local VM for development isn't on the network right now, and `stack new` is broken without network access.

aleator:
:+1:  Also, optionally giving templates as URLs would also be great!

GregorySchwartz:
I'll take a stab at this, the local templates.

snoyberg:
Closing in favor of PR #1095.

PR #1095, opened 2015-10-03
Now can specify a template in any local subdirectory to use

GregorySchwartz:
In reference to #783, added the ability to use any local template in the current directory or any subdirectory, for instance

`stack new project bar` or `stack new project ./foo/bar`

I could not include the ability to get a local template from a non local or subdirectory where the project was not located (e.g. ~/my_local_templates/bar) because the TemplateName type is for relative paths only. Is there a reason for this restriction? Why not have absolute paths for everything?

Also, loadTemplate is now, thanks to me, on the verge of getting a bit messy. Let me know if you would like me to clean it up.

### DIFF
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -110,12 +110,18 @@ loadTemplate name = do
     req <-
         parseUrl (defaultTemplateUrl <> "/" <> toFilePath (templatePath name))
     config <- asks getConfig
-    let path = templatesDir config </> templatePath name
-    _ <- catch (redownload req path) (throwM . FailedToDownloadTemplate name)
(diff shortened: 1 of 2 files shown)


snoyberg:
This looks good to me, but I'm a bit concerned about the relative path restriction. @chrisdone can you comment on that?

GregorySchwartz:
Yeah, I didn't want to change the TemplateName type (TemplateName !Text !(Path Rel File)) as it's used in many places and may require the relative path for one reason or another.

chrisdone:
Pushed commits that merges this in and allows absolute paths too. Ta!

PR #1112, opened 2015-10-06
Allow stack path to release the lock immediately.  Fixes #872

### LINKED ISSUE #872: `stack path` tries to acquire lock
Our build environment uses `stack path` to figure out where to get the binaries to archive. Right now this causes builds to be slow since this step waits for other stack instances in concurrent build jobs to release their lock.

I'd expect `stack path` to not require any locking at all since it just prints some path information, but I might be overlooking something here.

rrnewton:
@snoyberg vouched for stack path definitely not needing locking.  (Which makes intuitive sense, but I was conservative initially, and only opted out commands where after someone vouched for them or I read the relevant code.)

Note that as of the current release this only affects runs with `STACK_LOCK=true`.


snoyberg:
Thanks

PR #1115, opened 2015-10-06
Make the failed-to-grab-lock message obey --verbosity=silent.  Fixes #873.

### LINKED ISSUE #873: `stack path` outputs diagnostics to stdout
When I run `stack path` while another stack instance is running, it outputs information about acquiring a lock to stdout. This makes it harder for build scripts to deal with the output of `stack path`, I'd expect such diagnostics to go to stderr. 

Also, adding `--verbosity SILENT` doesn't make a difference, but I'm not sure if that's an issue. If so, I'd be happy to create a separate issue for that.

Note that this issue is related to #872, if no locks are needed for `stack path` there might not be an issue here at all.

rrnewton:


snoyberg:
Thanks!

Issue #965, opened 2015-09-09
Add --bare flag to "stack new" (similar to "yesod init --bare")


sfilipov:
I just checked the yesod quickguide and it has been updated recently. It now recommends "stack new my-project template-name" as the way to initialize a new yesod project. What is missing is the option to do so in the current folder. This was previously done using "yesod init --bare" followed by "stack init".

Is this an intentionally missing feature or just a lower priority one?

chrisdone:
I hadn't planned on it, but we can add it.

snoyberg:
Resolved by #1119

PR #1119, opened 2015-10-07
Add flag "--bare" to stack new #965

sebastianpoeplau:
Following #965, I added a new option "--bare" to stack new that causes the new project to be created in the current directory rather than in a subdirectory.


snoyberg:
Thanks!

PR #1130, opened 2015-10-09
Recommend extra-dep in yaml_configuration.yml

rimmington:
Now that valid-wanted has been deprecated, it should be removed from the documentation.


snoyberg:
Good catch, thanks!

PR #1133, opened 2015-10-09
Compile custom Setup.hs instead of interpreting them (fixes #1041)

### LINKED ISSUE #1041: Stack is extremely slow for Cabal, due to custom setup script
Stack uses runghc to run custom Setup scripts, which for Cabal means that builds are unusably slow as all of Cabal is interpreted. In these cases, I just stop having Cabal bootstrap itself because Stack is so slow in this case. I don't know, can Stack maybe build the script or something?

snoyberg:
Pinging @mgsloan for review. I just (ab)used your existing code for GHCJS, please sanity check that I didn't do anything terrible.

Also pinging @ezyang, who's probably interested in this.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -30,6 +30,7 @@ Other enhancements:
 * Docker: download or override location of stack executable to re-run in container [#974](https://github.com/commercialhaskell/stack/issues/974)
 * Docker: when Docker Engine is remote, don't run containerized processes as host's UID/GID [#194](https://github.com/commercialhaskell/stack/issues/194)
 * Docker: `set-user` option to enable/disable running containerized processes as host's UID/GID [#194](https://github.com/commercialhaskell/stack/issues/194)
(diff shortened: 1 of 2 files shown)


mgsloan:
LGTM!  Such a concise change :D

snoyberg:
Fortunately someone did all the hard work for me already

ezyang:
Heh. Back in Cabal land, we actually just axed the custom Setup.hs because it was more headache than it was worth. But this looks reasonable. (Question: when does Setup get rebuilt in this setup?)

snoyberg:
We ask GHC to build each run, and trust it to be smart enough to not
recompile unnecessarily

ezyang:
(This actually would work pretty poorly for old-bootstrapped Cabal; you'd end up rebuilding the setup script every time you make a Cabal change, but once again, we don't really care anymore!)

PR #1138, opened 2015-10-10
Fix issue links

joneshf:
The extra space was breaking the markdown.


snoyberg:
Can you clarify where you see broken links? Things are working just fine on the Github display for me.

joneshf:
That's odd, it's working now for me as well.

Maybe it was something temporary with the markdown parser?

Ignore this then :)

PR #1139, opened 2015-10-10
Fix `awaiting pr` label link

joneshf:
The previous link was giving back no results.
It seems like the label was changed.


snoyberg:
Thanks!

Issue #1109, opened 2015-10-06
Make compilation cycles in --file-watch mode visually separated


ethercrow:
Currently it's pretty hard to see in a wall of output where the previous compilation ended and current compilation began.

I suppose stack could make ExitFailure and ExitSuccess lines colored red and green when stdout goes to an interactive terminal. Alternatively, some ascii-art page breaks would work too.

Attaching output of `stack install --file-watch` and `sos -p '\.hs' -c 'stack build'` for comparison.
![screen shot 2015-10-06 at 14 55 46](https://cloud.githubusercontent.com/assets/222467/10304207/2c5ca344-6c32-11e5-81a6-a4dc3d01a3f7.png)
![screen shot 2015-10-06 at 14 56 03](https://cloud.githubusercontent.com/assets/222467/10304211/34915a64-6c32-11e5-8adb-92e5017b76c4.png)

snoyberg:
Sounds like a good change to me, and is in a very isolated part of the codebase. Interested in taking a stab at this?

ethercrow:
Sure, I'll probably look into it during the weekend if noone jumps in before that.

snoyberg:
Implemented by #1142, closing.

PR #1142, opened 2015-10-11
Colored build status in filewatch mode

ethercrow:


ethercrow:
See https://github.com/commercialhaskell/stack/issues/1109

snoyberg:
Thanks!

Issue #1071, opened 2015-09-28
solver not considering cabal flags


eallik:
when I add this

```yaml
  reactive-banana-wx: 
    buildExamples: true
```

to the `flags` section of `stack.yaml`, `stack build` starts complaining about missing dependencies (`random` and `executable-path`) and proposes adding `random-1.1` and `executable-path-0.0.3` to `extra-deps`, however, the next run of `stack build` complains about `random-1.1` and suggests something closer to `1.0.*` is needed. So I remove the 2 newly added items from `extra-deps` and try to make `stack solver` figure it out on its own. However, `stack solver` surprisingly responds with "no changes needed".

— does this mean `stack solver` doesn't take into account the `flags` section in `stack.yaml`?

snoyberg:
Correct, that's the current behavior, since part of the solver step is to determine flags for us. It could be argued that explicit flag settings present in stack.yaml should be respected. It shouldn't be hard to implement this, want to take a crack at such a change and report how it works on your project?

sebastianpoeplau:
I've tried an implementation (see referenced merge request). @eallik, would you mind testing how it works for your project?

snoyberg:
Implemented by #1143.

PR #1143, opened 2015-10-11
Consider user-specified package flags in stack solver #1071

sebastianpoeplau:
Pass the user-specified package flags to Cabal's "--constraint" argument.


snoyberg:
Thanks!

PR #1164, opened 2015-10-15
Use platform-specific search path separators

lukexi:
This patch ensures we use the correct platform-specific search path separators in the `stack ide` command.

With this we're very close to a fully working stack-ide on Windows - just [#92](https://github.com/commercialhaskell/stack-ide/issues/92) remains.


snoyberg:
Thanks!

PR #1185, opened 2015-10-17
Move toFilePathNoTrailingSlash to new module Path.Extra

sjakobi:


sjakobi:
If there is a better way to dedupe `toFilePathNoTrailingSlash` please tell me!

sjakobi:
I've found even more duplicates of that function. 

Closing until I've fixed the others.

borsboom:
I like `Path.Extra`.  There might be a few functions in Path.IO that are not actually I/O related that would more properly belong in this new module.

sjakobi:
> There might be a few functions in Path.IO that are not actually I/O related that would more properly belong in this new module.

I'll have a look at those!

PR #1193, opened 2015-10-18
Create Path.Extra for Path utilities unrelated to IO

sjakobi:
Superseding #1185.


borsboom:
Thanks, that's a nice bit of cleanup.

PR #1194, opened 2015-10-18
Replace libVersionFromInstalled with installedVersion

sjakobi:


borsboom:
Thanks!  I missed that one too...

PR #1196, opened 2015-10-19
Simplify the Appveyor before_test setup

ndmitchell:
Appveyor guarantees `curl` and `7z` on the PATH, so use the default tools, makes it easier to understand for non-Appveyor experts. I stole the appveyor code for HLint from Stack, then modified it in HLint, so this is me returning my changes.

cc @snoyberg who seems to be your Appveyor guy.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,11 +5,8 @@ cache:
 build: off
 
 before_test:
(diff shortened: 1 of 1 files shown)


ndmitchell:
Oh, it also grabs the latest stack always, instead of being hardcoded to 0.1.4.0.

snoyberg:
Wow, much nicer, thanks! Figuring out powershell to make that work was painful, I'm glad I can safely forget everything I learned ;)

PR #1197, opened 2015-10-19
Use Path type in Data.Binary.VersionTagged functions

sjakobi:


borsboom:
@snoyberg Thoughts on whether we want Data.Binary.VersionTagged to use Path?  Only reason I can think of not to would be if we want to break this out into a separate library at some point, and perhaps don't want to have a dependency on the `path` package?

snoyberg:
I'm fine with it going either way. We've had a track record of leaving non-Stack-specific modules using `FilePath`, but I don't think that's necessary.

PR #1199, opened 2015-10-19
yaml_configuration.md: Add information about template parameters

sfilipov:
Currently there is no documentation about the template parameters that `stack new` suggests if not set, i.e. `author-email`, `author-name`, `category`, `copyright` and `github-username`.

I've added some information based on my understanding of these parameters. I haven't looked at any code to see if my description is exhaustive of all effects of the parameters. It would be nice if someone who worked on the stack templates could proofread what I wrote.

I am open to suggestions of any kind.


borsboom:
Thanks!

PR #1201, opened 2015-10-20
Refactor findTransitiveGhcPkgDepends

sjakobi:
Regarding the FIXME:
res' is entirely contained in res''. Therefore the Map.union is
redundant.

### DIFF
--- a/src/Stack/GhcPkg.hs
+++ b/src/Stack/GhcPkg.hs
@@ -34,7 +34,6 @@ import           Control.Monad.Trans.Control
 import qualified Data.ByteString.Char8 as S8
 import           Data.Either
 import           Data.List
-import qualified Data.Map as Map
 import           Data.Maybe
 import           Data.Set (Set)
 import qualified Data.Set as Set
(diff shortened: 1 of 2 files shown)


src/Stack/GhcPkg.hs
```diff
@@ -193,29 +193,37 @@ findTransitiveGhcPkgDepends
...
+                     -> GhcPkgId
+                     -> m (Maybe PackageIdentifier)
+getPackageIdentifier menv wc pkgDbs ghcPkgId = do
+    let s = ghcPkgIdString ghcPkgId
+    mname <- findGhcPkgField menv wc pkgDbs s "name"
+    mversion <- findGhcPkgField menv wc pkgDbs s "version"
+    pure $ do
+        name <- mname >>= parsePackageName . T.encodeUtf8
+        version <- mversion >>= parseVersion . T.encodeUtf8
+        Just $ PackageIdentifier name version
```
sjakobi: BTW, why isn't the `PackageIdentifier` simply parsed from the`GhcPkgId`?

borsboom: That's how it used to work, but a GHC package IDs have become opaque and we can no longer rely on them having any particular format (even if their current format happens to the contain the PackageIdentifier).

borsboom:
Going to hold off merging this until some major refactoring-in-progress (which will remove much of the code modified in this PR) is done.

sjakobi:
Closing as `findTransitiveGhcPkgDepends` was deleted.

PR #1202, opened 2015-10-20
Stack config add resolver command

drwebb:
config selects the projects stack.yaml, regardless of directory

Parsing of config add command

Integration test helper (will be used to compare stack.yaml files)

Check to see if snapshot exists before writing

TODO:
Adding differnet fields
Additional subcommands for git

### DIFF
--- a/src/Stack/ConfigCmd.hs
+++ b/src/Stack/ConfigCmd.hs
@@ -1,0 +1,71 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | Make changes to the stack yaml file
+
+module Stack.ConfigCmd
+       (ConfigCmdSet(..)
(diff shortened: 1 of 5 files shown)


src/Stack/ConfigCmd.hs
```diff
@@ -0,0 +1,71 @@
...
+    -- We checking here that the snapshot actually exists
+    snap <- parseSnapName newResolverText
+    _ <- loadMiniBuildPlan snap
+
+    let projectYamlConfig' =
+            HMap.insert
+                "resolver"
+                (Yaml.String newResolverText)
+                projectYamlConfig
+    liftIO
```
snoyberg: More efficient: just use `Yaml.encodeFile`. Definitely not necessary to convert from a strict `ByteString` to a lazy `ByteString` via a `Builder`.

snoyberg:
I had one comment about `Yaml.encode`, plus the GHC 7.8.4 build failure. Once those are addressed, this is good to merge!

PR #1204, opened 2015-10-20
Fix stack ide crash when not using stty

lukexi:
The fix mentioned here https://github.com/lukexi/stack-ide-sublime/issues/17#issuecomment-142949632 still caused crashes since undefined was being evaluated. This makes it work as intended by changing it to `return ()` (and lets stack-ide-sublime work again)


rvion:
Is this ready ? :)

Issue #1195, opened 2015-10-19
Ignore cabal_macros.h as a dependency


lukexi:
As Michael says in #1134:
> Also somewhat tangential, but: we may want to always ignore cabal_macros.h as a dependency. That file should only change meaningfully when our configure options have changed, and therefore recompiles triggered by that file should be meaningless.

I seem to get endless spurious rebuilds of my local packages due to cabal_macros.h — nearly every time I make a change. If I run stack build a few more times it eventually settles down.

It's of course particularly pernicious when the cabal_macros.h change is detected in package A, which is depended upon by B C D E F, yet I'm only actually making changes in package G that depends on all of the above.

snoyberg:
PR was merged, closing. Thanks!

PR #1205, opened 2015-10-20
Filter the cabal_macros.h file when checking for dirty files

lukexi:
Re: #1195 — let me know if there's a better place to do this!

(I also noticed a couple of .git files in the list returned from `getPackageFilesSimple`, maybe we should filter those too?)


borsboom:
@snoyberg, would you mind reviewing this?  It looks good to me, but that's not a bit of the code I've touched yet.

> (I also noticed a couple of .git files in the list returned from getPackageFilesSimple, maybe we should filter those too?)

If you're using the `gitrev` package (as Stack does), the `.git` files are legitimate dependencies (changing the commit ID should cause the `stack executable` to build, since it includes the commit ID in `--version` output).

snoyberg:
Thanks!

lukexi:
@borsboom ah, that explains why there were just those two files, thanks : )

Issue #1052, opened 2015-09-25
stack prefers ghc from .stack-work\install\…\bin over the one from PATH


mwu-tow:
I have a package that compiles to the executable named `ghc.exe`. 
When building first time, everything works fine. The install step places the built `ghc.exe` in the `.stack-work\install\x86_64-windows\lts-2.22\7.8.4\bin` directory. As expected.

However, any subsequent `stack build` command will fail:
```
    setup-Simple-Cabal-1.22.4.0-x86_64-windows-ghc-7.8.4.exe: The program 'ghc' version >=6.4 is required but the version of C:\z\.stack-work\install\x86_64-windows\lts-2.22\7.8.4\bin\ghc.exe could not be determined.
    setup-Simple-Cabal-1.22.4.0-x86_64-windows-ghc-7.8.4.exe: fd:5: hGetContents: invalid argument (invalid byte sequence)
```
To built it again I have to delete `.stack-work` directory.

It is because stack tried to use `.stack-work\install\x86_64-windows\lts-2.22\7.8.4\bin\ghc.exe` — that is not an actual GHC but the locally built package that was just named so. 

I expect that `stack` should prefer ghc.exe from PATH or its `ghc-paths` over the built executables. The results of build should not affect subsequent builds, and overwriting the compiler executable should not be implicit.

snoyberg:
I've pushed a patch to master to pass in `--with-ghc` to the configure step, which should fix this. Can you give it a shot?

In general with build tools, however, we _want_ to use them for building things (e.g., a Stack-built alex should be used instead of whatever's on the PATH). I'm OK special-casing this for GHC, but in general having an executable shadow the name of another build tool should be considered an error.

mwu-tow:
Actually I'm a little surprised that stack-built build tools should be preferred over the PATH. I'd expect that `stack build` simply builds things and should not affect how subsequent `stack build`. Doesn't it weaken the rule of reproducible builds?  If I wanted, I'd try using something like `stack exec stack build` (does this makes sense?) or just `install` the build tools. 
[Not saying you should change how things work, just thought it might me worth of noting.]

Is there any way I can prevent build artifacts from affecting the build? 

Anyway, with the latest master everything works fine for me. Many thanks! :-)

snoyberg:
Consider the case of a tool like alex. Stack needs to build it in order for
other packages to use it, and we want to prefer that version of alex to
whatever happens to be on the PATH

mwu-tow:
[...]

PR #1210, opened 2015-10-21
Ignoring stack-built `ghc` executables when looking for compiler for Setup.hs (#1052).

mwu-tow:
The issue #1052 was once fixed but got broken again when compiling custom `Setup.hs` has been implemented — see https://github.com/commercialhaskell/stack/issues/1052#issuecomment-149281657. This PR fixes this by excluding local extra bin directories from lookup when searching for GHC to compile `Setup.hs` with.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -692,14 +692,18 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
 
     withCabal package pkgDir mlogFile inner = do
         config <- asks getConfig
-        menv <- liftIO $ configEnvOverride config EnvSettings
-            { esIncludeLocals = taskLocation task == Local
-            , esIncludeGhcPackagePath = False
-            , esStackExe = False
-            , esLocaleUtf8 = True
-            }
-        getGhcPath <- runOnce $ liftIO $ join $ findExecutable menv "ghc"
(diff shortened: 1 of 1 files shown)


src/Stack/Build/Execute.hs
```diff
@@ -692,14 +692,18 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
...
+                , esIncludeGhcPackagePath = False
+                , esStackExe = False
+                , esLocaleUtf8 = True
+                }
+        menv <- liftIO $ configEnvOverride config envSettings
+        -- When looking for ghc to build Setup.hs we want to ignore local binaries, see:
+        -- https://github.com/commercialhaskell/stack/issues/1052
+        menvWithoutLocals <- liftIO $ configEnvOverride config envSettings { esIncludeLocals = False }
+        getGhcPath <- runOnce $ liftIO $ join $ findExecutable menvWithoutLocals "ghc"
+        getGhcjsPath <- runOnce $ liftIO $ join $ findExecutable menvWithoutLocals "ghcjs"
```
snoyberg: @mgsloan Can you confirm if the local database should or shouldn't be used for finding the `ghcjs` executable?

mgsloan: Right, it makes sense to exclude locals for ghcjs as well.

snoyberg:
Merged, thanks!

Issue #817, opened 2015-08-18
stack ghci should be usable with minimal dependencies


rpglover64:
After upgrading to the newest version of stack (prompted by @chrisdone in the thread for #665), I ran `stack ghci` in the `turtle` project directory. Rather than failing (which was the previous undesirable behavior), it started building the dependencies for the test and bench suites. While this is clearly an improvement, I expect to be able to run ghci without necessarily installing the dependencies for the test and bench suites, and I think that should be the default behavior.

chrisdone:
Running it with `stack ghci turtle:lib` loads up just the library. The default behaviour for `stack build` means "all targets", `stack ghci` behaves the same, I think this is consistent.

chrisdone:
Hm, or it's supposed to. It seems not to be working properly:

```
chris@ubuntu:~/Repos/turtle$ stack ghci turtle:lib
Configuring GHCi with the following packages: turtle
GHCi, version 7.10.2: http://www.haskell.org/ghc/  :? for help
<command line>: cannot satisfy -package criterion-1.1.0.0
    (use -v for more information)
```

chrisdone:
I made a spin-off issue here: #818

rpglover64:
> The default behaviour for `stack build` means "all targets", `stack ghci` behaves the same, I think this is consistent.

Makes sense. I don't agree with this default, though. (Has it always been this way?) I believe that people will be using stack as package end-users (e.g. for local installs of packages) more than as package developers, so I'd optimize for the former case while keeping the latter case in easy reach.

Here's behavior I'd expect (I don't know how much of it is already implemented but I couldn't find in the documentation):

* `stack build` and `stack ghci` only build the package itself; that is, they default to only the lib target
* There is more documentation about targets; the section of the FAQ that deals with them doesn't mention anything that suggests `turtle:lib` as a target
* There is a command to list available targets for a given package
* There is a virtual target, e.g. "all", that builds all targets
* `stack build --test`, `stack test`, `stack bench`, etc. build the targets necessary for them to work

chrisdone:
Okay, #818 is fixed.

chrisdone:
> stack build and stack ghci only build the package itself; that is, they default to only the lib target

Actually, thinking about it, `stack build` differs in that it does only build executables and libraries at the moment. Then `--test` and `--bench` will build benchmarks. We can enable this behaviour for `stack ghci`. I think that would satisfy you and be consistent with `stack build`.

[...]

PR #1228, opened 2015-10-25
'stack ghci': build project before launching GHCi

### LINKED ISSUE #1065: Stack ghci before stack build fails
C:\Neil\hlint>stack ghci
    Using main module: Package `hlint' component exe:hlint with main-is file: C:\Neil\hlint\src\Main.hs
    Configuring GHCi with the following packages: hlint
    GHCi, version 7.10.2: http://www.haskell.org/ghc/  :? for help

    <command line>:
        Could not find module `Paths_hlint'
    Failed, modules loaded: none.
    Prelude> :q
    Leaving GHCi.

That's after deleting my .ghci file and removing my .stack-work.

My guess is `stack ghci` should generate the paths file if it needs it. I thought it used to do a whole build before launching ghci? Using 06a4b270de2d2dcbd3ee8abe98078151c4a8f075.

### LINKED ISSUE #1180: 'stack ghci' doesn't notice that a module got deleted
If you remove a module and then run `stack ghci` before `stack build`, it fails with a `Could not find module` error.

Hypothesis:

The build doesn't run before `stack ghci` reads the modules, so it doesn't get a chance to re-write the .dump-hi files with the updated information (that the module does not exist), so `stack ghci` is still reading the out-of-date information.

Notes from @chrisdone:

> as a heads up the stack ghci code uses the build info, but runs the equiv. of `stack build —only-dependencies` to ensure that dependencies are available
> at a guess, it seems that the build cache doesn’t get invalidated by simply reading it in the manner that stack ghci does, but i didn’t get niklas to test with `stack build` to see if running that would trigger a cache invalidation
> the building with `—only-dependencies` was enabled to avoid ghci not being able to start when there’s a build error. i don’t like that fix, i’d prefer if the build just ran and if it fails then it should catch the exception and print the compile error but still launch ghci anyway in an optimistic way (edited)

It's likely that fixing #1065 will also fix this, but tracking separately since the symptoms are different.

(PAT-125)

borsboom:
(closes #1065, fixes #1180)

Optimistically, launch GHCi anyway even if the build fails.

@chrisdone, can you review?

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -197,7 +197,7 @@ ghciSetup mbuildFirst mainIs stringTargets = do
                 return (Just targets')
     let bopts = makeBuildOpts targets
     econfig <- asks getEnvConfig
-    (realTargets,_,_,_,sourceMap) <- loadSourceMap AllowNoTargets bopts
+    (realTargets,_,_,_,sourceMap) <- loadSourceMap AllowNoTargets (bopts BSAll)
     menv <- getMinimalEnvOverride
     (installedMap, _, _, _) <- getInstalled
         menv
@@ -219,15 +219,15 @@ ghciSetup mbuildFirst mainIs stringTargets = do
                              Nothing -> return Nothing
                     else return Nothing
     -- Try to build, but optimistically launch GHCi anyway if it fails (#1065)
-    case (mbuildFirst, M.null realTargets) of
-        (Just _, False) -> do
-            eres <- tryAny $ build (const (return ())) Nothing bopts
+    case mbuildFirst of
+        Just buildFirst -> do
+            eres <- tryAny $ build (const (return ())) Nothing (bopts buildFirst)
             case eres of
                 Left err -> do
                     $logError $ T.pack (show err)
                     $logWarn "Warning: build failed, but optimistically launching GHCi anyway"
                 Right () -> return ()
-        _ -> return ()
+        Nothing -> return ()
     -- Load the list of modules _after_ building, to catch changes in unlisted dependencies (#1180)
     let localLibs = [name | (name, (_, target)) <- locals, hasLocalComp isCLib target]
     infos <-
@@ -236,7 +236,7 @@ ghciSetup mbuildFirst mainIs stringTargets = do
              makeGhciPkgInfo sourceMap installedMap localLibs name cabalfp component
     return (realTargets, mainIsTargets, infos)
   where
-    makeBuildOpts targets =
+    makeBuildOpts targets buildFirst =
         base
         { boptsTargets = stringTargets
         , boptsTests = any (hasLocalComp isCTest) elems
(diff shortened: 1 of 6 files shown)


src/Stack/Ide.hs
```diff
@@ -43,7 +43,7 @@ ide
     -> [String] -- ^ GHC options.
     -> m ()
 ide targets useropts = do
-    (_realTargets,_,pkgs) <- ghciSetup Nothing targets
+    (_realTargets,_,pkgs) <- ghciSetup (Just BSAll) Nothing targets
```
borsboom: @chrisdone @mgsloan: Is building the project (and continuing even if it fails) the correct behaviour for `stack ide`?

src/main/Main.hs
```diff
@@ -884,7 +884,7 @@ packagesCmd () go@GlobalOpts{..} =
 targetsCmd :: Text -> GlobalOpts -> IO ()
 targetsCmd target go@GlobalOpts{..} =
     withBuildConfig go $
-    do (_realTargets,_,pkgs) <- ghciSetup Nothing [target]
+    do (_realTargets,_,pkgs) <- ghciSetup (Just BSAll) Nothing [target]
```
borsboom: @chrisdone: Is building the project (and continuing even if it fails) the correct behaviour for `stack targets`?

borsboom:
I've rebased this to resolve merge conflicts.  @chrisdone / @mgsloan: still not sure what the right behaviour for `stack ide` and `stack targets` should be.

src/main/Main.hs
mgsloan: Not building at all is fine for the targets command.

src/Stack/Ide.hs
mgsloan: Actually, I think the right behavior for stack ide is to only build dependencies.

mgsloan:
LGTM!  should the `--dependencies-only` alias be part of the BuildSubset parser?

src/Stack/Ghci.hs
```diff
@@ -215,12 +218,22 @@ ghciSetup mainIs stringTargets = do
                                  return (Just (name, (cabalfp, simpleTargets)))
                              Nothing -> return Nothing
                     else return Nothing
+    -- Try to build, but optimistically launch GHCi anyway if it fails (#1065)
+    case (mbuildFirst, M.null realTargets) of
```
mgsloan: I think the check for (null realTargets) should be removed.  See point 1 on #1249 for why.  See also #817 for why the check was there in the first place.

src/Stack/Ide.hs
borsboom: @mgsloan: Should it stay with the old behaviour of not starting if the dependencies fail to build, or would the new behaviour of continuing anyway be better?

mgsloan: The new behavior of continuing is better, I think.  At least that way the user ends up with a GHCI session that's somewhat setup.

borsboom:
@mgsloan I've updates based on your comments, please give this a final review and merge if good.

mgsloan:
LGTM!

PR #1229, opened 2015-10-25
Support for extra-package-dbs in 'stack ghci'

vigoo:
An addition to the `extra-package-dbs` option (https://github.com/commercialhaskell/stack/pull/990), to pass the extra package databases in the `stack ghci` command too.

### DIFF
--- a/src/Stack/GhcPkg.hs
+++ b/src/Stack/GhcPkg.hs
@@ -170,10 +170,11 @@ getCabalPkgVer menv wc =
         maybe (throwM $ Couldn'tFindPkgId cabalPackageName) return
 
 -- | Get the value for GHC_PACKAGE_PATH
-mkGhcPackagePath :: Bool -> Path Abs Dir -> Path Abs Dir -> Path Abs Dir -> Text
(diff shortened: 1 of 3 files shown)


borsboom:
LGTM.

@chrisdone, can review this, since you're more familiar with `stack ghci`?

mgsloan:
I've taken a look at this, and I think it should be done a different way.

Note that currently nothing about the package databases is passed to GHCi, yet it works as it should.  This is because we're essentially running it via `stack exec` which sets various environment variables.  In particular, it sets `GHC_PACKAGE_PATH`.

So, I think the best way to fix this (and fix other things along with it!) is to include the extra-package-dbs in `GHC_PACKAGE_PATH`.  This will involve adding an extra argument to [`mkGhcPackagePath`](https://github.com/fpco/stack/blob/5ba00a7fd29be20d49fdcce8758ae6f2e4d168e0/src/Stack/GhcPkg.hs#L174).

vigoo:
I've rewritten this PR based on @mgsloan 's comment. Please review.

mgsloan:
Looks good to me, thanks!

PR #1230, opened 2015-10-25
Pad timestamps in `--verbose` output

sjakobi:


borsboom:
Did you consider using `%T.%q` (note lowercase `q`) instead of using `%T%Q` (uppercase).  From my reading of the `formatTime` docs, `%q` should always be padded to 12 digits.  I think then you could remove all the extra padding logic (and if you only want six digits, just truncate the last six digits).

Another minor think I noticed while reading the docs: I think you could use `%F` instead of `%Y-%m-%d`.

sjakobi:
> Did you consider using %T.%q (note lowercase q) instead of using %T%Q (uppercase).

No! I had looked at the docs too but somehow overlooked that bit. Thanks!

borsboom:
Another suggestion: make `timestampLength` a top-level function, because that way it'll only have to be evaluated once (currently it's evaluated for every call, and since `length` has to iterate the whole list of characters that's significant).  

Also, to let you share most of the format string rather than hard-coding the expected format, you could do something like this (sharing a constant for the `%F %T` part, however):
```
timestampLength = length (formatTime defaultTimeLocale "%F %T.000000" (UTCTime (ModifiedJulianDay 0) 0))
```

sjakobi:
> currently it's evaluated for every call

I'm curious. How can you tell?

borsboom:
One way is to see is by prefixing the function body with `trace "evaluating timestampLength" $`, so it prints a message to the console every time it's evaluated.  There's some explanation about this and related topics in [this StackOverflow question](http://stackoverflow.com/questions/3951012/when-is-memoization-automatic-in-ghc-haskell).

sjakobi:
> One way is to see is by prefixing the function body with trace "evaluating timestampLength" $, so it prints a message to the console every time it's evaluated. There's some explanation about this and related topics in this StackOverflow question.

I did the tracing and the outcome was - of course - as you said. And thanks for the link! Learned something about Haskell again! :grinning:

borsboom:
Thanks!

PR #1234, opened 2015-10-26
Speed up generateDepsHaddockIndex

sjakobi:
This change reduces the time needed to compute the transitive dependencies for the `stack` project from ~393ms to ~~~27ms~~ ~11ms on my computer.


borsboom:
Thanks!

PR #1237, opened 2015-10-26
Control/Concurrent/Execute.hs : redundant do

Heather:
sorry for minor request :/

for future requests - is it better to make conceptual or split and small :S


snoyberg:
Thanks!

PR #1242, opened 2015-10-27
various minor QA improvements using ghc-mod's lint

Heather:
- replace when not into unless
 - remove some redundant `do`s, brackets, LANGUAGE mods
 - few other

changes should not anyhow affect logics


borsboom:
Thanks!

PR #1281, opened 2015-11-03
Improved efficiency for getPackageCaches

drwebb:
This PR covers issues uncovered by profiling the noopt build case. Results found that `Stack.PackageIndex.getPackageCaches`
was a costly function taking approx `%8-9` of total execution time. In additional
it was called three times by the following functions:
- `Stack.Build.ConstructPlan.constructPlan`
- `Stack.Build.Source.loadSourceMap`
- `Stack.Fetch.withCabalLoader`
taken together this would mean that the total time is around `%25` spent in this function. This PR makes improvements to the efficiency of this operation.

This PR makes two changes:

1. runOnce for withCabalLoader action in build config.
2. Safe invocations of getPackagesCaches should instead get a stored copy of the package caches from the build config. These are:
- `Stack.Build.ConstructPlan.constructPlan`
- `Stack.Build.Source.loadSourceMap`
- `Stack.BuildPlan.resolveBuildPlan`

Some functions which don't use the build config packageCaches introduced with this change are in `Stack.Upgrade.upgrade` `Stack.Fetch.resolvePackagesAllowMissing` , which relies on an environment that has indices updated; `Stack.Fegtch.withCabalLoader` which invalidates caches;

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -29,14 +29,15 @@ import           Control.Monad.Trans.Resource
 import           Data.Aeson (Value (Object, Array), (.=), object)
 import           Data.Function
 import qualified Data.HashMap.Strict as HM
+import           Data.IORef.RunOnce (runOnce)
 import qualified Data.Map as Map
 import           Data.Map.Strict (Map)
 import           Data.Set (Set)
 import qualified Data.Set as Set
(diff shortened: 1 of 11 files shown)


drwebb:
@borsboom did you have a chance to review the full PR? It's ready for merge.

PR #1283, opened 2015-11-04
Do all-in-one package builds when there are no cyclic dependencies

mgsloan:
Third swing at this worked out well!

One issue this has is that in the presence of cyclic dependencies, it will rebuild each time.  I'm looking into this, but really it's not so bad since it's pretty rare to do this.

Here's the output of running `stack test` on my [https://github.com/mgsloan/multi-test-suite](multi-test-suite) repo, which originally was for testing test coverage with multiple test suites.  It now also includes an example of a cyclic dependency.  Specifically, `multi-test-suite` depends on `cyclic`, and `cyclic`'s test-suite depends on `multi-test-suite`.

```
cyclic-0.1.0.0: unregistering (dependencies changed)
multi-test-suite-0.1.0.0: unregistering (missing dependencies: cyclic)
cyclic-0.1.0.0: configure
cyclic-0.1.0.0: build
cyclic-0.1.0.0: copy/register
multi-test-suite-0.1.0.0: build
multi-test-suite-0.1.0.0: copy/register
multi-test-suite-0.1.0.0: test (suite: multi-test-suite-test)
multi-test-suite-0.1.0.0: test (suite: multi-test-suite-test-2)
sub-package-0.1.0.0: build (exe + test)
cyclic-0.1.0.0: configure (test)
cyclic-0.1.0.0: build (test)
Warning: module not listed in sub-package.cabal for 'sub-package-test' component (add to other-modules): Lib2
sub-package-0.1.0.0: copy/register
sub-package-0.1.0.0: test (suite: sub-package-test)
cyclic-0.1.0.0: test (suite: cyclic-test-suite)
Completed all 7 actions.
```

### DIFF
--- a/src/Control/Concurrent/Execute.hs
+++ b/src/Control/Concurrent/Execute.hs
@@ -24,6 +24,7 @@ import           Stack.Types
 
 data ActionType
     = ATBuild
+    | ATBuildFinal
     | ATFinal
     deriving (Show, Eq, Ord)
 data ActionId = ActionId !PackageIdentifier !ActionType
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -21,12 +21,6 @@ module Stack.Build.Cache
     , setTestSuccess
     , unsetTestSuccess
     , checkTestSuccess
-    , setTestBuilt
-    , unsetTestBuilt
-    , checkTestBuilt
-    , setBenchBuilt
-    , unsetBenchBuilt
-    , checkBenchBuilt
     , writePrecompiledCache
     , readPrecompiledCache
     ) where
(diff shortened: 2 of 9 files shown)


mgsloan:
This is split into 3 commits in an attempt to make it easier to figure out where things went wrong:

* The first commit should result in identical behavior, it's a refactoring that ought to preserve logic

* The second one has the test / bench builds use the same build code as library builds, but still with the test / bench build coming after the library build.

* The third adds detection of cyclic dependencies and support for all-in-one builds

mgsloan:
The reason it's doing a rebuild is that it's considering the dependencies of a plain build against the dependencies of a test build (which was the last thing to cause a configure).

I can work around this by allowing the new dependencies set to be a subset of the old one.  I can't think of any scenarios where this will cause problems.  The potential cases I can think of are non-issues because editing the cabal file causes a reconfigure due to file dirtiness.  @snoyberg / @manny-fp thoughts on this?  Maybe it's better to accept the redundant rebuilds than accidentally introducing inconsistency into the config cache.

borsboom:
LGTM (albeit with limited understanding of big chunks of the code being modified).  For the cyclic dependencies, how often would redundant builds happen currently (e.g., is it every time you run `stack test`)?  Also, can your multi-test-suite repo be adapted to an integration test?

mgsloan:
I've been using this branch for a while and only found one issue with it, which turned out to be an existing issue (https://github.com/commercialhaskell/stack/issues/1305#issuecomment-154818531).  So, here goes nothing, pulling the trigger on the merge.

Yes, multi-test-suite should probably become an integration test, good idea!  I've added that to my TODO list.

Issue #1161, opened 2015-10-14
Forwarding DYLD_LIBRARY_PATH


bitemyapp:
https://www.reddit.com/r/haskell/comments/3ooxu4/library_not_loaded_libmariadb2dylib_os_x/

Ah, I got the solution myself. It seems that I needed to do a

```
export DYLD_LIBRARY_PATH=/usr/local/lib/mariadb/
```

to add it to the search paths for dylds and it seems stack isn't forwarding env variables like `DYLD_LIBRARY_PATH`. I noticed because I tried to run the hsc2hs generated executable manually and it worked as expected.

Is this intended behaviour for stack?

snoyberg:
Just saw this issue after responding on Reddit. I'm not aware of anything Stack is doing to make `DYLD_LIBRARY_PATH` be handled specially, perhaps it's the OS itself doing something funny?

alexbiehl:
```
Spawning children processes of processes restricted by System Integrity Protection, such as by
launching a helper process in a bundle with NSTask or calling the exec(2) command, resets the Mach
special ports of that child process. Any dynamic linker (dyld) environment variables, such as
DYLD_LIBRARY_PATH, are purged when launching protected processes.
```
Source: https://forums.developer.apple.com/thread/9233

snoyberg:
Well, that explains that. If someone has an idea for how to resolve this issue, please send a PR, but my lack of OS X experience will prevent me from getting involved.

borsboom:
Is this specific to Stack, or does the same thing happen with cabal-install?  I just wonder if a workaround is even possible, because it would presumably require getting the subprocess to set `DYLD_LIBRARY_PATH` itself.  But if cabal-install doesn't have this problem then it should be fixable in Stack.  Could it be related to the fact that `ghc` is actually a shell script wrapper rather than a binary?

I'm not a Mac developer, just a Un*x developer that happens run OS X on the desktop, so I don't have much understanding of System Integrity Protection at this point.  I'm avoiding upgrading to El Capitan because I don't want to break my own development environment, but probably I should bite the bullet so I can look at these issues...

drbo:
Sounds like the only solution is to avoid going through a shell script wrapper, or distributing a custom shell.

alexbiehl:
I can have a look at it when I'm back home. The idea is: 

[...]

Issue #911, opened 2015-09-03
Docker-built binaries should live in a different path


rrnewton:
In the same sense that we separate binaries based on architecture, shouldn't we separate them based on whether they are built by a docker image?

When I built my first docker-based executable, it landed in a directory like this:

```
Linking .stack-work/dist/x86_64-linux/Cabal-1.18.1.5/build/docker-hello-world/docker-hello-world ...
.stack-work/install/x86_64-linux/lts-2.22/7.8.4.20141229/bin/
Installing executable(s) in
/tmp/docker_hello_world/.stack-work/install/x86_64-linux/lts-2.22/7.8.4.20141229/bin
```

This happens to be a different directory than if I build without docker, but only because the GHC version happens to be `7.8.4.20141229`.  

The above path doesn't give an indication that it was built for docker and in fact may not run at all on the host platform!  (It may depend on dynamic libraries inside the docker image the selected, right?)  And in principle the GHC version used in the docker build could be the same as in the non-docker build causing these two *different* kinds of files to collide, no?

rvion:
+1

borsboom:
Is this something you've experienced difficulty with?  In most cases, the binaries should be compatible since the directory they go into is based on the GHC bindist, so if your host system needs libgmp.so.3 but the Docker image uses libgmp.so.10, they'll end up in different directories (a `-gmp4` suffix for the former).  I've been able to switch between Docker/non-Docker with success in the past, although it's been a while since I tried that.

That said, what you suggest makes sense.  Would have to think on exactly how to construct the directory, though.  Using like the Docker image ID wouldn't be great because then if you switch to a slightly different Docker image (say, an extra layer that adds a distro package) you don't want it to throw out everything it's built already.  Perhaps a label value can be attached to images that'll be used as part of the path, that way image creators can control it.

borsboom:
This should also ensure that ~/.stack/setup-exe-cache binaries are not shared (as reported in #1367).

PR #1285, opened 2015-11-04
Support reproducible builds via Nix

YPares:
This pull request implements the capability for stack to launch the build inside a nix-shell, in order to specify dependencies to libraries out of Hackage that exist in Nixpkgs.

Simply modify your stack.yaml so it contains for instance:

```yaml
nix-shell:
    enable: false
    packages: [glpk]
    #shell-file: shell.nix
```

Build this project with `stack --nix build`. It will open a Nix-shell that will first download the `glpk` dependency, and then launch the build. Of course, this requires `nix` to be installed on your system, and it requires your `stack.yaml` to depend on an LTS resolver that has been mirrored in the nix packages (see for instance in https://github.com/NixOS/nixpkgs/tree/master/pkgs/development/haskell-modules the `configuration-lts-X.Y.nix` files).
If you know nix, you can also specify a shell.nix that creates a derivation that will be used to set the build environment. You cannot specify `packages` and `shell-file` at the same time!
For now, this works much like the docker backend, by relauching the same stack executable inside a new environment. It requires GHC to be installed by Nix.

It's not possible to use `stack ghci`, because GHCi fails to find the extra libraries, but this is not specific to stack, as the command:
`nix-shell --pure -p ghc extra-libs --command "stack ghci"`
without enabling nix support in stack already fails.

Theoretically, activating both Nix and Docker could be possible (resulting in a nix-shell running inside a container), but that requires a Docker image providing Nix, of course.


YPares:
In some cases, getting a fully working build environment can be a tedious process, especially if you're working in a team on a project with lots of dependencies inside and outside Hackage and need to get everybody up to speed and able to build.
Compared to containerization tools like Docker, Nix is much more lightweight. It does not require you to download full images. Nix is also much better in terms of sharing binaries and libraries between various projects and environments. The unit of granulatity in Docker is the FS layer. In Nix, every derivation (the Nix term for "package") can be shared between several projects definitions, so copying is limited to a minimum. Also, the lazy nature of Nix keeps the number of commands to issue to get a working environment very low. For instance a simple nix-shell -p ghc cabal-install haskellPackages.hasktags downloads everything needed to fire up a bash giving you a mini Haskell development environment, whatever libraries are already present on your system.

YPares:
[Disregard initial draft for motivations. Here they are below.]

The motivation is the following:

* Docker works well under Linux, but is not a cross platform solution. We wanted a cross platform, lightweight solution for repeatable builds beyond Linux for projects that depend on specific system libraries being present on the system (math libraries, R, GUI libraries, etc).
* Nix works well as a cross platform solution for repeatable builds: all system dependencies are installed locally and only visible within the scope of a project, without conflicting with any other global resources installed on the system.
* Nix works on OS X as much as it does on Linux. Possibly also on Windows.
* Docker provides better isolation than Nix, but Nix is more lightweight and cross platform. In this sense they are two complementary isolation mechanisms for two different classes of use cases: Docker is useful for projects that are deployed on a single platform but where strong isolation guarantees are required. Nix provides far weaker isolation guarantees, but does not require special kernel support. Nix is also appropriate for memory/storage constrained environments, due to better sharing between instances (or "containers" in docker parlance).

borsboom:
@YPares I'm giving this a try, but hit this error right off the bat:

```
$ stack -v --nix build
Version 0.1.7.0, Git revision 83d87796c50dc1516d26e910c62f701509ac681c (dirty) (2507 commits) x86_64
2015-11-09 06:53:48.571245: [debug] Checking for project config at: /Users/manny/fpco/tmp/stack1188x/stack.yaml @(stack_5HhO82aVZM8JgYWESH4I9g:Stack.Config src/Stack/Config.hs:553:9)
2015-11-09 06:53:48.577096: [debug] Loading project config file stack.yaml @(stack_5HhO82aVZM8JgYWESH4I9g:Stack.Config src/Stack/Config.hs:576:13)
2015-11-09 06:53:48.578503: [debug] Trying to decode /Users/manny/.stack/build-plan-cache/x86_64-osx/nightly-2015-10-15.cache @(stack_5HhO82aVZM8JgYWESH4I9g:Data.Binary.VersionTagged src/Data/Binary/VersionTagged.hs:55:5)
2015-11-09 06:53:48.588051: [debug] Success decoding /Users/manny/.stack/build-plan-cache/x86_64-osx/nightly-2015-10-15.cache @(stack_5HhO82aVZM8JgYWESH4I9g:Data.Binary.VersionTagged src/Data/Binary/VersionTagged.hs:64:13)
2015-11-09 06:53:48.588347: [debug] Run process: ghc --info @(stack_5HhO82aVZM8JgYWESH4I9g:System.Process.Read src/System/Process/Read.hs:272:3)
2015-11-09 06:53:48.686597: [debug] Run process: ghc --numeric-version @(stack_5HhO82aVZM8JgYWESH4I9g:System.Process.Read src/System/Process/Read.hs:272:3)
2015-11-09 06:53:48.713455: [debug] Run process: ghc-pkg --no-user-package-db field --simple-output Cabal version @(stack_5HhO82aVZM8JgYWESH4I9g:System.Process.Read src/System/Process/Read.hs:272:3)
2015-11-09 06:53:48.773741: [debug] Run process: ghc-pkg --no-user-package-db list --global @(stack_5HhO82aVZM8JgYWESH4I9g:System.Process.Read src/System/Process/Read.hs:272:3)
2015-11-09 06:53:48.798502: [debug] Using a nix-shell environment with nix packages: ghc, gnused, coreutils, glibcLocales @(stack_5HhO82aVZM8JgYWESH4I9g:Stack.Nix src/Stack/Nix.hs:127:6)
2015-11-09 06:53:48.798934: [debug] Run process: /Users/manny/.nix-profile/bin/nix-shell --pure -p ghc gnused coreutils glibcLocales --command "export STACK_IN_NIXSHELL=1 ; export LD_LIBRARY_PATH=`echo -n $NIX_LDFLAGS | tr ' ' $'\n' | sed -n '/-L/{s/-L//; p}' | tr $'\n' ':'` ; /Users/manny/.local/bin/stack --internal-re-exec-version=0.1.7.0 -v --nix build" @(stack_5HhO82aVZM8JgYWESH4I9g:System.Process.Run src/System/Process/Run.hs:92:5)
Version 0.1.7.0, Git revision 83d87796c50dc1516d26e910c62f701509ac681c (dirty) (2507 commits) x86_64
security: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

Any ideas?  Note: I just installed Nix on my Mac for the first time, so I know nothing about it beyond having run through the commands in the [quickstart](http://nixos.org/nix/manual/#chap-quick-start).

YPares:
Hello,
Yes, I also noticed it (I'm not a regular OSX user). It happens only on
OSX. I'm still trying to figure out why...

2015-11-09 15:59 GMT+01:00 Emanuel Borsboom <notifications@github.com>:

borsboom:
Ok, it's working better on Linux.  Next problem I ran into: if `ghc` is not on the PATH or installed in `~/.stack/programs`, this happens:

```
$ stack --verbose --nix build
Version 0.1.7.0, Git revision 83d87796c50dc1516d26e910c62f701509ac681c (2507 commits) x86_64
2015-11-09 17:20:41.884074: [debug] Checking for project config at: /home/vagrant/tmp/nix/stack.yaml @(stack_FHzONQT9HI3ATX2xQH198L:Stack.Config src/Stack/Config.hs:553:9)
2015-11-09 17:20:41.884437: [debug] Loading project config file stack.yaml @(stack_FHzONQT9HI3ATX2xQH198L:Stack.Config src/Stack/Config.hs:576:13)
2015-11-09 17:20:41.885223: [debug] Run process: ldd /home/vagrant/.local/bin/stack @(stack_FHzONQT9HI3ATX2xQH198L:System.Process.Read src/System/Process/Read.hs:272:3)
2015-11-09 17:20:41.895325: [debug] Trying to decode /home/vagrant/tmp/nix/.stack/build-plan-cache/x86_64-linux/lts-3.7.cache @(stack_FHzONQT9HI3ATX2xQH198L:Data.Binary.VersionTagged src/Data/Binary/VersionTagged.hs:55:5)
2015-11-09 17:20:41.900657: [debug] Success decoding /home/vagrant/tmp/nix/.stack/build-plan-cache/x86_64-linux/lts-3.7.cache @(stack_FHzONQT9HI3ATX2xQH198L:Data.Binary.VersionTagged src/Data/Binary/VersionTagged.hs:64:13)
No compiler found, expected minor version match with ghc-7.10.2 (x86_64) (based on resolver setting in /home/vagrant/tmp/nix/stack.yaml).
Try running "stack setup" to install the correct GHC into /home/vagrant/tmp/nix/.stack/programs/x86_64-linux/
```

As I understand it, `stack --nix` shouldn't require any GHC installed outside of the Nix environment it creates, correct?  I'd guess the cause is likely that it's loading the BuildConfig before re-running itself in the nix-shell, and loading the BuildConfig requires GHC to exist.

YPares:
@borsboom Yes, you're right. We load the BuildConfig to know which resolver is being used, to get GHC from the mirrored resolver in the nixpkgs.
I didn't know loading the BuildConfig required a GHC to work.
But this behaviour came from a problem we had when getting GHC able to access system libraries. On Ubuntu, it works if GHC is provided by Nix, but I discovered that on OSX there's no way around explicitely passing every dir to --extra-{lib,include}-dirs. Normally this should end the workaround and allow stack to work with its own GHC even inside a Nix-shell. So GHC would stop to be added as a Nix dependency.

borsboom:
You shouldn't need the BuildConfig to determine the resolver, since that's loaded in the Project config, which loaded by `loadConfig`.

YPares:
@borsboom So the recent commits fix the following:
- Stack no longer load BuildConfig, so it builds with --nix even when there is no local GHC installed
- We explicitely pass every dependency from nix to the inner stack with --extra-{lib,include}-dirs. This is necessary on OSX.
- It builds our test project (https://github.com/YPares/test-stack-nix) on OSX alright, but for now the shell is no longer pure (or else the command 'security' cannot be found).

Note that stack manual hasn't been updated to reflect this new feature.

YPares:
@borsboom So I resolved the conflicts with the master branch, if you want to give it a new try.

YPares:
@borsboom OK. So the import for Data.Monoid is required by 7.8 but not by 7.10. Should I solve it with a CPP pragma to test which version of GHC is being used?
I don't see this already being done elsewhere in Stack codebase.

YPares:
Ok. There's still a warning on GHC 7.8, but it's from `Network/HTTP/Download/VerifiedSpec.hs`, which I haven't modified.

YPares:
We're being bit by this issue ( https://github.com/commercialhaskell/stack/issues/1161 ) on OSX when trying to build something that uses `hsc2hs`. @borsboom when setting DYLD_LIBRARY_PATH it works (hsc2hs doesn't complain about not finding libraries). So do you think we should add it explicitely or wait for the aforementioned issue to be fixed?

borsboom:
I think #1161 is a different problem, because if setting DYLD_LIBRARY_PATH is working, that indicates you're not running into the #1161 (which is about setting DYLD_LIBRARY_PATH _not_ helping).  Are you El Capitan, by any chance?

It does seem like explicitly setting DYLD_LIBRARY_PATH for your case would make sense but, based on my somewhat limited understanding, it may not work on El Capitan.

src/Stack/Nix.hs
```diff
@@ -0,0 +1,180 @@
...
+    :: M env m
+    => m (FilePath,[String])
+    -> IO ()
+    -> m ()
+execWithOptionalShell getCmdArgs inner =
+  do config <- asks getConfig
+     inShell <- getInShell
+     isReExec <- asks getReExec
+     if | inShell && not isReExec ->
+            throwM OnlyOnHostException
```
borsboom: I don't think Docker's exception for this should be re-used.  Better to define a new one specific to Nix with a more accurate error messages.

Also, am I misreading this, or is it indicating that you can't re-run `stack` within a `stack exec` shell (e.g. `stack --nix exec bash` followed by `stack build` in the sub-shell isn't allowed)?  If so, why is that?

src/Stack/Nix.hs
```diff
@@ -0,0 +1,180 @@
...
+                                                         ," --extra-include-dirs=", "${"++p++"}/include"])
+                                           pkgsInConfig), ["' ;"
+                               ,"'';"
+                               ,"} \"\""]]]
+                    -- glibcLocales is necessary on Linux to avoid warnings about GHC being incapable to set the locale.
+         fullArgs = concat [ -- ["--pure"],
+                            map T.unpack (nixShellOptions (configNix config))
+                           ,nixopts
+                           ,["--command", ("export " ++ inShellEnvVar ++ "=1 ; ")
+                                          ++ intercalate " " (cmnd:args)
```
borsboom: This looks like it may not handle arguments with spaces or shell-interpreted characters properly.

src/Stack/Nix.hs
```diff
@@ -0,0 +1,180 @@
...
+                                          ++ " $STACK_IN_NIX_EXTRA_ARGS"]
+                           ]
+     $logDebug $ T.pack $
+         "Using a nix-shell environment " ++ (case mshellFile of
+            Just filePath -> "from file: " ++ filePath
+            Nothing -> "with nix packages: " ++ (intercalate ", " pkgsInConfig))
+     e <- try (callProcess'
```
borsboom: Could you use `Stack.Exec.exec` instead?  It doesn't look like this needs to do anything after the nix-shell exits, so replacing the current process would be better than spawning a sub-process.

src/Stack/Nix.hs
```diff
@@ -0,0 +1,180 @@
...
+                 (if isTerm then id else \cp -> cp { delegate_ctlc = False })
```
borsboom: I doubt the extra logic to prevent delegating Ctrl-C is necessary in this case.  For Docker, it's there because of some specific weird stuff that `docker run` and `docker start` do with container terminals, but this nix-shell looks like a regular process that uses the existing `std*` file handles and TTY normally.

src/Stack/Options.hs
```diff
@@ -308,6 +312,25 @@ configOptsParser hide0 =
...
+nixOptsParser :: Bool -> Parser NixOptsMonoid
+nixOptsParser showOptions =
+  NixOptsMonoid
+  <$> pure False
+  <*> maybeBoolFlags nixCmdName
+                     "using a Nix-shell"
+                     hide
+  <*> pure []
+  <*> pure Nothing
+  <*> many (textOption (long "nix-shell-options" <>
```
borsboom: Use `Options.Applicative.Args.argsOption` instead of `many . textOption`, which will handle options within the value properly.

src/Stack/Options.hs
```diff
@@ -308,6 +312,25 @@ configOptsParser hide0 =
...
+                        metavar "OPTION" <>
+                        help "Additional options passed to nix-shell" <>
+                        hide))
+  where
+    hide = if showOptions
```
borsboom: Use `hideOpts` here.

src/main/Main.hs
```diff
@@ -865,7 +880,10 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
                     -- Unlock before transferring control away, whether using docker or not:
                     (Just $ munlockFile lk)
                     (runStackTGlobal manager (lcConfig lc) go $
-                        exec plainEnvSettings cmd args)
+                        Nix.execWithOptionalShell
```
borsboom: You might be able to use `reexecWithOptionalShell` here.  The Docker case has switched to always using `reexecWithOptionalContainer` (even with `stack exec --plain`) and you can probably do the same.  After switching, that may open the opportunity to simplify `Nix.rexecWithOptionalShell` and `Nix.execWithOptionalShell` a bit.

borsboom:
When I tried this with a resolver that hadn't been mirrored to nix yet, I got the following error:

    error: attribute ‘lts-3_14’ missing, at (string):1:163
    (use ‘--show-trace’ to show detailed location information)

After reading the docs I understood what this meant, but is it possible to make this more user-friendly?  It would be nice if the error at least indicates why this would happen and perhaps suggests running `nix-channel --update` (if that makes sense).  Perhaps it could even run the update automatically and retry, kind of like how Stack will download a snapshot automatically the first time it's encountered (not sure if Nix users would like automatic update though; I defer to your judgement about whether this is a good idea).

borsboom:
Thanks for all the work on this, it looks really close to being ready to merge!  Don't forget to add a changelog entry.  It should go in the major changes section; this deserves lots of attention!

src/Stack/Nix.hs
YPares: Actually it's possible to run `stack build` inside a `stack --nix exec bash`. Although nix doesn't like when a sub-bash is launched within its shell (mangled output, arrow keys not working...)

src/Stack/Nix.hs
YPares: You mean the kind of arguments that could be passed after `stack exec -- `?

src/Stack/Nix.hs
YPares: Okay, fixed.

src/Stack/Nix.hs
YPares: Fixed along with previous comment.

YPares:
@borsboom I pushed a new commit addressing your comments.

Although, regarding the missing attribute error, to provide a more informative error message I would need to ask nix about which attributes are available, which is not easy to automate as I pointed out in the manual.

The real fix would be to stop to depend on a GHC provided by Nix and use a GHC downloaded by Stack, but for now on Linux this results in failure to find libs in the nix store.

borsboom:
Fair enough, I don't think the error message should hold up merging this, although it would be nice to improve at some point.  I'm not surprised that trying to use a non-Nix GHC causes problems, so leaving that as-is is reasonable.

Do you think this may end up with similar problems as Docker in #911 and #1367?  I'm beginning work on a fix for those two now, and most likely the same approach can be used for Nix as well.

src/Stack/Nix.hs
```diff
@@ -0,0 +1,146 @@
...
+                           ,nixopts
+                           ,["--command", intercalate " " (map (\a -> "'"++a++"'") (cmnd:args))
+                                          ++ " $STACK_IN_NIX_EXTRA_ARGS"]
+                           ]
+     $logDebug $
+         "Using a nix-shell environment " <> (case mshellFile of
+            Just filePath -> "from file: " <> (T.pack filePath)
+            Nothing -> "with nix packages: " <> (T.intercalate ", " pkgsInConfig))
+     e <- try (exec
+                 (EnvSettings {esIncludeLocals = False
```
borsboom: You can use `Stack.Config.Types.minimalEnvSettings` instead.

src/Stack/Nix.hs
```diff
@@ -0,0 +1,146 @@
...
+                               ,      (map (\p -> T.concat
+                                                  ["--extra-lib-dirs=${",p,"}/lib"
+                                                  ," --extra-include-dirs=${",p,"}/include "])
+                                           pkgsInConfig), ["'' ;"
+                               ,"} \"\""]]]
+                    -- glibcLocales is necessary on Linux to avoid warnings about GHC being incapable to set the locale.
+         fullArgs = concat [ -- ["--pure"],
+                            map T.unpack (nixShellOptions (configNix config))
+                           ,nixopts
+                           ,["--command", intercalate " " (map (\a -> "'"++a++"'") (cmnd:args))
```
borsboom: Surrounding by `'` isn't quite enough either; what if the value contain a `'` itself?  A reliable way to quote all characters, at least for bourne shell derivatives, is to surround the values in `'`s (like you did) and also replace all occurrences of `'` within each value with `'"'"'`.  Example code in Propellor's [shellEscape](http://hackage.haskell.org/package/propellor-2.14.0/docs/Propellor-Property-Cmd.html#v:shellEscape).

Of course it would be much nicer if `nix-shell` gave you a way to do this without having to escape it at all, but I guess that's not possible.

src/Stack/Nix.hs
borsboom: Ok, then I'm thinking it would be better to put the test back, since I'd rather have something clearly fail than exhibit unexpected behaviour.

This shouldn't hold up the merge, but a way around this might be be to have Stack check for the STACK_IN_NIX_EXTRA_ARGS environment variable at startup and implicitly insert its contents in front of the command-line arguments (using `withArgs`), instead of passing the value of $STACK_IN_NIX_EXTRA_ARGS to the in-nix-shell `stack` command explicitly.  This could get a bit hairy though so I'd rather defer tacking this unless we need to.

YPares:
@borsboom I corrected the escaping of the command line and re-merged the master.
For coherence's sake, I also made a slight adjustment: the section in the yaml is just called `nix:`, now (no longer `nix-shell:`).
That's because the CL option was already `--nix` and because soon nix will adopt a git-like CL interface. I.e. the `nix-shell` tool will become just another subcommand of a big `nix` command.

borsboom:
LGTM, thanks for this awesome new feature!

mgsloan:
Awesome :D

I think this warrants a `Major changes` entry in the ChangeLog

YPares:
@mgsloan @borsboom Thanks guys!

YPares:
I will publish a blog post about this feature shortly.

codygman:
@YPares looking forward to it, will you post it back here?

borsboom:
http://www.tweag.io/blog/stack-nix-portable-reproducible-builds

PR #1298, opened 2015-11-06
Show package suggestions at the end of output

mrkkrp:
Not sure whether `UnknownPackageNames` should have been extended too. Tell me if something can be improved.


mrkkrp:
Here I added commit to avoid using `show` on package names (see commit message). This can be removed or squashed into the previous one depending on your opinion.

mgsloan:
LGTM, thanks!!

PR #1310, opened 2015-11-08
Simplify a part of conduitDumpPackage a little bit

DNNX:
`fromMaybe x (Map.lookup y z)` is equivalent to `Map.findWithDefault x y z`


mgsloan:
Thanks!

PR #1318, opened 2015-11-09
Refactorings using mapMaybeA and forMaybeA

sjakobi:


mrkkrp:
Note that won't work with GHC 7.8 because applicative is not superclass of monad there. See the log.

borsboom:
We do still need to support building Stack on GHC 7.8.

rvion:
@borsboom Sorry in advance if this has been discussed before, but should we really support stack on GHC 7.8 ?
I ask the question because stack is not really a library one will depend on,

sjakobi:
Thanks for the heads up regarding GHC 7.8!

I've force-pushed a fix.

sjakobi:
Thanks for the heads up regarding GHC 7.8!

I've force-pushed a fix.

borsboom:
Thanks!

PR #1319, opened 2015-11-09
Some refactorings in Stack.Build.Haddock

sjakobi:
Among other things:

* Fix a comment that breaks haddocks generation for stack

* Rename the `generate*HaddockIndex` functions to `generate*Haddocks` as they generate both the index as well as the actual haddocks

### DIFF
--- a/src/Stack/Build/Haddock.hs
+++ b/src/Stack/Build/Haddock.hs
@@ -34,11 +34,11 @@ import           Data.Set                       (Set)
 import qualified Data.Set                       as Set
 import           Data.Text                      (Text)
 import qualified Data.Text                      as T
+import           Data.Time                      (UTCTime)
 import           Path
 import           Path.Extra
 import           Path.IO
 import           Prelude
-import           Safe                           (maximumMay)
 import           Stack.Types.Build
 import           Stack.PackageDump
 import           Stack.Types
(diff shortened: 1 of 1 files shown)


borsboom:
I don't think `generate*HaddockIndex` should be renamed, as it actually doesn't generate any of the haddocks themselves, it only creates the index and contents and puts dependencies' haddocks in the right place.  Stack.Build.Execute.singleBuild is where the haddocks are generated.

sjakobi:
Thanks for clearing that up, @borsboom!

src/Stack/Build/Haddock.hs
```diff
@@ -209,16 +208,17 @@ generateHaddockIndex descr envOverride wc dumpPackages docRelFP destDir = do
                                 , srcInterfaceModTime
                                 , srcInterfaceAbsFile
                                 , destInterfaceAbsFile )
+    tryGetModificationTime :: FilePath -> IO (Either () UTCTime)
+    tryGetModificationTime = tryJust (guard . isDoesNotExistError) . getModificationTime
+    copyPkgDocs :: (a, UTCTime, Path Abs File, Path Abs File) -> IO ()
     copyPkgDocs (_,srcInterfaceModTime,srcInterfaceAbsFile,destInterfaceAbsFile) = do
-        -- | Copy dependencies' haddocks to documentation directory.  This way, relative @../$pkg-$ver@
+        -- Copy dependencies' haddocks to documentation directory.  This way, relative @../$pkg-$ver@
```
sjakobi: In order to avoid problems like this one, would it be a good idea to let Travis build the haddocks too?

borsboom: Haddocks are built as part of the integration tests, but those are run less frequently to try to keep a decent turnaround time for the travis tests.

PR #1320, opened 2015-11-09
Enhance haddocks for Stack.Build.Execute

sjakobi:


sjakobi:
[`withSingleContext`](https://github.com/commercialhaskell/stack/pull/1320/files#diff-890614cc04afa326f5d48aac897e4aa8R672) still lacks documentation too, but I couldn't easily tell what it does.
Maybe somebody else can help?

borsboom:
Regarding `withSingleContext`, I guess you could say it sets up the environment in which an action executes, or something like that.

PR #1326, opened 2015-11-10
Some housekeeping in Stack.Types.Config

sjakobi:


sjakobi:
Should the "command-specific" types for eval, exec and setup possibly be moved to their own `Stack.Types.X` modules?

drwebb:
We are hoping to bring together config commands into one place so we can improve the entire config system, so distributing the types out to their different modules probably isn't the best idea. If anything we might want to refactor the command specific types to a submodule of `Stack.Types.Config`, like `Stack.Types.Config.Setup` for example.

sjakobi:
> We are hoping to bring together config commands into one place so we can improve the entire config system

Ok, that sounds very reasonable! Looking forward to the results!

Issue #1183, opened 2015-10-17
stack check - verify dist tarball


wereHamster:
Cabal has a command `check` which I used to verify the tarball before uploading it to hackage. Stack doesn't have such a command, and neither `sdist` nor `upload` indicate that they would run any such checks.

borsboom:
Seems like that would be nice to add.  Perhaps `stack sdist --check` and `stack upload --check` options?  A pull request would be welcome.

mrkkrp:
Is anyone working on this? I could try to implement it. I like the idea of `stack sdist --check` (print results of checking when creating a tarball), but how `stack upload --check` should work? Should it reject uploading if there are warnings or even worse upload package silently even if there are warnings (that's not what people would want, I guess)? Not clear.

mgsloan:
@mrkkrp I don't think anyone is working on this, feel free to have at it!

I think both `stack sdist --check` and `stack upload --check` should refuse to perform their action if build + tests fail.  In order to reject uploading when there are warnings, we can use the `--pedantic` flag.

However, this brings up a tricky part of this - should `sdist` and `upload` get all of the flags that can be passed to `build`?  Otherwise, you won't be able to customize the build and test.  Other commands like `ghci`, which involve a build step, are likely to accept build flags in the future (https://github.com/commercialhaskell/stack/issues/1186).

However, it seems weird to have a bunch of flags in the docs which are only relevant when `--check` is passed.  Maybe the way to go is have `--check` and `--check-with "--pedantic"` - the latter form takes a list of arguments passed to the build and test steps.

Good news is that we don't have to figure out the full details of the build/test flags in order to implement  the main functionality!

mgsloan:
See the [haddocks for Distribution.PackageDescription.Check](https://hackage.haskell.org/package/Cabal-1.22.4.0/docs/Distribution-PackageDescription-Check.html):

> This code gets uses when configuring a package when we consider only basic problems. The higher standard is uses when when preparing a source tarball and by Hackage when uploading new packages. 

I'm not sure what exactly this "higher standard" entails - does it mean that any `PackageCheck` result in the package is rejected?

If hackage is always running these checks, then we may as well have `stack upload` always run them too.  So, I think this shouldn't even require an extra `--check` flag, and perhaps that flag can instead be used for #1350.

mrkkrp:
[...]

Issue #717, opened 2015-08-05
Run tests on sdist tarballs with `stack sdist` and `stack upload`


mgsloan:
Currently, in order to test that all the necessary files are in a source distribution tarball, you need to manually try building it.   This feature has precedent in [cabal-src](https://github.com/yesodweb/cabal-src) and possibly elsewhere.

I'm thinking something like `stack sdist --test` to both generate and test the tarballs.  It'd also be nice to have `stack upload --test`, which would test the tarballs, and upload if they pass.

The first step is to extract the tarballs into a directory, and the next step is to build + test the code with the current project's stack.yaml configuration.

I'm not sure what the ideal implementation of this would look like.  One possibility would be to have the extraction directory mimic the structure of the mega-repo, and then copy the stack.yaml into it.  This would be rather costly in rebuild time and disk space, though.

For my own testing, I've just modified my `stack.yaml` to instead point at the extracted directories.  This suggests an implementation approach where `test` is run, but with portions of the packages list substituted.  This approach's downside is that it stomps on your old local install root.  I don't see this as a very large downside.

mgsloan:
Note: I'm thinking that building and testing the sdists is the correct default.  While it's a change in behavior, we should default to best-practices, and have a flag to turn this behavior.  This would also make things like https://github.com/commercialhaskell/stack/issues/1568 reasonable, as it's more likely that the sdist will actually be tested against its full range of deps.

seagreen:
>  I'm thinking that building and testing the sdists is the correct default.

I agree. When writing a library sdist is the only thing one cares about. Application authors might get annoyed at having to keep their cabal files up-to-date, but I think correctness is more important than convenience here.

Let me know if you start working on this, I'd be happy to help test it or help with some of the work.

In the meantime perhaps the [travis-with-caching](http://docs.haskellstack.org/en/stable/GUIDE/#travis-with-caching) script should be updated to test the sdists? If I'm reading it right it doesn't right now.

seagreen:
FYI I just found a nixpkgs problem this would have prevented: https://github.com/NixOS/nixpkgs/issues/15935

Though come to think of it I've written tests before that required checking out a git submodule before being run (also breaking on nixpkgs) so it's easy to do.

mgsloan:
> Let me know if you start working on this, I'd be happy to help test it or help with some of the work.

Feel free to implement this, it should be pretty straightforward.  Let me know if you need any pointers.

sjakobi:
[...]

PR #1334, opened 2015-11-11
Addition of ‘--check’ option for ‘upload’ and ‘sdist’

mrkkrp:
See #1183. Here I've added `--check` switch. Previously simple tuple was used to hold options for `upload` and `sdist` commands. Now with the addition of `--check` option, I switched to `(,,)`. Is it OK? Should some data structure be defined instead of tuple? This is very beginning of my work, so I create this PR early to ask questions along the way.

### DIFF
--- a/src/Stack/SDist.hs
+++ b/src/Stack/SDist.hs
@@ -62,10 +62,11 @@ type M env m = (MonadIO m,MonadReader env m,HasHttpManager env,MonadLogger m,Mon
 -- While this yields a 'FilePath', the name of the tarball, this
 -- tarball is not written to the disk and instead yielded as a lazy
 -- bytestring.
-getSDistTarball :: M env m
-                => Maybe PvpBounds -- ^ override Config value
-                -> Path Abs Dir
-                -> m (FilePath, L.ByteString)
+getSDistTarball
+  :: M env m
+  => Maybe PvpBounds            -- ^ Override Config value
+  -> Path Abs Dir               -- ^ Path to local package
+  -> m (FilePath, L.ByteString) -- ^ Filename and tarball contents
 getSDistTarball mpvpBounds pkgDir = do
     config <- asks getConfig
     let pvpBounds = fromMaybe (configPvpBounds config) mpvpBounds
@@ -149,7 +150,7 @@ gtraverseT f =
                  Nothing -> gtraverseT f x
                  Just b  -> fromMaybe x (cast (f b)))
 
--- Read in a 'LocalPackage' config.  This makes some default decisions
+-- | Read in a 'LocalPackage' config.  This makes some default decisions
 -- about 'LocalPackage' fields that might not be appropriate for other
 -- usecases.
 --
@@ -219,8 +220,8 @@ getSDistFileList lp =
 
 normalizeTarballPaths :: M env m => [FilePath] -> m [FilePath]
 normalizeTarballPaths fps = do
-    --TODO: consider whether erroring out is better - otherwise the
-    --user might upload an incomplete tar?
+    -- TODO: consider whether erroring out is better - otherwise the
+    -- user might upload an incomplete tar?
     unless (null outsideDir) $
         $logWarn $ T.concat
             [ "Warning: These files are outside of the package directory, and will be omitted from the tarball: "
--- a/src/Stack/SDist.hs
+++ b/src/Stack/SDist.hs
@@ -4,9 +4,12 @@
 {-# LANGUAGE RankNTypes            #-}
 {-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE ViewPatterns          #-}
+{-# LANGUAGE DeriveDataTypeable    #-}
 -- Create a source distribution tarball
 module Stack.SDist
     ( getSDistTarball
+    , checkSDistTarball
+    , checkSDistTarball'
     ) where
 
 import qualified Codec.Archive.Tar as Tar
@@ -14,8 +17,8 @@ import qualified Codec.Archive.Tar.Entry as Tar
 import qualified Codec.Compression.GZip as GZip
 import           Control.Applicative
 import           Control.Concurrent.Execute (ActionContext(..))
-import           Control.Monad (unless, void)
-import           Control.Monad.Catch (MonadMask)
+import           Control.Monad (unless, void, liftM)
+import           Control.Monad.Catch
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
 import           Control.Monad.Reader (MonadReader, asks)
@@ -26,6 +29,8 @@ import qualified Data.ByteString.Lazy as L
 import           Data.Data (Data, Typeable, cast, gmapT)
 import           Data.Either (partitionEithers)
 import           Data.List
+import           Data.List.NonEmpty (NonEmpty)
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (fromMaybe)
 import           Data.Monoid ((<>))
@@ -35,6 +40,9 @@ import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Encoding as TLE
 import           Data.Time.Clock.POSIX
 import           Distribution.Package (Dependency (..))
+import qualified Distribution.Package as P
+import qualified Distribution.PackageDescription as PD
+import qualified Distribution.PackageDescription.Check as Check
 import           Distribution.PackageDescription.PrettyPrint (showGenericPackageDescription)
 import           Distribution.Version (simplifyVersionRange, orLaterVersion, earlierVersion)
 import           Distribution.Version.Extra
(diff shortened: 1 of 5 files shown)


mrkkrp:
Is it better to throw exception to stop current action from checking function or from its caller (in `sdistCmd` or `uploadCmd` i.e. depending on results of check)? Checking function can be coded to return collection of `PackageCheck` things, which then can be analyzed. Another option is to print warnings inside of this function and throw exception if necessary. What is preferable?

mrkkrp:
I've done some progress, I guess. I do have more questions, without answers I won't be able to proceed (or it will take more time):

* how to get `PackageConfig`;
* how to get `PackageDescription`.

see `Stack.Check`.

Also please review what I already have.

mgsloan:
Thanks for working on this!  Sorry for the delayed feedback.

RE the first question, I think the check should just throw exceptions.  I figure that it may as well throw, since at that point, execution isn't going to continue beyond letting the user know how it failed.

What you've got looks pretty good so far.  It might be tricky to get "checkPackageContent" to work properly, though.  Instead of doing existence checks based on the working tree, they should be based on the tarball contents.  This should be possible with the in-memory representation of the sdist, but maybe it's easier to just unpack it.

Here's how to get `PackageConfig`: https://github.com/commercialhaskell/stack/blob/15dda17fb1b42d6961caea95a0b24c493758c04c/src/Stack/SDist.hs#L156

Note the TODO - ideally instead of copying this code yet a third time it would be put in a utility somewhere.

`PackageDescription` isn't as straightforward, since our datatypes extract all the data we need from the cabal datatypes, and don't store the cabal datatypes.  It's already getting computed when doing `readPackageDir`, which ends up invoking [resolvePackage](https://github.com/commercialhaskell/stack/blob/09991d98ed034e036668721de966c199e2e40642/src/Stack/Package.hs#L180) on a `gpkg` value.  I think `resolvePackageDescription packageConfig gpkg` is what you want.

Considering that `readPackageDir` isn't actually used anywhere else, I'd just rename it to `readPackageDescriptionDir`, and have it return a `PackageDescription`.

mrkkrp:
Thanks for the explanation, I think I will finish it in a couple of days now!

mrkkrp:
*(Rebased on current master.)* Please check this out. It works, but it may be kinda too sensitive (here I added `-O2` flag for testing):

```
~/projects/programs/haskell/stack $ stack sdist --check
Checking package 'stack' for common mistakes
Package check reported the following errors:
'ghc-options: -O2' is rarely needed. Check that it is giving a real benefit and not just imposing longer compile times on your users.
```

You see, this aborts the whole thing because of this. I don't know where should be the “threshold” where failed checks start to terminate current operation. `upload` is not currently affected. In `checkPackage` I used existing directory tree (uncompressed) because it's logical to check files before the whole thing goes into archive (also easier). Should be possible to do the same for `upload` (?)

I think I made the changes and refactoring you mentioned, see for yourself.

mgsloan:
Looking pretty good so far!  It looks like travis is failing build it due to a scope error, though.

Also, when reviewing the issue / PR initially, I misunderstood what the checks entail (I thought it was like #717) .  I think these checks would be a good thing to do by default.  How about instead adding a `--ignore-check` flag?  This would still do the check, but ignore if it fails.

Feel free to continue using exceptions for this (and catch them), but this would probably be a little cleaner if they're just returned.

Also, maybe this is getting a little nitpicky, I'll merge either way, but: for the most part we've tended towards rather large modules in stack.  How about just putting it in the SDist module?  It can be moved out if commands other than sdist / upload end up using it.

>  I don't know where should be the “threshold” where failed checks start to terminate current operation.

I'd say it should be the same threshold as hackage (I'm not sure what this is), and otherwise emit warnings.

> In checkPackage I used existing directory tree (uncompressed) because it's logical to check files before the whole thing goes into archive (also easier)

The issue with this is that the files in the existing directory tree aren't necessarily included in the archive.

> Should be possible to do the same for upload (?)

I'm definitely in favor of this!

So, in summary:

- [ ] Default to checking, have `--ignore-check` flag
- [ ] Return results instead of throwing?
- [ ] Move checkPackage into SDist?
- [ ] Same error threshold as hackage
- [ ] Check file existence based on existence within the tar
- [ ] Use it for upload

mrkkrp:
Since other functions work with paths outside of tar, I think I will temporarily unpack archive, check it, and delete temporary directory. Is this acceptable?

mgsloan:
Yup, that's acceptable!

Based on what I saw of cabal's API last time I looked, I think you might be able supply it with the "does this file exist" / "does this dir exist" predicates.

mrkkrp:
OK, here it is. Currently we decompress archives into temporary directories that are deleted after checking. What's worse, in `upload` command, if we are given directories, we get archives as lazy byte strings, which makes me temporarily save them *and only then unpack*. I don't like this at all.

I thought, maybe I could do everything in memory. We could use `gzip` library to get in-memory tar archive and then there is [interface](http://haddock.stackage.org/nightly-2015-11-17/tar-0.4.2.2/Codec-Archive-Tar-Entry.html) (although I've never worked with it) to get files from it (it's a bit hairy of course). This would mean rewriting of all helpers to work in this environment (because normally they work in file system), which I'm afraid I cannot do right now.

All in all, this should do the trick for now, tell me if there is something wrong with it.

src/Stack/Check.hs
```diff
@@ -0,0 +1,27 @@
+-- | Check package for common mistakes.
```
mgsloan: This file shouldn't get added

src/main/Main.hs
```diff
@@ -817,20 +820,23 @@ uploadCmd (args, mpvpBounds) go = do
...
-            liftIO $ forM_ files (canonicalizePath >=> Upload.upload uploader)
-            forM_ dirs $ \dir -> do
-                pkgDir <- parseAbsDir =<< liftIO (canonicalizePath dir)
-                (tarName, tarBytes) <- getSDistTarball mpvpBounds pkgDir
-                liftIO $ Upload.uploadBytes uploader tarName tarBytes
+    withBuildConfigAndLock go $ \_ -> do
+      uploader <- getUploader
+      unless ignoreCheck $
+        mapM_ (liftIO . canonicalizePath >=>
+               parseAbsFile              >=>
```
mgsloan: `liftIO . canonicalizePath >=> parseAbsFile` is the same as `Path.IO.parseRelAsAbsDir`

src/main/Main.hs
```diff
@@ -817,20 +820,23 @@ uploadCmd (args, mpvpBounds) go = do
...
+               checkSDistTarball)
+              files
+      liftIO $ forM_ files (canonicalizePath >=> Upload.upload uploader)
+      unless (null dirs) $
+        forM_ dirs $ \dir -> do
+          pkgDir <- parseAbsDir =<< liftIO (canonicalizePath dir)
```
mgsloan: Same here, this can also be `parseRelAsAbsDir`

mgsloan:
Great stuff, thanks for addressing all my points.  Here's all that's left: 

- [ ] Use `parseRelAsAbsDir`
- [ ] Remove `Stack/Check.hs`
- [ ] Use 4-space indentation - I'd actually merge it anyway since it's not like indentation is consistent throughout the codebase.  However, most code in stack does use 4 spaces.

mrkkrp:
@mgsloan, Done. I've come to using 2-space indentation almost everywhere (except for Python). Maybe it's because I've been writing Common and Emacs Lisps for a while in the past. Another consideration is when you have 2-space indentation you can put more code on a line (if you don't want to exceed 80 character width) and levels are still very visible, but that's a holy-war topic :-)

mgsloan:
LGTM, thanks!

src/Stack/SDist.hs
```diff
@@ -246,6 +260,61 @@ dirsFromFiles dirs = Set.toAscList (Set.delete "." results)
...
+checkSDistTarball tarball = withTempTarGzContents tarball $ \pkgDir' -> do
+    pkgDir  <- (pkgDir' </>) `liftM`
+        (parseRelDir . FP.takeBaseName . FP.takeBaseName . toFilePath $ tarball)
+    --               ^ drop ".tar"     ^ drop ".gz"
+    cabalfp <- getCabalFileName pkgDir
+    name    <- parsePackageNameFromFilePath cabalfp
+    config  <- getPackageConfig name
+    (gdesc, pkgDesc) <- readPackageDescriptionDir config pkgDir
+    $logInfo $ "Checking package '" <>
+               (T.pack . P.unPackageName . P.pkgName . PD.package $ pkgDesc) <>
```
sjakobi: The `unPackageName` accessor has existed only since Cabal-1.22.

I ran into this when I tried to build stack using the stack-7.8.yaml.

Maybe there should be an additional Travis build using the stack-7.8.yaml?

mgsloan: Good point!  I've resolved this with https://github.com/commercialhaskell/stack/commit/44da1c1bdfa3fb6d0f62a54f05859d80b0efd34b and opened https://github.com/commercialhaskell/stack/issues/1398

PR #1338, opened 2015-11-12
disallow package names with "words" consisting solely of numbers

### LINKED ISSUE #1336: "stack new" allows invalid package names
Repro:

1) stack new "1234-abc"

2) Fails to initialize a stack.yaml due to `Unable to parse cabal file /home/mgsloan/fpco/stack/1234-abc/1234-abc.cabal: NoParse "name" 1`

Looks like the [packageNameParser](https://github.com/fpco/stack/blob/24a461c950c5c8719ff427ab5a3b1494530b62e5/src/Stack/Types/PackageName.hs#L89https://github.com/fpco/stack/blob/24a461c950c5c8719ff427ab5a3b1494530b62e5/src/Stack/Types/PackageName.hs#L89) doesn't restrict package names enough.  From the [cabal users guide](https://www.haskell.org/cabal/users-guide/developing-packages.html#package-names-and-versions):

> A package is identified by a globally-unique package name, which consists of one or more alphanumeric words separated by hyphens. To avoid ambiguity, each of these words should contain at least one letter.

kadoban:
Fixes #1336, but now allows package names that cause a different error to happen when you try to build:

```
~/projects/test$ stack new "a-3o"
Downloading template "new-template" to create project "a-3o" in a-3o/ ...
[...]
Wrote project config to: [...]/projects/test/a-3o/stack.yaml
~/projects/test$ cd a-3o/
~/projects/test/a-3o$ stack build
Error parsing targets: Directory not found: a-3o
```

The error on attempting to build is of course not ideal.

It's caused by a "word" starting with a number, which is now allowed in my code. Cabal itself doesn't seem to care, but stack apparently does internally somewhere. It's related to the `PackageIdentifier` stuff of course, but I haven't looked deeper yet.

So, my question: should I just disallow that case, even though Cabal doesn't seem to care? Or should I actually attempt to fix the build error instead?

~~On another, only tangentially related note: are project names with underscores in them meant to be forbidden? They were by the old code and are by my change as well, but Cabal itself doesn't seem to care at all, and there's quite a few on Hackage that have underscores as well so it seems a bit odd to disallow them.~~
Edit: nevermind the strikeout part, turns out I was hallucinating the ones on hackage

### DIFF
--- a/src/Stack/Types/PackageName.hs
+++ b/src/Stack/Types/PackageName.hs
@@ -37,6 +37,7 @@ import qualified Data.ByteString.Char8 as S8
 import           Data.Char (isLetter)
 import           Data.Data
 import           Data.Hashable
+import           Data.List (intercalate)
 import           Data.Map (Map)
 import qualified Data.Map as Map
 import           Data.Text (Text)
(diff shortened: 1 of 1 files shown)


mgsloan:
Good stuff, thanks!  I can't reproduce the issue with building an `a-3o` package.

kadoban:
Ah, yes the problem with `a-3o` was apparently user error (using v0.1.6.0 to try to build, instead of the `master` branch version). Thanks.

Issue #1284, opened 2015-11-04
enable-split-objs


tinkyholloway:
**Feature request**
 
(Apparently) there is a `cabal configure` option `--enable-split-objs` that passes `--split-objs` to `ghc` for project sources and dependencies. This creates one object file per function and therefore can reduce the size of resulting binaries as the linker only includes necessary objects files (functions).

The `cabal configure --enable-split-objs` is cited as a solution to the "large binary" problem (http://stackoverflow.com/questions/9198112/haskell-unnecessary-binary-growth-with-module-imports/9198223#9198223, http://stackoverflow.com/questions/6115459/small-haskell-program-compiled-with-ghc-into-huge-binary/6115703#6115703, http://stackoverflow.com/questions/12019072/compilation-dependencies-and-executable-size-with-ghc-in-linux/12019307#12019307) and there are deployments where a reduced size, statically linked binary is preferred.

borsboom:
I agree, this would be a good enhancement.  A pull request would be welcome.

mgsloan:
Should split-objs be on by default?  AFAIK the tradeoff is between compilation time and output size.  I wonder how large the impact to compiletime is.

mgsloan:
Initial support done, with the following caveats:

> Building with --split-objs is enabled. Note that this feature is EXPERIMENTAL, and its behavior may be changed and improved. You will need to clean your workdirs befure use.  If you want to compile all dependencies with split-objs, you will need to delete the snapshot (and all snapshots that could reference that snapshot.

This will hopefully be improved by the resolution of the cluster of issues around https://github.com/commercialhaskell/stack/issues/1265 .

I actually got a segmentation fault out of `stack` built with `--split-objs` (but not its deps), so I'm not sure if this stuff is fully reliable (another reason for the EXPERIMENTAL marking).  Doing another `--split-objs` build did not reproduce the issue.

mgsloan:
Till #1265 is addressed, if you want to use split-objs all the time and don't want to worry about clearing out snapshots so that they can be rebuilt, then a good approach is to add the following to your `config.yaml`:

```yaml
build:
  split-objs: true
```

Issue #846, opened 2015-08-24
Allow more command-line options to be specified within stack.yaml


drvink:
Unless I'm missing something, it doesn't seem to be able to set defaults within a stack.yaml for things like `--library-profiling`, `--haddock`, etc. Furthermore, some flags that were available in cabal install are gone entirely (like `--enable-executable-stripping`, just as a random example), and it's not obvious whether there's any workarounds available.

drwebb:
@drvink You could make a PR with additional fields you think are good to have in the `stack.yaml` file. You'll want to start by editing the `Config` datatype in `src/Stack/Types/Config.hs`

Some flags present in cabal-install aren't present in stack, which is because we haven't made it a design goal to replicate every functionality. Starting other issues are good for tracking adding these.

lukexi:
As mentioned in #757, it would be great to have the `modify-code-page` flag available in a config file as it's generally something one needs to set persistently.

mgsloan:
How about having a field called `configure-options`, which has the same format as [ghc-options](https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md#ghc-options)?

mgsloan:
@snoyberg @borsboom Are y'all on board with "configure-options"?  See also the discussion here: https://github.com/commercialhaskell/stack/pull/1341#issuecomment-156983972

I wish there was something better for doing this with snapshot dependencies, but that seems like a rather tricky problem to solve.

borsboom:
@mgsloan I think you're mis-reading this issue.  This one is about taking *existing* options (e.g., the various build options) and making it possible to set defaults in stack.yaml.  Adding additional _new_ options (like configure-options) should be a separate issue.

borsboom:
@drwebb, I think you're working on this issue?

borsboom:
Discussion about adding `cabal configure` options: #1438

mgsloan:
I'm imagining this looking like:

```yaml
stack-options: --library-profiling
```

charles-cooper:
@mgsloan Could that result in redundant ways to set command line options? E.g.
```yaml
nix:
  enable: true
```
vs
```yaml
stack-options: --nix
```

mgsloan:
[...]

PR #1341, opened 2015-11-12
 enable-split-objs #1284

tinkyholloway:
Issue #1284 enable-split-objs

This PR adds an `enable-split-objs` option to `config` which takes a `boolean` value (defaulting to `false`).

When set `true`, `stack` builds new libraries passing the `--enable-split-objs` option to `cabal`. 

The result is that these libraries are built with multiple object files per module which results in unused functions not being linked into final executables. It's been reported that this can result in significantly reduced size binaries which can be advantageous in some deployments.

The correctness of the change can be confirmed by running `ar -t` on the resulting new library builds with the option set `true`, set `false`, or removed from `config.yaml`.

Use:
1.  start from a clean `~/.stack/` (previously built libraries need to be removed and rebuilt)
2.  add `enable-split-objs: true` to a newly created `~/.stack/config.yaml`
3.  in project directory: `stack clean`, `stack build`

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -653,10 +653,12 @@ ensureConfig newConfigCache pkgDir ExecuteEnv {..} announce cabal cabalfp = do
             return $ case mpath of
                 Nothing -> []
                 Just x -> return $ concat ["--with-", name, "=", toFilePath x]
+        enableSplitObjs <- asks $ configEnableSplitObjs . getConfig
         cabal False $ "configure" : concat
             [ concat exes
             , dirs
             , nodirs
+            , if enableSplitObjs then ["--enable-split-objs"] else []
             ]
         writeConfigCache pkgDir newConfigCache
         writeCabalMod pkgDir newCabalMod
@@ -886,6 +888,7 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
                             Ghc -> []
                             Ghcjs -> ["-build-runner"])
                     return (outputFile, setupArgs)
+
             runExe exeName $ (if boptsCabalVerbose eeBuildOpts then ("--verbose":) else id) fullArgs
 
     maybePrintBuildOutput stripTHLoading makeAbsolute pkgDir level mlogFile mh =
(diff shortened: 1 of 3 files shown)


mgsloan:
Cool, thanks for working on this!

Maybe it would be good to handle flags like this similarly to ghc-options?  See my last comment on https://github.com/commercialhaskell/stack/issues/846

We should also consider whether this flag should be part of the config cache, and instead get added [here](https://github.com/fpco/stack/blob/a4869efb1642f51ee3b03b5a9e75e778e184df2a/src/Stack/Types/Build.hs#L670), so that the packages get reconfigured properly when this flag changes.  Since this adds the flag directly to the invocation of `configure`, using `--enable-split-objs` doesn't force a reconfigure for all the packages.

tinkyholloway:
Thanks @mgsloan for reviewing my PR.

> Maybe it would be good to handle flags like this similarly to ghc-options? See my last comment on 
> #846

I can do this. Are you saying that you see this as being a "per-package" flag? It seems to me that `--enable-split-objs` only makes sense if enabled globally.

> We should also consider whether this flag should be part of the config cache, and instead get added 
> here

I'm not sure I understand this enough to comment. Does this mean that if the flag state changes in the project, then the project is rebuilt with the new flag state? Can it be forced to rebuild dependencies if the flag state changes?

Would you like both of these implemented or is the second one

mgsloan:
> Are you saying that you see this as being a "per-package" flag? It seems to me that --enable-split-objs only makes sense if enabled globally.

Right, I imagine that `--enable-split-objs` would usually be used like this:

```yaml
configure-options:
  "*": --enable-split-objs
```

I think adding something like this makes sense, otherwise, to cover everyone's usecases we'll need to re-wrap every flag `runhaskell Setup.hs configure" can take.

> I'm not sure I understand this enough to comment. Does this mean that if the flag state changes in the project, then the project is rebuilt with the new flag state? Can it be forced to rebuild dependencies if the flag state changes?

Yes, that's what it means.  However, it does not force a rebuild of snapshot dependencies.  I'm not sure what we should do about that.  The snapshot packages are shared among multiple projects, and some might want --enable-split-objs while others do not.  I think this deserves some further consideration.

However, for now, it's fine to just go with the same semantics as ghc-options - see the caveat [mentioned in the docs](https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md#ghc-options).

borsboom:
@tinkyholloway Thanks for your work on this so far.  It's clear from the discussion here that we need to make some deeper design decisions, so I'd like to move the discussion from here to #1438.

I'm going to close this PR for now, and we can revisit it once Stack's design can handle this case better.

mgsloan:
Since we haven't yet done the revamp of ghc-options / cabal-options, I figured it'd be good to revisit this.  Sorry for rejecting the PR, I've added the `--split-objs`, but marked with EXPERIMENTAL and a number of caveats.

Issue #1351, opened 2015-11-16
deepseq-1.4


phadej:
Generic derivation in `deepseq <1.4` is
```hs
rnf x = seq x ()
```
which is not what we want.
    
Previously the deepseq-generics package was used, but not everywhere.  I unfortunately dropped the dependency. OTOH there were fair amount of wrong instances defined without using the deepseq-generics. It's less error-prone to simply bump lower bound. But that's not that simple, as there are packages in `lts-2` (`Cabal` and `bytestring`) with `deepseq <1.4` upper bound.

What would be the way to tackle this. We can either do a
- `stack-7.8.yaml` with lots of overrides or
- introduce `deepseq-generics` back

I can execute the solution, when we agree on the approach. I prefer `deepeq >=1.4`

phadej:
Related: https://github.com/commercialhaskell/stack/issues/953

borsboom:
It looks like one unfortunate effect of upgrading `deepseq` is that, in some cases, it causes problems when building on ghc-7.8.4.  That's because upgrading `deepseq` requires upgrading `containers`, and since `template-haskell` depends on `containers`, `template-haskell` has to be reinstalled, and that sometimes triggers [this issue](https://ghc.haskell.org/trac/ghc/ticket/10192):

```
During interactive linking, GHCi couldn't find the following symbol:
  templatezmhaskell_LanguageziHaskellziTHziSyntax_zdfMonadQzuzdczgzgze_closure
```

I haven't encountered this myself, and it also doesn't seem to happen on Travis, but it does happen on the [Hackage Matrix Builder](http://matrix.hackage.haskell.org/package/stack#GHC-7.8/stack-0.1.10.0).  Not sure if this is worth fixing, but thought I'd put this here in case anyone runs into it.

borsboom:
I found a report that [using a newer version of cabal-install resolves this error in another case](https://github.com/yesodweb/yesod/issues/721), so maybe the matrix builder is just using an older version of cabal-install with GHC 7.8.

PR #1352, opened 2015-11-16
Bump lower bound of deepseq to >=1.4

phadej:
Generic derivation in <1.4 is

    rnf x = seq x ()

which is not what we want.

Previously the deepseq-generics package was used, but not everywhere.  I
unfortunately dropped it. OTOH there were fair amount of wrong instances
defined without using the deepseq-generics. It's less error-prone to
simply bump lower bound.

One way to resolve https://github.com/commercialhaskell/stack/issues/1351

### DIFF
--- a/stack-7.8.yaml
+++ b/stack-7.8.yaml
@@ -1,4 +1,4 @@
-resolver: lts-2.17
+resolver: lts-2.22
 extra-deps:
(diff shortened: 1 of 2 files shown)


borsboom:
Thanks, I think this was the right approach.

PR #1357, opened 2015-11-17
Slightly less confusing language in the documentation for `ghc-options`.

benjamin-wagon:
Less confusing. The documentation for the `apply-ghc-options` uses "targets" to mean something else.


mgsloan:
Thanks!

Issue #1198, opened 2015-10-19
Executables with the same names from different projects are clobbered


jberryman:
```
» stack --version
Version 0.1.5.0 X86_64
```

We have a `stack.yaml` starting like:

```
flags: {}
packages:
- foo/
- bar/
- baz/
```

It appears that if both `foo/foo.cabal` and `bar/bar.cabal` have an executable of the same name, with a section

```
executable some-exec
   ...
```

That the built executables in e.g. `.stack-work/install/x86_64-linux/lts-3.3/7.10.2/bin/` will be clobbered by whichever executable was just build. 

This should perhaps be a warning, or executables should be placed in sub-directories.

borsboom:
I think a warning would be appropriate.  Putting them in separate sub-directories still wouldn't help the `stack exec some-exec` case, since all the subdirectories would have to be on the PATH for `stack exec`.    A pull request would be welcome.

sjakobi:
I'm working on this on my [`1198-warnings` branch](https://github.com/sjakobi/stack/commits/1198-warnings).

I have added a warning to the `stack init` command and would add another one for `stack build`.
Does this sound about right?

Regarding the warning during `init`: Should we recommend renaming the executables or simply point out the problem?

mgsloan:
Cool, looks good to me so far!  Yes, warning on `stack init` and `stack build` should be sufficient.

I'd say it's reasonable to mention that renaming one of the executables is the way to resolve the problem.

PR #1373, opened 2015-11-19
Warn if there are several executables with the same name

sjakobi:
...in a multi-package project because any of them could be the one that
ends up being used by `stack exec` or locally installed.

Fixes https://github.com/commercialhaskell/stack/issues/1198.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -29,8 +29,10 @@ import           Data.Aeson (Value (Object, Array), (.=), object)
 import           Data.Function
 import qualified Data.HashMap.Strict as HM
 import           Data.IORef.RunOnce (runOnce)
+import           Data.List ((\\))
 import qualified Data.Map as Map
 import           Data.Map.Strict (Map)
+import           Data.Monoid
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Data.Text (Text)
@@ -106,6 +108,8 @@ build setLocalFiles mbuildLk bopts = fixCodePage' $ do
                            liftIO $ unlockFile lk
       _ -> return ()
 
+    warnIfExecutablesWithSameNameCouldBeOverwritten locals plan
+
     when (boptsPreFetch bopts) $
         preFetch plan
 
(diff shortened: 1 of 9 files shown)


sjakobi:
I have added an integration test to check for the presence/absence of the warnings.
Checking whether the literal warning messages appear in stderr output might not have been the best way to do so.
If there is a nicer way to do this please let me know!

sjakobi:
There's an issue with the Travis build here, that [I've already reported](https://github.com/travis-ci/travis-ci/issues/5137).

Could somebody with the proper permissions please restart the [GHC-HEAD job](https://travis-ci.org/commercialhaskell/stack/jobs/91958092)?

src/Stack/Build.hs
```diff
@@ -111,12 +113,14 @@ build setLocalFiles mbuildLk bopts = fixCodePage' $ do
...
     if boptsDryrun bopts
         then printPlan plan
-        else executePlan menv bopts baseConfigOpts locals
-                         globalDumpPkgs
-                         snapshotDumpPkgs
-                         localDumpPkgs
-                         installedMap
-                         plan
+        else do
+            warnIfExecutablesWithSameNameCouldBeOverwritten locals plan
```
mgsloan: How about moving this before the `if` so that the warnings are also printed for `--dry-run`?

sjakobi:
Thanks!

src/Stack/Init.hs
```diff
@@ -113,6 +117,31 @@ initProject currDir initOpts = do
     liftIO $ L.writeFile dest' $ B.toLazyByteString $ renderStackYaml p
     $logInfo $ "Wrote project config to: " <> T.pack dest'
 
+-- | See https://github.com/commercialhaskell/stack/issues/1198.
```
mgsloan: I didn't realize that this would require another function.  While it's not very complicated, we should avoid extra code when possible.  I think it'd be fine to only emit the warnings for `build`.

src/Stack/Build.hs
```diff
@@ -128,6 +132,65 @@ allLocal =
     Map.elems .
     planTasks
 
+-- | See https://github.com/commercialhaskell/stack/issues/1198.
+warnIfExecutablesWithSameNameCouldBeOverwritten
+    :: MonadLogger m => [LocalPackage] -> Plan -> m ()
+warnIfExecutablesWithSameNameCouldBeOverwritten locals plan = do
+    let localDuplicateExes = Map.filter (\ps -> length ps > 1) (localExes locals)
```
mgsloan: The mixture of let and where is a little confusing to me.  How about moving everything to the where clause, and:

* Inline `exesToBuild`.  The alternative is to take the `Plan` parameter off `plannedExes`, but I like how that scopes the usage of RecordWildCards.

* Remove the `locals'` parameter from `localExes`, and just use `locals`.  Along with this change, may as well move `pkgExePairs` to the where clause

* Possibly combine the definition of localExes and `localDuplicateExes`.  It's fine either way.

mgsloan:
I've restarted the job.

mgsloan:
Looks pretty good to me so far, thanks for working on this and even adding an integration test! :)

I've added a couple comments.  It surprised me a little that which exes are planned would make a difference.  Let's say package `a` and package `b` both have the same executable name.  If the user does `stack build a` followed by `stack build b`, they don't get a warning, despite overwriting an exe.  Is there a good reason to take the planned exes into account?  This would also simplify the code a bit.

src/Stack/Build.hs
mgsloan: Actually, nevermind on the `exesToBuild` / `plannedExes` stuff, maybe those should just be deleted.  See my comment [here](https://github.com/commercialhaskell/stack/pull/1373#issuecomment-158235789)

sjakobi:
Thanks for your comments! I agree that the warning for `stack init` might be too much and that the warning code for `stack build` should be less complicated.

> If the user does stack build a followed by stack build b, they don't get a warning, despite overwriting an exe.

Is this what you think what this code does or what you think this code should rather do?

Currently the command line dialog would be like this:
```
$ stack build a
Building executable 'a:exe'.
Other executables with the same name might be overwritten: 'b:exe'
...
$ stack build b
Building executable 'b:exe'.
Other executables with the same name might be overwritten: 'a:exe'
...
```

Do you think that we should only have warnings when we're building several executables that might overwrite each other? Even then we would have to look at `plannedExes` to know which executables are being built.

mgsloan:
> Is this what you think what this code does or what you think this code should rather do?

Oh, that was based on what I thought the code does.  Turns out I misread the code, and the purpose of using the planned executables is to give clearer warnings.

So, nevermind, feel free to leave the logic as is, maybe with a comment explaining the reasoning of the `case` being used for the function passed to mergeWithKey .

sjakobi:
> Oh, that was based on what I thought the code does. Turns out I misread the code, and the purpose of using the planned executables is to give clearer warnings.

Ok, I'll try to make that part a bit clearer! :)

sjakobi:
I hope I didn't overdo it with the comments now! ;)

mgsloan:
Awesome, thanks!

Issue #508, opened 2015-07-03
Print latest *applicable* version of packages on conflicts


bergmark:
Extension of #450, showing the latest isn't helpful when there is an upper bound, e.g.

```
--  Failure when adding dependencies:
      classy-prelude-conduit: needed (<0.12), latest is 0.12.0, but 0.12.0 found
      xml-conduit: needed (<1.3), latest is 1.3.0, but 1.3.0 found
```

Here it would be more useful to show the latest version of classy-prelude-conduit that's < 0.12. Perhaps change the message to say "latest applicable is ..."

snoyberg:
Want to take a stab at this? May be faster than iterating together, and
this code is pretty well isolated.

bergmark:
I might take a look over the weekend

snoyberg:
Updating with awaiting-pr label

snoyberg:
Bump

PR #1381, opened 2015-11-20
Print latest applicable version of packages on conflicts #508

sjakobi:


sjakobi:
So, I'm pretty sure this works as intended, but I've had a hard time finding an actual test case.

@bergmark, could you possibly give this one a try or maybe point me at the project you used in the example for #508?

bergmark:
Looks good to me, thanks!

You can test it by putting `foo-1.0` in your stack.yaml and adding an upper bound `foo < 1` in a project's cabal file. This simulates that a package hasn't updated to a newer version of a dependency.

sjakobi:
Thanks for the explanation, @bergmark!

Seems to work alright:
```
$ stack build

      vector: needed (<0.10), couldn't resolve its dependencies
    needed for package: appl-0.1.0.0

      primitive: needed (>=0.5.0.1 && <0.6), 0.6.1.0 found (latest applicable is 0.5.4.0)
    needed for package: vector-0.10.12.2
```

mgsloan:
LGTM, thanks!

PR #1382, opened 2015-11-20
Simplify calculation of build warnings

sjakobi:


sjakobi:
Sorry for coming up with this just after #1373 has been merged but it just so happened that I had a closer look at the documentation for Data.Map.
I realized that the combination of `mapMaybe` and `intersectionWith` would probably be more intuitive than the more complex `mergeWithKey`.
The usage of `NonEmpty` removes the case where `pkgsToBuild` could in principle be empty.

PR #1389, opened 2015-11-21
Improve calculation of build warnings

sjakobi:
* Using NonEmpty instead of [] removes one case to check in warnings
* Map.intersectionWith and Map.mapMaybe are more intuitive than
  Map.mergeWithKey
* localExes is computed with a single list comprehension instead of two

As a side effect the order of the executables in the warnings has changed.


mgsloan:
LGTM, thanks!

PR #1397, opened 2015-11-21
Support for dependent packages in Mercurial

DaveCTurner:
Hi,

The ability to refer to a dependency on a git commit is useful, and I'd quite like to add support for the same thing for Mercurial repositories.

Firstly, would that be useful?

Secondly, I think all I'd have to do would be to add a new constructor to `Stack.Types.Config.PackageLocation`, implement `ToJSON` and `FromJSON`, and add a new case to `Stack.Config.resolvePackageLocation` that would look very similar to the git one except would run `hg clone` and then `hg update -C` instead of `git clone` and `git reset --hard` respectively. I've not actually tried it so perhaps there's more, but it'd be useful to know I'm on the right lines.

Thirdly, the cases for `PLHttpTarball` and `PLGit` are already rather similar and I'd feel bad adding a third case just like them. Would it be ok to extract the common code into one place too?

Cheers,

David

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,8 @@ Other enhancements:
 
 * Print latest applicable version of packages on conflicts
   [#508](https://github.com/commercialhaskell/stack/issues/508)
+* Support for packages located in Mercurial repositories
+  [#1397](https://github.com/commercialhaskell/stack/issues/1397)
 
 Bug fixes:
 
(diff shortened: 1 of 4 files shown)


DaveCTurner:
Something like this, perhaps?

borsboom:
Looks great, and thanks for the very easy to read PR.  Can you just add an entry to the change log, and add "supported since 0.1.9.0" for the hg parts of yaml_configuration.md?

DaveCTurner:
Ok, I've done that.

borsboom:
Thanks; merged!

Issue #1398, opened 2015-11-22
Have the travis 7.8 build use stack-7.8.yaml


mgsloan:
Travis uses cabal to build, to ensure that we can always bootstrap.  However, this means that we can end up pulling in different versions than used by stack-7.8.yaml.  In particular, this can cause issues because it isn't build against `Cabal-1.18.*` - see [sjakobi's comment](https://github.com/commercialhaskell/stack/pull/1334#discussion_r45548676).

borsboom:
Travis is intentionally configured to use cabal-install instead of `stack`, to ensure that Stack can always be bootstrapped that way.  Instead, I'll get our Jenkins server to run the integration tests using stack-7.8.yaml as well, so that'll catch these problems pretty quickly.

phadej:
I'll setup travis to /also/ build with stack.

PR #1404, opened 2015-11-23
Build on travis with stack too

phadej:
Resolves https://github.com/commercialhaskell/stack/issues/1398

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -2,9 +2,21 @@
 
 set -eux
 
+case "$BUILD" in
+  stack)
+    mkdir -p ~/.local/bin;
+    if [ `uname` = "Darwin" ]; then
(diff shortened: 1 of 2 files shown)


phadej:
Done. ping @snoyberg @borsboom 

- Uses compiler hack to have separate caches for each matrix row
- osx build as an extra

mgsloan:
Thanks!  LGTM, I'm in favor of merging.  Seem good, @borsboom ?

borsboom:
It looks like this also re-enables running the integration tests.  In the past, we've had trouble with Travis timing out the builds when those are enabled.  I'll merge as-is and we'll see how it goes, but may have to disable the integration tests if they cause problems.

phadej:
@borsboom it doesn't enabled integration tests, they use "comment trick"
```sh
% true echo foo
# does nothing
```

borsboom:
Ah I see, I saw they were uncommented but didn't look more closely.

PR #1420, opened 2015-11-24
Updated FAQ link

kberger:
Link 'the FAQ' in the Questions, Feedback, Discussion section was pointing to faq.html instead of faq.md


borsboom:
This was an unfortunate thing I had to do to get links working with readthedocs.org, which is now being used to generate http://docs.haskellstack.org/.  We're moving references to the documentation to point there instead of at the GitHub repo.

PR #1424, opened 2015-11-25
Only run benchmarks specified as build targets (closes #1412)

### LINKED ISSUE #1412: Only run benchmarks specified as build targets
Currently which benchmark targets are selected is ignored when actually running them - see [this FIXME](https://github.com/fpco/stack/blob/master/src/Stack/Build/Execute.hs#L559).  Should be as simple as providing the benchmarks as component parameters [here](https://github.com/fpco/stack/blob/master/src/Stack/Build/Execute.hs#L1281), and testing that it works properly.

luigy:


mgsloan:
LGTM, thanks!

PR #1423, opened 2015-11-24
[RFC] Send output of building setup to stderr (fixes #1410)

rvion:
this PR
- rename `runIn` to `runCmd`
- add a new `runCmd'` function taking and extra ```(CreateProcess -> CreateProcess)``` param
- fixes #1410 by using ```runCmd' (\cp -> cp { std_out = UseHandle stderr })``` in `getSetupExe`
- define a new `data CMD = CMD directoryToRunIn commandToRun envOverride commandLineArguments`
- refactor `runIn`, `callProcess`, `callProcess'` to take a CMD arg to ease code exploration, and argument passing

PR #1430, opened 2015-11-25
[RFC] gitignore test.dot and test.html

rvion:
made after https://github.com/commercialhaskell/stack/pull/1423


mgsloan:
Hmm, where do these files come from?

rvion:
well, I don't know.
I thought they were generated by stack test, but if you tell me they're not, this PR shouldn't be

Issue #1425, opened 2015-11-25
stack rebuilds `Stack.Upgrade` when you change the git index at all


kadoban:
When you're developing on `stack` itself and modify the git index (aka the cache or staging area), it causes the `Stack.Upgrade` module to need to be rebuilt.

It's not exactly the end of the world, but it adds an extra delay in certain cases, like when you are running tests and it lengthens the time between iterations of what you're trying to mess with.

```
~/projects/stack (upstream)$ stack build
~/projects/stack (upstream)$ touch bogus.txt
~/projects/stack (upstream %)$ git add bogus.txt 
~/projects/stack (upstream +)$ stack build
stack-0.1.9.0: unregistering (local file changes: .git/index)
stack-0.1.9.0: build
Preprocessing library stack-0.1.9.0...
[73 of 76] Compiling Stack.Upgrade    ( src/Stack/Upgrade.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/Stack/Upgrade.o ) [~/projects/stack/.git/index changed]
In-place registering stack-0.1.9.0...
Preprocessing executable 'stack' for stack-0.1.9.0...
[2 of 2] Compiling Main             ( src/main/Main.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/stack/stack-tmp/Main.o ) [~/projects/stack/.git/index changed]
Linking .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/stack/stack ...
stack-0.1.9.0: copy/register
Installing library in
~/projects/stack/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/lib/x86_64-linux-ghc-7.10.2/stack-0.1.9.0-K5i01b4VNwNJsmjPkKylJf
Installing executable(s) in
~/projects/stack/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/bin
Registering stack-0.1.9.0...
~/projects/stack (upstream +)$ stack build
~/projects/stack (upstream +)$ 
```

Since the `bogus.txt` file doesn't matter at all, I would have ideally expected the second `stack build` to do nothing, but it spends some time rebuilding instead.

As pointed out in a short discussion about it on IRC, this is caused by the use of `$gitHash` [here](https://github.com/commercialhaskell/stack/blob/09991d98ed034e036668721de966c199e2e40642/src/Stack/Upgrade.hs#L45)

[...]

PR #1431, opened 2015-11-25
add flag to allow disabling compile-time git status checks

kadoban:
fixes issue #1425

Without the new flag, stack rebuilds a couple of modules when the git index changes:
```
~/projects/stack (flag_use_git_status_in_version)$ stack build
~/projects/stack (flag_use_git_status_in_version)$ touch bogus.txt                                                   
~/projects/stack (flag_use_git_status_in_version %)$ git add bogus.txt                                               
~/projects/stack (flag_use_git_status_in_version +)$ stack build
stack-0.1.9.0: unregistering (local file changes: .git/index)
stack-0.1.9.0: build
Preprocessing library stack-0.1.9.0...
[73 of 76] Compiling Stack.Upgrade    ( src/Stack/Upgrade.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/Stack/Upgrade.o ) [/home/mud/projects/stack/.git/index changed]
In-place registering stack-0.1.9.0...
Preprocessing executable 'stack' for stack-0.1.9.0...
[2 of 2] Compiling Main             ( src/main/Main.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/stack/stack-tmp/Main.o ) [/home/mud/projects/stack/.git/index changed]
Linking .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/stack/stack ...
stack-0.1.9.0: copy/register
Installing library in
/home/mud/projects/stack/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/lib/x86_64-linux-ghc-7.10.2/stack-0.1.9.0-K5i01b4VNwNJsmjPkKylJf
Installing executable(s) in
/home/mud/projects/stack/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/bin
Registering stack-0.1.9.0...
```

With the flag, the git index changing doesn't cause rebuilding on its own (and commits or changing branches don't necessarily either):
```
~/projects/stack (flag_use_git_status_in_version +)$ stack build --flag stack:disable-git-info
~/projects/stack (flag_use_git_status_in_version +)$ touch bogus.txt                                                 
~/projects/stack (flag_use_git_status_in_version +)$ git add bogus.txt                                               
~/projects/stack (flag_use_git_status_in_version +)$ stack build --flag stack:disable-git-info                       
~/projects/stack (flag_use_git_status_in_version +)$ git commit -m "temp commit"
[flag_use_git_status_in_version 7f5194e] temp commit
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 bogus.txt
~/projects/stack (flag_use_git_status_in_version)$ stack build --flag stack:disable-git-info
~/projects/stack (flag_use_git_status_in_version)$
```

Of course changing a file that stack actually *is* dependent on still correctly causes rebuilding:
```
~/projects/stack (flag_use_git_status_in_version)$ stack build --flag stack:disable-git-info
~/projects/stack (flag_use_git_status_in_version)$ echo "-- nonsense" >> src/main/Main.hs                            
~/projects/stack (flag_use_git_status_in_version *)$ stack build --flag stack:disable-git-info
stack-0.1.9.0: unregistering (local file changes: src/main/Main.hs)
stack-0.1.9.0: build
Preprocessing library stack-0.1.9.0...
In-place registering stack-0.1.9.0...
Preprocessing executable 'stack' for stack-0.1.9.0...
[2 of 2] Compiling Main             ( src/main/Main.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/stack/stack-tmp/Main.o )
Linking .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/stack/stack ...
stack-0.1.9.0: copy/register
Installing library in
/home/mud/projects/stack/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/lib/x86_64-linux-ghc-7.10.2/stack-0.1.9.0-3UM6IwMtAE82KY5FnDpl4M
Installing executable(s) in
/home/mud/projects/stack/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/bin
Registering stack-0.1.9.0...
```

`stack --version` clearly shows that the flag was used when building, and if for some odd reason someone submits an issue with that version, it will be obvious (and they can be told not to do that if necessary):
```
~/projects/stack (flag_use_git_status_in_version)$ stack exec -- stack --version
0.1.9.0 (git version info disabled) x86_64
```

`stack upgrade` will still work fine, the only caveat there is that if stack was built with the new flag and they use `--git`, it will always build stack anew, since there's no way to know if it should be or not. I added a warning in that case.

### DIFF
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -15,7 +15,6 @@ import           Data.Monoid                 ((<>))
 import qualified Data.Monoid
 import qualified Data.Set                    as Set
 import qualified Data.Text as T
-import           Development.GitRev          (gitHash)
 import           Network.HTTP.Client.Conduit (HasHttpManager)
 import           Path
 import           Path.IO
@@ -35,14 +34,16 @@ import           System.Process.Run
 upgrade :: (MonadIO m, MonadMask m, MonadReader env m, HasConfig env, HasHttpManager env, MonadLogger m, HasTerminal env, HasReExec env, HasLogLevel env, MonadBaseControl IO m)
         => Maybe String -- ^ git repository to use
         -> Maybe AbstractResolver
+        -> Maybe String -- ^ git hash at time of building, if known
         -> m ()
-upgrade gitRepo mresolver = withCanonicalizedSystemTempDirectory "stack-upgrade" $ \tmp -> do
+upgrade gitRepo mresolver builtHash =
+  withCanonicalizedSystemTempDirectory "stack-upgrade" $ \tmp -> do
     menv <- getMinimalEnvOverride
     mdir <- case gitRepo of
       Just repo -> do
         remote <- liftIO $ readProcess "git" ["ls-remote", repo, "master"] []
         let latestCommit = head . words $ remote
-        if latestCommit == $gitHash then do
+        if builtHash == Just latestCommit then do
           $logInfo "Already up-to-date, no upgrade required"
           return Nothing
         else do $logInfo "Cloning stack"
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE TemplateHaskell       #-}
 module Stack.Upgrade (upgrade) where
 
+import           Control.Monad               (when)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
@@ -11,6 +12,7 @@ import           Control.Monad.Reader        (MonadReader, asks)
 import           Control.Monad.Trans.Control
 import           Data.Foldable               (forM_)
 import qualified Data.Map                    as Map
+import           Data.Maybe                  (isNothing)
 import           Data.Monoid                 ((<>))
 import qualified Data.Monoid
 import qualified Data.Set                    as Set
@@ -43,6 +45,11 @@ upgrade gitRepo mresolver builtHash =
       Just repo -> do
         remote <- liftIO $ readProcess "git" ["ls-remote", repo, "master"] []
         let latestCommit = head . words $ remote
+        when (isNothing builtHash) $
+            $logWarn $ "Information about the commit this version of stack was "
+                    <> "built from is not available due to how it was built. "
+                    <> "Will continue by assuming an upgrade is needed "
+                    <> "because we have no information to the contrary."
         if builtHash == Just latestCommit then do
           $logInfo "Already up-to-date, no upgrade required"
           return Nothing
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -34,7 +34,7 @@ import qualified Data.Text.IO as T
 import           Data.Traversable
 import           Data.Typeable (Typeable)
 import           Data.Version (showVersion)
-import           Development.GitRev (gitCommitCount)
+import           Development.GitRev (gitCommitCount, gitHash)
 import           Distribution.System (buildArch)
 import           Distribution.Text (display)
 import           GHC.IO.Encoding (mkTextEncoding, textEncodingName)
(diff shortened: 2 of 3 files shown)


mgsloan:
Thanks for working on this.  Seems a bit of a complex solution, though, I'd prefer to avoid CPP when possible.

Would it be a good-enough solution to just make the gitHash into a parameter of upgrade?  This way only Main.hs will need to be recompiled when git changes.

kadoban:
Well, the only problem is that when it has to recompile anything it's really quite disruptive to messing around with tests and I assume some other workflows that I'm not thinking of. I'd have to time it, but subjectively it takes quite an appreciable amount of time even if it only has to build Main and then link.

That *would* be an improvement over the status quo though, if this approach isn't desired.

I can time it out a bit and see what the difference is if that'll help. Other than that, I guess it's just a matter of figuring out how important this particular workflow is in the scheme of things, compared to the added code complexity.

borsboom:
Having Stack.Upgrade take the gitHash as an argument seems like a good
change regardless.

>

src/main/Main.hs
```diff
@@ -123,6 +128,12 @@ main = withInterpreterArgs stackProgName $ \args isInterpreter -> do
                                                     commitCount /= ("UNKNOWN" :: String)]
             , [" ", display buildArch]
             ]
+#else
+     let versionString' = concat [ showVersion Meta.version
+                                 , " (git version info disabled)"
```
borsboom: I think you can just omit the ` (git version info disabled)` part.  Similar to when Stack is built using `cabal install stack` (where no git info is available), which also just leaves it out.

src/Stack/Upgrade.hs
```diff
@@ -42,7 +45,16 @@ upgrade gitRepo mresolver = withCanonicalizedSystemTempDirectory "stack-upgrade"
       Just repo -> do
         remote <- liftIO $ readProcess "git" ["ls-remote", repo, "master"] []
         let latestCommit = head . words $ remote
-        if latestCommit == $gitHash then do
+#ifdef USE_GIT_INFO
+            builtHash = $gitHash
+#else
+            builtHash = "git information disabled by flag"
```
borsboom: Maybe this warning should also be displayed for the case when no git version information is available (e.g. if you built from hackage directly or from an sdist).

borsboom: gitRev sets `$gitHash` to `"UNKNOWN"` in that case, by the way, so you could do the same if !USE_GIT_INFO.

kadoban:
Timing info, just for information (all times are median of three attempts, of `time stack build`'s "real" time, with specified circumstances, rounded to the nearest second):
- `Stack.Upgrade` and `Main` need to be rebuilt: **37s**
- Only `Main` needs to be rebuilt: **32s**
- Neither need to be rebuilt: **2s**

This is on a pretty crappy old laptop I use, so might overstate the case, but for me you can see it's quite a big difference between not having to rebuild anything and having to rebuild `Main`. The difference between having to rebuild both or only one is minor but appreciable.

Thanks for the good suggestions, I'll attempt to integrate those.

kadoban:
I've incorporated the suggestions from above; the result in quite pleasant.

I also attempted to make as many of the changes on top of a clean state, without the addition of the flag, and then re-introduced the flag addition afterwards. This was done so that if the flag part isn't desired, it's easier to just include the other changes and scrap that piece (though I personally think the entire change, including the flag, is the right way to go).

Sorry if the history is a bit messy, I'm not really clear on the protocol with respect to github PRs, so I wanted to avoid changing history (unless that's recommended?)

If this is judged worthy of merging, in whole or in part, history should probably be cleaned up a bit first (mostly removing the initial commit and the subsequent immediate revert of it). I'd be happy to do that upon request.

I've only done a couple of superficial tests on these modifications, more should be done before it's merged. (I plan to do more testing, but haven't had a chance yet)

mgsloan:
> Sorry if the history is a bit messy, I'm not really clear on the protocol with respect to github PRs, so I wanted to avoid changing history (unless that's recommended?)

Seems fine to me!  Changing history can indeed get tricky when iteratively reviewing larger changes.  For smaller changes to me it's fine to rebase.

>  If this is judged worthy of merging, in whole or in part, history should probably be cleaned up a bit first (mostly removing the initial commit and the subsequent immediate revert of it). I'd be happy to do that upon request.

Sure, we can merge this after those first two commits are removed.

kadoban:
Okay, I've fixed up history and tested the end result enough that I'm confident in it. I tested the build semantics without and without the flag, with and without a git repo from it to draw info from. I also tested upgrade semantics and it seems great.

So as far as I know it's all fixed up and ready to go.

borsboom:
Thanks!

PR #1434, opened 2015-11-26
Add executable fall-through

### LINKED ISSUE #1433: Plugin support with git-style "fall-through" of the executable call
So with `git`, if I have say a binary `git-branch-fzf` on my PATH and type `git branch-fzf`, it'll call my binary.

This is a nice way of providing some extension to the tool and I'd like if `stack` did it. Potentially it could give precedence to the outside binary, so that people could write experimental wrappers over built-in commands.

Why I think this matters:
- Being extensible is nice, this is an easy way of letting other developers add new commands or extend existing ones
- It's not enough to let people modify the `stack` project and ask users to rebuild it, because there's much more friction in doing this
- Good things come out of people experimenting with things and it's easier to write software when there's user feedback (at least for me)

Seems like it's pretty easy, not sure if it's wanted (maybe it's even already supported somehow).

yamadapc:
Uses `System.Directory.findExecutable`. Adds a `mOnInvalidArgs` callback to the
`complicatedOptions` function on `Options.Applicative.Complicated` the rest of
the logic is centralized on `Main`.

I didn't use `System.Process.Read.findExecutable` because I'd then need
`getMinimalEnvOverride`, which would require `GlobalOpts` I think. I'm not sure,
but I don't understand what'd give at the moment.

Will rebase after discussion.

This closes #1433.

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -37,7 +37,7 @@ import           Data.Version (showVersion)
 #ifdef USE_GIT_INFO
 import           Development.GitRev (gitCommitCount, gitHash)
 #endif
-import           Distribution.System (buildArch)
+import           Distribution.System (buildArch, buildPlatform)
 import           Distribution.Text (display)
 import           GHC.IO.Encoding (mkTextEncoding, textEncodingName)
 import           Network.HTTP.Client
@@ -84,6 +84,7 @@ import           Stack.Types.StackT
 import           Stack.Upgrade
 import qualified Stack.Upload as Upload
 import           System.Directory (canonicalizePath, doesFileExist, doesDirectoryExist, createDirectoryIfMissing)
+import qualified System.Directory as Directory (findExecutable)
 import           System.Environment (getEnvironment, getProgName)
 import           System.Exit
 import           System.FileLock (lockFile, tryLockFile, unlockFile, SharedExclusive(Exclusive), FileLock)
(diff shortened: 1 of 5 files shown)


yamadapc:
Updated to conform to 2 spaces indentation and not run `stylish-haskell` on the files.

src/main/Main.hs
```diff
@@ -138,6 +140,16 @@ main = withInterpreterArgs stackProgName $ \args isInterpreter -> do
          "stack - The Haskell Tool Stack"
          ""
          (globalOpts False)
+         -- When there's an invalid argument
+         (Just $ \parseResult as -> do
+           mExternalExec <- Directory.findExecutable ("stack-" ++ head as)
+           case mExternalExec of
+             Just ex -> do
+               ph <- spawnProcess ex (tail as)
```
borsboom: It would be better if this replaces the current process instead of spawning a sub-process and just waiting for it to finish.  `Stack.Exec.exec` will do that, but it's another function that needs an EnvOverride (this time via a Config).  It's possible to create an EnvOverride using `System.Process.Read.mkEnvOverride`, and `exec` could be modified to take an EnvOverride as an argument instead of needing a Config.

yamadapc:
It seems like I'd still need a config to know the platform... I made some changes, but couldn't get this to work.

borsboom:
Using Distrubition.System.buildPlatform will suffice for this case.

yamadapc:
Cool, thanks for the tip; I should have asked sooner. Updated and rebased against `origin/master`. I made it so that the callback I added to `Complicated.hs` was run on every parser failure; not just arguments. This seemed better.

Not sure if this was the `exec` modification you mentioned?

borsboom:
Thanks, this is looking just about ready for merge.  I tested it and everything seems to be in order.  Please add an entry to the changelog, and maybe squash it into one commit.

yamadapc:
Okie, coolness! Squashed, changelog entry added and rebased against `master`.

Issue #1178, opened 2015-10-16
Provide an option to configure the location  `.stack-work` directory


Gabriel439:
Right now I'm working on implementing a Haskell plugin for Twitter's build tool, `pants`, and one thing we'd like to have is the ability to configure the location of the `~/.stack` and `.stack-work` directories under directories managed by the `pants` tool.

If this is already possible, just let me know, but the last time I checked I couldn't figure out how to do this.

borsboom:
You can set the `STACK_ROOT` environment variable to override the location of `~/.stack`.  `.stack-work` is not currently overridable.

borsboom:
I don't think overriding `.stack-work` would be too hard to implement, and a PR would be welcome.

Gabriel439:
Thanks for the `STACK_ROOT` tip!  That helps a lot

feuerbach:
I would also like to see this implemented.

My use case is that I need to produce two build flavors: with and without hpc. The best way (for incremental compilation) to do this is to build them in two different build trees. I plan to work around this for now by making `.stack-work` an alternating symlink.

mrkkrp:
In case nobody is doing this already, I try could to implement the feature (once my current PR is merged). We could make `STACK_LOCAL` environment variable override `.stack-work`. If you have alternative ideas in mind, tell me.

Gabriel439:
Shouldn't this be passed on the command line?  Programming using environment variables is very error-prone: if you make a typo in the environment variable name nothing will catch the error

mgsloan:
@Gabriel439 Makes a good point, this makes more sense as a commandline argument, since it's probably something local to your current build.  `STACK_ROOT` is shared globally, which is why it makes sense to have it as an environment variable.

@mrkkrp Feel free to add this!

mrkkrp:
@Gabriel439, @feuerbach, this is solved now. You can add `work-dir` parameter with relative directory to use instead of `.stack-work` in `stack.yaml` or override it on command line via `--work-dir=some-directory` thing.

Here is the PR: #1442.

@mgsloan, Looks like this may be closed.

mgsloan:
@mrkkrp Thanks for doing this!

Gabriel439:
Thank you!  This is fantastic :)

magthe:
Brilliant!

abailly:
Thanks for providing this option. Is there a specific why it is not possible to pass an absolute directory?

mrkkrp:
[...]

PR #1442, opened 2015-11-28
Addition of ‘--work-dir’ option to override working directory, #1178

mrkkrp:
Here I've started to work on this. Currently I've added `--work-dir` option represented by `globalWorkDir` in `GlobalOpts`. Please correct me if this is a wrong place to put this option.

Now my plan is:

1. Store this option is Stack monadic environment (`Config`?), so functions that need to know stack working directory could query it. I think `HasConfig env, MonadReader env` can be used.

2. Instead of `workDirRel` constant, use that value from config, this will probably add `HasConfig, MonadReader env` or similar constraint to some functions that previously didn't need it.

3. Take care of `ignoreDirs` which specifies which directories shouldn't be searched for `.cabal` files. It currently lists `".stack-work"`. With this change it would need to construct the list more “dynamically”.

My questions:

1. Is `GlobalOpts` is the right place to add `--work-dir` option?

2. In case `GlobalOpts` is the right place, where is the best place to pass the value `globalWorkDir` from `GlobalOpts` to `Config` so the rest of Stack can query it?

Well, the rest seems to be pretty easy to do (at least for now).

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -91,7 +91,7 @@ import           System.Process.Run
 import           System.Process.Internals (createProcess_)
 #endif
 
-type M env m = (MonadIO m,MonadReader env m,HasHttpManager env,HasBuildConfig env,MonadLogger m,MonadBaseControl IO m,MonadCatch m,MonadMask m,HasLogLevel env,HasEnvConfig env,HasTerminal env)
+type M env m = (MonadIO m,MonadReader env m,HasHttpManager env,HasBuildConfig env,MonadLogger m,MonadBaseControl IO m,MonadCatch m,MonadMask m,HasLogLevel env,HasEnvConfig env,HasTerminal env, HasConfig env)
 
 -- | Fetch the packages necessary for a build, for example in combination with a dry run.
 preFetch :: M env m => Plan -> m ()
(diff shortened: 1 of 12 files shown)


mgsloan:
> Take care of ignoreDirs which specifies which directories shouldn't be searched for .cabal files. It currently lists ".stack-work". With this change it would need to construct the list more “dynamically”.

I don't think this is necessary.

> 1. Is GlobalOpts is the right place to add --work-dir option?
>
> 2. In case GlobalOpts is the right place, where is the best place to pass the value globalWorkDir from GlobalOpts to Config so the rest of Stack can query it?

How about adding it directly to `Config` / `ConfigMonoid` ?  This'd mean that it'd also be a stack.yaml field, which could certainly have its uses (ensuring that different dist dirs are used for two different configs)

mrkkrp:
Good, I'll take that direction.

mrkkrp:
Seems to work! That was fun, please review.

mgsloan:
LGTM, thanks!

lukexi:
Hmm, I'm confused as to why this is restricted to child directories of the current directory. In my use-case I'd like to use ".." as the .stack-work directory for a set of uncoupled but related projects, but this is disallowed by parseRelDir since:
https://github.com/chrisdone/path/commit/87a56a93b8dcec8274041b4bc7020b54c60fb4ec

mrkkrp:
@lukexi, See [this discussion](https://github.com/commercialhaskell/stack/issues/1178#issuecomment-162167351).

PR #1444, opened 2015-11-28
fix broken link in documentation

massysett:


borsboom:
See https://github.com/commercialhaskell/stack/pull/1420#issuecomment-159594363

PR #1446, opened 2015-11-28
fixed broken link

joshcough:


mrkkrp:
See #1444, #1420 (maybe there are more attempts).

@borsboom, Maybe you should add explicit message telling people that it doesn't work… *on purpose*. It seems this issue has the ability to generate several PRs every week! Maybe some sort of script that would edit links (to work around that bug) could be written?

joshcough:
Oh, I'm sorry about this. I was just trying to read docs and it seemed like
an easy thing.

PR #1465, opened 2015-12-03
Eyecandy update for nix_integration.md

luigy:
fixed a couple of outdated link anchors too


mgsloan:
Thanks!

YPares:
@luigy Thanks, I actually closely mirrored docker_integration.md which didn't use language flags, but that's prettier like that indeed.

PR #1467, opened 2015-12-03
add example on how to install packages from git

eallik:
this could possibly be amended with other options to install non-local non-hackage packages.


mgsloan:
Thanks!

PR #1470, opened 2015-12-05
Support multi line interpreter directive comments

### LINKED ISSUE #1394: Support multiline script interpreter directive for better readability
When we have many packages to be specified on the embedded stack command the comment line can become too long. It would be nice to support a multiline comment so that it can formatted for better readability. Like this:

    {- stack --verbosity silent 
             --resolver lts-3.14 
             --install-ghc runghc 
             --package system-filepath 
             --package system-argv0 
             --package random 
             --package argparser
             ...
    -}

Further comments from the stack mailing list:

>>Perhaps {-# OPTIONS_STACK ... #-} would be a more standard syntax for it?

>I'd worry about that being confused with compiler pragmas, and people expecting to to be used in all cases (i.e. by `stack build` as well).  Since this only applies when using `stack` as a script interpreter, that would be misleading.

harendra-kumar:
Multi line block comments can now be used for additional stack arguments
when stack is used as script interpreter. Like this:

```
{- stack
  --verbosity silent
  --resolver lts-3.14
  --install-ghc
  runghc
  --package random
  --package system-argv0
-}
```

Newlines are ignored in the block comments. This allows for better
formatting of longer commands.

Previous single line format is supported as usual. The single line
syntax now allows a missing space between comment marker and the start
of command i.e. ```--stack ...``` too works now.

Updated the guide with complete syntax and semantics.

Closes #1394


src/Data/Attoparsec/Args.hs
```diff
@@ -48,15 +97,41 @@ argsParser mode = many (P.skipSpace *> (quoted <|> unquoted)) <*
     nonquote = P.satisfy (not . (=='"'))
     naked = P.satisfy (not . flip elem ("\" " :: String))
 
+-- | Parser to extract the stack command line embedded inside a comment
+-- after validating the placement and formatting rules for a valid
+-- interpreter specification.
+
+interpParser :: String -> Parser Text String
```
mgsloan: Minor nitpick: How about `interpreterArgsParser`?  While we do abbreviate things here and there, it's a bit arbitrary to shorten interpreter to `interp`.

mgsloan:
LGTM!  Since all I've got is that minor nitpick, I'm going to just merge and add a commit with the rename

mgsloan:
I ended up making the following changes: https://github.com/commercialhaskell/stack/commit/e44d57c41943a5536fcf645ab8f5f36812162bf2

harendra-kumar:
Great! I was looking for the missing piece i.e. `sinkParser` but was not aware of its existence. Thanks for cleaning it up.

PR #1474, opened 2015-12-05
Apply various hlint hints

markus1189:
Appy various hlint hints:
- Attoparsec.Args: Replace not . (==) with (/=)
- Stack.Build: redundant brackets
- Stack.Build.Installed: use `forM_`, redundant bracket
- Stack.BuildPlan: fuse mapM and map, use const
- Stack.Build.Source: redundant bracket, eta reduce
- Stack.Config: Redundant brackets
- Stack.Config.Nix: use when
- Stack.Coverage: use (/=)
- Stack.Docker.GlobalDB: use unless
- Stack.Docker: redundant brackets
- Stack.Ghci:  use mapMaybe
- Stack.Image: reduce duplication

-- 24pullrequests
[](http://24pullrequests.com/)

### DIFF
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -84,7 +84,7 @@ import qualified System.FilePath as FP
 data BuildPlanException
     = UnknownPackages
         (Path Abs File) -- stack.yaml file
-        (Map PackageName (Maybe Version, (Set PackageName))) -- truly unknown
+        (Map PackageName (Maybe Version, Set PackageName)) -- truly unknown
         (Map PackageName (Set PackageIdentifier)) -- shadowed
     | SnapshotNotFound SnapName
     deriving (Typeable)
@@ -406,8 +406,7 @@ instance FromJSON Snapshots where
     parseJSON = withObject "Snapshots" $ \o -> Snapshots
         <$> (o .: "nightly" >>= parseNightly)
         <*> (fmap IntMap.unions
-                $ mapM parseLTS
-                $ map snd
+                $ mapM (parseLTS . snd)
                 $ filter (isLTS . fst)
                 $ HM.toList o)
       where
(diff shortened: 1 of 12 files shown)


src/Stack/BuildPlan.hs
```diff
@@ -638,7 +637,7 @@ shadowMiniBuildPlan :: MiniBuildPlan
 shadowMiniBuildPlan (MiniBuildPlan cv pkgs0) shadowed =
     (MiniBuildPlan cv $ Map.fromList met, Map.fromList unmet)
   where
-    pkgs1 = Map.difference pkgs0 $ Map.fromSet (\_ -> ()) shadowed
+    pkgs1 = Map.difference pkgs0 $ Map.fromSet const shadowed
```
mgsloan: Can you please leave this one out?  This actually causes the type of this map to change.  `(\_ -> ())` isn't the same as `const`.  Personally, I don't agree with some of hlint's suggestions.  Using `const` is one thing I don't like.  Compare:

`const ()`
`\_ -> ()`

The lambda version is shorter and more straightforward and naturally scales:

`const (const ())`
`\_ _ -> ()`

src/Stack/Docker/GlobalDB.hs
```diff
@@ -76,11 +76,11 @@ getDockerImagesLastUsed config =
...
-  where go = do l <- selectList [] []
-                forM_ l (\(Entity k (DockerImageProject{dockerImageProjectImageHash = h})) ->
-                           if h `elem` existingHashes
-                             then return ()
-                             else delete k)
+  where
+    go = do
+      l <- selectList [] []
+      forM_ l (\(Entity k DockerImageProject{dockerImageProjectImageHash = h}) ->
+                   unless (h `elem` existingHashes) $ delete k)
```
mgsloan: How about `when (h `notElem` existingHashes)`?

I find `when` much clearer than `unless` in general, it reads way better.

mgsloan:
Thanks!  I'll merge when my comments are addressed

markus1189:
Thanks for the thorough, feedback, agreed in all points!

PR #1481, opened 2015-12-09
Add unit tests for interpreter argument parser

harendra-kumar:
Added functions to generate input strings comprehensively for the
interpreter argument parser. The generator combines the alternatives in
all possible ways.

After running the tests, found a couple of bugs of theoretical interest
and fixed those so that the parser behaves in a consistent manner even
in those cases.


mgsloan:
Awesome, thanks!

PR #1487, opened 2015-12-10
make verbosity silent by default in script interpreter mode

### LINKED ISSUE #1472: Assume verbosity as silent when invoked in interpreter mode
Please refer to #1394 for the original discussion.

* Currently stack prints some output of its own before running the script:

> Run from outside a project, using implicit global project config
> Using resolver: lts-3.14 from implicit global project's config file: /var/home/harendra/.stack/global-project/stack.yaml

When used as interpreter no other output is desirable other than the script's output itself. In fact this should hold true for any _stack exec_ command. That might interfere with the processing of the script output.

We can fix this by assuming `--verbosity silent` when invoked as interpreter.

kadoban:
fixes #1472

### DIFF
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -565,12 +565,12 @@ execOptsExtraParser = eoPlainParser <|>
                            help "Use an unmodified environment (only useful with Docker)")
 
 -- | Parser for global command-line options.
-globalOptsParser :: Bool -> Parser GlobalOptsMonoid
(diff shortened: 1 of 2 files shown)


kadoban:
This defaults verbosity to silent only when being used as a script interpreter. You can manually set to other verbosities still, in case someone wants to do that for whatever reason.

mgsloan:
LGTM, thanks!

Issue #1453, opened 2015-11-30
Better errors - git fails with a confusing error message when a commit doesn't exist in a git repo, used in a stack.yaml config


athanclark:
Hello everybody, this is a casual issue and not immediately detrimental.

In a `stack.yaml`, if you set a source code location to a git commit that doesn't exist on the referenced repo, then you get the "could not parse object" error, which led me to believe I had somehow pasted my commit hash wrong. For instance, if you have this in your `stack.yaml`:

```yaml
packages:
- '.'
- location:
    git: https://github.com/athanclark/foo
    commit: 8e22ca206e9fc8fd7b903afa8b073120f8897793
  extra-dep: true
```

and you tried to `stack build`, you would get the error:

```
Running /usr/bin/git reset --hard 8e22ca206e9fc8fd7b903afa8b073120f8897793 exited with ExitFailure 128

fatal: Could not parse object '8e22ca206e9fc8fd7b903afa8b073120f8897793'.
```

Which is misleading.

mgsloan:
It'd make sense to me to add a message like `Please ensure that commit 8e22ca206e9fc8fd7b903afa8b073120f8897793 exists within https://github.com/athanclark/foo` when that command fails (but also provide the command output).

mrkkrp:
@mgsloan, Where this thing is processed? Where should one start to look to add this message?

mgsloan:
@mrkkrp [here](https://github.com/fpco/stack/blob/159c10a6c6eb3028a5d4665a93a9dbb52488608f/src/Stack/Config.hs#L475)

mgsloan:
It's fixed, thanks @mrkkrp 

```
Running /usr/bin/git reset --hard 8e22ca206e9fc8fd7b903afa8b073120f8897793 exited with ExitFailure 128

fatal: Could not parse object '8e22ca206e9fc8fd7b903afa8b073120f8897793'.

Please ensure that commit 8e22ca206e9fc8fd7b903afa8b073120f8897793 exists within https://github.com/ekmett/lens
```

PR #1489, opened 2015-12-10
Show a message when resetting git commit fails

mrkkrp:
See #1453. It should be as simple as this, but it doesn't work. According to actual message the whole `Left` pattern in `readInNull` is not executed (otherwise error message would be different), so that additional message is not logged. I grepped all places where `ProcessExistedUnsuccessfully` is caught, everywhere it's handled similarly to how it's done in `readInNull`. Are you sure it ever works?

Perhaps `Left (ProcessExitedUnsuccessfully _ ec)` is not correct approach here? Shouldn't we get existentially qualified `SomeException` with actual typeable exception inside it which then can be extracted with help of `fromException`? Maybe that's why pattern matching fails?

Maybe I'm wrong, I'll return to this later.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -477,7 +477,8 @@ resolvePackageLocation menv projRoot (PLRemote url remotePackageType) = do
                     Nothing
                 readInNull dirTmp commandName menv
                     (resetCommand ++ [T.unpack commit])
-                    Nothing
+                    (Just $ "Please ensure that commit " <> commit <>
+                      " exists within " <> url)
 
         case remotePackageType of
             RPTHttpTarball -> do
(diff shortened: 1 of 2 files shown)


mgsloan:
Thanks!

mrkkrp:
@mgsloan, Actually it was rather WIP, I didn't expect you merge this instantly. What about that problem with error reporting in `readInNull`?

mgsloan:
Ah, my bad!  The change looked good so I merged.  Fixed the `readInNull` thing here.  [`sinkProcessStdout` rethrows `ProcessExitedUnsuccessfully` as `ReadProcessException`](https://github.com/fpco/stack/blob/984271f4cdc88406188bb7638387193b419f528f/src/System/Process/Read.hs#L247).

PR #1490, opened 2015-12-10
Remove superfluous bang pattern

Jubobs:
See http://stackoverflow.com/questions/34200998/is-a-bang-pattern-on-unit-really-needed-here.


mgsloan:
Good point!

PR #1492, opened 2015-12-10
improve unicode handling

### LINKED ISSUE #1337: unicode in project names is very poorly handled (in 'stack new')
I apologize in advance for not diligently following the `CONTRIBUTING.md` format, but I'm not actually sure what the expected result of these commands should be (read below for why). But regardless of what it should be, there's a problem.

To preface, 'ば', '日' and '本' are all letters (according to `isLetter` from `Data.Char`). Now let's see what happens when we try to create a new project with each of them as a name:

This one just completely gives you an incorrect project name:
```
$ stack new "ば"                      
Downloading template "new-template" to create project "p" in p/ ...
```

This one fails with a fairly bizarre-to-the-user error:
```
$ stack new "日"
Cannot decode byte '\xe5': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```

And this one is a complicated ball of I'm not sure what:
```
$ stack new "本"
Expected valid package name, but got: 本
```
- '本' is a letter, according to `isLetter` from `Data.Char` at least, so it should be allowed
- But are cabal package names in ASCII, or can you actually use unicode?
- If it *is* supposed to fail, it only accidentally fails. Turns out it's truncating '本' down to ',' via `Data.ByteString.Char8.pack` and then fails because ',' isn't a valid package name.

What's happening is:
[the optparse-applicative parser](https://github.com/commercialhaskell/stack/blob/455181cd6f2b536e6cd6ae0f3fd20f60d424caee/src/Stack/Types/PackageName.hs#L173) is calling [`parsePackageNameFromString`](https://github.com/commercialhaskell/stack/blob/455181cd6f2b536e6cd6ae0f3fd20f60d424caee/src/Stack/Types/PackageName.hs#L115) which uses `Data.ByteString.Char8.pack`, truncating any `Char`s outside of the `Word8` range. Then [`packageNameParser`](https://github.com/commercialhaskell/stack/blob/455181cd6f2b536e6cd6ae0f3fd20f60d424caee/src/Stack/Types/PackageName.hs#L90) is used to parse the result, yielding some pretty odd outcomes.

The `decodeUtf8` error message on "日" is via `packageNameText` in some logging in the [`New` command code](https://github.com/commercialhaskell/stack/blob/455181cd6f2b536e6cd6ae0f3fd20f60d424caee/src/Stack/New.hs#L104)

So, my question is: what should actually be done to fix this? In my mind, there's two options:

1. Detect package names outside of ASCII range and reject them outright with a sane error message.
2. Correctly handle unicode characters, which I'm not sure how extensive the changes required would be. Probably a lot of code paths would have to at least be looked at, and `PackageName` would have to change at least a moderate amount.

I suspect that the first choice is the correct one, but can anyone confirm?

The above behavior is identical between the following two versions:
```
$ stack --version
Version 0.1.7.0, Git revision 2f00f7bd350192cef1c61a8f07cbe7341c1e735f (2555 commits) x86_64
```
```
$ stack --version
Version 0.1.6.0, Git revision e22271f5ce9afa2cb5be3bad9cafa392c623f85c (2313 commits) x86_64
```

kadoban:
This is a WIP attempt to fix #1337.

It's a refactor with very minor changes of my previous attempt (mostly cleaned up history and manually reviewed the changes I'd done to make sure they ... actually made sense).

Comments welcome, and I'm still testing and working on adding integration tests and such.

### DIFF
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -19,7 +19,6 @@ import           Control.Monad.IO.Class
 import           Control.Monad.Logger (MonadLogger, logWarn)
 import           Control.Monad.RWS.Strict
 import           Control.Monad.Trans.Resource
-import qualified Data.ByteString.Char8 as S8
 import           Data.Either
 import           Data.Function
 import           Data.List
@@ -180,7 +179,7 @@ constructPlan mbp0 baseConfigOpts0 locals extraToBuild0 localDumpPkgs loadPackag
         , combinedMap = combineMap sourceMap installedMap
         , toolToPackages = \ (Dependency name _) ->
           maybe Map.empty (Map.fromSet (const anyVersion)) $
-          Map.lookup (S8.pack . packageNameString . fromCabalPackageName $ name) toolMap
+          Map.lookup (T.pack . packageNameString . fromCabalPackageName $ name) toolMap
         , ctxEnvConfig = econfig
         , callStack = []
         , extraToBuild = extraToBuild0
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -24,19 +24,20 @@ import           Control.Concurrent.MVar.Lifted
 import           Control.Concurrent.STM
 import           Control.Exception.Enclosed (catchIO, tryIO)
 import           Control.Exception.Lifted
-import           Control.Monad (liftM, when, unless, void, join, guard, filterM, (<=<))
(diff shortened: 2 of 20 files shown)


kadoban:
The travis build failure appears to be because `Text` isn't an instance of `Binary` in old `base`. Not really quite sure what to do about that yet. Might have to implement `Binary` manually for `PackageName` and whatever else to support old `base`.

mgsloan:
LGTM, thanks!

PR #1499, opened 2015-12-12
Delete verbosity option guidance

harendra-kumar:
After the fix for #1472 the guidance about --verbosity silent is not needed.


mgsloan:
Thanks!

PR #1500, opened 2015-12-12
unicode integration tests

kadoban:
Create some integration tests to exercise some unicode usage in as many ways as I can think of.

Note: currently broken, possibly due to a cabal bug ... or a stack bug, needs investigation.


mgsloan:
Cool, LGTM so far!

kadoban:
Hmm, I didn't expect the CI checks to pass. Is there no CI check that the integration tests actually pass? At least on my home build, the new ones I wrote fail (I get a weird cabal error when I rebuild a unicodified package after changes).

borsboom:
We don't run the integration tests on Travis because they take too long and cause timeouts.  We do have a separate process that runs the integration tests on the master branch, so we at least catch breakages pretty quickly.

kadoban:
Oh I see, thanks. That makes much more sense, I was starting to think the error was only on my side. Back to investigating it then :)

kadoban:
It seems like the problem I'm seeing is Cabal's issue, as far as I can tell. I've removed the problematic final commit of this PR, and what's left at least runs correctly for me. I'll make a new PR with further tests at some point in the future, once I do some work trying to figure out what's up with Cabal's issues with unicode.

What's left can IMO be merged, if it seems like a useful test to you. The unicode handling is really fragile, but it might be worth at least making sure it doesn't get worse with future changes.

It's unfortunate that stack being able to basically handle unicode won't be practically useful in the meantime, but at least the base of it is there.

mgsloan:
Looks great, thanks!

Maybe an issue should be opened on the cabal tracker?  Seems like there are [a few issues involving unicode](https://github.com/haskell/cabal/search?q=unicode&state=open&type=Issues&utf8=%E2%9C%93).  In particular, [this one](https://github.com/haskell/cabal/issues/2557) seems likely to cause problems.

PR #1505, opened 2015-12-12
update link in doc/README.md

keent:


mgsloan:
Thanks, but the links are for http://docs.haskellstack.org/en/stable/README.html .  I've opened up #1506 for implementing a systemic resolution of this issue.

PR #1507, opened 2015-12-13
give a sane error when path containing ':' is added to $PATH

### LINKED ISSUE #1387: Directories with colon `:` corrupt path - add graceful error
Stack version:
```
$ stack --version
Version 0.1.8.0, Git revision e33dd995e292efcf8ee42e1e4c649b96232abf2e (2603 commits) x86_64
```

Steps to reproduce:

```
mkdir some:dir
cd some:dir
stack new simple
stack init
stack build
stack exec -- simple-exe
```

Expected output:

```
someFunc
```

The actual output is something like this:

```
Executable named simple-exe not found on path: ["/home/int-index/Documents/some","dir/.stack-work/install/x86_64-linux/lts-3.14/7.10.2/bin", ... <omitted>
```

Possible solution: escape colons before adding the current dir to `$PATH`.

kadoban:
closes #1387

### DIFF
--- a/src/System/Process/Read.hs
+++ b/src/System/Process/Read.hs
@@ -356,15 +356,15 @@ instance Exception PathException
 instance Show PathException where
     show (PathsInvalidInPath paths) = unlines $
         [ "Would need to add some paths to the PATH environment variable \
-          \to continue, but they would be invalid because they contain a \
-          \colon."
+          \to continue, but they would be invalid because they contain a "
+          ++ show FP.searchPathSeparator ++ "."
         , "Please fix the following paths and try again:"
         ] ++ paths
 
 -- | Augment the PATH environment variable with the given extra paths.
 augmentPath :: MonadThrow m => [FilePath] -> Maybe Text -> m Text
 augmentPath dirs mpath =
-  do let illegal = filter (':' `elem`) dirs
+  do let illegal = filter (FP.searchPathSeparator `elem`) dirs
      unless (null illegal) (throwM $ PathsInvalidInPath illegal)
      return $ T.intercalate (T.singleton FP.searchPathSeparator)
             $ map (T.pack . FP.dropTrailingPathSeparator) dirs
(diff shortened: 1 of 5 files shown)


mgsloan:
Thanks!  Ideally, this exception would get thrown in all the places that we're adding things to `PATH` in stack, and not just in the setup code.  Here's the first two I found: [(1)](https://github.com/fpco/stack/blob/628a5670fe0c919969068184a77b02aa9a200551/src/Stack/Docker.hs#L291) and [(2)](https://github.com/fpco/stack/blob/628a5670fe0c919969068184a77b02aa9a200551/src/System/Process/Read.hs#L360).  Maybe all PATH additions should be made with `augmentPathMap`?  Alternatively, you could have a function which checks every component of the PATH.

kadoban:
Ah hmm, good catch. I'll take another peek and see if I can figure out the right way to handle those too ideally without duplicating a bunch of code.

kadoban:
How does that attempt look?

(Note: history needs cleanup to remove the first commit and the revert of it before merging).

I'm not really sure if this was the right way to go or not. It seems to work and it's not phenomenally ugly or anything, but ... just not sure if changing `augmentPath` (and `augmentPathMap`) to be able to fail with that exception is the best way to go or if there's another that I'm missing.

src/System/Process/Read.hs
```diff
@@ -349,17 +349,32 @@ getEnvOverride platform =
...
+        ] ++ paths
+
 -- | Augment the PATH environment variable with the given extra paths.
-augmentPath :: [FilePath] -> Maybe Text -> Text
+augmentPath :: MonadThrow m => [FilePath] -> Maybe Text -> m Text
 augmentPath dirs mpath =
-    T.intercalate (T.singleton FP.searchPathSeparator)
-    $ map (T.pack . FP.dropTrailingPathSeparator) dirs
-   ++ maybe [] return mpath
+  do let illegal = filter (':' `elem`) dirs
```
borsboom: Will this work on Windows, e.g. if a path like `C:\Foo` is given?

kadoban: Uh oh; I suspect not! I'll try to make sure and/or figure out what to do instead.

mgsloan:
LGTM!  I'll merge after the mentioned rebasing is done.

kadoban:
Cool, thanks. Rebase cleanup done.

Issue #1509, opened 2015-12-13
Stack docker does not work with GHC HEAD (`ghc-pkg dump` format change)


CRogers:
I've made a docker image of the optimised release build of GHC HEAD (which I intend to trigger/push nightly builds of) but it sadly doesn't work with stack. It seems that `ghc-pkg` lists dependencies in a different format than stack is expecting.

**Steps to reproduce:**

* Clone this gist: https://gist.github.com/CRogers/85ad65c454753daade55 (`git clone https://gist.github.com/85ad65c454753daade55.git`)
* Run `stack --skip-ghc-check -v build`

**Expected:**

* The project builds, using the HEAD version of ghc inside the docker container.

**Actual:**

The version of stack I'm running:

```
$ stack --version
Version 0.1.10.0 x86_64
```

I'm on OS X running with docker-machine and docker 1.9.1. The command fails with this output:

```
$ stack --skip-ghc-check -v build
Version 0.1.10.0 x86_64
2015-12-13 17:36:38.330166: [debug] Checking for project config at: /Users/callumr/Dropbox/Programming/Haskell/ghc-head/stack-test/stack.yaml @(stack_15maE5vLiHaHnQnC7u59lN:Stack.Config src/Stack/Config.hs:576:9)
2015-12-13 17:36:38.330435: [debug] Loading project config file stack.yaml @(stack_15maE5vLiHaHnQnC7u59lN:Stack.Config src/Stack/Config.hs:599:13)
2015-12-13 17:36:38.332695: [debug] Run process: docker --version @(stack_15maE5vLiHaHnQnC7u59lN:System.Process.Read src/System/Process/Read.hs:273:3)
2015-12-13 17:36:38.340489: [debug] Run process: docker inspect 2be8010a53ad @(stack_15maE5vLiHaHnQnC7u59lN:System.Process.Read src/System/Process/Read.hs:273:3)
[...]

PR #1510, opened 2015-12-13
Deal with `ghc-pkg dump` format change in GHC HEAD.

CRogers:
See #1509.

Only has a failing test for now - I'm not sure how to fix it.


mgsloan:
Thanks!  Fixed the parsing issue in 9ee38c8fbee591160189d545eb6697620abc0ca0

PR #1511, opened 2015-12-13
Restrict commands allowed in interpreter mode

### LINKED ISSUE #1504: Interpreter allows arbitrary commands to be executed via the comment
For example:

```
cueball$ cat test.hs 
#!/usr/bin/env stack
-- stack exec ghci

main = putStrLn "Hello!"

cueball$ stack test.hs
GHCi, version 7.10.2: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( test.hs, interpreted )
Ok, modules loaded: Main.
*Main> 
```

It can confuse and surprise if one writes a wrong comment.

I have a fix which restricts the commands to `runhaskell` and `runghc` in the interpreter mode.

harendra-kumar:
Stack interpreter comment annotation allows arbitrary commands to be
written and executed when the file is run. It can lead to confusing
and surprising behavior if mistakes are made in writing a proper
comment.

This change restricts the interpreter mode commands to runghc and
runhaskell. This change moves add-commands to a separate function.

Closes #1504

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -9,6 +9,7 @@ module Stack.Docker
   ,CleanupAction(..)
   ,dockerCleanupCmdName
   ,dockerCmdName
+  ,dockerHelpOptName
   ,dockerPullCmdName
   ,entrypoint
   ,preventInContainer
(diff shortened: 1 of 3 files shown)


mgsloan:
Thanks!

Issue #1486, opened 2015-12-10
Prevent extra-deps / packages from overriding GHCJS boot packages


mgsloan:
Related to https://github.com/ghcjs/ghcjs/issues/434

tolysz:
My solution would be to:
redo `ghcjs-boot` with the `extra` packages set if they are used in the boot; then filter them from the extra section in `stack.yaml`
Still looking hard how to boot ghcjs purely on the packages from stackage (+ patches); 

For sure we might need some extra resolver just for booting `ghcjs-boot-lts-3.17.yaml`
So we can keep sane collections of bootable set.

This will require splitting the setup phase into 2 sections:
1) install ghcjs
2) boot it with a suitable `-B` directory.

mgsloan:
@tolysz Thanks for the input.  I should have made the issue clearer.

The idea is to detect the circumstance where something bad will happen due to using an unpatched ghcjs package, and error out in this case.  It might need to be just a noisy warning when it's a local package, though, as the user might have a patched version in that case.

mgsloan:
Here's how I think this should work:

* Ensure that the patched packages (+ deps) come from the global DB.  This is the list from [this closed PR](https://github.com/commercialhaskell/stack/pull/1514).  Throw errors if they're found in other DBs.

* Throw errors if the user attempts to override these packages with hackage versions.  Perhaps warn when they attempt to override some other way (local package, git repo)?

* Allow any boot package to differ from the version specified by the snapshot.  This means that you can use the booted `ghcjs-base` without having it in your packages list.

I haven't fully thought out the ramifications of this change.  In particular, the case of installing a custom ghcjs-base and updating the other deps seems tricky.

snoyberg:
The next version of Stack will no longer have GHCJS support, closing.

PR #1514, opened 2015-12-14
Remove unnecessary pkgs from ghcjs-boot list

mgsloan:
See [this comment](https://github.com/ghcjs/ghcjs/issues/448#issuecomment-164260475) - turns out not all the members of [`ghcjsBootPackages`](https://github.com/fpco/stack/blob/e43d6bf8fb1463841ce57162986d2927b0e0d747/src/Stack/Constants.hs#L313) are necessary.  Instead, just the patched packages and their dependencies are needed.

### DIFF
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -315,44 +315,31 @@ ghcjsBootPackages =
     maybe (error "Parse error in ghcjsBootPackages") HashSet.fromList mparsed
   where
     mparsed = mapM parsePackageName
-      -- stage1a
-      [ "array"
+      -- This list consists of packages from stage 1 of ghcjs-boot -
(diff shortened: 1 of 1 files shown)


tolysz:
If we actually could query for this list the `lib/etc/boot.yaml` ;) this would be superb.

mgsloan:
Agreed, having such things hardcoded is ugly.

A tradeoff with making this change, pointed out [here](https://github.com/ghcjs/ghcjs/issues/448#issuecomment-164446570), is that it means every ghcjs project would need to have `ghcjs-base` listed in its packages.  This seems like too high of a cost, so I'd rather have [an approach that can do both](https://github.com/commercialhaskell/stack/issues/1486#issuecomment-164593702).

PR #1523, opened 2015-12-16
very minor code consistency fixup in Options

kadoban:
Noticed this one line doesn't fit in with the rest of the options parser


kadoban:
Err, looks like the travis CI build errored itself, doesn't look to be because of the code change.

borsboom:
Thanks!

PR #1525, opened 2015-12-16
Interpreter error messages

### LINKED ISSUE #1516: Provide better error reporting in interpreter mode
There are multiple ways in which we can fail while trying to execute a file using stack as interpreter. For example:
* The file may not exist 
* or it may not have a stack comment at expected position 
* or the comment may have an unterminated string 
* or a block comment may be unterminated
* or the command parsing may encounter an error

Currently for all the cases we just provide a single blanket error message. It would be helpful for the users if we could provide better error reporting pointing out the source of error.

harendra-kumar:
Add better diagnostics to interpreter options parsing. More details in the commit messages.

Its best to review the commits spearately. The first commit just splits a file, no other change. The second one is most of the changes.

Closes #1516

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1361,7 +1361,15 @@ ghc` or `stack exec runghc` for that. As simple helpers, we also provide the
 stack also offers a very useful feature for running files: a script
 interpreter. For too long have Haskellers felt shackled to bash or Python
 because it's just too hard to create reusable source-only Haskell scripts.
-stack attempts to solve that. An example will be easiest to understand:
+stack attempts to solve that.
+
+You can use `stack <file name>` to execute a Haskell source file or specify
+`stack` as the interpreter using a shebang line on a Unix like operating systems.
+Additional stack options can be specified using a special Haskell comment in
+the source file to specify dependencies and automatically install them before
+running the file.
+
+An example will be easiest to understand:
 
 ```
 michael@d30748af6d3d:~$ cat turtle.hs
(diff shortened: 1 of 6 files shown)


mgsloan:
Great, thanks for working on this!  Based on a quick look, the change looks good to me.

I'm not sure if want `stack File.hs` to imply `stack runhaskell File.hs`.  I tend to prefer being explicit than adding special defaults.  @snoyberg @borsboom Thoughts?

harendra-kumar:
Few more thoughts on this as I care a bit about it.

I see stack as a tool which can help greatly in removing obstacles in the way of Haskell adoption. To achieve that goal we have to make things really easy for users. If we provide a way to run scripts via stack then it better work all the time. If I can run some scripts using `stack File.hs` and not others then it is an inconsistency from the user's perspective. I want to be able to have __just one way__ to run any Haskell script so that I don't need to remember multiple ways and get confused about when can I use and when can I use that, especially so when non-geeky users want to run scripts written by others.

Comparing it with other scripting languages, take python or perl for example. You can always say `python file.py` or `perl file.pl` and that's it. We know stack is different it allows you to specify dependencies from a snapshot. But do the **regular users** care about that detail? No. From a user's perspective it will be a WTF moment. You can run some scripts using 'stack File.hs' but for others you need to use `runghc File.hs` and perhaps they will discover that with some frustration. In contrast if stack is popular and successful user's don't need to know any other way of running scripts just say `stack File.hs` - easy to learn and propagate the knowledge.

Let's make stack just awesome!

mgsloan:
> If I can run some scripts using stack File.hs and not others then it is an inconsistency from the user's perspective.

I see, I didn't realize it worked that way!  Yeah, let's go ahead and allow it in general.

> Let's make stack just awesome!

Agreed!  Thanks for your efforts in making it more awesome :)

harendra-kumar:
Great! Thanks!

PR #1526, opened 2015-12-17
Added pointers for creating your own .deb packages

DrDub:
Some pointers of where to look to create your own .deb file.

At the time of this writing the following instructions get the job done:

You will need to have fpm installed, though. As root:

apt-get install ruby-dev libffi-dev make build-essential rubygems
gem install fpm  --version '< 1.4.0'

(assuming a binary version of stack is in the path and the current directory is the root of the git version)

cd etc/scripts
stack --install-ghc-build
RELEASE_SCRIPT=`stack exec which stack-release-script`
cd ../..
RELEASE_SCRIPT --upload-label='Linux 64-bit, standard' --allow-dirty build-debian-8

the .deb file will be located under the _release folder.

(I didn't add these instructions to the documentation itself as it will get hopelessly out of date very quickly, but feel free to do so if you think there'll be value on it.)

### DIFF
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -140,6 +140,10 @@ Yosemite and Mavericks as well, and may also work on older versions (YMMV).
 
         sudo apt-get update && sudo apt-get install stack -y
 
+### Building your own package
+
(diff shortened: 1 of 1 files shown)


borsboom:
Thanks.  I wonder if it might be better to augment https://github.com/commercialhaskell/stack/blob/master/etc/scripts/README.md instead and then point there.  I'm also not really sure why someone installing/upgrading Stack would need to make their own .deb.  For someone making packages, the https://github.com/commercialhaskell/stack/blob/master/doc/MAINTAINER_GUIDE.md might be a more appropriate place to look.

borsboom:
@DrDub are you planning any revisions to this PR?

borsboom:
Closing for now since I haven't heard back.  Please re-open if you'd like to make any revisions.

Issue #1251, opened 2015-10-28
Running stack templates bumps into Github API limit


bitemyapp:
> Failed to download templates. The HTTP error was:

``` haskell
StatusCodeException (Status {statusCode = 403, statusMessage = "Forbidden"}) [("Server","GitHub.com"),("Date","Wed, 28 Oct 2015 23:39:51 GMT"),("Content-Type","application/json; charset=utf-8"),("Transfer-Encoding","chunked"),("Status","403 Forbidden"),("X-RateLimit-Limit","60"),("X-RateLimit-Remaining","0"),("X-RateLimit-Reset","1446077019"),("X-GitHub-Media-Type","github.v3; format=json"),("X-XSS-Protection","1; mode=block"),("X-Frame-Options","deny"),("Content-Security-Policy","default-src 'none'"),("Access-Control-Allow-Credentials","true"),("Access-Control-Expose-Headers","ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval"),("Access-Control-Allow-Origin","*"),("Strict-Transport-Security","max-age=31536000; includeSubdomains; preload"),("X-Content-Type-Options","nosniff"),("Content-Encoding","gzip"),("X-GitHub-Request-Id","9D82BA36:146C7:57FF7EF:56315CC7"),("X-Response-Body-Start","{\"message\":\"API rate limit exceeded for 157.130.186.54. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)\",\"documentation_url\":\"https://developer.github.com/v3/#rate-limiting\"}"),("X-Request-URL","GET https://api.github.com:443/repos/commercialhaskell/stack-templates/contents/")] (CJ {expose = []})
```

Comes by way of @stephenjbarr

I don't know how Github limits API requests here. I imagined it is partly keyed by the resource being accessed, but I don't know what else factors in (IP address?)

Does anyone know the usual way to cope with this without obligating the user to sign up for credentials?

borsboom:
[...]

PR #1532, opened 2015-12-17
improve 'stack new'

kadoban:
- You can now specify an explicit URL to download the template from
- I've improved the logging/output so that it doesn't say it's "Downloading" a local file
- You can specify a default-template in ~/.stack/config.yaml, which is nice if you have a favorite

I considered adding the ability to configure the template repository location, but I'm thinking I might leave that for a separate PR later if it sounds interesting to anyone. I think it'd be a decent feature, but it's a little complicated by the existence of the `stack templates` command, which I haven't really worked out how to deal with.

I'm working on some tests to exercise this functionality and make sure, but it basically seems to work fine to me.

### DIFF
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -150,7 +150,6 @@ loadTemplate name logIt = do
         $logDebug ("Opening local template: \"" <> T.pack (toFilePath path)
                                                 <> "\"")
         exists <- fileExists path
-        unless exists ($logDebug "Template file doesn't exist")
         if exists
             then liftIO (T.readFile (toFilePath path))
             else throwM (FailedToLoadTemplate name (toFilePath path))
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -30,6 +30,7 @@ import           Data.Conduit
 import           Data.List
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as M
+import           Data.Maybe (fromMaybe)
 import           Data.Maybe.Extra (mapMaybeM)
 import           Data.Monoid
 import           Data.Set (Set)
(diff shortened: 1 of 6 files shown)


src/Stack/New.hs
```diff
@@ -107,37 +118,53 @@ new opts = do
...
+                                                     (templateDir </> relFile)
+                        Nothing -> throwM e
+                )
   where
     loadLocalFile :: Path b File -> m Text
     loadLocalFile path = do
+        $logDebug ("Opening local template: \"" <> T.pack (toFilePath path)
+                                                <> "\"")
         exists <- fileExists path
+        unless exists ($logDebug "Template file doesn't exist")
```
mgsloan: Are these `$logDebug`s intended to be visible to the user?  They'll only be present in the `-v` output.

It probably makes sense to delete this one since it's a bit redundant with the exception.  But perhaps it helps resolve ambiguity when debugging?

mgsloan:
LGTM!

> I considered adding the ability to configure the template repository location, but I'm thinking I might leave that for a separate PR later if it sounds interesting to anyone. I think it'd be a decent feature, but it's a little complicated by the existence of the stack templates command, which I haven't really worked out how to deal with.

Yeah, this would be nice, particularly if a file URL would just use a folder for the templates.  I think it might make sense to change how templates are done, in order to support this + to work offline: https://github.com/commercialhaskell/stack/issues/1251#issuecomment-165654478

Of course, that's bonus points, merging this isn't predicated on doing that.

src/Stack/New.hs
kadoban: Ah yeah, that is quite redundant now. I'll nuke it. (I think it exists from when I was doing my own debug briefly and just forgot to remove it)

kadoban:
> LGTM!

Yay :)

> > I considered adding the ability to configure the template repository location, but I'm thinking I might leave that for a separate PR later if it sounds interesting to anyone. I think it'd be a decent feature, but it's a little complicated by the existence of the stack templates command, which I haven't really worked out how to deal with.

> Yeah, this would be nice, particularly if a file URL would just use a folder for the templates. I think it might make sense to change how templates are done, in order to support this + to work offline: #1251 (comment)

Yeah I agree, that sounds like quite a nice way to go. I'll ponder a bit and respond in that thread if anything interesting comes to mind. I'd be interested in working on improving `new` more in that direction if someone else doesn't get to it first (but they might, it'll probably be a few days before I have time).

> Of course, that's bonus points, merging this isn't predicated on doing that.

Cool. Yeah I'd say most/all of this should end up useful even once something in that direction gets implemented, if only because requiring git probably isn't ideal (so it has to work without it at least in some fashion).

PR #1537, opened 2015-12-18
doc: update instructions in nix_integration.md to work without nix-repl

peti:
Instead, we use nix-instantiate and nix-env, which are core tools available on every Nix installation.


YPares:
Good idea. I don't know if appveyor build failure should block the merge (amazon server timed out), though. Does someone know if the build is automatically retried?

YPares:
I will add a note about the existence of `nix-repl`, because in the long run it's really more convenient than `nix-instantiate`.

Issue #1536, opened 2015-12-18
The command "stack --nix exec cmd --" passes unexpected arguments.


peti:
See https://gist.github.com/peti/ed64f5a9ac193690ec10 for an easily reproducible test case. Basically, the command

    $ stack --nix build stack-bug
    $ stack --nix exec stack-bug
    []

executes the `stack-bug` command without any arguments -- like you'd expect. The command

    $ stack --nix exec stack-bug --
    ["--extra-lib-dirs=/nix/store/j1q555xv31177kg9ncc9y0p3xaqbq1gf-ghc-7.10.2/lib","--extra-include-dirs=/nix/store/j1q555xv31177kg9ncc9y0p3xaqbq1gf-ghc-7.10.2/include"]

however, does not. Instead, it adds additional flags that look like they're intended for `cabal configure`.

This means that it's basically impossible to pass command line options to programs executed via `stack --nix exec ...`.

YPares:
It is indeed the same bug as #1538

PR #1539, opened 2015-12-18
Nix: Pass $STACK_IN_NIX_EXTRA_ARGS before command args.

### LINKED ISSUE #1538: stack --nix exec -- bash fails
The following works:

```
$ stack new foo
$ cd foo
$ stack exec bash
```

However, the following does not:

```
$ stack exec -- bash
/nix/store/gb2dwphbdjk56h0lbbyma32rlid022ad-bash-4.3-p42/bin/bash: --extra-lib-dirs=/nix/store/gf16fp2wxbfq0c6r7j6386gkk4r3aqdr-apache-maven-3.3.3/lib : option non valable
```

The problem seems to be that if you use `--` to delimit arguments passed to the subcommand, Stack for some reason passes extra arguments to the subcommand, which it of course fails to recognize. Without `--` this seems to work fine.

luigy:
fixes #1538 #1536

### DIFF
--- a/src/Stack/Nix.hs
+++ b/src/Stack/Nix.hs
@@ -10,6 +10,7 @@ module Stack.Nix
   ) where
 
 import           Control.Applicative
+import           Control.Arrow ((***))
 import           Control.Monad
 import           Control.Monad.Catch (try,MonadCatch)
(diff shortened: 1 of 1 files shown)


YPares:
Thanks! I'm merging

PR #1542, opened 2015-12-19
Add zsh completion function

vapniks:
The files _completion_helpers and _stack provide much better completion
than the bash completer. They provide command and options descriptions, and
completion of subcommands and option arguments.
The completion function is dynamically generated based on --help output
for stack and it's subcommands, so that it should always be up to date
as long as the help output keeps the same format.

### DIFF
--- a/doc/_completion_helpers
+++ b/doc/_completion_helpers
@@ -1,0 +1,239 @@
+#autoload
+
+# Filename: _completion_helpers
+# Description: Extra utility functions to help create zsh completion functions
+# Author: Joe Bloggs <vapniks@yahoo.com>
+# Copyleft (Ↄ) 2015, Joe Bloggs, all rites reversed.
+# Created: 2015-12-19 19:00:00
+#           By: Joe Bloggs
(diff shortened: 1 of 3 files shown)


borsboom:
Thanks for this, but unfortunately I don't think we can guarantee that --help output will remain stable (in fact, we're planning changes to it).  A better approach would be to add support for generating zsh completions to optparse-applicative (it was mentioned in https://github.com/pcapriotti/optparse-applicative/issues/10, but looks like only bash was implemented).  Another advantage of that is that when we make improvements to shell autocompletion (e.g. autocompleting not just option names but also valid values), the zsh completion would also use that.   As such, I'm reluctant to add these completion scripts to the repo since I doubt we'll maintain them.  A better place would be the [wiki](https://github.com/commercialhaskell/stack/wiki).  You could still link to the wiki from from `doc/shell_autocompletion.md`, though.

vapniks:
OK, at some point in the future I might look into that. In the meantime if anyone wants a better stack zsh completer, check my fork. It shouldn't be too hard to fix if any changes are made to --help output in the future.

PR #1545, opened 2015-12-20
faq.md: Don't split a link across a line break

mineo:
At the moment, the link is not actually rendered as a link at http://docs.haskellstack.org/en/stable/faq.html#i-get-strange-ld-errors-about-recompiling-with-fpic, this should (hopefully, since I couldn't find out how to actually build the documentation) fix that.


mgsloan:
Thanks!

PR #1550, opened 2015-12-21
All test suites run even when only one is requested

DaveCTurner:
In a previous version of `stack`, running

    stack test my-project:test:test-suite-1

would only run `test-suite-1`. The behaviour now seems to be to run all test suites.

Furthermore, running

    stack clean
    stack test my-project:test:test-suite-1

only seems to _build_ the given test suite but then attempts to run them all, and obviously the rest of them fail with `executable not found`.


DaveCTurner:
This seems to work, although it's quite a local change so may have wider consequences?

mgsloan:
LGTM, thanks!

PR #1555, opened 2015-12-22
give better errors for broken templates

### LINKED ISSUE #1535: edge cases in broken templates give odd errors
These are rather dumb, but I could still see them coming up in practice and it might not be the worst idea to handle them:

##### Steps to Reproduce
```
~$ touch blah.hsfiles
~$ stack new blah blah
Downloading template "blah" to create project "blah" in blah/ ...
/home/mud/blah/: getDirectoryContents: does not exist (No such file or directory)
```
##### Expected
An error message about the template being empty.

##### Steps to Reproduce
```
~$ printf "{-# START_FILE BOGUS #-}\n" > blah.hsfiles
~$ stack new blah blah
Downloading template "blah" to create project "blah" in blah/ ...
Writing default config file to: /home/mud/blah/stack.yaml
Basing on cabal files:

In order to init, you should have an existing .cabal file. Please try "stack new" instead
```

##### Expected
Either an `init`ed package or (since that's probably not really sane without a .cabal file) a better error message about the template being broken. Edit: I just noticed that it works file with an *empty* .cabal file, so maybe it would make sense for stack.yaml to do the same thing when .cabal doesn't exist?

In particular it's kind of amusing, but not really helpful, that it tells you to run "stack new" when that's what you just tried.

If there's a real use for a template that doesn't include a .cabal file (I can't think of one, but maybe someone else can), it might be acceptable to just have a big warning about why init didn't do anything, and ideally remove the suggestion about "stack new".

##### Version
```
Version 0.1.10.1, Git revision 83ec40fc5fd6d300c73d23deb1f41d30168fc236 (2783 commits) x86_64
```

kadoban:
fixes #1535

### DIFF
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -197,6 +197,9 @@ applyTemplate project template nonceParams dir templateText = do
                    throwM (InvalidTemplate template (show e)))
     when (M.null files) $
          throwM (InvalidTemplate template "Template does not contain any files")
+    unless (any (".cabal" `isSuffixOf`) . M.keys $ files) $
+         throwM (InvalidTemplate template "Template does not contain a .cabal\
+                                          \ file")
     liftM
         M.fromList
         (mapM
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -167,7 +167,7 @@ loadTemplate name logIt = do
 
 -- | Apply and unpack a template into a directory.
 applyTemplate
-    :: (MonadIO m, MonadThrow m, MonadReader r m, HasConfig r, MonadLogger m)
+    :: (MonadIO m, MonadThrow m, MonadCatch m, MonadReader r m, HasConfig r, MonadLogger m)
     => PackageName
     -> TemplateName
     -> Map Text Text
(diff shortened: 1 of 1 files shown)


kadoban:
Upon review, the error messages sound rather badly worded; I'll fix them up in the morning.

kadoban:
I think I'm happier with these errors, and it seems to work fine:
```
~/testing$ touch blah.hsfiles
~/testing$ stack new blah blah                 
Loading local template "blah" to create project "blah" in blah/ ...
The template "blah" is invalid and could not be used. The error was: "Template does not contain any files"
~/testing$ printf "{-# START_FILE BOGUS #-}\n" > blah.hsfiles
~/testing$ stack new blah blah
Loading local template "blah" to create project "blah" in blah/ ...
The template "blah" is invalid and could not be used. The error was: "Template does not contain a .cabal file"
~/testing$ printf "a b c d e" > blah.hsfiles
~/testing$ stack new blah blah
Loading local template "blah" to create project "blah" in blah/ ...
The template "blah" is invalid and could not be used. The error was: "InvalidInput "a b c d e""
```

Note that although the empty template error is technically redundant (it'll obviously be missing a .cabal file too in that case), I think it's probably still useful since it probably points to an *empty* template rather than one that someone forgot to add a .cabal file to.

mgsloan:
LGTM, thanks!

PR #1562, opened 2015-12-25
[docs] Small fix in 'How to contribute' section.

xor-xor:
Following steps from "How to contribute" section, I've noticed that `stack --version` doesn't launch newly-built stack executable and probably should be changed to `stack exec stack -- --version`.


mgsloan:
Thanks!

PR #1567, opened 2015-12-27
stack.cabal: remove the redundant build-dependency on email-validate

vu3rdd:
The build dependency on email-validate from bcf73ec0 seem to be redundant.


borsboom:
Thanks!

PR #1578, opened 2015-12-30
Include the current working directory in process failure exceptions

ndmitchell:
This makes reproducing bugs much easier. Now I see:

   Running C:\Program Files (x86)\Git\cmd\git.exe fetch --tags --depth=1 in directory C:\Users\ndmit_000\AppData\Roaming\stack\indices\Hackage\git-update\all-cabal-hashes\ exited with ExitFailure 128

Before I just saw a git, but the directory it was in took ages to find.


mgsloan:
Makes sense, thanks!

Issue #1533, opened 2015-12-18
Improve stack init


harendra-kumar:
Our goal should be to be able to init/build/install as many packages as possible with little or no intervention by the user. After trying `stack init` a little bit here are a few suggestions to make the experience better.

1) When `stack init` fails there is an intimidating amount of output. We can perhaps keep it shorter and have more details in a verbose mode. We can put a suggestion towards using verbose to get more details.

2) When no snapshot satisfying the dependencies could be found, we should sort the output showing snapshot options putting the best fit first.

3) Going even beyond that - we should automatically pick the best fit and put the rest of the packages in extra-deps. The user can be warned about the fact that some of the packages are not from a stack snapshot.

4) When automatically picking the best snapshot  we have to make sure that we pick only a snapshot which satisfies the compiler wired-in packages (base, ghc-prim, integer-gmp, integer-simple). They cannot go in extra-deps.

5) stack init --solver spits an almost incomprehensible message if the wrong ghc is in the path. To make things better we should be able to choose and install (if necessary) the right compiler. The output comes from cabal. If required we can decode the output and provide a simpler message to the user. Here is a sample of the output:
```
cueball:/vol/hosts/cueball/workspace/halberd$ stack init --solver
Writing default config file to: /vol/hosts/cueball/workspace/halberd/stack.yaml
Basing on cabal files:
- /vol/hosts/cueball/workspace/halberd/halberd.cabal

Asking cabal to calculate a build plan, please wait
Running /var/home/harendra/.local/bin/cabal --config-file=/tmp/cabal-solver13660/cabal.config install --enable-tests --enable-benchmarks -v --dry-run --only-dependencies --reorder-goals --max-backjumps=-1 --package-db=clear --package-db=global /vol/hosts/cueball/workspace/halberd/ exited with ExitFailure 1
/var/home/harendra/.stack/snapshots/x86_64-linux/lts-3.14/7.10.2/bin/cpphs --version
/usr/bin/gcc -dumpversion
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin/haddock --version
/var/home/harendra/.stack/snapshots/x86_64-linux/lts-3.14/7.10.2/bin/happy --version
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin/hpc version
looking for tool hsc2hs near compiler in
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin
found hsc2hs in
[...]

Issue #1529, opened 2015-12-17
Stack picks up the wrong compiler when using stack init --solver


harendra-kumar:
stack is setup to use ghc-7.8.4 from lts-2.22
```
cueball:/vol/hosts/cueball/workspace/halberd$ stack exec which ghc
Run from outside a project, using implicit global project config
Using resolver: lts-2.22 from implicit global project's config file: /var/home/harendra/.stack/global-project/stack.yaml
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.8.4/bin/ghc
```

But cabal invoked by stack init --solver is picking up the one which is in the user path:

```
cueball:/vol/hosts/cueball/workspace/halberd$ stack init --solver
Writing default config file to: /vol/hosts/cueball/workspace/halberd/stack.yaml
Basing on cabal files:
- /vol/hosts/cueball/workspace/halberd/halberd.cabal

Asking cabal to calculate a build plan, please wait
Running /var/home/harendra/.local/bin/cabal --config-file=/tmp/cabal-solver6697/cabal.config install --enable-tests --enable-benchmarks -v --dry-run --only-dependencies --reorder-goals --max-backjumps=-1 --package-db=clear --package-db=global /vol/hosts/cueball/workspace/halberd/ exited with ExitFailure 1
/var/home/harendra/.stack/snapshots/x86_64-linux/lts-3.14/7.10.2/bin/cpphs --version
/usr/bin/gcc -dumpversion
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin/haddock --version
/var/home/harendra/.stack/snapshots/x86_64-linux/lts-3.14/7.10.2/bin/happy --version
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin/hpc version
looking for tool hsc2hs near compiler in
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin
found hsc2hs in
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin/hsc2hs
/var/home/harendra/.stack/programs/x86_64-linux/ghc-7.10.2/bin/hsc2hs --version
/var/home/harendra/.stack/snapshots/x86_64-linux/lts-3.14/7.10.2/bin/HsColour -version
[...]

Issue #1265, opened 2015-11-01
Implicit snapshots (aka: Unify extra-deps and extensible snapshots)


borsboom:
There's a lot of overlap between extra-deps and extensible snapshots.  Both give you a to add some "extra stuff" to a snapshot.  The main difference is that extra-deps are installed into `<project>/.stack-work` and treated as slightly weird local packages, while extensible snapshots will go into `~/.stack/snapshots` and are immutable.  There's a fair amount of confusion around extra-deps (and packages that have `extra-dep: true`) , so instead of having two ways to do similar things, they should be unified into a single coherent approach.

Particular issues currently:
- Most people don't want extra-deps to behave like local packages, especially as regards GHC options
- extra-deps cannot be shared between projects (or copies of the same project)
- The current stack.yaml configuration is confusing

Now that snapshots can share packages between them, there's no longer a good reason why extra-deps have to be local to the project.  Instead, they could form an implicit extended snapshot.

So here, after discussing with @snoyberg, is my proposal for steps to take:
1. Ensure that extensible snapshots (#863) can support everything required to support extra-deps.  That would include:
   - Setting custom ghc-options for packages
   - Getting package source from a remote Git repository ~~or local directory~~ [edit: local directory for extended snapshot would not be a good idea, since contents could change]
2. Modify stack so that an _implicit_ custom snapshot is created for extra-deps, without changing the existing stack.yaml configuration.  The implicit custom snapshot's name would be generated by taking a hash the package source location, name, version, flags, and ghc-options (at least) for each package
3. Rationalize the stack.yaml configuration.  In particular, do away with the `extra-deps` section.  All additions to the custom snapshot will be added using a flag in the `packages` section (similar to the existing `extra-dep` flag, but perhaps with a different name like `extend-snapshot`).  The existing configuration would be deprecated but still supported for some time.
   
   Edit: alternatively, keep a separate section for the extended snapshot (akin to current `extra-deps` section), but have it accept exactly the same syntax as the `packages` section (but not accepting local directories).  Pro: a bit more clear; Con: more work to switch packages between local and extended snapshot (you have to move stuff around in stack.yaml instead of just toggling a flag).

ping @mgsloan @zudov @bitemyapp

3noch:
:+1:

mgsloan:
For (3), I'm thinking doing the following makes sense:

* Add a new `dependencies` section, which uses the same syntax as `packages`.

* Deprecate use of `extra-deps`.

[...]

PR #1583, opened 2015-12-31
Stack init and solver overhaul

harendra-kumar:
These commits are part of a series of commits to enhance the `stack init` and `solver` functionality. The goal is to be able to automatically init any project which has a working cabal file. Similarly the `stack solver` command should be able to verify and fix the configuration in an optimal manner. Also, to provide a better user experience by providing informative error messages and next steps at every failure. In other words, init should just work and be smooth and painless and always work unless the package is really broken.

See also #1529 and #1533.

The commit messages have more details about the problems and fixes. I will be adding more commits to the branch very soon. I am pushing these out for an early review. The changes are logically self contained, add incremental functionality and should not break any existing functionality.

**Please review individual commits in order and see the commit messages for an overview of each commit.**

### DIFF
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -695,8 +695,9 @@ selectBestSnapshot gpds snaps = do
           | (Map.size e1) <= (Map.size e2) = (s1, e1)
           | otherwise = (s2, e2)
 
-        reportResult (BuildPlanCheckOk _) snap =
+        reportResult (BuildPlanCheckOk _) snap = do
             $logInfo $ "* Selected " <> renderSnapName snap
+            $logInfo ""
 
         reportResult (BuildPlanCheckPartial _ errs) snap = do
             $logWarn $ "* Partially matches " <> renderSnapName snap
@@ -705,9 +706,9 @@ selectBestSnapshot gpds snaps = do
         reportResult (BuildPlanCheckFail compiler errs) snap = do
             $logWarn $ "* Rejected "
                        <> renderSnapName snap
-                       <> " due to conflicting compiler ("
+                       <> " due to conflict of compiler ("
                        <> compilerVersionText compiler
-                       <> ") requirements"
+                       <> ") packages"
             displayDepErrors errs
 
 displayDepErrors :: MonadLogger m => DepErrors -> m ()
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -27,7 +27,7 @@ module Stack.BuildPlan
 
 import           Control.Applicative
 import           Control.Exception (assert)
-import           Control.Monad (liftM, forM)
+import           Control.Monad (liftM, forM, when)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
@@ -621,9 +621,10 @@ checkBundleBuildPlan platform compiler pool flags gpds =
         dupError _ _ = error "Bug: Duplicate packages are not expected here"
 
 data BuildPlanCheck =
-      BuildPlanCheckFail CompilerVersion DepErrors
-    | BuildPlanCheckOk (Map PackageName (Map FlagName Bool))
+      BuildPlanCheckOk      (Map PackageName (Map FlagName Bool))
     | BuildPlanCheckPartial (Map PackageName (Map FlagName Bool)) DepErrors
+    | BuildPlanCheckFail    (Map PackageName (Map FlagName Bool)) DepErrors
+                            CompilerVersion
 
 -- | Check a set of 'GenericPackageDescription's and a set of flags against a
 -- given snapshot. Returns how well the snapshot satisfies the dependencies of
@@ -651,7 +652,7 @@ checkSnapBuildPlan gpds flags snap = do
     else if Map.null cerrs then do
             return $ BuildPlanCheckPartial f errs
         else
-            return $ BuildPlanCheckFail compiler cerrs
+            return $ BuildPlanCheckFail f cerrs compiler
     where
         compilerErrors compiler errs
             | whichCompiler compiler == Ghc = ghcErrors errs
@@ -683,7 +684,7 @@ selectBestSnapshot gpds snaps = do
             result <- checkSnapBuildPlan gpds Nothing snap
             reportResult result snap
             case result of
-                BuildPlanCheckFail _ _ -> loop bestYet rest
+                BuildPlanCheckFail _ _ _ -> loop bestYet rest
                 BuildPlanCheckOk _ -> return $ Just snap
                 BuildPlanCheckPartial _ e -> do
                     case bestYet of
@@ -699,20 +700,24 @@ selectBestSnapshot gpds snaps = do
             $logInfo $ "* Selected " <> renderSnapName snap
             $logInfo ""
 
-        reportResult (BuildPlanCheckPartial _ errs) snap = do
+        reportResult (BuildPlanCheckPartial f errs) snap = do
             $logWarn $ "* Partially matches " <> renderSnapName snap
-            displayDepErrors errs
+            displayDepErrors f errs
 
-        reportResult (BuildPlanCheckFail compiler errs) snap = do
+        reportResult (BuildPlanCheckFail f errs compiler) snap = do
             $logWarn $ "* Rejected "
                        <> renderSnapName snap
                        <> " due to conflict of compiler ("
                        <> compilerVersionText compiler
                        <> ") packages"
-            displayDepErrors errs
+            displayDepErrors f errs
 
-displayDepErrors :: MonadLogger m => DepErrors -> m ()
-displayDepErrors errs =
+displayDepErrors
+    :: MonadLogger m
+    => Map PackageName (Map FlagName Bool)
+    -> DepErrors
+    -> m ()
+displayDepErrors flags errs =
     F.forM_ (Map.toList errs) $ \(depName, DepError mversion neededBy) -> do
         $logInfo $ T.concat
             [ "    "
@@ -731,7 +736,17 @@ displayDepErrors errs =
             , " requires "
             , T.pack $ display range
             ]
+        F.forM_ (Map.toList neededBy) $ \(user, _) ->
+            maybe (return ()) (printFlags user) (Map.lookup user flags)
         $logInfo ""
+    where
+        printFlags user fl = when (not $ Map.null fl) $
+            $logDebug $ T.concat
+                [ "    - "
+                , T.pack $ packageNameString user
+                , " configured with flags "
+                , T.pack $ (show fl)
+                ]
 
 shadowMiniBuildPlan :: MiniBuildPlan
                     -> Set PackageName
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -706,37 +706,46 @@ findBuildPlan
     => [GenericPackageDescription]
     -> [SnapName]
     -> m (Maybe (SnapName, Map PackageName (Map FlagName Bool)))
-findBuildPlan gpds = do
-    loop Nothing
+findBuildPlan gpds snaps = do
+    $logInfo $ "Selecting the best among "
+               <> T.pack (show (length snaps))
+               <> " snapshots...\n"
+    loop Nothing snaps
     where
         loop Nothing [] = return Nothing
         loop (Just (snap, f, e)) []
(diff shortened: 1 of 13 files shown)


mgsloan:
Wow, this is great stuff, thanks a bunch for working on it!

Along with my commit comments, I noticed the following things that could be enhanced / fixed:

* If I do `stack init` in a directory that has a stack.yaml in an ancestor directory, I get an error.  Instead, I'd prefer that it only checked the current directory for a stack.yaml.  For example, if `~/dir/stack.yaml` exists but `~/dir/subdir/stack.yaml` doesn't:

```
mgsloan@computer:~/dir/subdir$ stack init
Stack configuration file stack.yaml exists, use 'stack solver' to fix the existing config file or '--force' to overwrite it.
```

* I also split off this issue https://github.com/commercialhaskell/stack/issues/1593

* There's an extra space towards the end of this message `- Create pivot points for the solver by specifying some extra dependencies in stack.yaml and then use 'stack solver' to figure out the rest of the  dependencies.`  Also, maybe instead of "Create pivot points for the solver by", say something like "Help the solver out by"

* When some a local package is also in a snapshot, `--solver` doesn't work.  For example, I get [this output](https://gist.github.com/mgsloan/086bfb570a592773d598) when building the latest lens repo.  I think what needs to happen is that the local packages need to get filtered out of `availablePkgs`, used to write out the `cabal.config` file.

* With an internal project, for some reason `stack init --solver` sets a flag like `time-locale-compat:-old-locale` (in the yaml, `old-locale: false`).  `time-locale-compat` does not get added to `extra-deps`, even though it needs to be present in order to work, [though this will change in the future](https://github.com/commercialhaskell/stack/issues/1265#issuecomment-168455198).  Oddly enough, the build works even if I remove the flag.  Unfortunately, I haven't yet come up with a minimal way to reproduce this.

harendra-kumar:
1)
> If I do stack init in a directory that has a stack.yaml in an ancestor directory, I get an error. 

Can't reproduce this, code too looks fine. Maybe you really had a file in the curdir and you thought you didn't?

3) I am fixing the error message, I agree that it sounds too geeky and could be simpler.

4) Even if the local and snapshot package versions conflict, the retry should work since it relaxes the snapshot constraints to preferences. Its a good idea to fix it though.

Looking at the solver output for lens, it seems there is a conflict between the base package of the snapshot and the constraints of gloss required by lens-examples. We need a snapshot with the correct base package - in other words we need a different GHC.

The compiler compatibility check that I have put in is pretty naive as it only works for the packages we are building, it cannot check whether the dependencies of the package are compiler compatible. We will have to parse the cabal error messages for that (yucky!) - or cabal itself can provide a distinctive error.

5) I found a bug with the cabal flags parsing in the solver. I have fixed it, will push the commit soon. Maybe its related to that, try and see if the problem persists with that fix.

mgsloan:
> Can't reproduce this, code too looks fine. Maybe you really had a file in the curdir and you thought you didn't?

Ah, I'm not sure how much the parent stack.yaml affects things (other than potentially overriding some non-project options).  However, here's a repro where if you have an invalid parent stack.yaml, things don't work: https://gist.github.com/mgsloan/2beeac1531b2ef07c651

mgsloan:
> Looking at the solver output for lens, it seems there is a conflict between the base package of the snapshot and the constraints of gloss required by lens-examples. We need a snapshot with the correct base package - in other words we need a different GHC.

Right, that's after it falls back to using preferred versions.  When using hard constraints, it yields `rejecting: lens-4.13.1 (global constraint requires ==4.13)`.  In other words, it's writing down the lens constraint from the snapshot in the cabal config, but that mismatches with our local version.

I found this issue by running `stack init` on all the repos cloned by [diagrams-ci](https://github.com/diagrams/diagrams-ci) (I had to modify the build-universe.sh script to just clone the repos and do nothing else).

I ran into a couple miscellaneous issue while looking into this: If I delete the `version:` field in the cabal file, I get this error:

```
cabal: Error parsing config file /tmp/cabal-solver11856/cabal.config:910:
Parse of field 'constraint' failed (expected a package name followed by a
constraint, which is either a version range, 'installed', 'source' or flags):
lens==
```

> The compiler compatibility check that I have put in is pretty naive as it only works for the packages we are building, it cannot check whether the dependencies of the package are compiler compatible. We will have to parse the cabal error messages for that (yucky!) - or cabal itself can provide a distinctive error.

How about always using hard constraints for wired in packages?

> 5) I found a bug with the cabal flags parsing in the solver. I have fixed it, will push the commit soon. Maybe its related to that, try and see if the problem persists with that fix.

Cool, I do get different behavior - more flags now (`text:-integer-simple` in addition to `time-locale-compat:-old-locale`).  They don't seem to be necessary, as I can delete them.

harendra-kumar:
> here's a repro where if you have an invalid parent stack.yaml

This warrants a separate issue. Perhaps we should avoid loading project config for init. Currently it uses `withConfigAndLock` which parses existing stack.yaml. We don't need to parse it for init, we only need to check the existence of it.

harendra-kumar:
> If I delete the version: field in the cabal file, I get this error:

I did not realize that version could remain empty for source packages. Thanks for catching this! I uploaded the fix. The fix assumes that if the package is in the Hackage index as well cabal will do the right thing and pick up the source version instead of the index version.

harendra-kumar:
> How about always using hard constraints for wired in packages?

Cabal anyway does that. It puts the wired-in packages in global constraints before solving.

harendra-kumar:
I am sorry, I never compiled with GHC 7.8.4. I have a fix for the CI build failure, should I rebase and squash with the last commit or just push another commit? Let me know if there is anything else to be taken care of as well. Many thanks for spending time on reviewing, testing and providing useful feedback.

harendra-kumar:
I found a couple of issues with flags. 

When solver returns flag changes for packages which are in snapshot, we update the flags in stack.yaml but we do not put those packages in extra dependencies. When we run `stack build` it complains about it.

We are not comparing the snapshot package flags (we compare only the version) when determining extra dependencies. If the snapshot flags are different than those the solver arrived at then we need to put those packages in extra dependencies. Otherwise the plan will not build because the status of flags can change dependencies.

This can be tested on stack itself. If we run `stack init --resolver lts-2.22 --solver` for stack itself the resulting plan is broken due to this problem. I will fix it tomorrow.

mgsloan:
> This warrants a separate issue. Perhaps we should avoid loading project config for init. Currently it uses withConfigAndLock which parses existing stack.yaml. We don't need to parse it for init, we only need to check the existence of it.

I've opened https://github.com/commercialhaskell/stack/issues/1604 . Yeah, I think we should be able to avoid using `withConfigAndLock`

> I am sorry, I never compiled with GHC 7.8.4. I have a fix for the CI build failure, should I rebase and squash with the last commit or just push another commit? Let me know if there is anything else to be taken care of as well. Many thanks for spending time on reviewing, testing and providing useful feedback.

No problem, that's what CI is for!  Pushing another commit is fine, we have lots of "Fix build on GHC 7.8" commits..

Many thanks for spending time on implementing this!  The next steps you've described sound good.

harendra-kumar:
It now gives a workable plan for stack itself using lts-2.22 and solver. Can you try it again now and see if you encounter any issues?

harendra-kumar:
We will have to update the documentation as well. I guess I can do that via a separate PR?

mgsloan:
Looks good!  Feel free to do documentation in this PR, or a different one.  I suppose I'd prefer including it with this one, but it's no big deal.

I'm still running into the issue where there are hard constraints which disagree with local package versions.  Could you please try this repro?

```
git clone https://github.com/mgsloan/diagrams-ci
cd diagrams-ci
git checkout stack-repro-for-1583
./build-universe.sh    # Doesn't actuall build, due to my one commit on the repo
cd build-tmp
stack init --solver
```

mgsloan:
Also with that repro, I get all the local packages having entries in `flags`, if I `stack init --solver --force`:

```
flags:
  diagrams-contrib: {}
  diagrams-lib: {}
  ...
```

It'd be best to leave these out!

harendra-kumar:
I will try the diagrams-ci test. For documentation, I was only worried about merge conflicts if this PR stays longer. I will assess how many changes are needed and then we can decide to postpone or not.

I thought I filtered out the empty flags but might have missed something, will check.

harendra-kumar:
The empty flags were getting through only in case of `--force` I fixed that. Another nice-to-have is to not write the flags if they are unchanged from defaults. Currently we are writing the default values of manual flags.

The solver errors are due to conflicting requirements of source packages in the diagrams suite. Here is the output I am getting:

```
cabal: Could not resolve dependencies:
trying: active-0.2.0.8 (user goal)
trying: diagrams-input-0.1 (user goal)
next goal: diagrams-lib (user goal)
rejecting: diagrams-lib-1.3.0.8, 1.3.0.7, 1.3.0.6, 1.3.0.5, 1.3.0.4, 1.3.0.3,
1.3.0.2, 1.3.0.1 (global constraint requires ==1.3)
trying: diagrams-lib-1.3
next goal: linear (dependency of active-0.2.0.8)
rejecting: linear-1.19.1.3, 1.20.3, 1.20.2, 1.20.1, 1.20, 1.19.1.2, 1.19.1.1,
1.19.1, 1.19 (conflict: diagrams-input => linear>=1.11.3 && <1.19)
rejecting: linear-1.18.3, 1.18.1.1, 1.18.1, 1.18.0.1, 1.18, 1.17.1, 1.17,
1.16.4, 1.16.2, 1.16.1, 1.16, 1.15.5, 1.15.4, 1.15.3, 1.15.2, 1.15.1,
1.15.0.1, 1.15, 1.14.0.1, 1.14 (conflict: diagrams-lib => linear>=1.20.1 &&
<1.21)
rejecting: linear-1.13, 1.12.1, 1.11.3, 1.11.2, 1.11.1, 1.11, 1.10.1.2,
1.10.1.1, 1.10.1, 1.10, 1.9.1, 1.9.0.1, 1.9, 1.8.1, 1.8, 1.7, 1.6, 1.4,
1.3.1.1, 1.3.1, 1.3, 1.2, 1.1.4, 1.1.2, 1.1.1, 1.0.1, 0.9.2, 0.9.1, 0.9, 0.8,
0.7, 0.6.1, 0.6, 0.5, 0.4.2.2, 0.4.2.1, 0.4.1, 0.2.0.2, 0.2 (conflict: active
=> linear>=1.14 && <1.21)
Dependency tree exhaustively searched.
```

Here source packages active, diagrams-input and diagrams-lib have a conflict on linear. Among themselves the dependencies are `diagrams-input->diagrams-lib->active`. So we can perhaps remove diagrams-input and anything that depends on that from our source packages and it should work. It indeed worked when I removed it.

So maybe a future enhancement can be to automatically suggest which user packages can be removed to get a working plan.

harendra-kumar:
A very clean stack.yaml is generated now. I have added logic to remove flags with default values as well. 

I might push one more commit to report compiler incompatibility errors for each individual source package rather than once for the whole set. That way the user can easily identify exactly which package has problems.

mgsloan:
This is a great set of changes, thanks for all your excellent work!

I [fixed a warning](https://github.com/commercialhaskell/stack/commit/f050d497f2f3e87e2ad517710006649564ebb604) that cropped up due to [now reporting more warnings](https://github.com/commercialhaskell/stack/commit/0b1b9c21fe1dc1f58f036867fcb0774a98193bc7).

mgsloan:
RE the diagrams-haddock thing, you're probably right, I was just getting confused by the wording of cabal-install's output.

harendra-kumar:
Great. Thanks! I will try to fix the user guide in the next couple of days.

borsboom:
Looks like there's a syntax error in the haddocks:
```
src/Stack/Solver.hs:306:25:
    parse error on input ‘-- ^ src package constraints’
     (src/Stack/Types/Internal.hs:56)
    HasSupportsUnicode (src/Stack/Types/Internal.hs:59)
     (src/Stack/Types/Internal.hs:62)
    Sticky (src/Stack/Types/Internal.hs:65)
    HasSticky (src/Stack/Types/Internal.hs:69)
     (src/Stack/Types/Internal.hs:72)
```

harendra-kumar:
Can we include haddock in CI?

mgsloan:
Sure! Added in https://github.com/commercialhaskell/stack/commit/92da37a463a444dcdc94cb64f9bac14c74cb7e8a

mgsloan:
Unfortunately, it doesn't work: https://github.com/commercialhaskell/stack/issues/1623

harendra-kumar:
I see you already fixed the haddock issues. Thanks!

PR #1589, opened 2016-01-02
Warn if newly installed executables won't be available on the PATH 

### LINKED ISSUE #1362: Make it clearer when stack installs to a spot that's not on the PATH or shadowed on the PATH
We already have a warning about this, but it gets lost in the mix:

```
Installing library in
/tmp/stack-upgrade17612/stack/.stack-work/install/x86_64-linux/lts-3.7/7.10.2/lib/x86_64-linux-ghc-7.10.2/stack-0.1.7.0-1qCQzKMpF5B2dxtC0o3K2h
Installing executable(s) in
/tmp/stack-upgrade17612/stack/.stack-work/install/x86_64-linux/lts-3.7/7.10.2/bin
Registering stack-0.1.7.0...
Completed all 2 actions.
Installation path /home/mgsloan/.local/bin not found in PATH environment variable
Copying from /tmp/stack-upgrade17612/stack/.stack-work/install/x86_64-linux/lts-3.7/7.10.2/bin/stack to /home/mgsloan/.local/bin/stack

Copied executables to /home/mgsloan/.local/bin:
- stack
```

There's a `Installation path /home/mgsloan/.local/bin not found in PATH environment variable` hiding in there.  I'd rather see:

```
Completed all 2 actions.
Copying from /tmp/stack-upgrade17612/stack/.stack-work/install/x86_64-linux/lts-3.7/7.10.2/bin/stack to /home/mgsloan/.local/bin/stack

Copied executables to /home/mgsloan/.local/bin:
- stack

WARNING: Installation path /home/mgsloan/.local/bin not found in PATH environment variable.
```

There's also the case where the installation path is on the PATH, but some earlier dir in PATH also contains an executable with this name (shadowing).  In this case, I think it makes sense to output something like `WARNING: The "stack" executable found on the PATH environment variable is /usr/bin/stack, and not the version that was just installed.  This means that "stack" calls on the command line will not use this version.`

sjakobi:
Fixes #1362.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -420,7 +420,15 @@ executePlan menv bopts baseConfigOpts locals globalPackages snapshotPackages loc
                                 , exe
                                 , "\" calls on the command line will not use this version."
                                 ]
-                    Nothing -> return ()
+                    Nothing -> do
+                        $logWarn ""
+                        $logWarn $ T.concat
+                            [ "WARNING: Installation path "
+                            , T.pack destDir'
+                            , " is on the PATH but the \""
+                            , exe
+                            , "\" executable that was just installed could not be found on the PATH."
+                            ]
             else do
                 $logWarn ""
                 $logWarn $ T.concat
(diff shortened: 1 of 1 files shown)


mgsloan:
Cool, thanks for working on this!  I'll merge once my comments are addressed.

sjakobi:
> Indeed, I went through the same reasoning, and would rather that that detail not exist. So yeah, how about using PATH directly (via `liftIO $ lookupEnv "PATH")?

Done (using the more convenient `System.FilePath.getSearchPath`).

> Also, I think the Nothing -> case should do some logging.

Done, too!

Thanks for the review and discussion!

mgsloan:
Thanks for working on this, LGTM!

PR #1598, opened 2016-01-03
Added interactive Main selection

### LINKED ISSUE #1068: Stack ghci should start with a project, even if there are two
C:\Neil\ghcid>stack ghci
    * * * * * * * *
    The main module to load is ambiguous. Candidates are:
    Package `ghcid' component exe:ghcid with main-is file: C:\Neil\ghcid\src\Ghcid.hs
    Package `ghcid' component test:ghcid_test with main-is file: C:\Neil\ghcid\src\Test.hs
    None will be loaded. You can specify which one to pick by:
     1) Specifying targets to stack ghci e.g. stack ghci ghcid:exe:ghcid
     2) Specifying what the main is e.g. stack ghci --main-is ghcid:exe:ghcid
    * * * * * * * *

This error is a little annoying - it says it's not going to do anything useful, and the two ways of picking the useful behaviours are both quite annoying - they require me to remember command line flags every time. Instead, how about:

* Always default to whichever is first in the `.cabal` file lexically, then people can reorder their `.cabal` file to indicate a preference.
* Add a facility to add an entry to `stack.yaml` to record a default.

achernyak:
The message has been changed to the below, and now includes interactive loading. If nothing or anything other than the options provided under `3)` is chosen, no main is loaded.

```
* * * * * * * *
The main module to load is ambiguous. Candidates are:
1. Package `testP' component exe:testP with main-is file: C:\Users\artem\dev\haskell\stack\testP\app\Main.hs
2. Package `testP' component test:testP-test with main-is file: C:\Users\artem\dev\haskell\stack\testP\test\Spec.hs
You can specify which one to pick by:
 1) Specifying targets to stack ghci e.g. stack ghci testP:exe:testP
 2) Specifying what the main is e.g. stack ghci --main-is testP:exe:testP
 3) Choosing 1, 2
* * * * * * * *
```

Resolves #1068

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -53,6 +53,7 @@ import           Stack.Package
 import           Stack.Types
 import           Stack.Types.Internal
 import           System.Directory (getTemporaryDirectory)
+import Text.Read (readMaybe)
 
 #ifndef WINDOWS
 import qualified System.Posix.Files as Posix
@@ -156,7 +157,7 @@ ghci GhciOpts{..} = do
 -- is none, sometimes it's unambiguous, sometimes it's
 -- ambiguous. Warns and returns nothing if it's ambiguous.
 figureOutMainFile
-    :: (Monad m, MonadLogger m)
+    :: (Monad m, MonadLogger m, MonadIO m)
     => BuildOpts
     -> Maybe (Map PackageName SimpleTarget)
     -> Map PackageName SimpleTarget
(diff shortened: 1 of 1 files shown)


mgsloan:
Thanks for working on this!  I think the following should be addressed before merge:

1) Squash all of these commits into one.  This way, the added + removed files don't go into the repo.

2) Replace the `1)`, `2)`, `3)` with just `*` bullet points for each one.  I think it's confusing that numbers are used for both lists

3) Instead have the text for the choice option be `Choose from the candidates above [1..N]?`, where N is the number of choices.

4) It's unclear that it's asking for input, because the cursor doesn't have anything in front of it (I'm imagining `* Choose from the candidates above [1..N]? |`, where `|` is the user's cursor.  Unfortunately, this isn't easily done with `$logWarn`.  How about using `hPutStr stderr "* Choose from the candidates above [1..N]? "`.  Alternatively, it might make sense to convert all that warning stuff to use `putStr` / `putStrLn`.

5) A bonus change would be telling the user how it interpreted the input.  Like this:

```
* * * * * * * *
The main module to load is ambiguous. Candidates are: 
1. Package `stack' component exe:stack with main-is file: /home/mgsloan/fpco/stack/src/main/Main.hs
2. Package `stack' component test:stack-test with main-is file: /home/mgsloan/fpco/stack/src/test/Test.hs
You can specify which one to pick by: 
 * Specifying targets to stack ghci e.g. stack ghci stack:exe:stack
 * Specifying what the main is e.g. stack ghci --main-is stack:exe:stack
 * Choose from the candidates above [1..2]?
* * * * * * * *

Loading main module from candidate 2, --main-is /home/mgsloan/fpco/stack/src/test/Test.hs
```

Or, if no input is provided:

```
Not loading any main modules, as no option selected
```

6) Another bonus change would be to output relative paths for the "main-is" options

achernyak:
Thank you so much for the detailed input. 

I was actually really struggling with getting the input to look properly with all the $logWarn messages, so I might end up switching all of them to `putStr`/`putStrLn. However I will see how far I can get with  $logWarn.

mgsloan:
Welcome!

Note that I only mean switching over the `logWarn` stuff that has to do with telling the user about the ways in which their configuration might not work for GHCi.  That stuff looked rather ugly in the verbose logging, anyway.

I think the only option for making the input work right with logWarn would be to send some tty cursor motion commands.  That seems like overkill compared to just using `putStr`.

achernyak:
I think I got it. Hope this is closer to what you were looking for. I made a slight change with where the request for input is displayed. 

It is outside the warning block completely. Please let me know your thoughts.

achernyak:
I think I got it. I did put the request for input outside of the warning window, but everything else is as you suggested.

Let me know your thoughts.

mgsloan:
LGTM, thanks!

mgsloan:
I've split off this issue https://github.com/commercialhaskell/stack/issues/1605

and added some tweaks: https://github.com/commercialhaskell/stack/commit/20423684e5376936406208b965545c8795ad4b55

PR #1607, opened 2016-01-05
GHCJS doc update

sheganinans:
Cheers!


mgsloan:
Seems reasonable!  I really look forward to having proper releases of improved-base.

PR #1617, opened 2016-01-06
Add note in debugging section on parameter passing

pavolzetor:
The -- can be used to pass parameters to the executable instead of
stack when stack exec is used. We added a short note in the debugging
section as an alternative.


borsboom:
Thanks!

PR #1618, opened 2016-01-06
Fix version mismatch in Travis-CI doc

lubomir:
The installed version did not match with what was added to PATH,
effectively ignoring installed GHC.


borsboom:
Thanks!

Issue #1593, opened 2016-01-03
Have "stack init" accept a list of sub-directories


mgsloan:
Currently, there's no way to selectively consider cabal files for `init`.  I think it makes sense to have it take a list of sub-directories to search in, defaulting to `.`.  This'd be consistent with `stack build` taking a list of sub-directories and building all projects in those sub-directories.

The `--ignore-subdirs` flag would still work - it'd mean that cabal files in sub-directories of the specified dirs are ignored.

I'd hope that this would work with bash expansion, so that if you do `stack init --ignore-subdirs *`, it'd build all the packages that are immediate subdirs (to faciliate this, merely warn when a path to a file is specified).

harendra-kumar:
That's cool!

achernyak:
@mgsloan Is there any difference between how `build` would handle the `--ignore-subdirs` flab and the expected `init` handling of this flag?

If there are no differences, I will go ahead and get started on this one over the weekend.

mgsloan:
Well, `build` doesn't take `--ignore-subdirs`.  It's not very useful for build, because usually you only use directory arguments when you have a dir that contains lots of packages (but not all of them).  In the case where you'd use `--ignore-subdirs` with init, you may as well just specify the target normally, using the package name.

But, yes, how `build` 's handles directory arguments is the same as how this would.

Great, certainly feel free to work on this!  Particularly timely, as @harendra-kumar 's large set of changes to init / solver just got merged.

harendra-kumar:
Fixed by PR #1780 .

PR #1639, opened 2016-01-10
Have "stack init" accept a list of sub-directories

achernyak:
@mgsloan Is this the rough behavior referred to in #1593?

After a confirmation I will add list behavior instead of single argument, and some clean up to how the arguments are handled.

### DIFF
--- a/src/Stack/Init.hs
+++ b/src/Stack/Init.hs
@@ -13,15 +13,15 @@ import           Control.Monad
 import           Control.Monad.Catch             (MonadMask, throwM)
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
-import           Control.Monad.Reader            (asks, MonadReader)
+import           Control.Monad.Reader            (MonadReader, asks)
 import           Control.Monad.Trans.Control     (MonadBaseControl)
 import qualified Data.ByteString.Builder         as B
-import qualified Data.ByteString.Lazy            as L
 import qualified Data.ByteString.Char8           as BC
+import qualified Data.ByteString.Lazy            as L
+import qualified Data.Foldable                   as F
 import           Data.Function                   (on)
 import qualified Data.HashMap.Strict             as HM
 import qualified Data.IntMap                     as IntMap
-import qualified Data.Foldable                   as F
 import           Data.List                       (intersect, maximumBy)
 import           Data.List.Extra                 (nubOrd)
 import           Data.Map                        (Map)
@@ -35,13 +35,13 @@ import           Network.HTTP.Client.Conduit     (HasHttpManager)
 import           Path
 import           Path.IO
 import           Stack.BuildPlan
+import           Stack.Config                    (getSnapshots,
+                                                  makeConcreteResolver)
 import           Stack.Constants
 import           Stack.Solver
 import           Stack.Types
-import           Stack.Types.Internal            ( HasTerminal, HasReExec
-                                                 , HasLogLevel)
-import           Stack.Config                    ( getSnapshots
-                                                 , makeConcreteResolver)
+import           Stack.Types.Internal            (HasLogLevel, HasReExec,
+                                                  HasTerminal)
 import qualified System.FilePath                 as FP
 
 -- | Generate stack.yaml
@@ -57,8 +57,8 @@ initProject
 initProject currDir initOpts mresolver = do
     let dest = currDir </> stackDotYaml
 
+    dirs <- mapM (resolveDir' . T.unpack) (searchDirs initOpts)
     reldest <- toFilePath `liftM` makeRelativeToCurrentDir dest
-
     exists <- doesFileExist dest
     when (not (forceOverwrite initOpts) && exists) $ do
         error ("Stack configuration file " <> reldest <>
@@ -67,8 +67,12 @@ initProject currDir initOpts mresolver = do
 
     let noPkgMsg =  "In order to init, you should have an existing .cabal \
                     \file. Please try \"stack new\" instead."
-
-    cabalfps <- findCabalFiles (includeSubDirs initOpts) currDir
+    let findCabalFiles' = findCabalFiles (includeSubDirs initOpts)
+    cabalfps <- if null dirs
+                then
+                   findCabalFiles' currDir
+                else
+                  liftM concat $ mapM findCabalFiles' dirs
     (bundle, dupPkgs)  <- cabalPackagesCheck cabalfps noPkgMsg Nothing
 
     (r, flags, extraDeps, rbundle) <- getDefaultResolver dest initOpts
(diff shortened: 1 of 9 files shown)


src/Stack/Options.hs
```diff
@@ -670,8 +671,16 @@ globalOptsFromMonoid defaultTerminal GlobalOptsMonoid{..} = GlobalOpts
...
-    InitOpts <$> method <*> solver <*> overwrite <*> fmap not ignoreSubDirs
+    InitOpts <$> method <*>
+      solver <*>
+      overwrite <*>
+      fmap not ignoreSubDirs <*>
+      subDir
   where
+    subDir =
+      textArgument
+        (metavar "SUB-DIRECTORIES" <>
```
mgsloan: I think it'd be more consistent to just use `"DIR"`

mgsloan:
LGTM, other than the comment.

In general I'd prefer the formatting changes happening in a different commit, but it's no big deal.

achernyak:
Ha, I didn't even notice those. I run stylish-haskell and those changes just happened on save. I will back them out before the final commit.

mgsloan:
Pinging @achernyak , how's it going?

achernyak:
@mgsloan Sorry for the long delay. Things got a bit crazy at work the last couple weeks. I plan to wrap everything up this weekend with all your suggestions and get this pushed out.

mgsloan:
Sounds good!  I'm just taking a look at all the open PRs!

achernyak:
@mgsloan I rebased the branch to get rid of some of the big style changes present before.

Also, now a list of directories can be passed in.

mgsloan:
Hey, sorry for the review delay.  Are you sure this works?  To me it looks like it invokes init on each of the subdirectories, which isn't what we want.

achernyak:
No problem at all. I must have miss understood the exact ask here. I see two other options:

1. It only takes one subdirectory to search and runs init on it.
1. It searches each of the subdirectories for a .cabal file and only runs init on the directories that have a .cabal file. In this scenario not sure if it should search all subdirectories of those directories too.

Let me know which is more aligned with what is expected.

mgsloan:
I think the main thing is that we only want to generate one stack.yaml, and so should only call `initProject` once.

> It searches each of the subdirectories for a .cabal file and only runs init on the directories that have a .cabal file. In this scenario not sure if it should search all subdirectories of those directories too.

Right [here](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Init.hs#L71), we want to search in the directories listed by the user.  Searching in all subdirectories is controlled by `includSubdirs`

achernyak:
Got it, so we are only searching the specified directories and ignoring the subdirectories of the provided directories. Then we run initProject on the first `.cabal` file we find.

mgsloan:
Well, we're trying to get a list of all the cabal files in the project `cabalfps`.  Then, `initProject` tries to figure out a configuration that can build them all together.

harendra-kumar:
I guess you don't need to change `initCmd` at all. Just extract the list of directories passed by the user inside initProject, use `findCabalFiles` on all of them (currently we use it only on `CurrDir`) and concat the result. The rest should be taken care of by the existing code. It should create a stack.yaml with all the packages in the specified directories.

achernyak:
@harendra-kumar Thanks, that makes complete sense. Realy cleared up my understanding on this one.

achernyak:
Just tested and I think this version should meet the behaviour described by @harendra-kumar.

achernyak:
So I tried to merge the upstream and apparently two of the methods I was using got replaced. I will work on fixing this.

src/Stack/Init.hs
```diff
@@ -35,13 +35,15 @@ import           Network.HTTP.Client.Conduit     (HasHttpManager)
 import           Path
 import           Path.IO
 import           Stack.BuildPlan
+import           Stack.Config                    (getSnapshots,
+                                                  makeConcreteResolver)
+import           Stack.Config                    (getSnapshots,
```
harendra-kumar: This import is duplicated.

src/main/Main.hs
```diff
@@ -1173,15 +1173,19 @@ withMiniConfigAndLock go inner =
 -- | Project initialization
 initCmd :: InitOpts -> GlobalOpts -> IO ()
 initCmd initOpts go = do
-    pwd <- getCurrentDir
-    withMiniConfigAndLock go (initProject pwd initOpts (globalResolver go))
+    let selectedDirs = searchDirs initOpts
```
harendra-kumar: This file can be reverted. We are already passing the dirs to initProject as part of initOpts. We can extract this information there itself rather than passing it separately.

src/Stack/Init.hs
```diff
@@ -51,11 +53,13 @@ initProject
...
        , HasTerminal env)
     => Path Abs Dir
+    -> [Path Abs Dir]
     -> InitOpts
     -> Maybe AbstractResolver
     -> m ()
-initProject currDir initOpts mresolver = do
+initProject currDir searchDirs' initOpts mresolver = do
     let dest = currDir </> stackDotYaml
+        dest' = toFilePath currDir
```
harendra-kumar: Redundant expression.

src/Stack/Options.hs
```diff
@@ -682,7 +683,12 @@ initOptsParser :: Parser InitOpts
 initOptsParser =
     InitOpts <$> solver <*> omitPackages
              <*> overwrite <*> fmap not ignoreSubDirs
+             <*> searchDirs
   where
+    searchDirs =
+      many (textArgument
+              (metavar "SEARCH-DIRECTORIES" <>
```
harendra-kumar: `DIRS` would be concise.

src/Stack/Options.hs
```diff
@@ -682,7 +683,12 @@ initOptsParser :: Parser InitOpts
...
+               help "Directories which are searched for cabal files. If non specified, uses current directory."))
```
harendra-kumar: How about a more concise help like `Directories to include, default is current directory.`?

achernyak: I can change it to be more concise. I was going off what @mgsloan suggested before. But I do like the more concise version better.

harendra-kumar:
Functionally it looks good. I added a few minor comments. Can you address those and squash the changes into a single commit? 

Unrelated formatting changes are better done via separate commits. Though its fine for this one.

achernyak:
@harendra-kumar Thanks for the detailed review, working on those changes now. 

I really need to turn off my `stylish-haskell` when working on stack. It's the issue where all the style changes are coming from just auto formatting.

src/Stack/Options.hs
harendra-kumar: What you have is fine except for a spelling correction (non -> none). Since init does not have too many options, the help text is not too long therefore we do not need to be worried about making the help too concise.

PR #1642, opened 2016-01-11
Nix: Select the correct GHC version in more cases.

### LINKED ISSUE #1641: stack --nix --resolver doesn't work as expected
Passing a concrete resolver on the command-line correctly overrides the resolver found in the stack.yaml, but passing an abstract resolver does not.

Example:

```
$ stack new --bare --resolver lts-4.1 foo
$ stack --nix --resolver lts-2.22 build
...
```

Compared to:

```
$ stack --nix --resolver lts-2 build
Selected resolver: lts-2.22
Compiler version mismatched, found ghc-7.10.2 (x86_64), but expected minor version match with ghc-7.8.4 (x86_64) (based on resolver setting in /tmp/foo/stack.yaml).
```

mboes:
Even when abstract resolver is provided.

Fixes #1641.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -23,6 +23,10 @@ Bug fixes:
   [Mailing list discussion](https://groups.google.com/d/msg/haskell-stack/iVGDG5OHYxs/FjUrR5JsDQAJ)
 - Gracefully handle invalid paths in error/warning messages
   [#1561](https://github.com/commercialhaskell/stack/issues/1561)
+- Nix: select the correct GHC version corresponding to the snapshot
+  even when an abstract resolver is passed via `--resolver` on the
+  command-line.
+  [#1641](https://github.com/commercialhaskell/stack/issues/1641)
 
 ## 1.0.0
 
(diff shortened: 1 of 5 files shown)


mboes:
Basic approach not workable. Will resubmit once I get some time to think about this a bit more.

borsboom:
What turned out to be unworkable about the approach?  This is relevant for Docker as well.

mboes:
What didn't work is that I completely forgot to take into the stack.yaml resolver. The needed info is available in the `Project` datatype passed to `nixOptsFromMonoid` and `dockerOptsFromMonoid`, but not in `reexecWithOptionalShell`.

So what we need is this: create a closure in `nixOptsFromMonoid` that chooses a compiler package name based on the `Project` info by default. Store the closure as a field in `NixOpts`. Then, if in reality the user passed in an abstract resolver on the command-line, then concretize it in `reexecWithOptionalShell` and invoke the closure to get a final compiler package name. That's exactly what I did in #1646, which is a slight rewrite of this PR.

PR #1645, opened 2016-01-12
Update copyright year in config.yaml example

Irio:


borsboom:
Thanks!  For future reference, better to make these changes in the `stable` branch which will update the docs.haskellstack.org main documentation.  The `release` branch is normally just for the version that was actually released.

Irio:
@borsboom Noted! You could consider changing [the default branch](https://github.com/commercialhaskell/stack/settings/branches) to `stable` so others don't commit the same mistake again.

borsboom:
I don't think we want to change the default branch.  Most commits should go in `master` (which also would have been fine for this PR).  `stable` is only for minor patches or small documentation updates that effect the current release.   It also wouldn't have helped here, because `release` is not the default branch.

PR #1646, opened 2016-01-12
Nix: select the right GHC version in more cases.

### LINKED ISSUE #1641: stack --nix --resolver doesn't work as expected

mboes:
Stack now selects the correct GHC version in the following cases:

* an abstract resolver was given,
* a concrete resolver was given,
* a compiler version was given.

Previously, we could only deal with concrete resolvers correctly. So
e.g.

```
$ stack --resolver lts-2 build
```

should now use GHC-7.8.4 from the nixpkgs repo. I think exactly the same
approach should be workable for Docker too, but how exactly the use case
should work there is less clear to me.

Fixes #1641.

### DIFF
--- a/src/Stack/Types/Config.hs-boot
+++ b/src/Stack/Types/Config.hs-boot
@@ -1,0 +1,5 @@
+module Stack.Types.Config where
+
+data AbstractResolver
+data Resolver
+data Config
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -23,6 +23,10 @@ Bug fixes:
   [Mailing list discussion](https://groups.google.com/d/msg/haskell-stack/iVGDG5OHYxs/FjUrR5JsDQAJ)
 - Gracefully handle invalid paths in error/warning messages
   [#1561](https://github.com/commercialhaskell/stack/issues/1561)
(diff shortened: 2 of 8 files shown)


src/Stack/Types/Config.hs-boot
```diff
@@ -0,0 +1,5 @@
+module Stack.Types.Config where
```
mboes: Note: An alternative to using an .hs-boot file is to parameter `NixOpts` over the resolver type. I found the .hs-boot file to be less verbose in the end.

mboes:
cc @YPares

borsboom:
LGTM, and I'll try to apply this approach to Docker as well.

PR #1649, opened 2016-01-13
Added option for full working dir cleanup in command clean

panamiga:
Now this option do the work requested.
stack new test
cd test
stack build
stack clean --full

This will clean all the targets for complex project.

Also tested with single and multiple targets ie
stack clean target1 target2 --full

### DIFF
--- a/src/Stack/Clean.hs
+++ b/src/Stack/Clean.hs
@@ -20,7 +20,7 @@ import           Data.Typeable (Typeable)
 import           Path.IO (removeTreeIfExists)
 import           Stack.Build.Source (getLocalPackageViews)
 import           Stack.Build.Target (LocalPackageView(..))
-import           Stack.Constants (distDirFromDir)
+import           Stack.Constants (distDirFromDir, workDirFromDir)
 import           Stack.Types (HasEnvConfig,PackageName)
 
 
@@ -33,7 +33,7 @@ clean
     :: (MonadThrow m, MonadIO m, MonadReader env m, HasEnvConfig env, MonadLogger m)
     => CleanOpts
     -> m ()
-clean (CleanOpts targets) = do
+clean (CleanOpts targets doFullClean) = do
     locals <- getLocalPackageViews
     case targets \\ Map.keys locals of
         [] -> do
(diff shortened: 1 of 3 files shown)


src/Stack/Clean.hs
```diff
@@ -42,15 +42,19 @@ clean (CleanOpts targets) = do
                         then Map.elems locals -- default to cleaning all local packages
                         else mapMaybe (`Map.lookup` locals) targets
             forM_ lpvs $ \(LocalPackageView{lpvRoot = pkgDir},_) -> do
-                distDir <- distDirFromDir pkgDir
-                removeTreeIfExists distDir
+                let delDir =
+                          if doFullClean
+                              then workDirFromDir pkgDir
+                              else distDirFromDir pkgDir
+                removeTreeIfExists =<< delDir
```
borsboom: Will this clean the project-wide `.stack-work` if there's no package directly in the project root?

panamiga: Thank you point this out.
For the case if no package specified ie `stack clean --full` it would be obvious to just remove the project-wide .stack-work dir.
If package(s) specified the things became more interesting. I currently investigate the way to handle this case correctly.
If you could give a suggestions it would be much appreciated.

borsboom: There's a good chance that the per-package `.stack-work`s will be eliminated at some point (they only exist to work around a Cabal issue on Windows).  As such, I'd suggest making it a requirement that `stack clean --full` applies to the whole project (so it would be an error to specify packages).

panamiga:
New testcase is:
mkdir clean-test
cd clean-test
stack new pr1
stack new pr2
stack init
stack build
stack clean pr1
#check stack do it right
stack clean --full

mgsloan:
LGTM!

panamiga:
Thanks!

Zingam:
I accidently had .stack-work opened in Windows Explorer and I got the error message below. It is not very informative.

X:\Haskell\StackHello>stack clean --full
removeDirectoryRecursive: unsatisfied constraints (The directory is not empty.)

mgsloan:
@Zingam Thanks for reporting, opened a PR to have somewhat better exceptions here https://github.com/commercialhaskell/stack/pull/3317 .  Nothing fancy but should make it less obscure.

PR #1654, opened 2016-01-14
Add static flag to the stack executable

mitchty:
Add an optional flag, that is off by default to allow the stack binary to be created as a fully static executable.

This is primarily to enable fully static binaries on Alpine Linux which uses musl libc and is designed for static linking. The resultant binaries can run on almost any linux however, depending on libc used. (don't do this with glibc distributions unless you like weird behavior)

Example:
```
/tmp/stack-1.0.0 # cabal install -fstatic
Resolving dependencies...
In order, the following will be installed:
stack-1.0.0 (latest: 9.9.9) +static (reinstall) (changes: gitrev-1.1.0 added, optparse-simple-0.0.3 added)
Warning: Note that reinstalls are always dangerous. Continuing anyway...
Notice: installing into a sandbox located at /tmp/stack-1.0.0/.cabal-sandbox
Configuring stack-1.0.0...
Building stack-1.0.0...
Installed stack-1.0.0
/tmp/stack-1.0.0 # file ./.cabal-sandbox/bin/stack
./.cabal-sandbox/bin/stack: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
/tmp/stack-1.0.0 # rm ./.cabal-sandbox/bin/stack
/tmp/stack-1.0.0 # cabal install
Resolving dependencies...
In order, the following will be installed:
stack-1.0.0 (latest: 9.9.9) (reinstall) (changes: gitrev-1.1.0 added, optparse-simple-0.0.3 added)
Warning: Note that reinstalls are always dangerous. Continuing anyway...
Notice: installing into a sandbox located at /tmp/stack-1.0.0/.cabal-sandbox
Configuring stack-1.0.0...
Building stack-1.0.0...
Installed stack-1.0.0
/tmp/stack-1.0.0 # file ./.cabal-sandbox/bin/stack
./.cabal-sandbox/bin/stack: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, stripped
/tmp/stack-1.0.0 # ldd ./.cabal-sandbox/bin/stack
        /lib/ld-musl-x86_64.so.1 (0x55a845acf000)
        libz.so.1 => /lib/libz.so.1 (0x7f02ce69f000)
        libgmp.so.10 => /usr/lib/libgmp.so.10 (0x7f02ce43b000)
        libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x55a845acf000)
```

### DIFF
--- a/stack.cabal
+++ b/stack.cabal
@@ -41,6 +41,13 @@ flag disable-git-info
   -- should otherwise be avoided
   -- see: https://github.com/commercialhaskell/stack/issues/1425
 
+flag static
+  manual: True
+  default: False
+  description: Pass -static/-pthread to ghc when linking the stack binary.
+  -- Not intended for general use. Simply makes it easier to
+  -- build a fully static binary on Linux platforms that enable it.
+
 library
   hs-source-dirs:    src/
   ghc-options:       -Wall -fwarn-tabs -fwarn-incomplete-uni-patterns -fwarn-incomplete-record-updates
@@ -218,6 +225,8 @@ executable stack
   main-is:        Main.hs
   ghc-options:    -threaded -rtsopts -Wall -fwarn-tabs -fwarn-incomplete-uni-patterns -fwarn-incomplete-record-updates
   other-modules:  Paths_stack
+  if flag(static)
+      ld-options: -static -pthread
 
   build-depends:  base >=4.7 && < 5
                 , bytestring >= 0.10.4.0


tuncer:
Nice. This can also be used to provide a single linux binary to download for each hardware architecture for stack's binary downloads. I've successfully built and used ghc-7.10.3 with zero patches applied on voidlinux-musl, so the compiler could also be packaged as a statically built version, but that's unrelated to this support patch.

mgsloan:
Cool!  I wouldn't mind adding a flag for this.  However, is it necessary?  Maybe we should just do `stack build --ghc-options "-static -pthread"`?

mitchty:
This is more to make packaging stack easier and to make building a full static binary of stack via cabal simpler.

And that wouldn't work as -pthread isn't a ghc option. It should be -optl=-pthread if you're passing this in via ghc-options aka:
```
stack build --ghc-options "-optl=-static -optl=-pthread"
```

But, doing that with cabal tends to break builds as the -static ld flag makes shared libraries emit invalid assembly that isn't pie so that ld can't use the libraries to link to. I'll just patch the cabal file for now no worries.

mgsloan:
I see!  I think that's good enough reason to merge, thanks!

mitchty:
Thanks @mgsloan!

For the record what eventually happens if you would try to build the way noted is this:
```
/usr/lib/gcc/x86_64-alpine-linux-musl/4.9.2/../../../../x86_64-alpine-linux-musl/bin/ld: /usr/lib/gcc/x86_64-alpine-linux-musl/4.9.2/crtbeginT.o: relocation R_X86_64_32 against `__TMC_END__' can not be used when making a shared object; recompile with -fPIC
/usr/lib/gcc/x86_64-alpine-linux-musl/4.9.2/crtbeginT.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
```

If you pick apart the shared object you'll notice it contains static relocations that won't work with the c runtime as used on alpine linux which is compiled PIE. And no that message from ld doesn't fix things, it is a lie like cake.

PR #1669, opened 2016-01-17
Fix broken link in the FAQ.

ghorn:


borsboom:
Perhaps there was a blip while I was releasting stack-1.0.2.  In any case, thanks for contributing.

PR #1674, opened 2016-01-19
Stack init and solver finish up and doc update

harendra-kumar:
Provide, working out of the box experience, for `stack init` and `stack solver` commands. Important fixes in this PR include:

* Automatically omit packages (--omit-packages)
 *  which cannot be compiled by the resolver
 * when there is a dependency conflict
 * when there is a package name conflict
 * Omitted packages are kept as commented in the config file
* Try all major lts versions during init so that we do everything possible to compile a package
* When some packages cannot be compiled choose a resolver which compiles maximum number of packages
* Provide user messages on every config load when packages are omitted or extra-deps are added by init
* Report packages with mismatched filename and package name
* Update stack init and solver documentation in the guide

The documentation in the user guide provides more information and workflow examples.

With these fixes stack init experience must now become much better. We should be able to generate a working stack.yaml for most packages without any issues. If a package has working cabal files and cabal can build it but we are not able to generate a stack.yaml out of the box then it should be considered a bug!


mgsloan:
Looks awesome, thanks!  I'll come back with more detailed feedback sometime in the next couple days.

mgsloan:
Sorry for letting this languish so long!  I've taken a look at the commits, and they look great.  `--solver --omit-packages` does a fantastic job of producing a stack.yaml for a whole load of diagrams packages.

Since I only have a couple comments / tweaks, I'm going to go ahead and merge this and fix them myself.

mgsloan:
Pushed the following adjustments: https://github.com/commercialhaskell/stack/commit/3dd473b37a5c02a83f0709bc11351b55cf111d98 https://github.com/commercialhaskell/stack/commit/87037226d398e69033bb86d798d79fbac9ca18c2

harendra-kumar:
Thanks! Did you look at the doc changes?

mgsloan:
Yup, I did!  Also LGTM

PR #1690, opened 2016-01-21
Updated Travis example

snoyberg:
* Better caching via compiler setting
* A little more commenting
* OS X support built-in


mgsloan:
LGTM

PR #1724, opened 2016-02-01
clearer missing deps error message

wolftune:


mgsloan:
Yeah, the text is a bit confusing.  I've made some suggestions.  One good way to improve this output would be to suggest a modification to stack.yaml's extra-deps.  I haven't really thought through what that would look like, though.  Could be that the only reasonable way to do it is `stack solver`

mgsloan:
Thanks!

PR #1726, opened 2016-02-02
Do not repeat printing of package flags

harendra-kumar:
For `stack init` - print package flags just once per set of errors instead of spamming repeatedly.


mgsloan:
Thanks!

Issue #1388, opened 2015-11-21
`stack runghc --no-config` and `--standalone` options


massysett:
I am using `stack` as a script interpreter to run a script that generates the `.cabal` file.  Problem is, if there is no `.cabal` file or if the `.cabal` file is empty (as it will be if a script truncates it before writing the output) `stack` will complain.

Currently I move the `stack.yaml` out of the way and then run the script, then move `stack.yaml` back; I have also kept the script in a separate directory to avoid this problem, but both seem to be yucky solutions.  I don't understand why a `stack.yaml` influences the script interpreter behavior, especially when the script interpreter feature works fine with no `stack.yaml` at all.

borsboom:
When you run a script, it runs within the context of the current project.
If you're outside of an explicit project, Stack uses the implicit global
project in `~/.stack/global-project`.  So there is *always* a `stack.yaml`
(since without it Stack has no way to know which resolver to use).  I think
in most cases running a script in the context of the current project is the
right default behaviour, but maybe an option to force Stack to use the
implicit global project could come in handy for cases like this.

A workaround for now might be to pass the `--stack-yaml` argument in the
`-- stack` line to point it at a different project config file which
doesn't include any packages.

borsboom:
Huh, Github cut off a part of the comment I posted by e-mail.  Here's the rest:

A workaround for now might be to pass the `--stack-yaml` argument in the
`-- stack` line to point it at a different project config file which
doesn't include any packages.

mgsloan:
How about a switch that specifies to not load a `stack.yaml`, requiring that everything get passed in via CLI - something like `--no-config`.  This way you can't accidentally depend on your global config for your standalone script.

It'd also be good to have a flag for https://github.com/commercialhaskell/stack/issues/1208 , like maybe `--explicit-packages`.

Then, for ease of use, we might have a flag `--script` that's a synonym for `--no-config --install-ghc --explicit-packages`.

massysett:
> How about a switch that specifies to not load a stack.yaml, requiring that everything get passed in via CLI - something like --no-config.

This would be helpful.  Since `stack` already has the capability to work without a local project, some way to force it to that behavior would be helpful.

borsboom:
That all sounds pretty good to me.

borsboom:
[...]

Issue #1361, opened 2015-11-17
stack ghci <file>


mitchellwrosen:
Is it possible to load a single module into ghci, outside of a project?

```
> stack ghci foo.hs
Run from outside a project, using implicit global project config
Using resolver: lts-3.10 from implicit global project's config file: /Users/mrosen/.stack/global-project/stack.yaml
Error parsing targets: Directory not found: foo.hs
```

mgsloan:
Not currently, but it's a good idea.  I'd also like to support loading files from the project, and have it figure out which cabal components to load.

The way to do this is `stack ghci`, then within the repl, `:load foo.hs`.  Or, within a project, `stack exec ghci` followed by `:load foo.hs`.

mitchellwrosen:
Within a project, is `stack ghci` not equivalent to `stack exec ghci`?

mgsloan:
It isn't - with v0.1.6, `stack ghci` loads up every cabal component for every project.  On the latest dev branch, it now loads up all the library / executable components, consistent with `stack build`.  I did consider having `stack ghci` load up no components - see #1307.  If there's demand for this behavior, I wouldn't mind making the change.

mgsloan:
It'd also be good to be able to support `stack ghci Foo.Bar.Baz`, or `stack ghci --module Foo.Bar.Baz` (otherwise we might be overloading TARGET with a few too many ways to parse)

mgsloan:
This has been implemented!  One caveat is that you cannot specify both normal targets and file targets.  The logic for specifying both would have been complicated, so I took the easy way out.

Issue #1208, opened 2015-10-21
interpreter: Don't make cached packages available that are not specified with `--package`


soenkehahn:
Consider the following script:

``` haskell
#!/usr/bin/env stack
-- stack --resolver lts-3.10 --install-ghc runghc

import Safe

main :: IO ()
main = do
  print $ headMay "foo"
```
This script depends on the package `safe`. It currently works without problems on my machine, but only because I already have the package `safe` cached for the specified snapshot. On machines where the package is not installed it fails with
```
Test.hs:4:8:
    Could not find module ‘Safe’
    Use -v to see a list of the files searched for.
```
It would be great if `stack` scripts would either fail or succeed independent of cache state. Then I could create `stack` scripts on my machine, copy them to different machines and be fairly sure that they will run.

borsboom:
There's a tension between making scripting as convenient as possible for "one-offs" and making them reproducible.  My inclination is to lean toward the "one-off" case for this, and recommend making an actual project if you want proper reproducibility.  I already find the `-- stack ...` header line gets unwieldy, and having to list all dependencies there would be much worse.

If we *do* want proper reproducible scripts, I think we'd have to considering adding support for an inline `stack-yaml`-like configuration in the script's header, rather than reproducing a bunch of those features on the command-line.  But is it worth it?  I'm inclined to think not.

soenkehahn:
I didn't really consider the use-case of one-off scripts, but that's a very good point.

Ideally I think we would have both: reproducible and one-off scripts. (This could be done with a command line switch or a separate command.) Any thoughts on this?

borsboom:
No objection to finding a way to optionally improve reproducibility for scripts.  Simplest might be be to add a switch to `stack runghc` that requires dependencies to be explicitly listed.

borsboom:
Note discussion of multiline script directive in #1394, which should probably be done before this.

Also, related discussion about adding a `--script` flag in  https://github.com/commercialhaskell/stack/issues/1388#issuecomment-158703708.

rrnewton:
`--standalone` will be great.

In the meantime, is there any better way to test whether your script has undeclared dependencies rather than having to manually create a `stack.yaml` file and paste in the dependencies?  For long dependency lists, this is rather tiresome because it's a different textual format... i.e. reformatting `--package foo --package bar` to a list in the yaml file.

[...]

PR #1737, opened 2016-02-04
Change 'stack ghc' to provide package context

afcady:
This patch is necessary for me to use `flycheck` (which calls `stack ghc` internally) on a project that uses `cryptonite` with a package database that contains `cryptohash`.

Without it, I get flycheck errors like this:

```
    Ambiguous module name ‘Crypto.Hash’:
      it was found in multiple packages:
      cryptonite-0.10@crypt_9z0j8QI27Av2VIWw0mEkTO cryptohash-0.11.6@crypt_3Cwvwq9ssRY1dmVI1qI6C2
```

Related: https://github.com/flycheck/flycheck-haskell/issues/52

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -13,7 +13,6 @@ module Stack.Ghci
     , GhciPkgInfo(..)
     , GhciException(..)
     , ghciSetup
-    , ghciSetup'
     , ghci
     ) where
 
@@ -101,14 +100,6 @@ instance Show GhciException where
         [ "Not attempting to start ghci due to these duplicate modules."
         , "Use --no-load to try to start it anyway, without loading any modules (but these are still likely to cause errors)"
         ]
-
-ghciSetup' :: (MonadIO f, HasHttpManager r, MonadReader r f, MonadBaseControl IO f, MonadMask f, MonadLogger f, HasEnvConfig r, HasTerminal r, HasLogLevel r) => GhciOpts -> f [String]
-ghciSetup' opts = genOpts . extract <$> ghciSetup opts
-  where
-    extract (_, _, x) = x
-    genOpts :: [GhciPkgInfo] -> [String]
-    genOpts pkgs = nubOrd (concatMap (concatMap (oneWordOpts . snd) . ghciPkgOpts) pkgs)
-    oneWordOpts bio = bioOneWordOpts bio ++ bioPackageFlags bio
 
 -- | Launch a GHCi session for the given local package targets with the
 -- given options and configure it with the load paths and extensions
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -13,6 +13,7 @@ module Stack.Ghci
     , GhciPkgInfo(..)
     , GhciException(..)
     , ghciSetup
+    , ghciSetup'
     , ghci
     ) where
 
@@ -100,6 +101,14 @@ instance Show GhciException where
         [ "Not attempting to start ghci due to these duplicate modules."
         , "Use --no-load to try to start it anyway, without loading any modules (but these are still likely to cause errors)"
         ]
+
+ghciSetup' :: (MonadIO f, HasHttpManager r, MonadReader r f, MonadBaseControl IO f, MonadMask f, MonadLogger f, HasEnvConfig r, HasTerminal r, HasLogLevel r) => GhciOpts -> f [String]
+ghciSetup' opts = genOpts . extract <$> ghciSetup opts
+  where
+    extract (_, _, x) = x
+    genOpts :: [GhciPkgInfo] -> [String]
+    genOpts pkgs = nubOrd (concatMap (concatMap (oneWordOpts . snd) . ghciPkgOpts) pkgs)
+    oneWordOpts bio = bioOneWordOpts bio ++ bioPackageFlags bio
 
 -- | Launch a GHCi session for the given local package targets with the
 -- given options and configure it with the load paths and extensions
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -621,6 +621,10 @@ execOptsExtraParser = eoPlainParser <|>
                       ExecOptsEmbellished
                          <$> eoEnvSettingsParser
                          <*> eoPackagesParser
+                         <*> boolFlags False
+                                 "add-ghci-packages"
+                                 "prefix the executed command line with the same -package (etc.) args that would be used for GHCI"
+                                 idm
   where
     eoEnvSettingsParser :: Parser EnvSettings
     eoEnvSettingsParser = EnvSettings
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -383,6 +383,7 @@ data ExecOptsExtra
     | ExecOptsEmbellished
         { eoEnvSettings :: !EnvSettings
         , eoPackages :: ![String]
+        , eoGhciPackages :: !Bool
         }
     deriving (Show)
 
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -304,6 +304,10 @@ commandLineHandler progName isInterpreter = complicatedOptions
                     "Run ghc"
                     execCmd
                     (execOptsParser $ Just ExecGhc)
+        addCommand' "ghcCmd"
+                    "Run ghc with package context"
+                    ghcCmd
+                    (execOptsParser $ Just ExecGhc)
         addCommand' "ghci"
                     "Run ghci in the context of package(s) (experimental)"
                     ghciCmd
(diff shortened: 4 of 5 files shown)


mgsloan:
Thanks for working on this.  One issue is that since it uses the default build options, info from all the executable and library components will be used (and not from test / benchmarks).  Sometimes the info from two packages don't work well together, for example, if one imports `monads-tf` and the other uses `mtl`, then we'll get a lot of ambiguous module name errors.  So, the user needs to be able to specify which packages to use.

I actually don't think this is necessary, it should already be possible with `stack ghci --with-ghc flycheck --ghci-options my-extra-opt-1 --ghci-options my-extra-opt-2`, if flycheck accepts all the same args as ghc.

afcady:
Oh, I see.

mgsloan:
Hrm, actually `defaultBuildOptions` has `boptsTargets = []`.  I'm rather confused as to why this would work, as if there are no targets then `ghciSetup` won't do much cleverness, and we won't get any `-package-id` flags.

afcady:
Oh, I see.  Didn't read that yet when I replied.  I guess it would fail on certain projects.

As far as `--with-ghc`, I didn't know you could do that -- but flycheck isn't a command line app, it's elisp inside emacs which launches an external process via `stack ghc` on the file that you're editing in emacs (or, actually, a copy).  It just adds options to redirect output.

So I think what has to happen to fix what you're talking about is to go from a ``x.hs`` filename (the info available to flycheck) to a cabal target.

---

>  I'm rather confused as to why this would work, as if there are no targets then ghciSetup won't do much cleverness, and we won't get any -package-id flags.

Well, it produces the same package-id flags that would be produced with a simple ``stack ghci`` with no other args.  (Which is not an empty set, as I verified with `ps`.)  Actually `stack ghci --with-ghc echo` gets you those flags too :)

afcady:
I did a `git pull --rebase` and `git push -f` (editing git history) and I guess that caused github to hide this conversation which had previously been visible here:

https://github.com/afcady/stack/commit/7897a9da2178651401a9a7adb555e8ced4188126#commitcomment-15877233

mgsloan:
So is `stack ghci --with-ghc echo` sufficient?  I'm trying to avoid adding stuff when unnecessary.  One issue is that I don't like the `ghcCmd` name.  No other command is camel case. 

We already have `stack ghc` as an alias for `stack exec ghc`.  How about adding something like this as a subcommand of `query` - `stack query ghc-args`?  I'd want it to take all the arguments that `stack ghci` takes, but with the following conceptual defaults: `--no-load --no-build --no-interpreter`.  I say conceptual, because probably the code paths would be different beyond `ghciSetup`.

Sorry to push back on code you've worked on, but we want things to fit together nicely into a cohesive whole.

afcady:
> So is `stack ghci --with-ghc echo` sufficient?

No it isn't.  For two reasons:

1. The issue you raised, that the default cabal target is not going to work if different targets have conflicting packages.  So ultimately the package arguments are not even correct.

2. `ghci` isn't `ghc`, so the arguments need to be edited: `--interactive` needs to be stripped off, `-threaded` ideally needs to be *re-*added (if it was removed), `-optP/tmp/ghci29597/cabal_macros.h`doesn't look like it should be there, etc.  

  Of course, this could be done with some kind of external program instead of `echo`, that gets the `-package` args out of the command line and then constructs a new command line (which must be kept in sync with the version of `stack` that is calling it) and then calls it.  Ultimately such a program would have to solve problem #1 too.  Since it does need to be written, I'd say the logical place for it is therefore within the stack program itself.  Right?

-----

>  One issue is that I don't like the ghcCmd name.

There's a miscommunication going on here -- you're looking at the wrong patch!  Look at a diff between my master and the master of commercialhaskell -- there is no such name anymore.  I removed it before even submitting the pull request, although I didn't amend my git history.  (I guess that was a mistake, I should have submitted a single commit.)

Here is the pull request with all four commits combined:

https://github.com/commercialhaskell/stack/pull/1737/files

What I did in the final version of the pull request is modify the default behavior for `stack ghc` and the reason I did it that way is that, on my computer where I'm writing code, this fixes flycheck error checking, thus restoring my workflow to what it was (very pleasant!) before I added a dependency on `cryptonite` -- without me having to modify/reconfigure `flycheck` or otherwise write any elisp or anything like that.

But it does seem to me that modifying the default behavior of `stack ghc` is a good idea here, even beyond fixing flycheck my own box.  It seems to me that `stack ghc`'s behavior is simply not correct, unless it receives the package-id args (and the correct ones).

>Sorry to push back on code you've worked on, but we want things to fit together nicely into a cohesive whole.

That's OK.  Just make sure you look at the right patch :)

mgsloan:
Ah, I see!  More food for thought on this: https://github.com/commercialhaskell/stack/issues/1208 https://github.com/commercialhaskell/stack/issues/1388

> -optP/tmp/ghci29597/cabal_macros.hdoesn't look like it should be there, etc.

You actually probably do want that, if you want cabal CPP macros to work.

> I removed it before even submitting the pull request, although I didn't amend my git history. (I guess that was a mistake, I should have submitted a single commit.)

Yeah, it's better to leave out the previous attempt, to avoid confusion.

> But it does seem to me that modifying the default behavior of stack ghc is a good idea here, even beyond fixing flycheck my own box. It seems to me that stack ghc's behavior is simply not correct, unless it receives the package-id args (and the correct ones).

`stack ghc` simply runs ghc in stack's environment. These args come from knowing which package we're trying to build, which requires targets.

> The issue you raised, that the default cabal target is not going to work if different targets have conflicting packages. So ultimately the package arguments are not even correct.

Yeah, that's why we need to know which targets the user cares about.  https://github.com/commercialhaskell/stack/issues/1361 is relevant, we should be able to determine the target the user cares about, based on which file they are compiling.

afcady:
>>    -optP/tmp/ghci29597/cabal_macros.h

> You actually probably do want that, if you want cabal CPP macros to work.

Ha!  So it should.  Turns out that's another bug in flycheck+stack!  The cabal CPP macros indeed do *not* work in flycheck.

(Of course, just adding the argument isn't enough to fix it, because the file also has to be generated (and, ideally, removed) -- another problem for a `stack ghci --with-ghc=echo` kind of approach.)

> #1361 

Seems about right.  One kink though is that `flycheck` does not run `stack ghc` on the actual file, but instead saves from the editor buffer into a temporary file and runs `stack ghc` on that.

Probably it's too much for stack to guess the original filename --  flycheck should supply more info (although this breaks compatibility with existing flycheck installs).  But the functionality of mapping filename -> cabal target  still seems to fall into the purview of stack.

Seems to me like a possible plan would be something like this:

1. Add a flag `--no-interactive` to `stack ghci` which disables adding `--interactive` and also disables anything else that is done solely for `ghci` (like removing `-threaded` -- and maybe that's it?)

2. Add a parameter `--filename` (or whatever) to `stack ghci` that derives the target from a filename (but *does not* add the filename to the `ghc` command line).

  (Alternatively: add another syntax to the target parser.  Maybe `source-file:` to go along with `exe:`, `test:`, and `bench:`.)

3. Modify `flycheck` to call `stack ghci --no-interactive --filename <original filename in project> <temp file>` instead of `stack ghc <temp file>` (iff stack version is high enough)

(That creates a weirdness where `stack ghci --no-interactive` != `stack ghc` but the latter could be made an alias of the former if desired.)

----

Thoughts?

afcady:
So I got cabal_macros.h included, fixing cabal CPP in flycheck, but I had to change some things around to do so.  Specifically, that `ghci29597` in `optP/tmp/ghci29597/cabal_macros.h` changes on every compile, forcing a recompilation of everything.  I changed the filename to exist under `.stack-work/` and to use the hash of the file content as the basename.

https://github.com/afcady/stack/compare/ghci-no-interactive~2...afcady:ghci-no-interactive?diff=unified&name=ghci-no-interactive

mgsloan:
Hey, sorry for letting this languish!  Things got busy

> But the functionality of mapping filename -> cabal target still seems to fall into the purview of stack.

True!  How about "stack query targets-for FILE"  and "stack query default-target-for FILE`.  It doesn't really fit into query's current tree-of-data schema, but it's experimental, so we can change how it works (ideally maintaining compatibility with the existing queries).  As already mentioned, the implementation can be shared with #1361

> Add a parameter --filename (or whatever) to stack ghci that derives the target from a filename (but does not add the filename to the ghc command line).
>
> (Alternatively: add another syntax to the target parser. Maybe source-file: to go along with exe:, test:, and bench:.)

The targets parser is already pretty overloaded.  However, none of our existing target syntax allows ".hs" or ".lhs" (well except dirs, and in that case it's unambiguous what the user means).

> (That creates a weirdness where stack ghci --no-interactive != stack ghc but the latter could be made an alias of the former if desired.)

I think that's fine.  Retrospectively, using `repl`, like cabal, may have been a better choice.  Then, `stack ghci` could just be an alias of `stack exec ghci` (which is a rather useful thing independently of `stack ghci`!).

> Modify `flycheck` to call `stack ghci --no-interactive --filename <original filename in project> <temp file>` instead of `stack ghc <temp file>` (iff stack version is high enough)

Seems reasonable!  You could also just demand that the user updates their stack version since this will work quite a bit better.

mgsloan:
Going to go ahead and close this, thanks for the discussion!  I opened #2053 to track further work here

Issue #1515, opened 2015-12-14
stack setup on OSX El Capitan fails with --version: command not found


smoothdeveloper:
I'm trying to run stack setup to install ghc under OSX El Capitan but it results with this output:

```
$ stack setup
Run from outside a project, using implicit global project config
Using resolver: lts-3.18 from implicit global project's config file: /Users/tejasvinee/.stack/global-project/stack.yaml
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Already downloaded.                 
Configuring GHC ...Running /private/var/folders/nq/2r9cgdrn5d538hjt7dmydgfc0000gn/T/stack-setup526/ghc-7.10.2/configure --prefix=/Users/tejasvinee/.stack/programs/x86_64-osx/ghc-7.10.2/ exited with ExitFailure 1
checking for path to top of build tree... /private/var/folders/nq/2r9cgdrn5d538hjt7dmydgfc0000gn/T/stack-setup526/ghc-7.10.2
Build platform inferred as: x86_64-apple-darwin
Host platform inferred as: x86_64-apple-darwin
Target platform inferred as: x86_64-apple-darwin
GHC build  : x86_64-apple-darwin
GHC host   : x86_64-apple-darwin
GHC target : x86_64-apple-darwin
checking for perl... /opt/local/bin/perl
checking for a BSD-compatible install... /opt/local/bin/ginstall -c
checking whether ln -s works... yes
checking for gsed... /opt/local/bin/gsed
checking XCode version... 7.2
configure: XCode version component 1: 7
configure: XCode version component 2: 2
checking for gcc... /opt/local/bin/gcc
checking for llc-3.5... no
checking for llc... no
checking  is version 3.5... no
checking for opt-3.5... no
checking for opt... no
checking  is version 3.5... no
checking for ld... /opt/local/bin/ld
checking for gcc... /opt/local/bin/gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... 
/private/var/folders/nq/2r9cgdrn5d538hjt7dmydgfc0000gn/T/stack-setup526/ghc-7.10.2/configure: line 3861: --version: command not found
/private/var/folders/nq/2r9cgdrn5d538hjt7dmydgfc0000gn/T/stack-setup526/ghc-7.10.2/configure: line 4108: --version: command not found
[...]

PR #1740, opened 2016-02-04
note in install_and_upgrade regarding #1515

smoothdeveloper:
Small note prompting to run `xcode-select --install` if C compiler is not found while calling `stack setup`


sjakobi:
Please ignore the Travis build failure here. The PR is based on 964db1cd38915cc57f0310c5ff7ccf45e31e3c64 which introduced a compilation problem with GHC-7.8 which has since been fixed.

mgsloan:
Thanks!

PR #1743, opened 2016-02-04
Expose "other-modules" #1733

### LINKED ISSUE #1733: Export stack's `other-modules` so they aren't forgotten?
I just implemented a function that is really similar to [`findDirUp` in `Path.Find`](https://github.com/commercialhaskell/stack/blob/8b010060b0d79f8455562a2716c68b72796f25b6/src/Path/Find.hs#L28). I happened to find `Path.Find` only by chance after the fact.

I believe I would have known about `Path.Find` if it appeared in the generated haddocks for stack, which it doesn't because it's listed as an `other-module` in `stack.cabal`.

As stack doesn't intend to be a library anyways, would it hurt to export the `other-modules`?

If we don't end up exporting these modules I'd really like to have something like `stack haddock --other-modules`.

sjakobi:
This has the benefit that we now have haddocks for all of stack's internal
modules. This even includes "Paths_stack", assuming that others would be
as curious about its contents as me.

Fixes #1733.

### DIFF
--- a/stack.cabal
+++ b/stack.cabal
@@ -51,9 +51,32 @@ flag static
 library
   hs-source-dirs:    src/
   ghc-options:       -Wall -fwarn-tabs -fwarn-incomplete-uni-patterns -fwarn-incomplete-record-updates
-  exposed-modules:   Options.Applicative.Builder.Extra
+  exposed-modules:   Control.Concurrent.Execute
+                     Data.Aeson.Extended
+                     Data.Attoparsec.Args
+                     Data.Attoparsec.Combinators
+                     Data.Attoparsec.Interpreter
+                     Data.Binary.VersionTagged
+                     Data.IORef.RunOnce
+                     Data.Maybe.Extra
+                     Data.Set.Monad
+                     Distribution.Version.Extra
+                     Network.HTTP.Download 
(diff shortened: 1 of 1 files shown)


sjakobi:
One potential issue with including `Paths_stack` is that it shows some internal filepaths. On my system these are:

```haskell

bindir     = "/home/simon/src/stack/.stack-work/install/x86_64-linux/lts-5.0/7.10.3/bin"
libdir     = "/home/simon/src/stack/.stack-work/install/x86_64-linux/lts-5.0/7.10.3/lib/x86_64-linux-ghc-7.10.3/stack-1.0.3-8gbSHoEAZVg7jVpdBWh7Yj"
datadir    = "/home/simon/src/stack/.stack-work/install/x86_64-linux/lts-5.0/7.10.3/share/x86_64-linux-ghc-7.10.3/stack-1.0.3"
libexecdir = "/home/simon/src/stack/.stack-work/install/x86_64-linux/lts-5.0/7.10.3/libexec"
sysconfdir = "/home/simon/src/stack/.stack-work/install/x86_64-linux/lts-5.0/7.10.3/etc"
```

mgsloan:
Yeah, that's expected of `Paths_` modules, though.  Thanks!

Issue #1774, opened 2016-02-10
Stack setup fails for ghcjs project on windows


martin-kolinek:
`stack setup` fails when executed in a project setup for ghcjs like this:

```
compiler: ghcjs-0.2.0.20151230.3_ghc-7.10.2
compiler-check: match-exact
setup-info:
 ghcjs:
  source:
   ghcjs-0.2.0.20151230.3_ghc-7.10.2:
    url: "https://github.com/nrolland/ghcjs/releases/download/v.0.2.0.20151230.3/ghcjs-0.2.0.20151230.3.tar.gz"
```

The error is `Expected a single directory within unpacked ...`, and happens on windows.

**Steps to reproduce**
1. Have a windows system without ghcjs installed
2. Create a project with stack.yaml such as [this one](https://github.com/commercialhaskell/stack/files/124868/stack.yaml.txt) (based on the guide)
3. Run `stack setup`

**Expected:**
Stack sets up ghcjs

**Actual**
Stack fails with `Expected a single directory within unpacked ...`

Whole output:

```
Preparing to install GHCJS to an isolated location.
This will not interfere with any system-level installation.
Preparing to download ghcjs-0.2.0.20151230.3_ghc-7.10.2 ...
No sha1 found in metadata, download hash won't be checked.
Preparing to download ghcjs-0.2.0.20151230.3_ghc-7.10.2 ...
Already downloaded.
Unpacking GHCJS into C:\Users\kot\AppData\Local\Programs\stack\x86_64-windows\ghcjs-0.2.0.20151230.3_ghc-7.10.2\src\ 
Preparing to download 7z.dll ...
Already downloaded.
Preparing to download 7z.exe ...Already downloaded.
Expected a single directory within unpacked C:\Users\kot\AppData\Local\Programs\stack\x86_64-windows\ghcjs-0.2.0.20151230.3_ghc-7.10.2.tar.gz
```

Output of `stack --version`:

```
Version 1.0.2, Git revision fa09a980d8bb3df88b2a9193cd9bf84cc6c419b3 (3084 commits) x86_64
```

I [attached](https://github.com/commercialhaskell/stack/files/124883/out.txt) the output of `stack setup --verbose`

mgsloan:
Doh!  Thanks for finding and fixing this :D

RaoulSchaffranek:
Unfortunately, the bug seems to have sneaked in again with version 1.6.3:

~~~
stack --version
[...]

PR #1775, opened 2016-02-10
Fix withUnpackedTarball7z to find name of srcDir after unpacking #1774

martin-kolinek:


martin-kolinek:
Fixes issue #1774

mgsloan:
Thanks!

PR #1788, opened 2016-02-14
Fix path to FAQ

gasi:
Seems like the file extension was wrong.


gasi:
Never mind, saw the note about file extensions.

PR #1801, opened 2016-02-17
Add 'allow-different-user' flag and configuration option #471, fix #1777

### LINKED ISSUE #1777: Directory ownership test fails with docker
`stack build` in a repo that builds in a docker container results in

> You are not the owner of '/home/mgsloan/.stack/'. Aborting to protect file permissions.
> Retry with '--allow-different-user' to disable this precaution.

Despite me owning that folder.  `stack exec --allow-different-user -- whoami` reports `stack`, and `stack exec --allow-different-user -- ls -la /home/mgsloan` shows a `.stack` folder owned by `stack`, so I'm not sure what's going wrong here.  I haven't looked at the code yet, could be something simple.

Docker is discussed a bit in the PR - https://github.com/commercialhaskell/stack/pull/1599 - but I didn't test it.

sjakobi:
Users other than the owner of the ~/.stack directory are now prevented
from using a stack installation in order to avoid problems with file
permissions. To disable this precaution users can pass the
--allow-different-user flag or use the 'allow-different-user'
configuration option in their ~/.stack/config.yaml.

On Windows, the new flag and configuration options have no effect.

When stack is re-spawned in Docker, the ownership check is skipped
(see #1777).

This reverts commits 0a89c9c29d744a4955e62b59774036e98d6b1e8e and
218e7ddb90c346d814474cb6b8c5e9c67c4706f8.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE TemplateHaskell #-}
@@ -30,6 +31,8 @@ module Stack.Config
   ,getIsGMP4
   ,getSnapshots
   ,makeConcreteResolver
+  ,checkOwnership
+  ,getInContainer
   ) where
 
 import qualified Codec.Archive.Tar as Tar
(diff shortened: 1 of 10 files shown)


src/Stack/Config.hs
```diff
@@ -289,6 +296,11 @@ configFromConfigMonoid configStackRoot configUserConfigPath mresolver mproject c
          configAllowNewer = fromMaybe False configMonoidAllowNewer
          configDefaultTemplate = configMonoidDefaultTemplate
 
+     configAllowDifferentUser <-
+        case configMonoidAllowDifferentUser of
+            Just True -> return True
+            _ -> getInContainer
```
sjakobi: This disables the ownership check inside containers as per @borsboom's [recommendation](https://github.com/commercialhaskell/stack/issues/1777#issuecomment-183122824).

mgsloan:
Awesome, thanks!

PR #1804, opened 2016-02-18
Allow some build options to be set in stack.yaml

drwebb:
See configSpec.hs test for build options that can set in a stack.yaml.

Breaking up the parsing of buildOpts into a buildOptsMonoid and buildOptsCLI isn't ideal in my opinion, but it's a stepping stone to something better I'm sure

These changes touch a lot of code and probably should have more testing, though I haven't found anything broken. I'm tossing it out there now in a PR as it's been an off and on evening project and it's being going on long enough, so I would like to get it off my shoulders.

Please see commit messages for a more details.


borsboom:
As mentioned earlier, going to hold off merging until after the next Stack release.

PR #1815, opened 2016-02-20
Fix #1813: Check package.yaml and cabal file modification times

### LINKED ISSUE #1813: No way to disable hpack running
If I have used non-upstream version of hpack, now I'm forced to also use non-upstream version of stack - as I cannot disable regeneration of `cabal` files.

This now breaks many of our CI builds as they pull the latest stack binary in. Some of cabal files are generated with patched `hpack` or manually edited afterwards.

phadej:


mgsloan:
Hey, thanks for working on this, #1813 ought to be fixed one way or another.

I don't like this approach because if the user accidentally `touch`es the cabal file, it'll stop updating.  That seems very opaque to me, so closing.

phadej:
@mgsloan but that's why how e.g. `make` works. User can then `touch` the `package.yaml` file? Of course we could save the hash of cabal file contents to be even more sure, but IMHO this is a good first step.

As mentioned, currently stock `stack` is completely unusable for our company.

mgsloan:
Yeah, I considered this further - since you'd have to `touch` the `*.cabal` after the modifying `package.yaml`, this seems fine.

mgsloan:
@sol What do you think of making this part of hpack's behavior?  Or is it not so necessary there because you tend to explicitly run it.

mgsloan:
Considering that this also resolves #1828, I think i ought to merge this.  Thanks!

One minor reason to do this in hpack is that hpack determines the output cabal file name.  This isn't really an issue, though, since we enforce that there's only one cabal file per dir.

PR #1822, opened 2016-02-22
Adds hpack support for when initing projects.

silky:
This allows stack templates to define only a `package.yaml` file instead of a
`.cabal` file.

Let me know if this needs a bit more finesse around how it generates the cabal files.  I.e. I could probably unify how it walks the dir path generating as it goes if a package file is found, instead of walking twice.

### DIFF
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -205,8 +205,8 @@ applyTemplate project template nonceParams dir templateText = do
                    throwM (InvalidTemplate template (show e)))
     when (M.null files) $
          throwM (InvalidTemplate template "Template does not contain any files")
-    unless (any (".cabal" `isSuffixOf`) . M.keys $ files) $
-         throwM (InvalidTemplate template "Template does not contain a .cabal\
+    unless (any (\x -> or [".cabal" `isSuffixOf` x, x == "package.yaml"]) . M.keys $ files) $
+         throwM (InvalidTemplate template "Template does not contain a .cabal or `package.yaml`\
                                           \ file")
     liftM
         M.fromList
--- a/src/Stack/Solver.hs
+++ b/src/Stack/Solver.hs
@@ -52,6 +52,7 @@ import           Prelude
 import           Stack.BuildPlan
 import           Stack.Constants             (stackDotYaml)
 import           Stack.Package               (printCabalFileWarning
+                                             , hpack
                                              , readPackageUnresolved)
 import           Stack.Setup
 import           Stack.Setup.Installed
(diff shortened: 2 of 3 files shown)


harendra-kumar:
> I could probably unify how it walks the dir path generating as it goes if a package file is found, instead of walking twice.

I don't think we should care about that kind of optimization here.

src/Stack/New.hs
```diff
@@ -205,8 +205,8 @@ applyTemplate project template nonceParams dir templateText = do
                    throwM (InvalidTemplate template (show e)))
     when (M.null files) $
          throwM (InvalidTemplate template "Template does not contain any files")
-    unless (any (".cabal" `isSuffixOf`) . M.keys $ files) $
-         throwM (InvalidTemplate template "Template does not contain a .cabal\
+    unless (any (\x -> or [".cabal" `isSuffixOf` x, x == "package.yaml"]) . M.keys $ files) $
```
harendra-kumar: It will be nice if we can restrict the lines to 80 columns.

src/Stack/Solver.hs
```diff
@@ -468,12 +469,20 @@ checkResolverSpec gpds flags resolver = do
...
 
--- | Finds all files with a .cabal extension under a given directory.
+-- | Finds all files with a .cabal extension under a given directory. If
+-- a `hpack` `package.yaml` file exists, this will be used to generate a cabal
+-- file.
 -- Subdirectories can be included depending on the @recurse@ parameter.
 findCabalFiles :: MonadIO m => Bool -> Path Abs Dir -> m [Path Abs File]
-findCabalFiles recurse dir =
-    liftIO $ findFiles dir isCabal (\subdir -> recurse && not (isIgnored subdir))
+findCabalFiles recurse dir = do
```
harendra-kumar: We can rename findCabalFiles  to something like findProjectFiles and let it take a predicate like (== "package.yaml") or (".cabal" `isSuffixOf`). Then we can call it once for finding package.yaml in the first step (and then run hpack on the result) and then again for finding .cabal files in the next step. That will cleanly separate these two steps, making the code a tiny bit easier to read.

harendra-kumar:
LGTM. I have a few minor suggestions.

harendra-kumar:
Thanks for the PR. I added a few minor changes (17301a526adeffc5a4fbbebe3a7d0a7589472bf7) along with the merge.

silky:
thanks; looks great, much better, sorry i didn't get around to addressing your comments, it was a busy week :)

Issue #1391, opened 2015-11-21
`stack unpack` downloads deprecated package versions


borsboom:
##### Steps to reproduce:

Run `stack unpack stack`
##### Expected:

It unpacks the latest non-deprecated version (as of this filing, `stack-0.1.8.0`).
##### Actual:

It unpacks `stack-9.9.9`, which is deprecated and unbuildable.

```
$ stack unpack stack
stack-9.9.9: download
Unpacked stack-9.9.9 to /Users/manny/fpco/tmp/stack-9.9.9/
```

DNNX:
I just scanned hackage index trying to find what other packages are affected. As it turns out, at the moment 15 out of 9069 packages have `max deprecated-versions >= max non-deprecated-versions`. In other words, `stack unpack`ing any of those 15 packages would download a deprecated version:

```
call	0.2
fptest	0.2.2.0
hydrogen-cli-args	0.9
hydrogen-data	0.7.1
hydrogen-prelude	0.9
hydrogen-syntax	0.7.1
json-autotype	1.0.9
leksah-server	0.8.0.8
leksah	0.8.0.8
ltk	0.8.0.8
peyotls-codec	0.3.1.1
pipes-cereal	0.1.0.0
sasl	0.0.0.2
stack	9.9.9
word12	1.0.0
```

I hope it will be useful for anyone who decides to fix this bug (at least for testing). 

I initially wanted to fix it myself, but it seems that it is not that trivial. We need to either make a request to [Hackage API](https://hackage.haskell.org/package/stack/preferred) each time a user runs `stack unpack`, or enrich the package index with information about package deprecations in order to have it at our disposal at runtime. Currently package index just don't have this info. As far as I understand, package index is created by another tool (not stack), so maybe another issue should be created in that tool's Github repo as well.

mgsloan:
Hmm, it's surprising to me that `stack unpack` would unpack the latest version rather than the snapshot version.  Maybe that's a good way to fix this?  Granted, that doesn't help for the packages that aren't in the snapshot / when you're using a compiler resolver.

borsboom:
Yeah, I could go either way on this.  Not really clear to me whether it's preferable to use the latest version or the snapshot version, so I'd tend toward staying with the current behaviour unless there's a compelling reason to change it (and I'm not sure that this issue qualifies as compelling enough).

[...]

Issue #1441, opened 2015-11-28
Allow GHCi to load non-local packages by unpacking them


mgsloan:
It'd be cool if `stack ghci text` unpacked `text` and loaded it into `ghci`.  This is also relevant to the code recently added for also loading intermediate deps (#584) - it will need to be modified to handle the `PSUpstream` case.

StevenXL:
Is this still an issue? I can run `stack ghci --package text`. Are we talking about something else?

nekoniaow:
> Is this still an issue? I can run `stack ghci --package text`. Are we talking about something else?

This seems to be about packages which are referenced from the `extra-deps` entry in `stack.yaml`.
Example:

in `stack.yaml`:
```
extra-deps:
- webdriver-w3c-0.0.3
```

```
$ stack  ghci package webdriver-w3c

Warning: Some targets (package, webdriver-w3c) are not local packages, and so cannot be directly loaded. In future versions of stack, this might be supported - see
         https://github.com/commercialhaskell/stack/issues/1441 . It can still be useful to specify these, as they will be passed to ghci via -package flags.

Note: No local targets specified, so a plain ghci will be started with no package hiding or package options.
      
      You are using snapshot: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/19/29.yaml
      
      If you want to use package hiding and options, then you can try one of the following:
      
      * If you want to start a different project configuration than <path/to/project>stack.yaml, then you can use stack init
        to create a new stack.yaml for the packages in the current directory. 
        
      * If you want to use the project configuration at <path/to/project>stack.yaml, then you can add to its 'packages' field.
      
Configuring GHCi with the following packages: 
GHCi, version 9.0.2: https://www.haskell.org/ghc/  :? for help
```

And with the `--package` option:
```
$ stack  ghci --package webdriver-w3c
Using main module: 1. Package `<project-package>' component <project>:exe:<project>-exe with main-is file: <path/to/project>/app/Main.hs
The following GHC options are incompatible with GHCi and have not been passed to it: -threaded
Configuring GHCi with the following packages: <project>
```

As you can see, the `webdriver-w3c` package is totally ignored.

As I understand it, that's the problem pointed by this issue.

nekoniaow:
A more functional workaround in the case I described above would be to use the `stack unpack` command instead:

1. comment the package entry in `stack.yaml:extra-deps`,
[...]

Issue #1261, opened 2015-10-31
Allow targets to override snapshot versions?


mgsloan:
Thanks to zedbourbaki in #haskell IRC for bringing this up.  If you try to override a package version that's in the resolver, via `stack install`, it refuses:

```
mgsloan@computer:~$ stack install hmatrix-0.17.0.1
Run from outside a project, using implicit global config
Using resolver: lts-3.10 from global config file: /home/mgsloan/.stack/global/stack.yaml
Error parsing targets: Specified target version 0.17.0.1 for package hmatrix does not match snapshot version 0.16.1.5
```

This is a pretty good message, but particularly when using the global project, to me it makes sense to implicitly make this into an extra-dep.  Some of the checkboxes on #272 seems to suggest that this was intended at some point:

> - To force a specific version of a package to be installed, can run `stack install foo-1.2.3`
> - For ease of use, can provide `stack install foo:latest` to ignore whatever's in the snapshot

borsboom:
I agree, seems like that should work based on #272.

PR #1839, opened 2016-02-25
[rfc] only unpack package versions within preferred-versions #1391

luigy:
first pass at #1391

* fetches version present in snapshot otherwise falls back to hackage
* caches `preferred-versions` from index
* only versions within `preferred-versions` are fetched unless explicitly asked by package identifier
* added --latest flag to fetch latest version from hackage regardless of resolver

TODO:

- [ ] Don't copy `loadBuildPlan` from `Stack.BuildPlan` into `Stack.Fetch`
- [x] Update `Stack.Upgrade` to only consider non-deprecated versions.

New behavior:

```sh
λ cat stack.yaml
resolver: lts-5.4

λ stack unpack stack
Unpacked stack-1.0.2 to /Users/luigy/stack-1.0.2/

λ stack --resolver lts-3.0 unpack stack
Unpacked stack-0.1.3.0 to /Users/luigy/stack-0.1.3.0/

λ stack unpack --latest stack
Unpacked stack-1.0.4.1 to /Users/luigy/stack-1.0.4.1/

λ stack unpack stack-9.9.9
Unpacked stack-9.9.9 to /Users/luigy/stack-9.9.9/
```

### DIFF
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -293,7 +293,7 @@ resolvePackagesAllowMissing menv idents0 names0 = do
 
     toTuple' (PackageIdentifier name version) = (name, [version])
 
-    groupByPackageName = fmap Set.fromList . Map.fromListWith mappend . map toTuple' . Map.keys
+    groupByPackageName = fmap Set.fromList . Map.fromListWith (<>) . map toTuple' . Map.keys
 
     toVersionRange (_, PreferredVersionsCache raw) = fromMaybe anyVersion $ parse raw
       where parse = simpleParse . T.unpack . T.dropWhile (/= ' ')
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -55,7 +55,7 @@ import           Data.List.NonEmpty             (NonEmpty)
 import qualified Data.List.NonEmpty             as NE
 import           Data.Map                       (Map)
 import qualified Data.Map                       as Map
-import           Data.Maybe                     (maybeToList, catMaybes, fromMaybe)
+import           Data.Maybe                     (maybeToList, catMaybes)
 import           Data.Monoid                    ((<>))
 import           Data.Set                       (Set)
 import qualified Data.Set                       as Set
@@ -64,8 +64,6 @@ import           Data.Text.Encoding             (decodeUtf8)
 import           Data.Typeable                  (Typeable)
 import           Data.Word                      (Word64)
 import qualified Data.Yaml                      as Yaml
-import           Distribution.Text              (simpleParse)
-import           Distribution.Version           (anyVersion)
 import           Network.HTTP.Client            (checkStatus)
 import           Network.HTTP.Download
 import           Network.HTTP.Types.Status
@@ -269,12 +267,9 @@ resolvePackagesAllowMissing
     -> Set PackageName
     -> m (Set PackageName, Set PackageIdentifier, Map PackageIdentifier ResolvedPackage)
 resolvePackagesAllowMissing menv idents0 names0 = do
-    (caches, preferred) <- getPackageCaches menv
-    let preferredVersion = maybe anyVersion toVersionRange . flip Map.lookup preferred
-        latestApplicable name vs =
-            fromMaybe (Set.findMax vs) $ latestApplicableVersion (preferredVersion name) vs
-        versions = Map.mapWithKey latestApplicable $ groupByPackageName caches
-        (missingNames, idents1) = partitionEithers $ map
+    (caches, _preferred) <- getPackageCaches menv
+    versions <- getLatestApplicablePackageCache menv
+    let (missingNames, idents1) = partitionEithers $ map
             (\name -> maybe (Left name) (Right . PackageIdentifier name)
                 (Map.lookup name versions))
             (Set.toList names0)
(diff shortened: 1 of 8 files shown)


src/Stack/Fetch.hs
```diff
@@ -228,6 +289,18 @@ resolvePackagesAllowMissing menv idents0 names0 = do
                 , rpIndex = index
                 })
 
+    toTuple' (PackageIdentifier name version) = (name, [version])
+
+    groupByPackageName = fmap Set.fromList . Map.fromListWith mappend . map toTuple' . Map.keys
+
+    filterBy' pvs name vs =
```
mgsloan: Why not `filterBy` without the prime?

mgsloan:
LGTM so far!  Are those checkboxes supposed to be checked? (in the issue description)

It'd be cool, though not required, to have tests that the snapshot version is unpacked, that `--latest` works, and the preferred versions stuff works

mgsloan:
Also, note that it's failing to build for 7.8.  Without looking at the error, most of the time I get that it's due to needing to use `liftM` instead of `fmap` in some circumstances.

luigy:
>Are those checkboxes supposed to be checked? (in the issue description)

Maybe :)
For the first one I was opening it up for suggestions about the duplication of `loadBuildPlan`
I have taken care of the second one, but not sure about the name `getLatestApplicablePackageCache`

I have also added tests and have a, probably too fragile, way to get the latest non-deprecated from hackage to check against `stack unpack --latest stack` test result.
Maybe just make sure that it exits successfully ?

mgsloan:
Yeah, I'd be fine with just making sure that it exits successfully for the latest case.  It's sufficient to just check that `stack unpack --latest stack` does not download `stack-9.9.9`

mgsloan:
Thanks for adding a test!

For avoiding copying `loadBuildPlan`, how about moving the code for getting the versions from snapshot packages into `Main.hs`?  I don't mind having some logic in Main.hs, particularly if it's concerned with a CLI convenience.

I have an enhancement suggestion: how about only loading the configuration / build plan if there are package names to resolve.  So if the user does `stack unpack specific-version-1.0`, we don't even need a valid stack.yaml.  This is another reason to put this logic in `Main.hs`, since that's where the configuration gets loaded.

Ooh, and I just thought of something.  I think we want `stack unpack` to also pay attention to extra-deps.  This would mean using something like `loadSourceMap AllowNoTargets defaultBuildOpts`, I think.  This might be overkill, I don't think this is strictly necessary, but it would also be good to have clear messaging in the following case:

* If the user attempts to unpack a local package (like `stack unpack pkg`, and `pkg` is one of the project's packages):

  - If the package exists on hackage, and --latest is specified, warn that we're unpacking a version from hackage, and not using the local one.

  - If the package exists on hackage, but --latest is not specified, error out explaining that it's a local package and to download from hackage, a specific version or --latest is required.

  - If the package does not exist on hackage, error out explaining that the user tried to unpack a local package.

luigy:
I'm back! let's :ship: this one! :tada:  :clock1:

>For avoiding copying loadBuildPlan, how about moving the code for getting the versions from snapshot packages into Main.hs?

I like! done :heavy_check_mark: 

>I have an enhancement suggestion: how about only loading the configuration / build plan if there are package names to resolve.

Ommiting the buildplan! done :heavy_check_mark:

Hmm, omitting the configuration :heavy_multiplication_x: 
What about private package index? upcoming extensible snapshots?
It makes sense but, I have to think a bit more about this one. I'm in between convenience vs. being explicit about it.  We should look more into the global [`--no-config`](https://github.com/commercialhaskell/stack/issues/1388#issuecomment-158703708) flag for that :smile:

>Ooh, and I just thought of something. I think we want stack unpack to also pay attention to extra-deps. 

Meaning that it should fetch the extra-dep's version if it shadowed the snapshot one when only given a package name?

Makes sense!  This behavior goes more inline with what you were suggesting in #1441, right?

:bangbang: In any case, we should start trying to make sure that we are maintaining a consistent behavior across commands when taking into account the `config` :mag:
For example, should #1441 also accept `stack ghci text-1.2.0.0` similar to `stack unpack`? 
Similar to the expected behavior of #1261 

--

>If the package exists on hackage, and --latest is specified, warn that we're unpacking a version from hackage, and not using the local one.

:+1: 

>If the package exists on hackage, but --latest is not specified, error out explaining that it's a local package and to download from hackage, a specific version or --latest is required.

:+1: makes sense in order to show a consistent behavior when handling the `config`.  Currently, I was ignoring `extra-deps` and just going straight to the `snapshot`.  Also, once again, if they really want the snapshot version a `--no-config` flag can come in handy 

>If the package does not exist on hackage, error out explaining that the user tried to unpack a local package.

:+1:  but also considering the config's custom indices/snapshots, right? same goes for the first suggestion

mgsloan:
> What about private package index? upcoming extensible snapshots?

Good point!  The private package index option does mean that the config should always be loaded. 
 
> > Ooh, and I just thought of something. I think we want stack unpack to also pay attention to extra-deps.
> Meaning that it should fetch the extra-dep's version if it shadowed the snapshot one when only given a package name?

Yup

> For example, should #1441 also accept stack ghci text-1.2.0.0 similar to stack unpack? 
Similar to the expected behavior of #1261

Yes indeed!

mgsloan:
@luigy Hey!  I'm doing a bit of a PR cleanup sweep, how's it going?

I think potentially the discussion above is making this a bit complicated.  How about lets get an initial version of this in - just the bit with the code cleanup

mgsloan:
Since this has languished for a while, I've added a link to it from the issue.

PR #1840, opened 2016-02-25
alex, happy, and libgmp where necessary

choener:
the ghc (non-stack) builds will sometimes need alex or happy.
having gmp-dev is not enough, we need the actual libs as well.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1842,30 +1842,30 @@ matrix:
   include:
   # We grab the appropriate GHC and cabal-install versions from hvr's PPA. See:
   # https://github.com/hvr/multi-ghc-travis
-  - env: BUILD=cabal GHCVER=7.0.4 CABALVER=1.16
+  - env: BUILD=cabal GHCVER=7.0.4 CABALVER=1.16 HAPPYVER=1.19.5 ALEXVER=3.1.7
     compiler: ": #GHC 7.0.4"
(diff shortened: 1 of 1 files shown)


mgsloan:
Hm, I don't quite follow.  What is different for the builds that use cabal instead of stack?

I can believe that if any of our transitive dependencies require alex or happy, then we might need to install them explicitly for the cabal based travis script.  However I don't see any build failures in travis history [here](https://travis-ci.org/commercialhaskell/stack/builds) that would indicate that this would be a problem.  Is the issue somehow hidden by caching?

choener:
Hi,

happy/alex: This concerns ghc+cabal only builds, not stack builds.
gmp: this concerns stack nightly builds.

I originally copied the travis config from here:
https://www.fpcomplete.com/blog/2016/02/updated-haskell-travis-config?preview=true
and tried it with my 'bimaps' library:
https://travis-ci.org/choener/bimaps/jobs/110405080
line 443 complains about missing 'happy'
The updates are against the one in guide.md.

stack nightly dies like this:
https://travis-ci.org/choener/bimaps/jobs/110407882

Gruss,
Christian

mgsloan:
Ohh, sorry, I missed that this was being applied to the guide rather than stack's CI configuration.  Thanks!

mgsloan:
Ahh, sorry!  I didn't catch that this was applied to the docs rather than stack's CI.  Danke!

PR #1849, opened 2016-02-27
Allow --omit-packages even without --solver

harendra-kumar:
Sometimes you already know that dependencies will not be met and want to create
a config anyway which you will fix later. In that case we can skip using
--solver and directly use --omit-packages to create a config with packages
commented out in the config.


mgsloan:
LGTM, thanks!

PR #1850, opened 2016-02-27
Improve the generated stack.yaml

harendra-kumar:
Add more details in comments
Wrap lines after 80 columns.


src/Stack/Init.hs
```diff
@@ -176,65 +177,124 @@ renderStackYaml p ignoredPackages dupPackages =
...
+        , "will not be run. This is useful for tweaking upstream packages."
+        ]
+
+    footerHelp = commentHelp
+        [ "Control whether we use the GHC we find on the path"
+        , "system-ghc: true"
+        , ""
+        , "Require a specific version of stack, using version ranges"
+        , "require-stack-version: -any # Default"
+        , "require-stack-version: >= 1.0.0"
```
luigy: this one has always bothered me since it is not valid yaml the version range should be quoted :)
this seems to be updated manually with version bumps, but why not just grab it from Paths_stack dropping the minor version?

harendra-kumar: It should be quoted at least. Not sure if we really need to update this comment on version bumps. Though grabbing it from Paths_stack will be nice. I can try that.

mgsloan:
LGTM, thanks!

PR #1852, opened 2016-02-28
Clone recursively

### LINKED ISSUE #1764: Repos specified by git should have an option to initialize submodules
We probably don't want to wind up turning the `location:` block into a full git porcelain, but it would be very handy to at least have Stack to a `git submodule update --init` after checking out a particular revision, otherwise packages that have submodules won't build. I often run into this with FFI wrappers around a C library that don't have particularly sophisticated uses of submodules, so we might not need much configurability.

Maybe something like:

```
- location:
    git: https://github.com/acfoltzer/hs-ed25519-donna.git
    commit: 7e99a67cae6e3946873ca071536ff0ad3cb23948
    init-submodules: true
    extra-dep: true
```

If there's an equivalent for Mercurial, it would also be nice to support that while we're at it, I just am not familiar with it.

robrix:
`git` packages using submodules may need to be cloned recursively.

- Allows us to pass arguments to the `clone` command on a case-by-case basis.
- Adds the `--recursive` option to `git clone`.
- `hg clone` will [already clone subrepos](https://www.mercurial-scm.org/wiki/Subrepository#Update).

I have not added a configuration option to control this, because it doesn’t seem particularly useful in the context of repos which are Haskell packages.

Fixes #1764.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -576,13 +576,14 @@ resolvePackageLocation menv projRoot (PLRemote url remotePackageType) = do
     unless exists $ do
         ignoringAbsence (removeDirRecur dirTmp)
 
-        let cloneAndExtract commandName resetCommand commit = do
+        let cloneAndExtract commandName cloneArgs resetCommand commit = do
                 ensureDir (parent dirTmp)
                 readInNull (parent dirTmp) commandName menv
-                    [ "clone"
-                    , T.unpack url
+                    ("clone" :
+                    cloneArgs ++
+                    [ T.unpack url
                     , toFilePathNoTrailingSep dirTmp
-                    ]
+                    ])
                     Nothing
                 readInNull dirTmp commandName menv
                     (resetCommand ++ [T.unpack commit])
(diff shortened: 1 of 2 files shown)


src/Stack/Config.hs
```diff
@@ -576,10 +576,10 @@ resolvePackageLocation menv projRoot (PLRemote url remotePackageType) = do
...
         ignoringAbsence (removeDirRecur dirTmp)
 
-        let cloneAndExtract commandName resetCommand commit = do
+        let cloneAndExtract commandName cloneArgs resetCommand commit = do
                 ensureDir (parent dirTmp)
-                readInNull (parent dirTmp) commandName menv
-                    [ "clone"
+                readInNull (parent dirTmp) commandName menv $
+                    "clone" : cloneArgs ++ [
                     , T.unpack url
```
silky: typo here; don't want that leading `,` (this caused the builds to fail)

robrix:
Whoops—my bad!

robrix:
Oof. My bad again. Excuse me while I correct this.

robrix:
There we go, green builds.

mgsloan:
Thanks!  Can you please squash this into one commit?

robrix:
Consider it squashed.

mgsloan:
LGTM, Thanks!!

robrix:
:bow: Thank you!

acfoltzer:
Thanks!

Issue #1752, opened 2016-02-05
Stack requires presence of extra-deps even when not needed for target, or, cannot specify extra-deps at build-time


acfoltzer:
Minimal example: a project that includes two packages:
- a Haskell code generator `generator` that only depends on `base`
- a package `foo` that depends on `base` and the generated package `generated`

Since `foo` won't build without its dependencies listed in `stack.yaml`, I have to list `generated` in `stack.yaml`, but running Stack commands without that package in place fails even if that package is not needed for the requested target.
#### Steps to reproduce
- Set up the directory:

```
% tree
.
├── foo
│   ├── ...
│   └── foo.cabal
├── generator
│   ├── ...
│   └── generator.cabal
└── stack.yaml
% cat stack.yaml
resolver: lts-5.1
packages:
- location: foo/
- location: generator/
  extra-dep: true
- location: generated/
  extra-dep: true
% cat foo/foo.cabal
name:                foo
[...]
library
  build-depends:       base >=4.7 && <5,
                       generated
```
- `stack build generator --exec 'generator --build-me-some-code -o generated'`
- `stack build foo`
#### Expected

I expect Stack to build the `generator` executable, and then run it to produce a cabal package in `/generated`, so that I can then build `foo` now that its dependency exists.
#### Actual

```
% stack build generator --exec 'generator --build-me-some-code -o generated'
Could not resolve directory /private/tmp/ex/generated/
```

I can create the empty directory, but then Stack wants a cabal file:

```
% mkdir generated
% stack build generator --exec 'generator --build-me-some-code -o generated'
No .cabal file found in directory /private/tmp/ex/generated/
```

---
#### Alternate scenario

I could imagine not including `generated` in the `stack.yaml`, but instead being able to pass it as a dependency when running `stack build foo`.

harendra-kumar:
hmm, interesting use case. Currently stack requires all packages to have a valid cabal file present which is not possible in this case. We could ignore the missing cabal files as long as they are not needed for building the requested target. Once that is done your two step build should work as expected. 

This should be fixed. A related issue is #1722 , perhaps both of these can be fixed together.

Until this is fixed you can try a workaround using two stack.yaml files, one to only generate the code (without having foo in it) and another one using the generated code. You can use --stack-yaml to select the appropriate stack.yaml. For example `stack build foo --stack-yaml stack.foo.yaml` .

mgsloan:
[...]

Issue #1745, opened 2016-02-05
Add a proper system for stack warnings suppression / collection / display 


mgsloan:
From the discussion [here](https://github.com/commercialhaskell/stack/issues/1705#issuecomment-180062695):

>  It'd be nice is every warning message could be disabled (analogous to GHC's `-fno-warn-*` options), which would cover this particular issue as well. It'd also be nice if repeated warnings were automatically suppressed and they could be emitted together for a given package instead of being interspersed with other output.

> Another reason to have a system like this is so that they can be moved to the end of the output. Currently it's easy for them to get lost in the midst of output.

> We've also sometimes needed to restructure code to plumb warning / error info to the end of output. It'd be nicer to store away warnings / errors at the location that generates them.

mgsloan:
It'd also be nice to be able to accumulate non-fatal errors.  In this case, stack should continue work, but at the end output the errors and exit with a nonzero code.  (see https://github.com/commercialhaskell/stack/issues/2065)

mgsloan:
Another good warning case to present at the end of output is when the config file has unrecognized fields.

Issue #1722, opened 2016-02-01
Do not require .cabal for all packages in stack.yaml when running stack runghc


harendra-kumar:
I saw [this script](https://github.com/massysett/pinchot/blob/b945ab83c6e2b3f83e697bd755c8892f82fcfca1/buildprep) in the pinchot package. It is moving stack.yaml out of the way before invoking a script using `stack runghc`.

When we use runghc on a file inside a project dir do we need settings from `stack.yaml`? Or can we just avoid reading it? 

This will allow us to freely invoke scripts inside the project dir. For example the pinchot package is generating .cabal file using a script. When the script is run via `stack runghc` it finds a stack.yaml but no .cabal so it fails when trying to process the stack.yaml.

mgsloan:
It is necessary to read stack.yaml for `stack runghc`, because it determines the package databases which are used with it.  However, there isn't any reason for `stack runghc` to demand that all listed packages have cabal files.  It seems to me like the right solution to this is to only do this check (and similar checks, if any) when the packages list is actually needed.

harendra-kumar:
worth doing for use cases like this one?

mgsloan:
Yeah, it seems reasonable to me!  It also means that `stack exec` in general will work even if cabal files aren't there.

We will have to figure out how to draw the line of when things get validated.  I'd say that all validation for the stack.yaml that's purely based on its contents should happen up-front.  Validation that requires some filesystem state, such as the presence of cabal files, should happen lazily.

luigy:
```sh
#!/bin/sh

mv stack.yaml stack-old.yaml
stack --resolver=lts-3.14 runghc genCabal.hs --package cartel --install-ghc > pinchot.cabal
mv stack-old.yaml stack.yaml
```

In the example above this happens because `--package cartel` triggers a `Stack.Build.build`  which tries to load local packages and errors out even though they are not wanted.  I think this is another example that sometimes we need `loadSourceMap` to return warnings and not error out or for us to catch them and handle them accordingly.  An `exec` without `--package TARGET` should not trigger that behavior, though.

In a way, it seems that sometimes we want a way to force stack to use the global config or a way to tell it ignore local packages because we are only going to work with snapshot.  Maybe we can have `stack path --global-config-location` and then
`stack --stack-yaml "$(stack path --global-config-location) runghc args"`


[...]

PR #1858, opened 2016-02-29
Detect extra-deps missing from indices

### LINKED ISSUE #1521: Typos in extra-deps aren't detected
Noticed this while helping L8D in #haskell 

Repro:

1) `stack new issue1521`
2) `cd issue1521`
3) Add `hspec >= 2.2.1` to the test-suite build-depends
4) Add the following to `extra-deps`:
```
- hspec-2.2.1 
- hspec-core-2.2.1 
- hspec-discover-2.2.1 
- hspec-epectations-0.7.2 
```
(Note the mis-spelled `hspec-epectations`)
5) `stack test`
6) Get this error:
```

      hspec-core: needed (==2.2.1), couldn't resolve its dependencies
      hspec-expectations: needed (==0.7.2.*), 0.7.1 found (latest applicable is 0.7.2)
    needed for package: hspec-2.2.1

      hspec-expectations: needed (==0.7.2.*), 0.7.1 found (latest applicable is 0.7.2)
    needed for package: hspec-core-2.2.1

      hspec: needed (>=2.2.1), couldn't resolve its dependencies
    needed for package: issue1521-0.1.0.0
```

luigy:
Stack used to silently accept `extra-deps` from config or cli that were missing from indices.
It now errors out when they are not found 
(will add package suggestions for typos in another PR)

fixes #1521

### DIFF
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -439,7 +439,10 @@ extendExtraDeps :: (MonadThrow m, MonadReader env m, HasBuildConfig env)
                 -> Map PackageName Version -- ^ latest versions in indices
                 -> m (Map PackageName Version) -- ^ new extradeps
 extendExtraDeps extraDeps0 cliExtraDeps unknowns latestVersion
-    | null errs = return $ Map.unions $ extraDeps1 : unknowns'
+    | not (Map.null missing) =
+        throwM $ UnknownPackageIdentifiers (toPkgIdents missing) ""
+    | null errs =
+        return $ Map.unions $ extraDeps1 : unknowns'
     | otherwise = do
         bconfig <- asks getBuildConfig
         throwM $ UnknownTargets
@@ -447,7 +450,12 @@ extendExtraDeps extraDeps0 cliExtraDeps unknowns latestVersion
             Map.empty -- TODO check the cliExtraDeps for presence in index
             (bcStackYaml bconfig)
   where
-    extraDeps1 = Map.union extraDeps0 cliExtraDeps
+    (missing, extraDeps1) =
+        Map.partitionWithKey missingFromIndex $ Map.union extraDeps0 cliExtraDeps
+
+    missingFromIndex name _ = isNothing $ Map.lookup name latestVersion
+
+    toPkgIdents = Set.fromList . map fromTuple . Map.toList
 
     (errs, unknowns') = partitionEithers $ map addUnknown $ Set.toList unknowns
     addUnknown pn =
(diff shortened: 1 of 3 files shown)


mgsloan:
Awesome, thanks!

Can these please get turned into warnings?  Having it as an error like this is indeed more straightforward, however, having this as an error rules out the following use cases:

* Using an environment that does not have all of the required packages.  For example, if you usually run with a docker container that has proprietary packages, but would still like to be able to build some packages outside of the container.

* If the `stack.yaml` is required to build tools / run scripts which populate the package DB with some more packages, satisfying the extra-deps.

This may be a bit tricky when they're also accompanied by build plan failures as in the example in https://github.com/commercialhaskell/stack/issues/1521 .  Ideally they'd come at the end of the output in that case, since it's probably the issue.  I haven't looked at how the code would need to change to do that.  Feel free to keep this as an exception constructor even if it's never thrown.

luigy:
Oh, I see! yes, definitely this should be delayed till later.

>This may be a bit tricky when they're also accompanied by build plan failures as in the example in #1521 . Ideally they'd come at the end of the output in that case, since it's probably the issue.

stack can have `loadSourceMap` return these as warnings and catch build plan failures  in `Stack.Build.build` to display the messages in a better order.  A change like this to `loadSourceMap` could also be used by a possible solution to   #1752 #1722 that also depend, in a way, on `loadSourceMap` not throwing and holding off till wether the plan succeeds or not.  Stack will have to make it clear that is doing this 

-- 

Hmm should extra-deps coming through the cli still error out, though, to match `stack`'s original behavior? and only warn on the ones coming from the config?

luigy:
looks like something for #1745 could help with display order of warnings

mgsloan:
Yep, I encourage you to work on #1745 if you want to!  Particularly if it makes the implementation of this cleaner (it probably would).

I haven't thought about it thoroughly, but this could look like adding some `IORef [Message]` type to `Env`.  `runStackT` would always output these messages (even on exception), so that they don't get lost.

mgsloan:
@luigy Hey! I'm doing a bit of a PR cleanup sweep, how's it going?

This change seems a bit related to https://github.com/commercialhaskell/stack/pull/1839 .  If it makes it easier, feel free to merge them into one PR (potentially a new PR if you like)

mgsloan:
@luigy I pondered how to implement this given all the various concerns, and it was indeed quite tricky!  Turns out we needed to plumb the extraDeps info out of `loadSourceMap` and then check it against the installed packages.  Here's the commit: https://github.com/commercialhaskell/stack/commit/ced6ea8439aebff5a6225b6f9e3569416f08410e

Thanks for working on this!

PR #1860, opened 2016-03-01
Use hpack (wip)

sol:


sol:
This is an attempt to use `hpack` on `stack`.  It uses an unreleased version of `hpack` that adds has support for flags.

Before using this we would need to add two more features to `hpack`

- Support for `buildable: False`
- Support for `ld-options`

@mgsloan In case you are interested in this, can you cherry-pick 0757cae to `master` already, so that we will not run into conflicts later on.

mgsloan:
Hey!  Sorry for letting this languish so long that it got all conflict-ey.  It'd be really great to have a tool to generate hpack from cabal.

It would indeed be good to dogfood hpack so that we test its integration with stack more.  At once, I'm not sure if we want to do that at this time.  Thoughts, @borsboom ?

mgsloan:
Going to go ahead and close this.  That doesn't mean we aren't going to do it, just not this particular swing at it.

Issue #1862, opened 2016-03-01
Executing stack ghci with local packages defined in stack.yaml hides existing package?


beerendlauwers:
Ok, so the title might be a little confusing. Here's the issue:

* I have a package on Github called [`selenese`](https://github.com/beerendlauwers/haskell-selenese).
* I have another package on Github called [`selenese-intern`](https://github.com/beerendlauwers/haskell-selenese-intern), which has a dependency on `selenese`.
* I've defined the package dependency in `selenese-intern`'s `stack.yaml` file like so:

```YAML
packages:
- location:
    git: https://github.com/beerendlauwers/haskell-selenese.git
    commit: HEAD
```

* Running `stack ghci` on a freshly cloned `selenese-intern` on Windows 7, Stack Version `0.1.10.0, Git revision a25892f8374d4bcbed6dd0f0c3bcb6cb93df490c (2778 commits) x86_64` results in the following output:

```ghci
Setting codepage to UTF-8 (65001) to ensure correct output from GHC
Configuring GHCi with the following packages: selenese
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Test.WebDriver.Selenese ( D:\xxxx\xxxxx\xxxx\Seleniu
m\haskell-selenese-intern\.stack-work\downloaded\0706fe8cbc9917e86157eb4f07b385b
352d1bff6efd8df2635e2b96acf689f1a\src\Test\WebDriver\Selenese.hs, interpreted )
Ok, modules loaded: Test.WebDriver.Selenese.
*Test.WebDriver.Selenese Test.WebDriver.Selenese>
```

I am unable to load the module `Test.WebDriver.InternFormatter` from `selenese-intern`:

```
*Test.WebDriver.Selenese Test.WebDriver.Selenese> :l Test.WebDriver.InternFormatter

<command line>:
    Could not find module `Test.WebDriver.InternFormatter'
    Perhaps you meant Test.WebDriver.Internal
    Use -v to see a list of the files searched for.
Failed, modules loaded: none.
Prelude>
```

Am I doing something obviously wrong, here?

luigy:
selenese-intern's stack.yaml only knows about the [remote](https://github.com/beerendlauwers/haskell-selenese-intern/blob/master/stack.yaml#L7) package :p 

it is missing the package that is in the same directory as the stack config.
should be like this:

```yaml
packages:
- location: '.'
- location:
[...]

PR #1867, opened 2016-03-02
Clarified the role of the `- '.'` line.

beerendlauwers:
As mentioned in #1862 (specifically [this comment](https://github.com/commercialhaskell/stack/issues/1862#issuecomment-191173247)).

* Role of the `- '.'` line has been clarified
* A note has been added to the git-only example to reinforce this.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -42,7 +42,7 @@ packages:
 extra-deps: []
 ```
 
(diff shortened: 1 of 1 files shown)


mgsloan:
Thanks!  I further clarified that with https://github.com/commercialhaskell/stack/commit/ca685c7a1a93d2b07290689ab5df38db28ea4acf - there can only be one package per dir.

PR #1885, opened 2016-03-08
Fix use of package specified by a url to a tarball

### LINKED ISSUE #1884: Bug when a package is specified by a url to a tarball
This is a small, but serious, error introduced in https://github.com/commercialhaskell/stack/commit/99eb520bc3099afacb1403098838a5280ee76e32 (on 27 Jan).  I'll send a PR in a few minutes to fix it and provide a test case.

Using a reference like this in `stack.yaml` is now broken:
```
packages:
- https://hackage.haskell.org/package/half-0.2.2.3/half-0.2.2.3.tar.gz
```
It produces an error along the lines of "Archive extraction failed. We support tarballs and zip, couldn't handle the following URL, ...", but the underlying problem is that it is trying to use the tarball file name as the directory into which to unpack!

[This is nothing to do with the particular package, which has been chosen here simply because it is a small one for testing purposes.]

paul-rouse:
Fixes #1884 and adds integration test for it.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -601,7 +601,7 @@ resolvePackageLocation menv projRoot (PLRemote url remotePackageType) = do
                         liftIO $ withBinaryFile fp ReadMode $ \h -> do
                             lbs <- L.hGetContents h
                             let entries = Tar.read $ GZip.decompress lbs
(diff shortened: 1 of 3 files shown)


mgsloan:
LGTM, thanks for adding a test!

Issue #1587, opened 2016-01-02
Select appropriate msys2/MinGW shell on Windows


tejon:
msys2 provides three shells: the full-POSIX msys2 shell, plus 64- and 32-bit MinGW shells which provide a GNU tool set but use a lighter, more optimized translation layer. Matching this, there are up to three versions of any msys2 `pacman` package: the unprefixed msys2 version, and architecture-prefixed MinGW versions.

Currently, stack uses the msys2 shell in all cases. This is poor behavior; in fact, the msys2 docs suggest only using that shell for system upgrades and `pacman` itself, and in practice I've found that MinGW packages install fine from the bit-appropriate MinGW shell. Correct behavior would be to pick MinGW64 or MinGW32 to match the GHC version in the active build strategy, and only use the raw msys2 shell when performing automated installation/upgrades (e.g. as suggested in #1586).

This is not a mere matter of optimization or correctness for its own sake: [`sdl2`](https://github.com/haskell-game/sdl2) fails to build with `stack build` on the Windows commandline, as it doesn't even have an `msys2` package version; but with the MinGW64 `pkg-config` and `sdl2` packages installed, manually launching the `mingw64_shell.bat` installed by stack and running `stack --skip-msys build` works fine. (Note if you want to reproduce this: there's a buggy header in `sdl2-2.0.3`, to get a successful build you'll need the replacement [here](https://hg.libsdl.org/SDL/raw-file/e217ed463f25/include/SDL_platform.h)).

borsboom:
@snoyberg, do you have an opinion on this?  I don't know the history of how the current design was reached.

snoyberg:
I'm open to any suggested changes here from people who know better. I did not understand any of the details of these "multiple shells" when setting up the code. @tejon As you seem to have a pretty good understanding of this, would you like to take a crack at a PR on this?

tejon:
Apologies for being slow getting to this, but yes, I'm giving it a shot. My Haskell's still at the rough end of intermediate, though. I don't suppose you could give me a quick run-down of the relevant code paths? I.e. those involved in selecting a build strategy (particularly selecting x86 vs x64), determining the host is Windows and locating/invoking msys. Figuring out where to look for things is often the longest and most tedious part of the process for me, heh.

mgsloan:
[...]

PR #1909, opened 2016-03-13
Architecture-aware Windows MSYS2 path/env config 

tejon:
Addresses #1587.

Note: tested only on 64-bit Windows.


mgsloan:
Hey, sorry for letting this languish.  I don't follow the details 100%, but if this makes things better I'm all for it.

Issue #1914, opened 2016-03-15
Installation Instructions Keyserver


mpdehnel:
Hi,

The installation instructions state that you need to run the following command:
`sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 575159689BEFB442`

This is fine if your firewall doesn't block it, but this isn't the case for many people (port 11371 is often blocked). The workaround for this is to use the following key server instead:

`sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 575159689BEFB442`

If you could update the installation instructions to reflect this option, I'm sure it will save a lot of people a load of work!

Thanks,

Martin Dehnel-Wild

mgsloan:
Good idea!  Feel free to open such a PR changing the docs in the repo!

PR #1916, opened 2016-03-16
Change default keyserver URL to port 80 not 11371 (to prevent firewall issues in installation)

mpdehnel:
The current installation instructions state that you need to run the following command:
`sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 575159689BEFB442`

This is fine if your firewall doesn't block it, but this isn't the case for many people (port 11371 is often blocked). The workaround for this is to use the following url for the key server instead:

`sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 575159689BEFB442`

This is the pull request changing this, as suggested here: https://github.com/commercialhaskell/stack/issues/1914

Martin Dehnel-Wild

### DIFF
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -96,7 +96,7 @@ problems, see the linked FAQ entries:
 
  1. Get the FP Complete key:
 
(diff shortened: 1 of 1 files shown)


mgsloan:
Thanks!

Issue #1892, opened 2016-03-09
Avoid unnecessarily loading the hackage index


mgsloan:
Loading the hackage is expensive, because the encoded index cache is 18Mb.  On my fairly fast computer, it takes 0.3 seconds to decode it.  Search `getPackageCaches` to find the spots it's used.

There are two cases where loading the hackage currently happens unnecessarily:

1) In `loadSourceMap`, because the user might have specified a package target on the commandline which is not in the resolver.  In this case, it uses the latest from hackage.

2) In `constructPlan`, because plan construction errors get annotated with "latest applicable" versions.

For both of these, loading the hackage index is only necessary under specific and abormal circumstances.  However, currently this expensive operation is being done regardless of need.

It's also used in `resolveBuildPlan`, but that only seems to be used by the tests.

luigy:
>Loading the hackage is expensive, because the encoded index cache is 18Mb. On my fairly fast computer, it takes 0.3 seconds to decode it. 

Hmm, eventually we got to steal `hoogle` decoding tricks :zap:, but in the meantime lazy me just wants to write a separate cache like I do in #1839 since it looks like there are a few cases in which all we care about is the latest version.  I tested throwing this info in the same cache and the size is around ~505Kb.

>2) In constructPlan, because plan construction errors get annotated with "latest applicable" versions.

The smaller cache mentioned above can help with this and also for `show` command in #1614

```sh
$ stack show latest-version base
4.8.2.0
```

Also as another enhancement we can fork reading the cache as soon as we hit a command that might need it + using an MVar for `configPackageCaches`

mgsloan:
> Hmm, eventually we got to steal hoogle decoding tricks ￼, but in the meantime lazy me just wants to write a separate cache like I do in #1839 since it looks like there are a few cases in which all we care about is the latest version. I tested throwing this info in the same cache and the size is around ~505Kb.

Currently working on a fast serialization library - this is one of the applications we have in mind!  https://www.fpcomplete.com/blog/2016/03/efficient-binary-serialization

> Also as another enhancement we can fork reading the cache as soon as we hit a command that might need it + using an MVar for configPackageCaches

True!  In general we might be able to make things a little faster by running such blocking fetches in parallel.

luigy:
>Currently working on a fast serialization library - this is one of the applications we have in mind!

Excellent!
so if I want this now, I have to be join the fp force... hmm :stuck_out_tongue_winking_eye:

[...]

PR #1936, opened 2016-03-22
In 'loadSourceMap', load the package caches only when necessary

sjakobi:
See https://github.com/commercialhaskell/stack/issues/1892 for context.


mgsloan:
Awesome, thanks!

Issue #55, opened 2015-05-24
`stack hoogle` command


chrisdone:
I want this! It should:

1. Run hoogle with the right database according to the stack config in the directory in which it's run (so it's easily ran from a terminal or by IDEs).
2. Automatically download the hoogle database from stackage.org e.g. `https://www.stackage.org/snapshot/lts-2.1/db.hoo` for lts-2.1.
3. Put it in some reasonable place like `~/.stack/hoogle-databases/lts-2.1.hoo`.
2. Or when running inside docker, use the hoogle database provided by the docker.

The last docker point means we probably need to sync up some things.

With this command I could provide some nice Emacs support with a live hoogle search buffer, and I'd basically never have to leave Emacs to open my browser to read docs again. This closes off two big pain points for me: (1) always having to open the browser and leave my editor environment, (2) needing an internet connection to view docs. This will also basically make my whole `haskell-docs` package redundant which I'll be able to deprecate in favor of `stack`.

Pinging @manny-fp and @snoyberg for feedback/opinions.

ndmitchell:
I've just pushed something to Hoogle about 10 patches before the minimum-viable product, but it's enough to see where the holes are. Given Hoogle HEAD, you can do `hoogle generate --local` and it will use `ghc-pkg` to figure out which packages are installed and where there docs are, then `hoogle map` and it will find `map` or `hoogle server` and it will launch a server at `localhost` which you can search from. Clicking links will still go to the remote docs, but it is based on the local Hoogle info.

When doing that under Stack there are 122 packages, but only 23 have docs (those shipped with GHC), because Stack doesn't include docs by default. To make `stack hoogle` work reasonably then `stack hoogle` would have to generate docs for the immediate dependencies of a module (at least), either always as it was going, or when you typed `stack hoogle`.

borsboom:
Cool!  `stack build --haddock` (aka `stack haddock`) already generates docs incrementally (i.e. it will build them for dependencies that don't already have them), so doing that implicitly before `stack hoogle` (with an option to skip for the impatient) would probably work.

ndmitchell:
That works as a solution - it's just a shame that you have to reconfigure all the packages and that it's so slow. Also it fails on Windows, but I'm sure that's fixable: #1266.

borsboom:
One thing that will help with this is that we're planning to support settings defaults for build options on the configuration file.  That means you could configure Stack to build haddocks for dependencies by default.

[...]

PR #1939, opened 2016-03-22
Add stack hoogle command (#55)

chrisdone:
Pinging interested parties: @ndmitchell @borsboom @mgsloan @snoyberg 

## The feature

This is a regular `exec` interface to Hoogle, but it adds a `--database` argument. This saves me from judging a new interface to Hoogle. The basic use-case is self-explanatory:

``` haskell
$ stack hoogle Conduit
module Data.Conduit
type Conduit i m o = ConduitM i o m ()
module Data.CSV.Conduit
[...]
```

To generate a database, one runs:

    $ stack hoogle -- generate --local

If you run `stack hoogle foo` without a database file present, it will automatically run `stack hoogle --generate --local`:

```
$ stack hoogle Conduit
Hoogle database doesn't exist.
Running: stack hoogle -- generate --local
You can re-run this command later to update the database.
Starting generate
Reading ghc-pkg... 0.11s
[1/142] Cabal... 0.42s
[2/142] Glob... 0.01s
[...]
```

## Details & issues

The database argument is set to `$PROJECT_ROOT/.stack-work/hoogle` as suggested by @ndmitchell in https://github.com/commercialhaskell/stack/issues/55#issuecomment-155186311.

The issue is that this support is available in Hoogle 5, which is not released yet. The workaround for users is to run:

    $ git clone git@github.com:ndmitchell/hoogle.git
    $ cd hoogle
    $ stack install

With `hoogle` version 5 in the `~/.local/bin`, the above feature works.

### DIFF
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -386,6 +386,7 @@ data SpecialExecCmd
     = ExecCmd String
     | ExecGhc
     | ExecRunGhc
+    | ExecHoogle
     deriving (Show, Eq)
 
 data ExecOptsExtra
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -305,6 +305,10 @@ commandLineHandler progName isInterpreter = complicatedOptions
                     "Run ghc"
                     execCmd
                     (execOptsParser $ Just ExecGhc)
+        addCommand' "hoogle"
+                    "Search the local packages with Hoogle"
+                    execCmd
+                    (execOptsParser $ Just ExecHoogle)
         addCommand' "ghci"
                     "Run ghci in the context of package(s) (experimental)"
                     ghciCmd
@@ -999,11 +1003,13 @@ execCmd :: ExecOpts -> GlobalOpts -> IO ()
 execCmd ExecOpts {..} go@GlobalOpts{..} =
     case eoExtra of
         ExecOptsPlain -> do
+            (manager,lc) <- liftIO $ loadConfigWithOpts go
             (cmd, args) <- case (eoCmd, eoArgs) of
                  (ExecCmd cmd, args) -> return (cmd, args)
                  (ExecGhc, args) -> return ("ghc", args)
                  (ExecRunGhc, args) -> return ("runghc", args)
-            (manager,lc) <- liftIO $ loadConfigWithOpts go
+                 (ExecHoogle, args) ->
+                     return ("hoogle", makeHoogleArgs (lcConfig lc) args)
             withUserFileLock go (configStackRoot $ lcConfig lc) $ \lk ->
               runStackTGlobal manager (lcConfig lc) go $
                 Docker.reexecWithOptionalContainer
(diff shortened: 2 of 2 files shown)


chrisdone:
Oh, idea: we could have `stack haddock` automatically update the hoogle database, perhaps with a configuration entry.

mgsloan:
Awesome!

Why make this part of the exec command?  None of the exec options are relevant to hoogle.  I can't think of any use for `--plain`, `--no-ghc-package-path`, `--no-stack-exe`, or `--package`.  So, I'd rather keep the codepath separate (though perhaps using Stack.Exec.exec, which is used in a fair number of places).

As for the hoogle version thing, how about running `hoogle --numeric-version` and giving the user some instructions when it's < 5.

ndmitchell:
The fact that Hoogle 5 isn't released yet more than a year after I was aiming for does suck, and makes something like this harder, but I'm very grateful to @chrisdone for using the new version in preference (I owe :beers:).

One way to sidestep version issues would have to have `stack hoogle` install its own version of Hoogle, from git, into somewhere Stack controlled, and then use that - much like it does with GHC.

mgsloan:
Yep, auto-installing hoogle 5 would be cool!

chrisdone:
> Why make this part of the exec command?

I started a separate codepath but found myself repeating the code in that codepath, and saw that it uses `exec` to launch GHC specially, so figured it was an okay place to put it. The code is a bit hairy so I didn't feel like refactoring it.

> As for the hoogle version thing, how about when the hoogle invocation fails, run `hoogle --numeric-version` and giving the user some instructions when it's < 5.

That's a nice idea. I'd considered auto-installing but thought that might be a bit invasive.

> One way to sidestep version issues would have to have stack hoogle install its own version of Hoogle, from git, into somewhere Stack controlled, and then use that - much like it does with GHC.

Re-considering it, perhaps it could use one of the .tar.gz from Github (avoiding the need for the Git dependency for users who don't have it) and yeah, just install it in Stack's own directory structure. That's consistent with the use of GHC. That gets around the problem of when Hoogle 5 is going to be released. We'd like for this to just work out of the box for newbies and experts alike. We can announce the feature in the next Stack release as simply `stack hoogle blah`.

One nice part of this feature is that you can hoogle your own codebase.

A separate note: perhaps it should also run `stack haddock .` automatically before building a new hoogle database, to make sure it's there? I just realised I'd performed this step myself but a user probably won't have done this.

mgsloan:
I think we're going to a do a release soon, should we add this?  TODO as far as I'm concerned:

- [ ] Not take the `stack exec` options.

- [ ] Check for hoogle version

- [ ] Open an issue about potentially installing hoogle version

- [ ] One of:

    * Have it run haddock first
    * Have it detect when haddock needs to run, and tell the user that
    * Create an issue about the need for this, and document it

chrisdone:
AFK till Monday but I'd like to contribute to getting this in before the
release with auto download of hoogle5. Haddock running as a message sounds
good to me.

On 22 Apr 2016 09:18, "Michael Sloan" <notifications@github.com> wrote:
>
> I think we're going to a do a release soon, should we add this? TODO as
far as I'm concerned:
>
> Not take the stack exec options.
>
> Check for hoogle version
>
> Open an issue about potentially installing hoogle version
>
> One of:
>
> Have it run haddock first
> Have it detect when haddock needs to run, and tell the user that
> Create an issue about the need for this, and document it
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
<https://github.com/commercialhaskell/stack/pull/1939#issuecomment-213297849>

chrisdone:
Working on a new branch with the above things in mind.

PR #1950, opened 2016-03-25
mounts the host's TMPDIR as /tmp in the container

### LINKED ISSUE #1949: Dev tools like flycheck & flycheck-haskell need the host's TMPDIR and HOME mounted in the docker container
I'm using emacs (spacemacs with the default 'haskell' & 'syntax-checker' layer in this case) which means flycheck w/ flycheck-haskell.  While flycheck-haskell accounts for stack usage, it doesn't work for docker enabled stack projects.

flycheck-haskell expects to run `stack --verbosity silent runghc -- ....` on 2 files in the host's $HOME/.emacs.d/elpa/flycheck-haskell/ dir. Of course when you run this on a docker project your $HOME/.emacs.d dir is out of sight when stack runs.  [I'll work with the flycheck project to fix this in a stack-friendly way that accounts for diving into the stack docker container. Like maybe putting the haskell files it needs in $PROJECT/.stack-work/flycheck-haskell or something.]

flycheck-haskell also expects to write files in the temp dir before running stack. These files also disappear as soon as stack is called & it jumps into the container. Having the host's $TMPDIR mounted as the container's /tmp dir seems useful.  

It's a tiny patch. It seems harmless. I've submitted a pull request.  This solves part of the problem with flycheck-haskell in a way that's transparent to flycheck-haskell. (I don't have to ask them to change even more things for stack users.)

I do my work on Linux. How can I make the patch better for windows %TEMP_DIR% or Mac (if needed) ?

reference: 
- https://github.com/flycheck/flycheck-haskell/blob/master/flycheck-haskell.el#L92
- https://github.com/flycheck/flycheck-haskell/blob/master/flycheck-haskell.el#L70
- https://github.com/flycheck/flycheck-haskell/blob/master/flycheck-haskell.el#L118
- https://github.com/flycheck/flycheck-haskell/blob/master/flycheck-haskell.el#L192

dysinger:
This is for development tools like Emacs flycheck. Some tools expect to
be able to write temp files into /tmp and then run `stack ....` to
process them. Except, without this patch & using docker, the /tmp dir 
files are gone.

resolves #1949

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -65,7 +65,7 @@ import           Stack.Docker.GlobalDB
 import           Stack.Types
 import           Stack.Types.Internal
 import           Stack.Setup (ensureDockerStackExe)
-import           System.Directory (canonicalizePath,getHomeDirectory)
+import           System.Directory (canonicalizePath,getHomeDirectory,getTemporaryDirectory)
 import           System.Environment (getEnv,getEnvironment,getProgName,getArgs,getExecutablePath)
 import           System.Exit (exitSuccess, exitWith)
 import qualified System.FilePath as FP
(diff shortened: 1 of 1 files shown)


mgsloan:
Cool, LGTM!  How about using http://haddock.stackage.org/lts-5.9/path-io-0.2.0/Path-IO.html#v:getTempDir to have it be cross-platform?

PR #1951, opened 2016-03-25
mount $HOME in the container

### LINKED ISSUE #1949: Dev tools like flycheck & flycheck-haskell need the host's TMPDIR and HOME mounted in the docker container

dysinger:
This patch is for docker projects and specifically to assist tools (e.g.
Emacs) that read/write files in $HOME for dev work (flycheck & friends)

resolves #1949

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -320,6 +320,7 @@ runContainerAndExit getCmdArgs
           ,"-e","HOME=" ++ toFilePathNoTrailingSep sandboxHomeDir
           ,"-e","PATH=" ++ T.unpack newPathEnv
           ,"-e","PWD=" ++ toFilePathNoTrailingSep pwd
+          ,"-v",toFilePathNoTrailingSep homeDir ++ ":" ++ toFilePathNoTrailingSep homeDir
           ,"-v",toFilePathNoTrailingSep stackRoot ++ ":" ++ toFilePathNoTrailingSep stackRoot
           ,"-v",toFilePathNoTrailingSep projectRoot ++ ":" ++ toFilePathNoTrailingSep projectRoot
           ,"-v",toFilePathNoTrailingSep sandboxHomeDir ++ ":" ++ toFilePathNoTrailingSep sandboxHomeDir


mgsloan:
LGTM!

Does this have any potential interactions with the ssh dir mounting stuff? https://github.com/commercialhaskell/stack/blob/mount-home-in-container/src/Stack/Docker.hs#L331 - or perhaps this allows that simplified.  Probably not, since inside the container there's also the sandbox home dir.

Will everything that expects things in the homedir use absolute paths?  In other words, it seems like `~` will refer to the sandbox home dir, when the tools might expect to find the contents of the home dir that's outside the docker container.

@borsboom Thoughts?

dysinger:
I'll wait for @borsboom's thoughts.  Mounting $HOME could cause unexpected probs.

borsboom:
I think this should work fine with the `.ssh` mounting -- it means `.ssh` will be mounted both in the original home directory and in the sandbox-specific home directory.  You're right that `~/` paths in the container will refer to the sandbox-specific home directory rather than the host's home directory, but I've thought about this before and it needs to stay this way, as annoying as it is.  Otherwise it's just too likely that things like the host's `.bashrc` will cause problems in the container (especially when the host is running a different OS, like OS X).  A possible workaround for that could be to bind-mount specific files (like `.bashrc` etc.) into the host home directory location, but that might be even more confusing.  Worth thinking about doing, but IMO in no way blocks this PR.

borsboom:
By the way (since you mentioned it elsewhere), this should all work fine for directories with spaces, since the whole argument to `-v` is passed as a single argument regardless of it containing spaces (it's not parsed, except by docker to find the `:` separator).

gregwebs:
Seems like this would cause a problem if using a container where the user name in the container (home directory actually) is the same as on the host. Should this be checked first?

mgsloan:
@gregwebs Good point!  That seems like it could likely cause problems.  I've opened https://github.com/commercialhaskell/stack/issues/2108

borsboom:
I'm not sure how much it would cause problems, really.  Stack overrides $HOME in the container anyway, so I guess it would only matter if something you're running in Stack needs files that exist in a path in the home directory within the image.

mgsloan:
@borsboom Does that reasoning also cover the case where the user is named `stack`?

PR #1966, opened 2016-03-29
stack init: allow local packages to be deps of deps

### LINKED ISSUE #1965: Allow stack init when local packages also figure in the deps of deps
This problem was discovered from the question asked on haskell-stack regarding `stack init` of https://github.com/haskell-servant/servant-snap. Which is a bit complicated due to a lot of dependencies cloned in the same project and local packages figuring in the dependencies of dependencies as well.

We use `--only-dependencies` flag for cabal solver. When some local packages also figure in the dependencies of dependencies of other packages in the project then cabal refuses to proceed with the following message:

```
>>>> Cabal errors begin                                                                                                                                                           
cabal: Cannot select only the dependencies (as requested by the                                                                                                                   
'--only-dependencies' flag), the packages are required by a dependency of one                                                                                                     
of the other targets.                                                                                                                                                             
<<<< Cabal errors end 
```

I removed the `--only-dependencies` flag from the cabal invocation to allow this case. I could not think of any potential issues due to this but I might be missing something. Will send a pull request.

harendra-kumar:
Fixes #1965

### DIFF
--- a/src/Stack/Solver.hs
+++ b/src/Stack/Solver.hs
@@ -50,6 +50,7 @@ import           Path
 import           Path.Find                   (findFiles)
 import           Path.IO                     hiding (findExecutable, findFiles)
 import           Prelude
+import           Safe                        (headMay)
 import           Stack.BuildPlan
 import           Stack.Constants             (stackDotYaml)
 import           Stack.Package               (printCabalFileWarning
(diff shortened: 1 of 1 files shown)


mgsloan:
Thanks!

PR #1967, opened 2016-03-30
Update the packages section of yaml config guide

harendra-kumar:
It lacked some crucial information to understand the behavior unambiguously.
Also, organized it systematically.

Please review for correctness and style.


mgsloan:
Thanks!

Issue #1837, opened 2016-02-24
Build only some of the Docker images


snoyberg:
Currently, `stack image container` always builds all images specified in your stack.yaml file. It would be convenient in some use cases to give a list of Docker images that should be built instead.

mgsloan:
@dysinger I merged the PR before realizing that it takes these image names via an undecorated argument.  I don't think this is what we want, because in the future we might also want to take TARGET syntax like the normal build command.

How about putting this option under a flag like `stack image container --name` or `stack image container --image` or something?

dysinger:
How would multiple images look? `stack image container --image black --image orange --image red` ?

mgsloan:
Yup, or possibly also `stack image container --image "black orange red"`

dysinger:
It seems reordering tickets on waffle.io adds back the "in progress" label

borsboom:
It looks like this issue can be closed.  Please reopen if I'm incorrect.

PR #1972, opened 2016-03-31
options not args (& hindent johan-tibell 4 spaces)

dysinger:
`stack image container` image names as options (not arguments)


dysinger:
resolves #1837

Issue #1148, opened 2015-10-12
Need a way to change global-stack-root


rrnewton:
`stack path --global-stack-root` reports the root.  But as far as I can see this is hard coded?  There is no analog of `stack --local-bin-path` or an environment variable like `STACK_YAML` to control the root?

My use case is to pick between machine-local and NFS installations of stack.  That is, we may have `~/.stack` necessarily as an NFS-stored directory, for machines where we don't have any local mounts.  But then ones where we do, we could do something like the following

```bash
if [ -d /home.local/me/.stack ]; then
  export STACK_GLOBAL_ROOT=/home.local/me/.stack
...
```

borsboom:
The `STACK_ROOT` environment variable sets the location.  That could definitely be documented better (it's only in the installation instructions).  Not sure why the inconsistency between that name and the `stack path --global-stack-root` argument, might be nice to fix.  And having it overridable on the command-line would be nice too.

Issue #1546, opened 2015-12-20
stack path --ghc-paths broken


mboes:
From the documentation:

```
  --ghc-paths              Installed GHCs (unpacked and archives)
```

I'm not sure how to grok the explanation of the option, but at any rate I'd expect this variable is essentially pointing me to the local GHC install that stack is using for the current project (or indeed the system installed GHC if that's what is being used).

In reality, in the `stack` repository I get:

```
/home/mboes/.stack/programs/x86_64-linux
```

which is a path that in fact does not exist (on my system).

mgsloan:
`stack path --ghc-paths` tells you where `stack setup` will install GHC and other programs.  So, if `stack setup` has never been used, this path won't exist.  I agree that this is a confusing thing for `stack path --ghc-paths` to do.  I propose:

* Deprecating `--ghc-paths`, and adding `--programs-path` to provide the current functionality.
* Adding `--ghc-path`, which points to the GHC install used for the project.  Would you expect the value of `stack --ghc-path` to be something like `/usr/local/bin/ghc`, or just `/usr/local/`?

sjakobi:
> Deprecating --ghc-paths, and adding --programs-path to provide the current functionality. Or maybe just straight up removing it.

I think that the flag loses most of it's value when there's also `--ghc-path`. Removing it should be ok.

> Would you expect the value of stack --ghc-path to be something like /usr/local/bin/ghc, or just /usr/local/?

Personally, I would prefer the more precise result.

What should `stack path --ghc-path` return in the case of a GHCJS project? If it should point at the location of `ghcjs` in that case, should the option possibly be called `--compiler-path` instead?

mgsloan:
Good point!  I'd be in favor of having it be `--compiler-path`

sjakobi:
Fixed via #2016.

PR #1983, opened 2016-04-01
Deprecate 'stack path --global-stack-root', add '--stack-root' options – #1148

sjakobi:


sjakobi:
I have added a deprecation warning that shows up when someone uses the `stack path --global-stack-root` flag explicitely:
```
$ stack path --global-stack-root
'--global-stack-root' will be removed in a future release.
Please use '--stack-root' instead.

/home/simon/.stack
```

Should there also be a warning for `stack path` without any flags? I'm not sure whether it would be obnoxious or maybe even break things. On the other hand people might be confused that `stack path` lists both `global-stack-root` and `stack-root` with the same content:

```
$ stack path
global-stack-root: /home/simon/.stack
stack-root: /home/simon/.stack
...
```

sjakobi:
One aspect of #1148 was that the documentation on `$STACK_ROOT` was somewhat sparse.

I think the situation has already become a bit better with the new entry in `stack --help`:

```
  --stack-root STACK-ROOT  Absolute path to the global stack root directory
                           (Overrides any STACK_ROOT environment variable)
```

Yet should I possibly add an entry to the FAQ, e.g. `How can I change the location of the global stack root directory?`. Or is there a more appropriate place?

ChangeLog.md
```diff
@@ -16,7 +16,9 @@ Behavior changes:
 * For consistency with the `$STACK_ROOT` environment variable, the
   `stack path --global-stack-root` flag and the `global-stack-root` field
   in the output of `stack path` are being deprecated and replaced with the
-  `stack-root` flag and output field. See
+  `stack-root` flag and output field.
+  Additionally, the stack root can now be specified via the
+  `--stack-yaml` command-line flag. See
```
mgsloan: This should be `--stack-root`, right?

src/Stack/Options.hs
```diff
@@ -294,6 +303,15 @@ configOptsParser hide0 =
             hide
   where hide = hideMods (hide0 /= OuterGlobalOpts)
 
+readAbsDir :: ReadM (Path Abs Dir)
```
mgsloan: I can buy that this is needed for good errors.  However, `parseAbsDir` directly will also throw a not-so-bad looking error

mgsloan:
LGTM, thanks!  Just one comment that needs to be addressed.

ChangeLog.md
sjakobi: Of course. Fixed.

sjakobi:
I have added a commit that fixes an exception for permission-checking code that was still `$STACK_ROOT`-specific.

@mgsloan: Have you seen my two comments at the top or do you think these are non-issues?

src/Stack/Options.hs
sjakobi: `ReadM` sadly isn't an instance of `MonadThrow`, which `parseAbsDir` requires.

That might change with https://github.com/pcapriotti/optparse-applicative/pull/196 though.

mgsloan:
> Should there also be a warning for stack path without any flags? I'm not sure whether it would be obnoxious or maybe even break things. On the other hand people might be confused that stack path lists both global-stack-root and stack-root with the same content:

I think removing it from the list is fine.  True, someone could be parsing it.  However, I don't know of any tools that do this.  The breakage should be minimal at this time.

As long as we're doing such a deprecation / change for `--global-stack-root`, it'd be good to also address https://github.com/commercialhaskell/stack/issues/1546

> Yet should I possibly add an entry to the FAQ, e.g. `How can I change the location of the global stack root directory?`. Or is there a more appropriate place?

Thanks for considering docs!  I think the current level of docs is fine.  AFAIK there is little need for `STACK_ROOT` other than having shorter paths on windows, and this is well covered in the windows installation instructions

sjakobi:
> I think removing it from the list is fine.

Fixed with edaae29.

> As long as we're doing such a deprecation / change for --global-stack-root, it'd be good to also address #1546

I'll take a look!

sjakobi:
Rebased.

Issue #1595, opened 2016-01-03
Clone stack templates repo instead of using github API


mgsloan:
In order to resolve #1251, and support #1571, we need to clone the templates repository.  Here's how this would work:

1) If it doesn't exist yet, clone the templates repo to some location in `~/.stack/`.

2) If the repo is dirty, complain to the user about it and skip step (3).  This allows you to easily modify existing templates / add new ones.

3) Fetch and reset to the latest master.  If the user passes `--offline`, then errors while running these git commands will be ignored.  When `--offline` isn't passed and an error occurs, the exception will mention that you can use `--offline`.  See also this issue about `-offline`: #1594

4) Load the template from the file in the cloned repo.

Another detail here is windows newline normalization, as recently mentioned by Lokathor in the #haskell-stack IRC channel.  Since we're using the github API to fetch the file, we currently get `\n` rather than `\r\n`.  This is something that [git can deal with](https://help.github.com/articles/dealing-with-line-endings/#platform-all), at least if we prefer using `\r\n` on windows.

normenmueller:
Want to `stack new foobar new-template` w/o an internet connection but get failed internet connection while stack tries to download `new-template`. I already created several projects w/ stack, so the template is already there. Is there an option to run stack in offline mode or is what this issue is about? Is there a workaround w/ the current version?

mgsloan:
There is not, this issue is tracking the resolution of that problem.

bitemyapp:
>2) If the repo is dirty, complain to the user about it and skip step (3). This allows you to easily modify existing templates / add new ones.

I like this part.

@mgsloan seems like this would trigger a dep. Is there a preferred git client library? I'm having a hell of a time figuring out how to clone a repo with `gitlib`. Should I just invoke git via the command line?

mgsloan:
One downside of this is that it would cause us to really require git.  Stack doesn't currently require git - it's just used as an optimized way of fetching the hackage index, if available.

snoyberg:
The new approach to templates does not require `git` commands or API calls, but downloads from raw.githubusercontent.com instead. Closing.

Issue #1571, opened 2015-12-28
List available "new" templates along with descriptions


mgsloan:
It'd be nice to get a list of available templates, via something like `stack new --list` or `stack templates`.  Ideally, these templates would be accompanied by a short description.

Split off from #1564

kadoban:
Any thoughts on how that should be implemented? I'm attempting to implement some more changes to 'new' at the moment and if I could roll that in it sounds like a really nice addition.

I guess it'd have to require some kind of addition to the template format, ideally something backwards compatible? Maybe there could be a specially named file in the template that if it exists is displayed for that purpose and otherwise not used? That sounds a bit hacky, but should work and nothing else really comes to mind.

mgsloan:
Yeah, a file like `new-template.txt` or `new-template.desc` makes sense to me.

mgsloan:
I think the best implementation would be based on resolving https://github.com/commercialhaskell/stack/issues/1595 .  So, marking this as blocked on that issue.

mgsloan:
An alternative implementation not blocked on #1595 would be to have a single file in the templates repo that associated template names with descriptions / other metadata.  This might actually be preferable to putting the description in a separate file.

Marking as unblocked.

mgsloan:
This got addressed by @narrative.  Thanks!

PR #1985, opened 2016-04-03
Template descriptions

narrative:
Hi! This is my first stab at #1571. There didn't seem to be much of a consensus on how best to implement this feature so I went off of @mgsloan's suggestion

> An alternative implementation not blocked on #1595 would be to have a single file in the templates repo that associated template names with descriptions / other metadata.

Running `stack templates` now downloads and can display optional metadata from a .yaml file located in the stack-templates repository.

The file is setup like this:

``` yaml
quickcheck-test-framework:
  description: a library for random testing of program properties
servant:
  description: a set of packages for declaring web APIs at the type-level
ghcjs:
  description: Haskell to JavaScript compiler, based on GHC
```

Currently the data type I implemented supports `description` and `author` (not used for anything at the moment) fields, however this could be extended relatively easily.

I'm a newbie so criticism is definitely welcome and appreciated.

Thanks


narrative:
I'm going to clean this up and get it compiling with ghc-7.8.4.

PR #1988, opened 2016-04-03
retrieve template metadata from stack-template repo

narrative:
Hi! This is my first stab at #1571. There didn't seem to be much of a consensus on how best to implement this feature so I went off of @mgsloan's suggestion:

> An alternative implementation not blocked on #1595 would be to have a single file in the templates repo that associated template names with descriptions / other metadata.

Running stack templates now downloads and can display optional metadata from a .yaml file located in the stack-templates repository.

The file is setup like this:

``` yaml
quickcheck-test-framework:
  description: a library for random testing of program properties
servant:
  description: a set of packages for declaring web APIs at the type-level
ghcjs:
  description: Haskell to JavaScript compiler, based on GHC
```

Currently the data type I implemented supports description and author (not used for anything at the moment) fields, however this could be extended relatively easily.

I'm a newbie so criticism is definitely welcome and appreciated.

Thanks

Note: The above is copied and pasted from my last PR that I closed. I cleaned up the branch commit history and made a few minor changes to get it compiling with ghc-7.8.4. Still not sure why the AppVeyor build seems to fail randomly (Doesn't *appear* to be related to anything specific to this PR).

Example `stack templates` output:
```
PS C:\Users\erikn\stack> stack exec stack -- templates
Template                    Description
chrisdone
franklinchen
ghcjs                     - Haskell to JavaScript compiler, based on GHC
ghcjs-old-base
hakyll-template           - a static website compiler library
haskeleton                - a project skeleton for Haskell packages
hspec                     - a testing framework for Haskell inspired by the Ruby library RSpec
new-template
quickcheck-test-framework - a library for random testing of program properties
rubik
scotty-hello-world
scotty-hspec-wai
servant                   - a set of packages for declaring web APIs at the type-level
servant-docker
simple
simple-hpack
simple-library
tasty-discover
tasty-travis
unicode-syntax-exe
unicode-syntax-lib
yesod-hello-world
yesod-minimal
yesod-mongo
yesod-mysql
yesod-postgres
yesod-postgres-fay
yesod-simple
yesod-sqlite
PS C:\Users\erikn\stack>
```

### DIFF
--- a/src/Stack/New.hs
+++ b/src/Stack/New.hs
@@ -30,6 +30,7 @@ import qualified Data.ByteString.Lazy as LB
 import qualified Data.ByteString.Lazy.Char8 as L8
 import           Data.Conduit
 import           Data.Foldable (asum)
+import qualified Data.HashMap.Strict as HM
 import           Data.List
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as M
@@ -46,6 +47,7 @@ import qualified Data.Text.Lazy as LT
 import           Data.Time.Calendar
 import           Data.Time.Clock
 import           Data.Typeable
+import qualified Data.Yaml as Y
 import           Network.HTTP.Client.Conduit hiding (path)
 import           Network.HTTP.Download
 import           Network.HTTP.Types.Status
@@ -58,6 +60,7 @@ import           Stack.Types.TemplateName
 import           System.Process.Run
 import           Text.Hastache
 import           Text.Hastache.Context
+import           Text.Printf
 import           Text.ProjectTemplate
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mgsloan:
Thanks for working on this!  Nice output formatting!

The code looks reasonable, I have the following tweaks:

* For simplicity and consistency, please download the file the same way templates are downloaded.  See the code [here](https://github.com/commercialhaskell/stack/blob/5f7ba827e4e5cb8c50bc9e5322e660861a51f355/src/Stack/New.hs#L166).  True, this means that we'll be unnecessarily storing a file, but it's no big deal.

* Please use indentation style consistent with the surrounding code.  Stack isn't entirely consistent on spacing, but it'd be best to not mix 2-space / 5-space indents in with this file that uses 4-space indents.

Thanks!

narrative:
Hey I appreciate you looking at it.

Concerning your first tweak, the section of code you linked pertains to downloading or loading local copies of individual template files following failure.

Issuing `stack templates` calls `listTemplates`, which calls [`getTemplates`](https://github.com/commercialhaskell/stack/blob/5f7ba827e4e5cb8c50bc9e5322e660861a51f355/src/Stack/New.hs#L284), in which the downloaded list of all available templates is never saved locally so does not have a fallback. This is what I based my download function on.

Am I understanding correctly that you'd like both `getTemplates` and `getTemplatesInfo` to be updated to follow a similar convention as `loadTemplate`?

I do agree that it would be a good idea considering that currently `stack templates` completely fails without an internet connection, but just wanted to clarify. Perhaps the list of all templates not being saved locally was an oversight?

mgsloan:
Ah, I see!  How about this:

1) Remove `getTemplates` in favor of `getTemplatesInfo`

2) Have stack-templates test check that the templates metadata matches the repo contents

narrative:
I do think the current solution of parsing the GitHub API response is rather neat, but including a test to ensure nothing is missed shouldn't be very much trouble.

Perhaps this is a discussion more appropriately carried out on the stack templates repository, but if that is the route you'd like to take with this, I might also recommend the addition of some documentation which makes mention of the new requirement.

I'll see what I can do.

mgsloan:
True, the approach is clever.  I like having the test fail when the metadata file isn't updated, though.  It's a good reminder to add a description.

mgsloan:
On revisiting this, I don't think the download method needs to be improved.

Thanks for implementing this!

Issue #1396, opened 2015-11-21
A command `stack docs` that opens the locally installed documentation of packages?


sjakobi:
For example, from inside the current stack project:

* `stack docs Cabal` opens
  `~/.stack/snapshots/x86_64-linux/lts-3.14/7.10.2/doc/Cabal-1.22.4.0/index.html`
* `stack --stack-yaml stack-7.8.yaml docs Cabal` opens
  `~/.stack/snapshots/x86_64-linux/lts-2.22/7.8.4/doc/Cabal-1.18.1.5/index.html`

Given that stack already installs the documentation for ghc and haddock together with the binaries, `stack docs` could also open these. There needs to be a way to distinguish between the library documentation for haddock and the user guide though.

sjakobi:
On second thought, maybe `stack haddock --open [PACKAGE]` would be enough.

Without the `[PACKAGE]` argument I'd like the following behaviour:
* When inside a project, open `project/.stack-work/install/<platform/<snapshot>/<ghc-version>/doc/all/index.html`.
* Else, open `~/.stack/snapshots/<platform>/<global snapshot>/<ghc-version>/doc/index.html`.

mgsloan:
Sounds like a good idea to me!  One tricky thing is what command to use for the browser.  I suppose this could be part of `.stack/config.yaml`, and have it auto-try things like `sensible-browser` and `xdg-open`.

How about a `--url` option which makes it just put the URL to stdout instead of opening a browser.

> On second thought, maybe `stack haddock --open [PACKAGE]` would be enough.

`stack haddock` is a synonym for `stack build --haddock`, so I'd prefer a new `stack docs` command.

borsboom:
Sounds good to me as well.  A pull request would be welcome.

borsboom:
One problem with `stack docs` is that it conflicts with the `stack doc` command suggested in #1395.  Making `--open` a build option, as originally suggested, might not be so bad; then running `stack haddock --open <PACKAGE>` would build the package's haddocks and then open them.

borsboom:
> have it auto-try things like sensible-browser and xdg-open.

If there isn't already a package on Hackage that provides cross-platform support for opening a browser, one should be created for this issue.

yamadapc:
> If there isn't already a package on Hackage that provides cross-platform support for opening a browser, one should be created for this issue.

[...]

PR #1548, opened 2015-12-20
Add initial prototype of `stack haddock --open`

yamadapc:
This is related to #1396. Definatelly not ready for merge; is this the right
direction? It'd close #1396.

Issue #1372, opened 2015-11-19
Where do the haddocks for non-snapshot packages go?


sjakobi:
Running from outside any project:

```
$ stack haddock discrete-space-map --resolver lts-2.22
Run from outside a project, using implicit global project config
Using resolver: lts-2.22 specified on command line
discrete-space-map-0.0.4: download
discrete-space-map-0.0.4: configure
discrete-space-map-0.0.4: build
discrete-space-map-0.0.4: haddock
discrete-space-map-0.0.4: copy/register
```

So where can I find the haddocks now?

Turns out they installed in

```
/home/simon/.stack/global/.stack-work/install/x86_64-linux/lts-2.22/7.8.4/doc/discrete-space-map-0.0.4/
```

Sadly they are unlinked and there is no index file at

```
/home/simon/.stack/global/.stack-work/install/x86_64-linux/lts-2.22/7.8.4/doc/index.html
```

borsboom:
Right, there's no index of all the packages in the databases; only an index
for packages that are actually part of the project or snapshot is
generated.  Adding an index of all packages in the databases is a good idea.

mpilgrem:
I am closing this issue given the passage of time and because `stack haddock acme-missiles` outside of a Stack project does report the location of a Haddock index.

PR #2015, opened 2016-04-09
Add `haddock --open` flag. #1396

sjakobi:


sjakobi:
The behaviour of the flag is pretty close to [this short description](https://github.com/commercialhaskell/stack/pull/1548#issuecomment-168855569) by @mgsloan.
The one difference is that when we are in the implicit global project, using the flag with two or more targets opens the haddock index for the global project's snapshot, because the global project doesn't have its own haddock index. As the flag uses the targets for the build command, `stack haddock --open` in the implicit global project results in the following error message:

```
~ $ stack haddock --open
Using resolver: ghc-7.10 from implicit global project's config file: /home/simon/.stack/global-project/stack.yaml
Error parsing targets: The specified targets matched no packages.
Perhaps you need to run 'stack init'?
```

The flag is not very smart about figuring out the right haddock location when called with a single target. For example, running `stack haddock --open base` inside the `stack` project opens `/home/simon/.stack/snapshots/x86_64-linux/lts-5.3/7.10.3/doc/base-4.8.2.0/index.html` on my computer. I can put in more work here to make sure that (transitive) dependencies of the current project are opened in the project haddocks, if someone thinks that would be desirable.

I was a bit surprised by the behaviour of the command with non-project, non-snapshot targets. For example, when I run `stack haddock --open semiring-simple` in the `stack` project, `semiring-simple`, which isn't included in LTS-5.3, is locally built and its haddocks are locally generated and opened. I'm wondering whether I would prefer an error message in that case, but I may just be overlooking the use case for this behaviour.

mgsloan:
When I run this I get 

> gvfs-open: /home/mgsloan/fpco/stack/.stack-work/install/x86_64-linux/lts-5.3/7.10.3/doc/"stack-1.0.5"/index.html: error opening location: Error when getting information for file '/home/mgsloan/fpco/stack/.stack-work/install/x86_64-linux/lts-5.3/7.10.3/doc/"stack-1.0.5"/index.html': No such file or directory

I think there is an errant use of `show`.  Also, I don't see why this should need any special code for the global project.  It may need to special case compiler-only resolvers, though?

sjakobi:
> I think there is an errant use of show.

You're right! I didn't notice because xdg-open strips out the quotes automatically.

> Also, I don't see why this should need any special code for the global project.

As I've noted above, the global project doesn't have it's own haddock index file. ([This issue](https://github.com/commercialhaskell/stack/issues/1372) is related). So where I would open `all/index.html` in a local project, I can only offer the snapshot index while in the global project.

Would you prefer a different behaviour for a call like `stack haddock --open base containers` in the global project?

_Aside_: Not being able to simply call `stack haddock --open` in the global project with the expectation to get some useful index is pretty shitty. The reason (as noted in my first comment) is that `stack build` needs some kind of target. Maybe we should try to add some logic to create a special case for `stack haddock --open` calls?

> It may need to special case compiler-only resolvers, though?

Good point, I hadn't thought about that. `extra-deps` are already counted as local packages though and therefore opened locally, so it seems we can get away without a special case here.

sjakobi:
> I think there is an errant use of show.

I've fixed this in combination with a rebase.

mgsloan:
What do you think of having `--open` take a package identifier argument (version optional).  This argument would implicitly get treated as a target, and we'd open the docs.

Then, we can also have `stack haddock --open-index`, which could work even when the build step has nothing to build.

One reason I like this is that latter is that later on we can further overload it with opening the docs for a particular module, like `stack haddock --open Control.Lens`.  That said, I already plan to overload the target syntax for `stack ghci` with something similar.  Might consider putting it under `stack ghci --module Control.Lens` or something, though.

sjakobi:
> What do you think of having --open take a package identifier argument (version optional). This argument would implicitly get treated as a target, and we'd open the docs.
> 
> Then, we can also have stack haddock --open-index, which could work even when the build step has nothing to build.
> 
> One reason I like this is that latter is that later on we can further overload it with opening the docs for a particular module, like stack haddock --open Control.Lens. That said, I already plan to overload the target syntax for stack ghci with something similar. Might consider putting it under stack ghci --module Control.Lens or something, though.

I think these are very good ideas. The problem is that I'm not sure when I'll get around to implementing them. I think this PR as it stands is already useful although it certainly has some warts.

Would you consider merging the PR as is and tracking these ideas for improvement in a separate issue?

mgsloan:
Sure.  I still don't think there should be logic for the global project, but I'll fix it.

PR #2016, opened 2016-04-09
stack path: Rename --ghc-paths to --programs-path, add --compiler-path #1546

sjakobi:


sjakobi:
There's another reference to `--ghc-paths` in the guide where I'm not quite sure how to update it because I don't have a Windows system:

> On Windows, you will also need to delete `stack path --ghc-paths`

mgsloan:
I think that comment should be updated with `--programs-path`.  I think this is the usecase for the existence of such a flag.

sjakobi:
> I think that comment should be updated with --programs-path (and an implementation of it).

Done!

mgsloan:
Thanks!!

mgsloan:
FYI, I've made a few updates that bikeshed this a bit, while it's easy to change: https://github.com/commercialhaskell/stack/commit/2f7755d3a24d2b94d5f715a6cc97f7b18982e76f

Issue #1794, opened 2016-02-15
Add an option to specify build plan download address.


TerrorJack:
Currently, stack downloads stackage build plans from s3.amazonaws.com. Unfortunately, in many areas in China, this domain is blocked due to the Great Firewall (national network censoring system).

So I'm proposing that stack add an option to specify build plan download address. This allows us to set up a stackage mirror along existing hackage mirrors in China, and Chinese Haskellers won't have to endure Internet connection issues in their workflow. Thank you!

mgsloan:
Yeah, this would be good. Looks like the following URLs contain `amazonaws.com`:

![2016-02-17-200800_983x54_scrot](https://cloud.githubusercontent.com/assets/22570/13133668/4950f0e2-d5b2-11e5-92c6-d3f93f4c8e21.png)

So I suppose the solution here would be to allow configuration of something other than `https://s3.amazonaws.com`.  Something like `metadata-mirror: ...` in yaml configuration.

Is github also affected?  We also download things from github.

TerrorJack:
Yes, github is also affected.

mgsloan:
I see.  I suppose we could have something like:

```yaml
stack-urls:
    snapshots: https://s3.amazonaws.com/haddock.stackage.org/snapshots.json
    index: https://s3.amazonaws.com/hackage.fpcomplete.com/00-index.tar.gz"
    # etc
```

Alternatively, we could make it so that you can download some set of files and use those locally (see also https://github.com/commercialhaskell/stack/issues/1594 ).

igrep:
I wonder whether [package-indices](http://docs.haskellstack.org/en/stable/yaml_configuration/#package-indices) and [latest-snapshot-url](http://docs.haskellstack.org/en/stable/yaml_configuration/#latest-snapshot-url) is not for you...?

I'm also very interested in this topic.

TerrorJack:
@igrep package-indices affects Hackage indices/packages only; latest-snapshot-url won't do much good, since it's only used when stack needs to find a latest LTS/nightly, and does not affect how stack fetches a build plan.

igrep:
To make it clear, it is [rawGithubUrl](https://github.com/commercialhaskell/stack/blob/f2d362e157d9a7d22182e339b636950c54679808/src/Stack/Constants.hs#L173-L187) that you want to make configurable, right?

[...]

PR #2024, opened 2016-04-11
[WIP] Add feature to configure build plan download URL.

igrep:
Fixes https://github.com/commercialhaskell/stack/issues/1794

# Left TODOs

- [ ] Test (Really necessary?)
- [ ] Rename/Restructure the new configuration item if you can find better ones.
    - It may not have to be split into `lts` and `nightly`.
- [ ] Documentation (I want to write after you approve).
- [ ] ChangeLog.

Feel free to advice me on the source code structure, the new configuration item, and any others!


mgsloan:
Since we may want to configure quite a few different URLs, why not just call this `StackURLs`?  The fields can specify that they are prefixes.  This would correspond to the configuration file option

```
urls:
    lts-prefix: ...
    nightly-prefix: ...
    ...
```

Also, I don't see actual usage of this, just types and commandline parsing.

igrep:
> Since we may want to configure quite a few different URLs, why not just call this `StackURL`s?

You

> Also, I don't see actual usage of this in the diff, just types and commandline parsing.

OMG! :scream: Seems I've deleted the change by mistake! Sorry, fix soon.

igrep:
> Since we may want to configure quite a few different URLs, why not just call this `StackURL`s?

You mean I should unify the configurations like `download-prefix` in `package-indices` and `latest-snapshot-url` into `StackURL`?
Sounds good, but I have to consider `stack.yaml`'s compatibility doing that.
Can I make a separate pull request to make a new configuration item to unify them?
I guess the change would be very big.

> Also, I don't see actual usage of this in the diff, just types and commandline parsing.

OMG! :scream: Seems I've deleted the change by mistake! Sorry, fix soon.

igrep:
Fixed and rebased!

mgsloan:
We can just deprecate the `latest-snapshot-url` field and add it to this urls list.  I think the indices urls can be left where they are

mgsloan:
Oops sorry for letting this languish.  LGTM, thanks a bunch!

TerrorJack:
Thank you for this awesome work!

igrep:
Oh my... :astonished: I was actually in progress.
I was planning to push https://github.com/commercialhaskell/stack/commit/5874b9ee660e6ff997e4a1b78522fb56376b5dc4 ...
But, Okay! Thanks!

mgsloan:
Oh, sorry, I was in PR cleanup mode

PR #2061, opened 2016-04-24
Speed up build plan construction

sjakobi:
Collecting the package versions in a lazy HashMap instead of a strict
Map results in a speedup of ~10% for no-op builds in a small project.


sjakobi:
I've finally figured out how to use this `runOnce` magic in order to load the package caches from inside `Stack.Build.ConstructPlan.addPackageDeps`, rendering the change here redundant.

I'll open another PR soon.

PR #2070, opened 2016-04-27
Cabal file Git SHA

snoyberg:
This patch turned out to be much larger than expected, and mostly for tedious reasons. I recommend reviewing one commit at a time, as two of these are very mechanical (changing constraints and adding fields to some data types).

I added support to stackage-curator recently to include hashes of .cabal files for newly generated snapshots. This patch will ensure that, when using Git as a package index and a snapshot with this hash information, we get the same version of the .cabal file as was used when generating the snapshot, avoiding the possibility of Hackage edits from corrupting the snapshot for end users.

I recommend holding off on merging this for a few days so it can be properly tested with upstream Stackage Nightly snapshots that actually have this .cabal hash information.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,9 @@ Behavior changes:
 
 Other enhancements:
 
+* Grab Cabal files via Git SHA to avoid regressions from Hackage revisions
+  [#2070](https://github.com/commercialhaskell/stack/pull/2070)
+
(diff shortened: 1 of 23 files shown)


mgsloan:
@snoyberg LGTM!  I pushed a fix to the build for 7.8

mgsloan:
LGTM!

Issue #1924, opened 2016-03-18
config: remove nix.enable heuristic


zimbatm:
I have a project which gets compiled by nix and Ubuntu users. Both set of users expect to be able to type `stack build` and have the right thing to happen.

My current stack.yaml:
```yaml
---
packages:
- .
- location:
    git: https://github.com/motus/haskell-zookeeper-client.git
    commit: e724636886e3590624f79e1267bfb4d8eab96f86
resolver: lts-3.20
nix:
  enable: false
  packages:
    - gcc
    - git
    - pkgconfig
    - zlib
    - zookeeper_mt
```

I am forced to set `nix.enable` to `false` for Ubuntu users. If I don't set to anything the current heuristic is to set it to true because of the nix.packages existing.

From the doc, and this is what I have witnessed too:

```yaml
nix:
  # true by default when the nix section is present. Set
  # it to `false` to disable using Nix.
  enable: true
```

On nix I was hoping to set `nix.enalbe` to `true` in `~/.stack/config.yaml` but the per-project config has precedence over the global config (which makes sense to me).

So currently there is no way to support my scenario. While I see that the heuristic is trying to be helpful I think that it's counter-productive and should be removed.

mgsloan:
Interesting!  It makes some good sense to me to have a mode where nix isn't strictly required.  Perhaps that mode would even be a good default.

Pinging @YPares

chris-martin:
I don't understand why Nix can be enabled/disabled from `stack.yaml` at all. It seems to me that it's a system setting (i.e. every time I invoke stack one of my machines I need to enable Nix, and pretty much everyone else in the world always needs to disable it).

chris-martin:
And I've had to figure this out the hard way... https://github.com/yesodweb/wai/pull/520

YPares:
Sorry for the late answer. @chris-martin Indeed when I included this heuristic I didn't think about your use case.
And actually no, nix is by no means a system setting: it's definitely a per-project setting, as the non-haskell dependencies will be specified on a per-project basis.

YPares:
@zimbatm "On nix I was hoping to set nix.enalbe to true in ~/.stack/config.yaml"

YPares:
@borsboom Do you think the heuristic should be removed then?

chris-martin:
Under what circumstances would I *not* want to enable Nix when building a Stack project on NixOS? I've encountered very few nontrivial projects that can build successfully without including a zlib Nix dependency.

[...]

PR #2080, opened 2016-04-29
Removing nix enabled by default when nix: section is present

YPares:
In order to solve https://github.com/commercialhaskell/stack/issues/1924, I propose to no longer enable nix by default when a nix: section is present in the configuration.

This allows package developers to propose a nix configuration (list of nix packages) for their packages without _forcing_ the use of nix. I.e. now, if the user has a `nix.enable` flag set to `false` in their `$HOME/.stack/config.yaml`, this setting will not be overriden by the presence of a nix package list.


mgsloan:
LGTM

PR #2087, opened 2016-05-02
Use Cabal-1.24

phadej:
https://github.com/fpco/stackage/issues/1405

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -12,6 +12,7 @@ nix:
   packages:
     - zlib
 extra-deps:
+- Cabal-1.24.0.0
 - hpack-0.14.0
 - path-io-1.1.0
 - th-lift-instances-0.1.7


mgsloan:
Certainly something we could do.  I don't think we have much reason to do this.  The library is mostly just used to parse cabal files, and AFAIK that hasn't changed much.

I think the main thing to consider is adding support for setup-deps.  This could help make our "explicit-setup-deps" hack less fiddly.

Since we're going to do a release soon, we'd rather hold off on doing this change now.

phadej:
@mgsloan Optionally we could just relax bounds, and have a Travis job to build with `1.24` but have binary releases with default `Cabal`

mgsloan:
Looks like the bounds have been relaxed.  Merging https://github.com/commercialhaskell/stack/pull/2322 will result in updating to 1.24

PR #2107, opened 2016-05-05
always use --with-fingerprint

dysinger:
Otherwise we are assuming the user has set up their gpg.conf like
everyone else [bad assumption]


mgsloan:
LGTM

PR #2109, opened 2016-05-06
docs: Add permalinks to headings

sjakobi:
(including subheadings)


borsboom:
Thanks, I didn't know about that option!

sjakobi:
> Thanks, I didn't know about that option!

I had to google it too!

BTW would you mind cherry-picking the commit into the `stable` branch, so it comes into effect there, too?

Issue #1748, opened 2016-02-05
Typo detection in more places in stack


mgsloan:
See https://github.com/commercialhaskell/stack/issues/158 , where it will suggest near matches of the user's input.  This only works for things like `stack build pkg-0.0...`.  It'd be great to have fuzzy suggestions for all of the following:

- [ ] Targets.  Ideally this would handle fuzzy suggestions of package names and component names.  Bonus would also be supporting fuzzy suggestions of dirs, but that's not a big deal.

- [ ] extra-deps.  See also https://github.com/commercialhaskell/stack/issues/1521

- [ ] filepaths?  For example `--stack-yaml`

- [ ] more?

Ideally, these suggestions would share the same system as shell completion (see https://github.com/commercialhaskell/stack/issues/1343 and https://github.com/commercialhaskell/stack/issues/832 .  This way we can get checking and suggestions from the same mechanism!

(but of course, PRs adding any of these will be accepted, no need to do it all)

mgsloan:
- Fields in .cabal files.
    - [ ] Most important is `build-depends` (see https://github.com/commercialhaskell/stack/pull/2111#issuecomment-219292476)
    - [ ] It would also be good to support other various strings like paths and modules.

mgsloan:
- [ ] Unknown packages in dependencies - use fuzzy matching with all known hackage packages.  See https://github.com/commercialhaskell/stack/issues/2713

PR #2111, opened 2016-05-06
Mention that package names are case-sensitive

ismail-s:
If you feel this is a nitpick, or that I've phrased it badly, then by all means do what you want. But I just ran into this issue and it took me a while to realise that this might be the issue.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -430,7 +430,7 @@ Completed all 2 actions.
 This output means that the text package was downloaded, configured, built, and
 locally installed. Once that was done, we moved on to building our local package
 (helloworld). At no point did we need to ask stack to build dependencies — it
-does so automatically.
+does so automatically. Also, note that package names are case-sensitive.
 
 ### extra-deps
 


sjakobi:
Well, that's a bit of information that is more in the scope of `Cabal` than in the scope of `stack`.

What we could possibly do is cite the relevant rules in the section for `stack new`:

> A package is identified by a globally-unique package name, which consists of one or more alphanumeric words separated by hyphens. To avoid ambiguity, each of these words should contain at least one letter. Chaos will result if two distinct packages with the same name are installed on the same system.

(From https://www.haskell.org/cabal/users-guide/developing-packages.html#developing-packages)

What do you think, @mgsloan?

But tell me how exactly you ran into this problem, @ismail-s! Ideally you'd get a suggestion for similar, existing packages when you pass a misspelled package name to `stack`.

ismail-s:
Well, I was  working through the cis194 Spring 2013 exercises, and the last exercise on monads requires the use of the MonadRandom module (note the uppercase chars). So I added `monadrandom` to the build-depends section of the cabal file, ran something like `stack build/test`, and got something like:

```
ismail_s:~/workspace/cis194-sols (master) $ stack build

      monadrandom: needed (-any), stack configuration has no specified version
    needed for package cis194-sols-0.1.0.0

--  While attempting to add dependency,
    Could not find package monadrandom in known packages
```

Eventually, once I thought to try typing `MonadRandom` instead, stack build worked successfully.

If you want more info on the environment I was running these commands in, or want to try it out yourself, then go to https://ide.c9.io/ismail_s/cis194-homework, you should be able to access it with your github account.

mgsloan:
@sjakobi Sure, seems reasonable!

Thanks for the feedback, but I'm not sure how helpful it will be to point it out in this portion of the documentation. Pretty much everything in haskell is case-sensitive.

ismail-s:
Fair enough then, you guys can reject this PR. Ideally, though, stack would suggest similarly named packages in the error message.

mgsloan:
Alright, thanks for the suggestion!  Good idea about giving fuzzy suggestions for build-depends, I've noted it down on https://github.com/commercialhaskell/stack/issues/1748#issuecomment-219322184

PR #2114, opened 2016-05-08
Fix typo: persistent -> persist

igrep:
According to [here](https://github.com/commercialhaskell/stack/blob/47759559222cc501b5e2bbd0080963cc3229f489/src/Stack/Types/Docker.hs#L119) and [here](https://github.com/commercialhaskell/stack/blob/47759559222cc501b5e2bbd0080963cc3229f489/src/Stack/Types/Docker.hs#L241).


mgsloan:
Thanks!

PR #2097, opened 2016-05-04
Make hpack default to only outputting warnings #2011

mgsloan:
Not yet complete, requires updating to hpack version with
https://github.com/sol/hpack/pull/95 merged

PR #2117, opened 2016-05-10
Support hpack-0.14.0

phadej:
https://github.com/fpco/stackage/issues/1428

### DIFF
--- a/stack.cabal
+++ b/stack.cabal
@@ -218,7 +218,7 @@ library
                    , hastache
                    , project-template >= 0.2
                    , zip-archive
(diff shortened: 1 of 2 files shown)


mgsloan:
I imagine this will require updating and merging these changes to use the `hpackResult` function instead of `hpackSilent` https://github.com/commercialhaskell/stack/pull/2097/commits/cf61e326285fbeabed305f3ed3263b478d0ce649

mgsloan:
Actually, considering that further, it isn't required since the change was just an addition.  Would be good to get that PR updated and merged though

PR #2118, opened 2016-05-10
nix: fix generation of compiler attribute

bennofs:
We only want to filter the version, not the whole name. Otherwise, we end
up with things like "haskellpackagesghc7103" which is of course nonsense.

Also, when the stack.yaml file only specifes a major ghc version (like 7.10), we
need to supply a minor version so that we can look up the correct compiler in
nixpkgs.


mgsloan:
Thanks for fixing it!  I guess everyone has been using the nix support along with snapshots

PR #2143, opened 2016-05-15
Docs: Cite package name rules in section for `stack new`

sjakobi:
[ci skip]

Background: #2111

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -86,9 +86,19 @@ through the most common stack commands.
 
 ### stack new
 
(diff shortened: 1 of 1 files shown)


mgsloan:
Can we not have the "Chaos will result if two distinct packages with the same name are installed on the same system."?  This isn't true for stack.  Potentially also true for cabal-nix

sjakobi:
> Can we not have the "Chaos will result if two distinct packages with the same name are installed on the same system."?

Fixed!

mgsloan:
Thanks!  Nice commit message :)

PR #2152, opened 2016-05-17
Fix pattern overlap warning with GHC-7.8 and GHC-7.10

sjakobi:
`show (ErrorCall x) = x`, so the error message is as before.


mgsloan:
Thanks!

PR #2165, opened 2016-05-18
Ability to provide args to haddock command

yogsototh:
This is my first PR to this repository, I tried to follow the code usages.

Tell me if something is surprising.

This simply add the "--haddock-arguments" options which takes a string of arguments.
So you could for example use your own CSS:

~~~
stack haddock --haddock-arguments "--css $PWD/my.css"
~~~

### DIFF
--- a/src/Stack/Build/Haddock.hs
+++ b/src/Stack/Build/Haddock.hs
@@ -129,6 +129,7 @@ generateLocalHaddockIndex envOverride wc bco localDumpPkgs locals = do
         "local packages"
         envOverride
         wc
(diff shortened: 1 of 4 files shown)


mgsloan:
LGTM, thanks!

mgsloan:
Circling back to this because I was looking at the code while considering https://github.com/commercialhaskell/stack/issues/2229 .  I noticed that these arguments do not get provided to the actual running of haddock by cabal.  Is that intended?

Not sure, I'm not actually all that familiar with our haddock stuff, that's more @borsboom 's domain.

PR #2191, opened 2016-05-25
stack.cabal: add a 'debug' flag

da-x:
This change enables hacking on Stack with a shorter development cycle, using:

```
    stack build --flag stack:debug
```

It cuts down the full build of Stack on my setup from 3.3 minutes to 1 minute, and I get a functioning exe that is a bit slower (and I'm not doing optimization work most of time, so it's cool).

### DIFF
--- a/stack.cabal
+++ b/stack.cabal
@@ -49,9 +49,17 @@ flag static
   -- Not intended for general use. Simply makes it easier to
   -- build a fully static binary on Linux platforms that enable it.
 
+flag debug
+  manual: True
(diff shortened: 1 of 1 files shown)


mgsloan:
Good idea, I tend to just use `stack build --fast`, though

da-x:
Uh, somehow I missed on `--fast` option existing.

Does having this flag have any value now?

mgsloan:
Probably not, thanks for opening a PR anyway though!

PR #2202, opened 2016-05-27
listInstalled: move side-effects away

da-x:
It's odd to create an empty directory just to list it, from a function that merely lists the installed programs. Better to have that directory created at the same place we may be writing to it.


mgsloan:
Good point, thanks!

Issue #2211, opened 2016-05-31
Remove rest of Binary-code.


phadej:
I guess we have to wait few releases, before old `binary`-based caches are upgraded? I see that update code path was left in. ping @mgsloan

mgsloan:
Yup, re-using old precompiled cache info is a big win, prevents a lot of unnecessary recompilation.

snoyberg:
We've moved away from these with the move to SQLite, closing

PR #2213, opened 2016-05-31
Remove binary tagged

phadej:
Resolve https://github.com/commercialhaskell/stack/issues/2211

I made this already, so can use it locally, and try out (I have my caches busted while transferring to GHC-8.0 anyway)


mgsloan:
I actually see no reason to rush taking out the old compatibility code.  Certainly one day!

Issue #1634, opened 2016-01-10
WISH: sandboxing and/or stashing of compilations when switching git branches


wolftune:
Okay, this may be an unrealistic dream but could also revolutionize work for Haskell overall (being a compiled language)…

When I change back and forth between two git branches that may be distinct enough to require a large recompile for the project, I wish Stack would keep the compiled stuff around from each branch and use it accordingly. Thus, each branch would only need to compile the updates that are since the last compile of that branch rather than run a large recompile every time I switch.

Ideally, this would recognize that some files are totally unchanged and have unchanged dependencies among branches and somehow know to use that same compilation even when distinct files are sandboxed for each branch. That may not be feasible even dream sense, but it would be the ideal.

Perhaps as a practical matter, a "save this branch compilation" function could be added so that even if the whole wish idea above is impossible or unrealistic to automate, I could decide myself that I want to effectively *stash* a current compilation of the project, move to another branch, do whatever there, including recompiling, and then be able to go back to the first branch and *stash pop* so-to-speak to get back the stashed compilation and not have to recompile.

Ideally, I could stash with a name, like using memory registers to have a few stashed compilations, but a single stash at first would still be a useful step.

In the end, some sort of this working could end a lot of excessive recompiling for all Haskell development…

mgsloan:
This is actually a quite realistic dream :) .  I was thinking about the same thing a couple days ago, infact, in terms of per-branch compilation results.

I'm temporarily dubbing this a "build cache", dunno if that name should stick.

One tricky thing is that we can't actually move `dist` dirs, and then use them in a different location, because the build depends on the absolute path.  Probably not the only thing, but `Paths_*` module provide absolute paths.  This isn't a big problem, just inconvenient.  It means that when we copy a dist dir (forking the build cache), the only way we can use it is by moving it back into the spot it came from.  This works out fine for the usecase you describe, but I'd also like to use multiple build caches at once.  This'd allow you to simultaneously build the project with different settings.  Unfortunately, this conflicts with being able to copy the build cache.

[...]

Issue #1262, opened 2015-11-01
Relocatable data-files: support


lukexi:
I have some packages that require some accompanying data to be placed alongside any executables that depend on them; in a few cases that's some closed-source DLLs, and in another it's some shader files. I could use [file-embed](https://hackage.haskell.org/package/file-embed) as a workaround for the shaders, but the DLLs aren't as straightforward.

`cabal-install` has the ability to specify where the `data-files:` of a package are placed via `--datadir` and `--datasubdir`, which I think would solve this nicely alongside the existing `--local-bin-dir` option?

(I haven't needed this until I switched to Stack, but I assume `cabal-install`'s `--datadir` flag works alongside the `Paths_foo` module system to enable use of `getDataFileName` regardless of installation destination, for the shader example case)

borsboom:
@lukexi, will #848 do what you need?  If not, is there a way that proposal could be extended to support what you need?

borsboom:
For lack of feedback, closing as duplicate of #848.

lukexi:
Whoops, sorry I missed this!

I *think* #848 will do the trick — I'll explain exactly what I'm hoping for just in case and you can close again if #848 subsumes it : )

Say I have a dependency tree like this:
```
foo
|
bar baz
|    |
\   /
 bat
```

`bat` defines an executable, and all of the projects have `data-files` in their `.cabal`.


When I run `stack install --prefix releasedir`, I want to have all of `foo`, `bar`, `baz` and `bat`'s `data-files` along with `bat`'s executable installed into `releasedir`, and for `Paths_Foo`, `Paths_Bar`, and `Paths_baz` to be configured so that `bat`'s executable can load data from them at runtime.

Hope that makes sense : )

borsboom:
Pretty sure #848 will cover this case.  I'll reference your case over there to make sure.

Issue #848, opened 2015-08-25
Add `--prefix` option


borsboom:
This new `stack build` option will add a post-build step that rebuilds the targets with a `cabal configure --prefix=` option in a separate `dist` directory.  This will allow installing packages that use `data-files` in a fixed location on the file system outside of any stack-controlled sandbox.

borsboom:
In order to support installation into directories where the user doesn't have write permission (e.g., `/usr/local`), should support a flag to specify a command to prefix the `Setup.hs copy` step with (e.g., `sudo`).  Before starting, check whether writing to the destination works, and if it does not, display a friendly error with a suggestion to use the prefixing option.

borsboom:
Make sure this covers the case outlined in https://github.com/commercialhaskell/stack/issues/1262#issuecomment-157862165

mgsloan:
Seems like this issue is cropping up for multiple people.  Bumping to P2

borsboom:
Important use case for supporting a `--prefix` option: with it, it would be feasible for Homebrew to build their packages using Stack (for those packages that have a `stack.yaml`), which would aid reproducibility.  See https://github.com/Homebrew/homebrew-core/issues/1480 and https://github.com/Homebrew/legacy-homebrew/pull/49158 for related discussion.

borsboom:
Here's a more concrete example for Homebrew: https://github.com/Homebrew/homebrew-core/pull/1630

Blaisorblade:
Further rationale—sorry if already mentioned in some links: clearing snapshots (to recover space) leaves compiled binaries around, but not their data-files.

borsboom:
Many packages have `data-files` that aren't actually required for the package to work (like READMEs and Changelogs), and since this feature would require rebuilding those packages and all packages that depend on them, it should have a way to skip some or all dependencies.

mgsloan:
More demand for this feature, bumping priority.

jgm:
I don't understand why this hasn't been fixed by now.  It is a serious problem if stack can't be used to install data files to a location outside the stack work directory.  Cabal has been able to do this from the beginning.  Is there some serious conceptual problem with supporting it for stack?

mgsloan:
@jgm I think the reason this hasn't been resolved is that the data-files mechanism seems pretty clunky.  I've never used it, and I don't think most of the stack developers use it.

[...]

PR #2215, opened 2016-05-31
Make the paths stored in precompiled-cache Stack-root relative

da-x:
Because these paths always point inside the Stack root, better make them relative to the Stack root, making the Stack root more easily relocatable, in the future.

This change is backward-compatible.

### DIFF
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -66,6 +66,7 @@ import           Path
 import           Path.IO
 import           Stack.Constants
 import           Stack.Types
+import qualified System.FilePath as FilePath
 
 -- | Directory containing files to mark an executable as installed
 exeInstalledDir :: (MonadReader env m, HasEnvConfig env, MonadThrow m)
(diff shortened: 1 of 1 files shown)


mgsloan:
Makes sense.  Stack roots aren't going to be relocatable any time very soon though, unfortunately, cabal package build results have absolute paths.

da-x:
Apart from this, we have the Cabal conf files that you mentioned, the GHC pkgdb, some shell wrappers and RPATH in DSOs. All of these are easier and known formats that are easy to deal with. Only the precompiled-cache is very much Stack-specific.

BTW, I'm working on a 'stack relocate' command. I'm designing it to be used under `rpmbuild` mock chroot. @juhp

mgsloan:
That's awesome!  I'd love to be able to copy build directories for https://github.com/commercialhaskell/stack/issues/1634 , but I'm not sure how this would get around binary dependencies on the `Paths_*` module 

See these stack issues on relocatable data files:  https://github.com/commercialhaskell/stack/issues/1262 https://github.com/commercialhaskell/stack/issues/848

(compiled in absolute path dependencies on these data files is the issue)

da-x:
There's a partial working for binaries that assumes that the path you are relocating to is shorter or of the same length, and that there are no checksums or signatures. This is mostly the case in build servers, which tend to build under paths that quite long, e.g. `/build/rpmbuild/[sometimes-long-rpm-package-name-plus-version/build-root/...`, verifiably shorter than the usual install destination `/usr`. The trick is to assume null termination of the strings, pad the right side with zero characters, and do binary replacement inside the generated files. It's a bit nasty, though.

mgsloan:
Better hope some machine code doesn't magically resemble a filepath! :D

da-x:
The `package.cache` file proved stubborn with the zero padding, but I've realized that for the UNIX case I can pad the beginning of the path with `/`, since `/////usr` is as valid a path as `/usr`. So now `stack relocate` works ([implementation](https://github.com/commercialhaskell/stack/commit/de66db70857161c42f498b438d048630d6072dba)).

Issue #2224, opened 2016-06-01
Allow DownloadInfo be a file


cblp:
I'm trying to make fully offline environment for building a stack project.

I've managed to substitute all parts of stack, except GHC download source. It would be nice if it may be a local file instead of HTTP URL.

Bad solution: run a local web server.

Ugly solution: setup stack in /opt/stack, copy all bazillion files on other machine at /opt/stack, and use it as STACK_ROOT.

Good solution: copy ghc.tar.xz and give it to stack with stack-setup.yaml.

mgsloan:
> I've managed to substitute all parts of stack, except GHC download source.

What's that look like?  Ideally we'd have this in stack itself.  I would love to have a better offline mode: https://github.com/commercialhaskell/stack/issues/1594

> Good solution: copy ghc.tar.xz and give it to stack with stack-setup.yaml.

Definitely feasible.  It should be a straightforward change [here](https://github.com/commercialhaskell/stack/blob/7f7cead556f6bcb7ff3fe15af31e0f8c47149b71/src/Stack/Setup.hs#L631) which skips doing the download and instead provides the file directly to the install step.

PRs appreciated! :D

cblp:
If I understand right, #1594 is about _not_ building in case if connectivity is needed. My goal is to _do_ build even if there will be no connectivity.

> What's that look like?

It will be in https://github.com/cblp/stack-offline soon.

mgsloan:
The goal of #1594 is to have a flag to skip attempts at internet access.  It is true that most of the time when stack tries to download things, it needs to.  In the event that we add some stuff to check for a more recent stack version, then the offline flag will be more relevant.

We do want to have full support for offline use.  Would you be keen on merging this with stack, or deprecating it once stack has direct support?  That way more people will benefit rather than needing to know about `stack-offline`.

cblp:
My first milestone is a stand-alone utility, and the second milestone is to integrate it to the stack.

mgsloan:
Awesome, sounds good!  Thanks for working on this!

cblp:
#2232

Blaisorblade:
Should this be closed now that #2232 is merged, or do we still have sth. to track?

mgsloan:
I think it's done!

PR #2232, opened 2016-06-03
Install GHC from file

cblp:
Implements #2224.

Should I write tests?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,9 @@ Behavior changes:
 
 Other enhancements:
 
+* `setup-info` in stack.yaml and in stack-setup.yaml, also `--ghc-bindist` command-line option
+  now support absolute file path in `url` field.
+
 Bug fixes:
 
 ## 1.1.2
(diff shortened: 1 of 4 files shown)


mgsloan:
Thanks for working on this!

mgsloan:
Why not allow setting `content-length` and `sha1` for local files?  This way you could potentially test the stack.yaml configuration before uploading, and not need to test downloading.

Also, other places that use `chattyDownload` could benefit from this as well.  While it's an unlikely case, you could specify your own 7zip version for example.

I guess the resolution of this would be to have `chattyDownload` and `verifiedDownload` take a `Either (Path Abs File) DownloadRequest`, and return a `Path Abs File` for the local location of the file.

cblp:
> Why not allow setting `content-length` and `sha1` for local files? This way you could potentially test the stack.yaml configuration before uploading, and not need to test downloading.

OK, I'm removing warnings about them.

cblp:
I didn't change chattyDownload, because downloading to specified path is strange — it is actually copying and it's always better to use file from original location on disk than copying it to another directory. Or not always?

mgsloan:
Right, we'd want to call it something different.  I understand why you didn't make the change that way, it would be larger.

I don't think that this verification stuff should be coupled with downloading and should .  I can worry about that decoupling, though..

mgsloan:
Merged!  Sorry for letting this languish.

cblp:
Thanks!

Issue #1739, opened 2016-02-04
More concise, directed error messages when there are dependency errors


mgsloan:
See the beginning of the discussion of https://github.com/commercialhaskell/stack/issues/1735 .  Leaving this as a stub for now.

sjakobi:
For reference: There's some discussion in https://github.com/commercialhaskell/stack/issues/713 on this topic and https://github.com/commercialhaskell/stack/issues/1790 contains another example case.

mgsloan:
Done!

PR #2236, opened 2016-06-04
Pretty printer + improved plan construction errors

mgsloan:
Sometimes when you select the wrong version of a package that a lot of things depend on, you can get gobs and gobs of output.  For example, setting `containers-0.3.0.0` breaks pretty much every package in the build plan and you get hundreds of lines of output: https://gist.github.com/mgsloan/2a93bc5282f38d9b337e9571dbc8611a

I've omitted info about which specific inter-package dependencies have failed.  Instead, it tells you the root problem with color-coded info

2016-06-03-211018_472x97_scrot

The "needed due to" path gives the shortest path from a target to the package.  Hopefully it's clear that this is not necessarily the only path (in this case, just one of many)

### DIFF
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -29,15 +29,19 @@ import qualified Data.Map.Strict as Map
 import           Data.Maybe
 import           Data.Set (Set)
 import qualified Data.Set as Set
+import           Data.String (fromString)
 import           Data.Text (Text)
 import qualified Data.Text as T
 import           Data.Text.Encoding (decodeUtf8With)
 import           Data.Text.Encoding.Error (lenientDecode)
+import           Data.Typeable
 import qualified Distribution.Package as Cabal
+import qualified Distribution.Text as C
 import qualified Distribution.Version as Cabal
 import           GHC.Generics (Generic)
 import           Generics.Deriving.Monoid (memptydefault, mappenddefault)
 import           Network.HTTP.Client.Conduit (HasHttpManager)
+import           Path
 import           Prelude hiding (pi, writeFile)
 import           Stack.Build.Cache
 import           Stack.Build.Haddock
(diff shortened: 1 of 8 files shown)


src/Stack/Build/ConstructPlan.hs
```diff
@@ -800,3 +830,65 @@ pprintExceptions exceptions stackYaml =
...
+    chooseBest x y = if x > y then x else y
+    -- Extend a path to all its parents.
+    extendPath :: (PackageName, DepsPath) -> [(PackageName, DepsPath)]
+    extendPath (n, dp) =
+        case M.lookup n parentsMap of
+            Nothing -> []
+            Just (_, parents) -> map (\(pkgId, _) -> (packageIdentifierName pkgId, extendDepsPath pkgId dp)) parents
+
+data DepsPath = DepsPath Int Int [PackageIdentifier]
+    deriving (Eq, Ord, Show)
```
sjakobi: I think this type could be improved with record syntax and/or a comment.

sjakobi:
Awesome stuff! This will fix https://github.com/commercialhaskell/stack/issues/1739 too!

src/Stack/Build/ConstructPlan.hs
```diff
@@ -731,7 +749,8 @@ pprintExceptions exceptions stackYaml =
         vsep (map pprintExtra (Map.toList extras)) <>
         line <>
         line <>
-        "You may also want to try the 'stack solver' command" <> line <> line
+        "You may also want to try the 'stack solver' command" <> line <> line <>
+        fromString (show parentMap)
```
sjakobi: The debug output here should be removed eventually.

sjakobi:
Also, the line wrapping could be a bit cleaner IMO. Currently, readability at both 80 and 100 columns isn't that great. (Much better than before anyway, of course! :D)

80 columns:
pprint_80cols

100 columns:
pprint_100cols

I'd prefer the output to be optimized for 80 columns but clean output at 100 columns would make me happy, too! ;)

src/Stack/Build/ConstructPlan.hs
mgsloan: Good point! Done

Issue #1239, opened 2015-10-26
local packages with c-sources give runtime errors in stack ghci


vagifverdi:
if you point to a package source code located on the hard drive and that package has c source files, it will give runtime errors in stack ghci.

For example hdbc-odbc errors out in stack ghci when trying to connect to db. It gives following error:

/usr/local/lib/ghc-7.10.2/bin/ghc: symbol lookup error: /tmp/ghc25980_0/libghc_1.so: undefined symbol: SQLSetConnectAttr 

The function SQLSetConnectAttr is calling a c module hdbc-odbc-helper.c that is bundled with hdbc-odbc package.

Loading hdbc-odbc from hackage works fine in stack ghci.

Stack build also works fine.

borsboom:
@chrisdone, any ideas?

mgsloan:
I cannot reproduce this issue with another package which has C source files, `process`.  My steps:

1) Clone https://github.com/haskell/process
2) `autoreconf -i`
3) stack ghci

```
*Main> rawSystem "ls" []
aclocal.m4	cbits	      config.log     configure	   include  process.buildinfo  prologue.txt  Setup.hs	 System  tests
autom4te.cache	changelog.md  config.status  configure.ac  LICENSE  process.cabal      README.md     stack.yaml  test
ExitSuccess
```

@vagifverdi Can you please give a set of steps for reproducing the problem?

vagifverdi:
@mgsloan You need to create a separate project that refers to the cloned package.
Create a new folder test
In it add stack.yaml and refer to the package you cloned:

packages:
- ../process

Also, Not every call fails. Only the ones that call a code from the c files in the project.

mgsloan:
Even with process as a separate package, I don't get .  Note that being in a separate package really has no affect on GHCI, since it loads all the modules together anyway.

It'd be great to have a repro that's more minimal than a usage of hdbc.  Or if you could point me to your project (if it has instructions for setting it up).  As is, I keep on spending time attempting to construct a repro for this, and not getting anywhere.

Also, does this happen with the development version of stack (`stack upgrade --git` to install)?  There have been a number of fixes to GHCi.

vagifverdi:
i'll try to build a minimal example.

vagifverdi:
You are right. It is not related to c-sources generally. process package works for me too. But hdbc-odbc does not. It works fine if stack build or in cabal repl, but not in stack ghci

Here are the steps to reproduce:

[...]

PR #2242, opened 2016-06-04
Support .buildinfo files in stack ghci.

### LINKED ISSUE #2239: Honour .buildinfo files in stack ghci
A standard set of `Setup.hs` hooks is `autoconfUserHooks`. When using these hooks, or any other I suppose, the `Setup.hs` script is allowed to create a `.buildinfo` file, whose content will be merged with that of the `.cabal` file in any subsequent actions, as documented [here](https://www.haskell.org/cabal/users-guide/developing-packages.html#system-dependent-parameters).

Stack doesn't need to do anything special in the case of `stack build`, since building is outsourced to the Cabal library, which knows to source the `.buildinfo` file if present. But `stack ghci` doesn't use Cabal, so in that case we have to do it ourselves.

An example of package that does this: http://hackage.haskell.org/package/network. It's a package that can't currently be loaded into GHCi due to this issue. See also #1239.

cc @chrisdone.

mboes:
`Distribution.Simple` based `Setup.hs` sometimes create
`<package>.buildinfo` files as an artifact of the configure phase.
This is the case in particular when using the autoconf hooks. Since
this is an internal detail private to any given package's `Setup.hs`,
Stack shouldn't (and doesn't) need to care or be aware of this. But
`stack ghci` does, since in that use case we're not using Cabal to
build - we're taking care of the build ourselves. In particular,
ignoring the `<package.buildinfo` file can mean that a successful
build using `stack build` can't be replicated using `stack ghci`.

E.g. in the case of the `network` package, the set of C files to build
is decided at configure time, based on the current platform. So the
`c-files` stanza in the .cabal file is not complete. This in turn
means `stack ghci` isn't aware of all object files it has to link in,
leading to obscure link errors (see #1239).

See [1] for more on .buildinfo files.

Fixes #2239.

To test:

```
$ stack unpack network
$ cd network
$ stack ghci
```

Without this patch the above sequence of commands doesn't work.

[1]: https://www.haskell.org/cabal/users-guide/developing-packages.html#system-dependent-parameters

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -18,12 +18,12 @@ module Stack.Ghci
 
 import           Control.Applicative
 import           Control.Exception.Enclosed (tryAny)
-import           Control.Monad
+import           Control.Monad hiding (forM)
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
-import           Control.Monad.RWS.Strict
-import           Control.Monad.State.Strict
+import           Control.Monad.State.Strict (State, execState, get, modify)
+import           Control.Monad.Reader (MonadReader, asks)
 import           Control.Monad.Trans.Unlift (MonadBaseUnlift)
 import qualified Data.ByteString.Char8 as S8
 import           Data.Either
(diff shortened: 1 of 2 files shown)


mgsloan:
Cool, resolving this'd make a lot of people happy, I didn't realize it'd be so simple.

Are there any possible interactions between buildinfo files and Cabal versions?  Like if stack updates to building against Cabal-1.24, will it read buildinfo files from Cabal-1.18?

In other words, unlike cabal-install, the Cabal that stack is built against can vary with the one that's used for building.

mboes:
This patch doesn't depend on any new Cabal modules that Stack didn't depend on previously. The only new function it depends on is `readHookedInfo` and `updatePackageDescription`, both of which are part of Cabal-the-library (and I think mandated by Cabal-the-spec, though I haven't checked). The .buildinfo files are just text files that are meant to contain a subset of the syntax for .cabal files. It's up to the user to make sure (s)he writes both .cabal and .buildinfo files in a given project in a way that is compatible with the Cabal version that Stack uses.

Issue #2225, opened 2016-06-01
Can't stack exec a .bat file


ndmitchell:
Given a batch file `hello.bat` I can run it from the Windows command line with `hello`, `hello.bat`, `cmd /c hello` and `cmd /c hello.bat`. When running `stack exec -- ...` for any of those 4 variants, using `stack-1.1.2 i386`, I get one of:

```
C:\Neil\temp\stack-bat>stack exec -- cmd /c hello.bat
Run from outside a project, using implicit global project config
Using resolver: lts-2.14 from implicit global project's config file: C:\Users\ndmit_000\AppData\Roaming\stack\global\stack.yaml
C:\Users\ndmit_000\AppData\Local\Programs\stack\i386-windows\msys2-20150512\usr\bin\cmd: createProcess: invalid argument (Exec format error)
```

```
C:\Neil\temp\stack-bat>stack exec -- hello.bat
Run from outside a project, using implicit global project config
Using resolver: lts-2.14 from implicit global project's config file: C:\Users\ndmit_000\AppData\Roaming\stack\global\stack.yaml
Executable named hello.bat not found on path: ["C:\\Users\\ndmit_000\\AppData\\Roaming\\stack\\global\\.stack-work\\install\\1fc798c0\\bin","C:\\Users\\ndmit_000\\AppData\\Roaming\\stack\\snapshots\\0ac80986\\bin","C:\\Users\\ndmit_000\\App
... ]
```

The `.bat` suffix does not seem to make a difference.

mboes:
I get the same error, using

```
$ stack exec -- pacman -Sy
$ stack exec -- pacman -S autoconf
$ stack exec -- autoreconf -i
C:\Users\appveyor\AppData\Local\Programs\stack\i386-windows\msys2-20150512\usr\bin\autoreconf: createProcess: invalid argument (Exec format error)
```

mboes:
Note that `autoreconf` is a Perl script. I guess I could get around this by calling perl.exe and passing the full path to the `autoreconf` script to Perl, but that's pretty inconvenient. https://ghc.haskell.org/trac/ghc/ticket/365?cversion=0&cnum_hist=4#comment:7 suggests that rather than `createProcess`, `stack exec` should be using the `ShellExecute()` win32 call.

mboes:
cc @snoyberg in case this is actually a `process` package issue.

ndmitchell:
[...]

PR #2247, opened 2016-06-06
Support most executable extensions on Windows #2225

snoyberg:


snoyberg:
@ndmitchell just confirmed that this fixes the issue on his system, so this is likely safe to merge. It should have no impact on behavior on non-Windows systems.

mgsloan:
LGTM!

PR #2264, opened 2016-06-11
Add a config option to override gcc path

khanage:
This allows one to specify `with-gcc:` in stack.yaml.

Happy for any feedback - I just picked this up as it's marked with newcomer!

### DIFF
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -281,7 +281,7 @@ configOptsParser hide0 =
            <> hide
             )))
     <*> optionalFirst (textOption
-             ( long "with-ghc"
+             ( long "with-gcc"
             <> metavar "PATH-TO-GCC"
             <> help "Use gcc found at PATH-TO-GCC"
             <> hide
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -203,7 +203,7 @@ cleanOptsParser = CleanShallow <$> packages <|> doFullClean
 -- | Command-line arguments parser for configuration.
 configOptsParser :: GlobalOptsContext -> Parser ConfigMonoid
 configOptsParser hide0 =
-    (\stackRoot workDir buildOpts dockerOpts nixOpts systemGHC installGHC arch os ghcVariant jobs includes libs skipGHCCheck skipMsys localBin modifyCodePage allowDifferentUser -> mempty
+    (\stackRoot workDir buildOpts dockerOpts nixOpts systemGHC installGHC arch os ghcVariant jobs includes libs overrideGccPath skipGHCCheck skipMsys localBin modifyCodePage allowDifferentUser -> mempty
         { configMonoidStackRoot = stackRoot
         , configMonoidWorkDir = workDir
         , configMonoidBuildOpts = buildOpts
@@ -218,6 +218,7 @@ configOptsParser hide0 =
         , configMonoidJobs = jobs
         , configMonoidExtraIncludeDirs = includes
         , configMonoidExtraLibDirs = libs
+        , configMonoidOverrideGccPath = overrideGccPath
         , configMonoidSkipMsys = skipMsys
         , configMonoidLocalBinPath = localBin
         , configMonoidModifyCodePage = modifyCodePage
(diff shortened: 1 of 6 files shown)


src/Stack/Options.hs
```diff
@@ -279,6 +280,12 @@ configOptsParser hide0 =
            <> help "Extra directories to check for libraries"
            <> hide
             )))
+    <*> optionalFirst (textOption
+             ( long "with-ghc"
```
sjakobi: That should probably be `"with-gcc"` – not sure if there would be an advantage in using `configMonoidOverrideGccPathName` directly…

khanage: Not sure what you mean by `configMonoidOverrideGccPathName`?

sjakobi: In `Stack.Types.Config` you define `configMonoidOverrideGccPathName = "with-gcc"`, so you could potentially reuse that definition instead of duplicating the literal.

khanage: I just cargo culted that - I'm happy to make the change, although I think it should be done categorically, as it's "not the way it's currently done".

Thoughts?

sjakobi: > I think it should be done categorically, as it's "not the way it's currently done".

Good point. You can certainly just use a literal for this PR.

src/Stack/Types/Config.hs
```diff
@@ -840,6 +842,8 @@ data ConfigMonoid =
     -- ^ See: 'configExtraIncludeDirs'
     ,configMonoidExtraLibDirs        :: !(Set Text)
     -- ^ See: 'configExtraLibDirs'
+    , configMonoidOverrideGccPath    :: !(First Text)
```
sjakobi: Personally, I'd prefer a `First (Path Abs File)` here.

The main advantage I'd expect from that, would be better error messages than those that we currently relay from `cabal` or `ghc`.

I also think that this should be done "categorically", so for `configMonoidExtraIncludeDirs` and `configMonoidExtraLibDirs` too.

`Options.Applicative.Builder.eitherReader` would probably be a useful helper function for this.

sjakobi: I have opened https://github.com/commercialhaskell/stack/issues/2267 with regard to the overly permissive types.

Given that it's a rather tangential point, feel free to change `configMonoidOverrideGccPath` to `First (Path Abs File)` or not.

sjakobi:
Can you also add a paragraph to `doc/yaml_configuration.md` and an entry to the changelog?

I think this will be good to go then.

src/Stack/Types/Config.hs
khanage: I just had a quick play with this - there's a slight problem with there being no `FromJSON` instance for `Path Abs File` etc. I'd rather fix that as part of #2267 (as it is a little tangential). I'm happy to pick that one up as well once this is done though :).

sjakobi: I've already started working on #2267 so you may want to choose a different issue next. ;)

sjakobi:
LGTM!

Thanks for working on this, @khanage!

sjakobi:
I'm sorry to see this PR languishing for so long.

@mgsloan: Given that I've already reviewed this, are you ok with me simply merging this?

Alternatively, would anyone else be interested in giving this PR a second look?

borsboom:
I'm fine with this.  @sjakobi, go ahead and merge when ready.

simonjantsch:
Are dependencies also built with the specified gcc? I have one project where I need to build a dependency with a special gcc version, I just tried to add the 'with-gcc' option in the projects stack.yaml, it did not work though. The dependency does not have a stack.yaml at the moment, so I could not add the option there.

sjakobi:
@simonjantsch 

> Are dependencies also built with the specified gcc?

They certainly should. Did you make sure that the dependency was actually rebuilt? If so, please open an issue and include the output with `--verbose --cabal-verbose`, and ideally a project that we can reproduce the problem with, too.

simonjantsch:
@sjakobi I have opened an issue (see reference above)

PR #2268, opened 2016-06-14
Support -ferror-spans syntax in GHC error messages

danald:


sjakobi:
I have restarted the failing OS X build.

sjakobi:
> I have restarted the failing OS X build.

I've restarted the OS X job once again.

sjakobi:
@danald: I'm sorry that your PR has been languishing for so long. The core maintainers appear to be very busy right now.

I'm afraid I'm not very experienced with this part of the code and can't easily tell if it works or not.

Would you mind adding two short test cases for the two error formats in `src/test/Stack/Build/ExecuteSpec.hs`? A short comment on the formats in the code itself would be good too.

I'll simply merge the PR if the tests pass.

borsboom:
This looks good to me, so I agree with merging once the test cases are in place.

mgsloan:
Works for me, I think we can skip tests.  ExecuteSpec is essentially empty!  Might belong as an integration test if ghc actually gets invoked.

Issue #2259, opened 2016-06-08
Migrate --nix support to use haskell.compiler.ghcXYZ attributes instead of haskell.packages.lts_X_Y


peti:
The `haskell.packages.lts_X_Y` package sets will disappear from Nixpkgs in the foreseeable future and `stack` should  not rely on them any longer. Instead, please access the appropriate compiler using the `haskell.compiler.ghcXYZ` attributes instead. http://permalink.gmane.org/gmane.linux.distributions.nixos/20505 has more details about the motivation behind that upcoming change.

YPares:
Ok, so what would be the cleanest way to get the required GHC version from the LTS?

mgsloan:
I do not see the reason for this change, and it will break existing stack versions and build configurations.  This seems like a frustration with a variety of non-endemic issues.  Please consider this more carefully.
 
> 1. Stackage does not "maintain" any of its LTS releases. Rather, the
> Stackage volunteers compile a list of package versions, test and verify
> them to the best of their abilities, release that list, and then they
> never touch it again.

How about bringing up this issue on stackage, or integrating the nix feedback into the curation of stackage?

> 2. Following LTS strictly may deprive us of important security updates.

Ok, this seems to be more like "I don't want stable package sets".  It turns out that people do.

It sounds like you want the ability to retroactively patch a package set.  We tend not to allow this because it affects build determinism.  However, for careful patches I can see the value.  It's certainly something to consider adding to stackage, perhaps in the form of minor lts versions (`lts-5.1.1`)

> 3. Stackage Nightly is not a stable package set.

So we do want stable package sets!  :)

This seems like an impedence mismatch between how nightlies have been exposed to nix users.  If "haskellPackages" is always the latest nightly, then yes, you shouldn't do that if you want reliable builds.

If you want something stable, use lts.  If you want something unstable, use nightly.  It's that simple.

> Stackage releases come out only after a complete test build of all
> packages has succeeded. Unfortunately, those tests don't always catch
> all issues we might run into, because we compile packages in a different
> environment.

Ok, lets work with you guys to curate stack via feedback from nix.

IMHO, none of the reasons given justify this drastic action.  That said, I am not a nix user.

mgsloan:
> Ok, so what would be the cleanest way to get the required GHC version from the LTS?

The snapshot specifies a compiler version, so stack knows that once it's loaded.  Do you need to get the GHC version outside of stack itself?

peti:
[...]

Issue #2175, opened 2016-05-22
"Right Nothing" printed to the console during setup


silky:
with

````
>stack --version
Version 1.1.3, Git revision 8f7e62fbe3620a4b0ae5eb833cebfa50820f5050 x86_64 hpack-0.14.0
````

when i attempt a setup i see:
````
⚡ 05:37 PM noon ∈ ~>stack --resolver nightly-2016-05-13 setup
Run from outside a project, using implicit global project config
Using resolver: nightly-2016-05-13 specified on command line
Downloaded nightly-2016-05-13 build plan.    
Did not find .cabal file for servant-yaml-0.1.0.0 with Git SHA of d7c8a8eaf9e7c215433f8f39daac7fc6cf116325
Right Nothing
Did not find .cabal file for reducers-3.12.1 with Git SHA of bcebf11be1770ac480d9d0ee6238044f143a57f8
Right Nothing
Did not find .cabal file for math-functions-0.1.7.0 with Git SHA of 5870cda6259e8a838c8cb22043b9d3cbf55d9a8e
Right Nothing
Did not find .cabal file for hw-prim-0.0.0.10 with Git SHA of 3658e799df0b7fceb045899b7a3d686b960b5d01
Right Nothing
Did not find .cabal file for hspec-megaparsec-0.1.1 with Git SHA of 116c887f005a6e599dc9b6ce5627f1b69782f44b
Right Nothing
Did not find .cabal file for dixi-0.6.9.1 with Git SHA of 17cba908e23588d9e5b67541f53afd8590c5c30f
Right Nothing
Did not find .cabal file for cron-0.4.0 with Git SHA of d485877cb0e43ee1b8fcdaa0a7224acbd753186c
Right Nothing
stack will use a locally installed GHC
For more information on paths, see 'stack path' and 'stack exec env'
To use this GHC and packages outside of a project, consider using:
stack ghc, stack ghci, stack runghc, or stack exec
````

it's mildly unfortunate that `Right Nothing` is printed here.

introduced a few weeks ago by https://github.com/commercialhaskell/stack/commit/f483e820d23707ec4ddd3fcdf38385e36f769203#diff-2068556e565acc62de252ac25ed550c4R288

mgsloan:
Good point, fixed!

mgsloan:
Then again, perhaps the root issue is not addressed - why are these SHAs missing?

silky:
i'm not too sure about that one.

let me know if you want more information from me.

[...]

PR #2270, opened 2016-06-15
Nix-shell no longer called with a GHC from an LTS mirror

YPares:
As explained in https://github.com/commercialhaskell/stack/issues/2259
we will soon no longer be able to give to nix-shell indentifiers like `haskell.packages.lts-5_14.ghc`
since LTS mirroring will be removed from the Nixpkgs.
This PR proposes to retrieve the wanted GHC version before creating the nix shell, which will now
always be called with derivations like `haskell.compiler.ghc7103`.
Until the lts mirroring is removed from Nix, this also provides the advantage that now you don't have to pay attention to which LTS you use into your `stack.yaml`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,6 +18,9 @@ Other enhancements:
 
 * Add the ability to explictly specify a gcc executable.
   [#593](https://github.com/commercialhaskell/stack/issues/593)
+* Nix: No longer uses LTS mirroring in nixpkgs. Gives to nix-shell a derivation
+  like `haskell.compiler.ghc801`
+  See [#2259](https://github.com/commercialhaskell/stack/issues/2259)
 
 Bug fixes:
 
--- a/doc/nix_integration.md
+++ b/doc/nix_integration.md
@@ -40,37 +40,36 @@ a version of GHC matching the configured resolver. Enabling Nix
 support means packages will always be built using a GHC available
 inside the shell, rather than your globally installed one if any.
 
-Note that in this mode `stack` can use only those resolvers that have
-already been mirrored into the Nix package repository. The
-[Nixpkgs master branch](https://github.com/NixOS/nixpkgs/tree/master/pkgs/development/haskell-modules)
-usually picks up new resolvers such as Stackage nightlies and LTS
-versions within two or three days. Then it takes another two or three
+Note that in this mode `stack` can use only GHC versions than have
+already been mirrored into the Nix package repository.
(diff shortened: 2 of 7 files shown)


YPares:
Ok, actually it did work because I already had in my `$HOME/.stack` the versions of GHC I tested, but EnvConfig apparently cannot be loaded before a compiler can be found on the system.
(So I need a compiler to set up a compiler...)

YPares:
It's fixed now. I've tried it with several resolvers, both lts and fixed ghc versions.

YPares:
The CI error on OSX is rather weird, I'm getting right from the beginning of a fresh build:
```
Did not find .cabal file for yesod-core-1.4.20.2 with Git SHA of 5350ec333aa9ed6ae24e46142e33d2f851ace553
Right Nothing
Did not find .cabal file for wizards-1.0.2 with Git SHA of 525156becfe006b51f4e46b5e60c1bfbe3237336
Right Nothing
Did not find .cabal file for websockets-0.9.6.1 with Git SHA of 2f90616ee89447c142e535fecb5bdd407c14fa9a
Right Nothing
Did not find .cabal file for waitra-0.0.4.0 with Git SHA of 396179446c25a5f9e5ad931f09414f59e0bb4467
Right Nothing
Did not find .cabal file for wai-middleware-static-0.8.0 with Git SHA of 818e8bc0b3ebde0723a750aafb418e413437e78f
Right Nothing
Did not find .cabal file for uuid-1.3.12 with Git SHA of 79e931a867a5e3dd960b7e29426328fa73d69e00
Right Nothing
Did not find .cabal file for unix-compat-0.4.1.4 with Git SHA of e007b19ea18bb290e961f75812befc39dd1c3f49
Right Nothing
Did not find .cabal file for union-find-0.2 with Git SHA of d3241b27b99cf737fbad58f5cb9a4a3a27021ba0
Right Nothing
Did not find .cabal file for twitter-types-0.7.2.1 with Git SHA of 3be5081ac2a883993b7483ca9e44242102a36838
Right Nothing
```

However the build succeeds and `stack test` too. However here, the build of my three commits fails with the same "ghc not found in path" error.
Should I disregard this CI error?

mgsloan:
I think I've seen the "ghc not found in path" error before.  Some travis nondeterminism of some variety :/

The git sha thing isn't cause for concern, and is tracked by https://github.com/commercialhaskell/stack/issues/2175

mboes:
@mgsloan anything holding back the merge of this PR? Now that LTS snapshots have been removed from Nixpkgs, Stack --nix users are pretty stuck without this PR.

mboes:
@YPares branch force pushed to resolve conflicts.

PR #2271, opened 2016-06-15
docs: Add cabal build with ghc-8.0.1 to example .travis.yml

sjakobi:


sjakobi:
Ah, I forgot this should go into `stable`. @borsboom: I'll push it right in.

PR #2273, opened 2016-06-15
Run only 5 Travis jobs so they can all run in parallel

sjakobi:
…and finish more quickly

I deleted the jobs with ghc-7.10.1 and ghc-7.10.2 under the assumption that these are less important now that ghc-7.10.3 is out and that compatibility with them can be more or less concluded from the ghc-7.10.3 jobs.

Alternatively the cabal ghc-7.10.3 job could instead run with ghc-7.10.1.

The ghc-head/cabal-head jobs are IMO rather useless as they hardly ever succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,10 @@ cache:
 
 matrix:
   include:
+    - env: BUILD=cabal CABALVER=1.18 GHCVER=7.8.4
+      compiler: ": #GHC 7.8.4"
+      addons: {apt: {packages: [cabal-install-1.18,ghc-7.8.4], sources: [hvr-ghc]}}
+
     - env: BUILD=cabal CABALVER=1.22 GHCVER=7.10.3
       compiler: ": #GHC 7.10.3"
       addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.3], sources: [hvr-ghc]}}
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Yeah on all counts:
👍  on dropping support for 7.10.1, 7.10.2.
👍  on dropping 7.8.4-Cabal; among other reasons, https://github.com/commercialhaskell/stack/pull/2392 shows it is *already silently broken* (I'd argue), it only builds because of the build cache.
👍  on dropping HEAD, unless somebody tries to fix it—but the (few) failures I've looked at are on the ecosystem side.

borsboom:
I'm fine with dropping ghc-7.10.{1,2} and HEAD.  I'd like to keep ghc-7.8.4 with cabal-install, though, because we should make sure that case continues to work.  I wonder if we should get it to nuke the build cache periodically (every 24 hours?) in order to catch problems like the one currently masked.

sjakobi:
I have added the ghc-7.8.4 cabal job back.

> I wonder if we should get it to nuke the build cache periodically (every 24 hours?) in order to catch problems like the one currently masked.

That sounds like a good idea! Maybe we could have a cron job that runs `true | travis cache --delete`  or so. Here are the relevant docs: https://docs.travis-ci.com/user/caching/#Clearing-Caches

mgsloan:
LGTM!

Issue #2267, opened 2016-06-13
Use less permissive types in the config monoids


sjakobi:
`ConfigMonoid` (and probably other config monoids too) contains a few fields that have more permissive types than necessary. This has the effect that bad inputs go undetected or result in errors that are reported out of context or only after the inputs have been passed to ghc.

A few of the problematic fields:
* `configMonoidWorkDir`
* `configMonoidOs`
* `configMonoidExtraIncludeDirs`

Example interaction:
```
$ stack --extra-lib-dirs --oops build
these-0.7: unregistering (local file changes: Control/Monad/Chronicle.hs Control/Monad/Chronicle/Class.hs Control/Monad/Trans/Chronicle.hs Data...)
these-0.7: configure
Configuring these-0.7...
Warning: 'extra-lib-dirs: --oops' directory does not exist.
...
```

Compare this to the error message for an argument that must be a `Path Abs Dir`:
```
$ stack --stack-root --oops path
option --stack-root: Failed to parse absolute path to directory: '--oops'

Usage: stack [--help] [--version] [--numeric-version] [--hpack-numeric-version]
...
```

chrisdone:
Agreed, do you want to submit a PR for this?

sjakobi:
> Agreed, do you want to submit a PR for this?

Yes I've already started to work on this.

One thing that would be useful for this, are `FromJSON` instances for the four `Path` types. I assume you wouldn't want these in `path` itself.
Should I make a `path-aeson-instances` or even a `path-instances` package?

chrisdone:
I suppose I wouldn't mind including instances for `FromJSON` in `path` itself. It should have four instances, one for each possible path combination.

sjakobi:
> I suppose I wouldn't mind including instances for FromJSON in path itself.

Great! I'll make a PR.

chrisdone:
Here are some necessary ingredients to achieve what you'd like for e.g. `extra-lib-dirs`:

* (The option parser for extra lib dirs)[https://github.com/fpco/stack/blob/master/src/Stack/Options.hs#L276..L281]
* (How to define an optparse-applicative parser for a custom type)[https://github.com/commercialhaskell/stack/blob/master/src/Stack/Types/PackageName.hs#L153..L169]
* You already know where the config types are. 👍 

It may eventually be a nice idea to include such an optparse combinator in the path package itself eventually, but for now it can be put in `Options.hs`.

Do you have everything you need? ☝️ 😄

[...]

PR #2294, opened 2016-06-22
Use less permissive types in the config monoids (#2267)

sjakobi:
For a review it's probably best to look at each commit in sequence.

See #2267 for the original issue.


sjakobi:
Because I would prefer not to rebase this PR again and because apart from https://github.com/commercialhaskell/stack/pull/2294/commits/72e0fa32ae2cbddf0ffe974dcc580bbd3a915eb3 I'm pretty sure there's no change in behaviour*, I'll merge this myself if no one feels like reviewing this.

*: Of course, some bad inputs are caught earlier now!

PR #2295, opened 2016-06-22
Detect resolver change in `stack solver`

### LINKED ISSUE #2252: stack solver --update-config does not update the config if the resolver changes
I tried to use
```
stack solver --update-config --resolver lts-6
```
to change my `stack.yaml` from a different resolver. This works if `stack solver` has to apply other changes (`extra-deps` for example). But if the yaml requires no change but the `resolver` field, then `stack solver` says:
```
test -z "$STACK" || echo "resolver: lts-1.0" > stack.yaml # dummy value
test -z "$STACK" || stack solver --update-config --resolver "$STACK"
Selected resolver: lts-3.22
Using configuration file: stack.yaml
Using cabal packages:
- arbtt.cabal

No changes needed to stack.yaml
```

sjakobi:
Fixes #2252.

@harendra-kumar: Could you take a look?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,6 +18,8 @@ Bug fixes:
 
 * Support most executable extensions on Windows
   [#2225](https://github.com/commercialhaskell/stack/issues/2225)
+* Detect resolver change in `stack solver`
+  [#2252](https://github.com/commercialhaskell/stack/issues/2252)
 
 ## 1.1.2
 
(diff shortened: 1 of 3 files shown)


harendra-kumar:
Awesome! Looks good to me.

harendra-kumar:
I renamed moldResolver to mOldResolver (94c2cb2). without camel case it sounded like `mold resolver` rather than `m old resolver`.

sjakobi:
> I renamed moldResolver to mOldResolver (94c2cb2). without camel case it sounded like mold resolver rather than m old resolver.

Haha, yeah, that's better! :)

Thanks for the review!

mgsloan:
I actually have used the `mold` prefix a lot, I don't mind either way, though.

Issue #2289, opened 2016-06-20
GHC-Options flags being ignored in newer Stack builds


AaronFriel:
Arch Linux and other rolling distro users are experiencing errors due to the assembler using an unhandled relocation type.

See: 
https://github.com/DanielG/ghc-mod/issues/762
https://github.com/gibiansky/IHaskell/issues/636

One trend with these reports is that some users report success from putting in their `config.yaml` or `stack.yaml` the lines:

```
ghc-options:
  cryptonite: -opta-Wa,-mrelax-relocations=no
```

Or variations thereof (e.g.: `"*": ...`).

I'm experiencing this as well, even when I try to ghc-options all the things:

```
rebuild-ghc-options: true
apply-ghc-options: everything
ghc-options:
  cryptonite: -opta-Wa,-mrelax-relocations=no
  cipher-aes: -opta-Wa,-mrelax-relocations=no
  "*": -opta-Wa,-mrelax-relocations=no
```

Even when I do that, `stack build cipher-aes` won't rebuild unless I use ghc-pkg to remove it (what?!) and even when I do, I still get these errors when using ghc-mod:

```
ghc-mod: /home/developer/.stack/snapshots/x86_64-linux/lts-6.4/7.10.3/lib/x86_64-linux-ghc-7.10.3/cipher-aes-0.2.11-LCbQiUgBdfG1swKGFESvIB/libHScipher-aes-0.2.11-LCbQiUgBdfG1swKGFESvIB.a: unhandled ELF relocation(RelA) type 42
```

An interesting observation is that *some users* report that the ghc-options command worked for them, specifically, @shimuuar noted:

> stack-1.0.4.3, no ghc-options: unhandled ELF relocation
> stack-1.0.4.3 cryptonite: -opta-Wa,-mrelax-relocations=no: works
> stack-1.1.2 with or without ghc-options: unhandled ELF relocation

Since it looks like a GHC version that fixes this won't be released soon, I'm wondering if it might be possible to see why Stack 1.1.2 is not working with these options?

AaronFriel:
Edit: I have reproduced this, it took some time (and an intervening episode of a popular television series), but using Stack 1.0.4.3 and the config.yaml I showed above, I am able to compile and run ghc-mod against my source.

mgsloan:
[...]

Issue #2296, opened 2016-06-22
Stack unregisters and rebuilds all the extra-deps with each new (dev) stack version


sjakobi:
While developing on `stack`, I see the following whenever I build a new executable:

```
~/s/stack $ stack install --fast --ghc-options -j
stack-1.1.3: unregistering
store-0.1.0.1: unregistering (missing dependencies: th-lift-instances, th-orphans, th-utilities)
th-lift-instances-0.1.7: unregistering (old configure information not found)
th-orphans-0.13.1: unregistering (missing dependencies: th-lift-instances)
th-utilities-0.1.1.0: unregistering (missing dependencies: th-orphans)
th-lift-instances-0.1.7: configure
th-lift-instances-0.1.7: build
path-0.5.8: configure
path-0.5.8: build
th-lift-instances-0.1.7: copy/register
th-orphans-0.13.1: configure
path-0.5.8: copy/register
th-orphans-0.13.1: build
path-io-1.1.0: configure
path-io-1.1.0: build
path-io-1.1.0: copy/register
th-orphans-0.13.1: copy/register
th-utilities-0.1.1.0: configure
th-utilities-0.1.1.0: build
th-utilities-0.1.1.0: copy/register
store-0.1.0.1: configure
store-0.1.0.1: build
```

I think this is due to some versioning feature in `store`. It's really annoying though.

mgsloan:
Agreed, this is really annoying, I was aware of a more restricted version of this with ghci - https://github.com/fpco/store/issues/53 .  A simple fix is possible, but it would just be a patch to something that's a bit iffy anyway (good for original app, not so good for stack).  A more comprehensive fix is in progress.

I didn't realize it happened with each new dev version.  I thought it was just when dependencies involved in the type info change, or if you use ghci to load stack.

sjakobi:
Hypothesis from #2353:

> It seems like perhaps the git hash for the stack binary is causing such invalidation.

mgsloan:
I've got a branch.  Pending release of store / etc.

sjakobi:
Looks like this was fixed in #2425 and can be closed!?

Awesome work, @mgsloan! :clap:

mgsloan:
Yup!  Thanks!

PR #2305, opened 2016-06-25
Fix issue #2289, apply ghc-options to snapshot pkg

AaronFriel:
Commit c891a24 created a regression, ghc-options were no longer applied to 
packages in the snapshot. The "mini build plan" for missing snapshot packages
did not check the configuration's "ghc-options" settings at all, so even
configurations with "apply-ghc-options: everything" would not apply those
options.

This commit, combined with a change to config.yaml, fixes these issues:

https://github.com/DanielG/ghc-mod/issues/762
https://github.com/gibiansky/IHaskell/issues/636

To fix those issues, `config.yaml` (usually located in `~/.stack`) should
contain:

```
apply-ghc-options: everything
ghc-options:
  "*": -opta-Wa,-mrelax-relocations=no
```

And for snapshots already installed, the following command will force stack
to rebuild them with those options:

```
stack exec -- ghc-pkg unregister $PACKAGENAME --force
```

*Not tested:* Alternatively, add the line `rebuild-ghc-options: true` to
`config.yaml`. This may cause undesirable behavior in forcing all ghc-options
changes to rebuild the snapshot.

### DIFF
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -147,8 +147,9 @@ loadSourceMap needTargets boptsCli = do
                 let p = lpPackage lp
                  in (packageName p, PSLocal lp)
             , extraDeps3
-            , flip fmap (mbpPackages mbp) $ \mpi ->
-                PSUpstream (mpiVersion mpi) Snap (mpiFlags mpi) (mpiGhcOptions mpi) (mpiGitSHA1 mpi)
+            , flip Map.mapWithKey (mbpPackages mbp) $ \n mpi ->
+                let configOpts = getGhcOptions bconfig boptsCli n False False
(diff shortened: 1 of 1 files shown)


AaronFriel:
Pinging @mgsloan

mgsloan:
Thanks a bunch for figuring this out and opening a PR!!

CthulhuDen:
Is there any information on when this mix might land in releases?

sjakobi:
> Is there any information on when this fix might land in release?

I agree that this an important bugfix that should be released as soon as possible.

There's another open bug that will have to be fixed before a release is possible though: https://github.com/commercialhaskell/stack/issues/2296

vaibhavsagar:
New release soon, now that the other issue has been closed?

PR #2318, opened 2016-07-01
Allow global Stack arguments with a script

### LINKED ISSUE #2316: Nix and shebangs don't work together
Given a script of:

```haskell
#!/usr/bin/env stack
-- stack --nix --resolver lts-5.9 runghc

main = return ()
```

We get:

```
File does not exist or is not a regular file `--extra-lib-dirs=/nix/store/n962p1idgzh07wvhxam5r6icv894kj61-ghc-7.10.3/lib'
```

puffnfresh:
Stack has support for running Haskell scripts directly, when you pass a
filename as an argument.

Previously this worked by having a fallback: if any argument couldn't be
parsed as a valid option, check if the first argument was a file.

Sadly this broke for cases such as:

    stack --nix example.hs

And this means something different:

    stack example.hs --nix

Now Stack will split the arguments into two:

1. All arguments until the first existing file
2. The first existing file and all arguments afterwards

So the above example gets parsed as:

1. `["--nix"]`
2. `["example.hs"]`

The first section will be reparsed as options. The second section will
be used like it was previously.

This fixes the specific case of running interpreter scripts when having
the Nix option enabled globally. Stack would correctly start a
nix-shell. A nix-shell respawns Stack with some extra global
arguments (extra-lib-dirs and extra-include-dirs) but Stack would treat
them as file arguments, because just one of the arguments (the script)
was a file.

---

Fixes #2316

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -498,18 +498,32 @@ interpreterHandler
   -> ParserFailure ParserHelp
   -> IO (GlobalOptsMonoid, (GlobalOpts -> IO (), t))
 interpreterHandler args f = do
-  isFile <- D.doesFileExist file
-  if isFile
-  then runInterpreterCommand file
-  else parseResultHandler (errorCombine (noSuchFile file))
+  -- args can include top-level config such as --extra-lib-dirs=... (set by
+  -- nix-shell) - we need to find the first argument which is a file, everything
+  -- afterwards is an argument to the script, everything before is an argument
+  -- to Stack
+  (stackArgs, fileArgs) <- spanM (fmap not . D.doesFileExist) args
+  case fileArgs of
+    (file:fileArgs') -> runInterpreterCommand file stackArgs fileArgs'
+    [] -> parseResultHandler (errorCombine (noSuchFile firstArg))
   where
-    file = head args
+    firstArg = head args
 
-    -- if the filename contains a path separator then we know that it is not a
-    -- command it is a file to be interpreted. In that case we only show the
+    spanM _ [] = return ([], [])
+    spanM p xs@(x:xs') = do
+      r <- p x
+      if r
+      then do
+        (ys, zs) <- spanM p xs'
+        return (x:ys, zs)
+      else
+        return ([], xs)
+
+    -- if the first argument contains a path separator then it might be a file,
+    -- or a Stack option referencing a file. In that case we only show the
     -- interpreter error message and exclude the command related error messages.
     errorCombine =
-      if elem pathSeparator file
+      if elem pathSeparator firstArg
       then overrideErrorHelp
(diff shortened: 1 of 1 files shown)


sjakobi:
This does not appear to fix the my similar issue in https://github.com/commercialhaskell/stack/issues/2343.

Could this fix be generalized for commands like `stack -v script.hs`?

harendra-kumar:
@sjakobi that seems to be a different problem special to verbosity level. We turn down the noise when in interpreter mode so that we do not see unnecessary output along with the script output. But that seems to override even if we want to turn it on.

```
        globalOptsParser kind (if isInterpreter
                                then Just $ LevelOther "silent"
                                else Nothing)
```

See #1487 and #1472. I see your comments in 1472 for exactly the same problem earlier.

harendra-kumar:
@puffnfresh there are other alternatives to achieve the same thing but they are not as convenient:
* modify the command line in the script before running it.
* copy and paste the command line from the script and add to it before running. Not convenient and especially hard when the command in the script is a multiline command.

This fix, as I see it, adds value in conveniently augmenting the existing command line inside the script. There are a few things to note about it though:
* This is limited to prefixing options to the existing stack command inside the script i.e. add global arguments. Any subcommand arguments are still controlled only from within the script.
* If a global argument refers to an executable file, that file will be executed instead. This might come as a surprise to the users. That's the artifact of a hack. It was a hack already though, this fix builds upon that.

I am inclined towards merging it, except if someone has a concern about the second point.

src/main/Main.hs
```diff
@@ -498,18 +498,32 @@ interpreterHandler
...
 interpreterHandler args f = do
-  isFile <- D.doesFileExist file
-  if isFile
-  then runInterpreterCommand file
-  else parseResultHandler (errorCombine (noSuchFile file))
+  -- args can include top-level config such as --extra-lib-dirs=... (set by
+  -- nix-shell) - we need to find the first argument which is a file, everything
+  -- afterwards is an argument to the script, everything before is an argument
+  -- to Stack
+  (stackArgs, fileArgs) <- spanM (fmap not . D.doesFileExist) args
```
harendra-kumar: We can use `takeWhileM` from [monad-loops](https://hackage.haskell.org/package/monad-loops-0.4.3/docs/Control-Monad-Loops.html). `monad-loops` is already a dependency of stack via `monad-logger`.

The `spanM` function can be added to `monad-loops` package.

harendra-kumar:
We should also add a note about this new behavior in the `script interpreter` section of the [GUIDE](https://github.com/commercialhaskell/stack/blob/master/doc/GUIDE.md).

harendra-kumar:
Its been pending for a while, I am just going ahead and merging this.

Issue #2262, opened 2016-06-09
Allow --pvp-bounds to use multiple stack.yaml configurations


mgsloan:
This is split off from https://github.com/commercialhaskell/stack/issues/1568, which focused on having multiple configurations in a single stack.yaml file.

As @sjakobi suggests, it's quite reasonable to still have a `--pvp-bounds` flag which uses multiple configs:

> How about specifying the extra `stack.yaml`s to consider with an optional flag to `--pvp-bounds`, e.g. `--also-considering`?
> Example command: `stack sdist --pvp-bounds both --also-considering stack-7.8.yaml stack-8.0.yaml`.

Folks would probably want to automate this via a Makefile or script.

borsboom:
I'm very much in favour of this.  It'd be super useful when uploading `stack` itself to Hackage.

DanBurton:
+1 looks like a very sensible user interface to add needed flexibility to `--pvp-bounds`.

sjakobi:
I'm wondering if we could get away without having a separate `stack.yaml` for each dependency configuration that should be considered.

With the current feature idea, if I had a simple library, I'd create a directory `other-stack-yamls` that contains a bunch of `stack-lts-X.yaml` files containing only the `resolver` line with one `X` for each lts generation that I want to support and then run `stack sdist --pvp-bounds both --also-considering other-stack-yamls/stack*.yaml` when making a release.

So what I'm currently thinking, is that it should be possible to pass both `stack.yaml` file names and snapshot names to `--also-considering`, so you could say e.g. 

```
stack sdist --pvp-bounds both --also-considering stack-7.8.yaml lts-4 nightly-2016-06-19
``` 

Thoughts on that, anyone?

Ideally it would also be easy to run tests with the same list of dependency configurations.

mgsloan:
@sjakobi Seems like a good approach to me!

DanBurton:
Well, there can be a lot more config than just the `resolver` in the stack.yaml file; it doesn't make sense to me to allow a file for the `--resolver` flag.

sjakobi:
> Perhaps we should allow stack.yaml filepaths in generaal for --resolver? I do not see a downside. It is shorter and easier to type than --stack-yaml.

Assume that there's a file `stack.yaml`:

```
packages:
- pkg1
- pkg2
resolver: lts-6.4
```

and a file `stack-8.0-testing.yaml`:

```
packages:
- pkg1
resolver: nightly-2016-06-24
```

[...]

PR #2322, opened 2016-07-02
Add multi-config pvp-bounds generation (#2262)

sjakobi:
**PSA:** I don't intend to rebase this before at least some of the design issues below have been discussed. So please take a look and chime in!

### DIFF
--- a/src/Stack/Build/Installed.hs
+++ b/src/Stack/Build/Installed.hs
@@ -8,6 +8,7 @@ module Stack.Build.Installed
     ( InstalledMap
     , Installed (..)
     , GetInstalledOpts (..)
+    , defaultGetInstalledOpts
     , getInstalled
     ) where
 
@@ -53,10 +54,18 @@ data GetInstalledOpts = GetInstalledOpts
       -- ^ Require haddocks?
     }
 
+defaultGetInstalledOpts :: GetInstalledOpts
+defaultGetInstalledOpts = GetInstalledOpts
+    { getInstalledProfiling = False
+    , getInstalledHaddock = False
+    }
+
 -- | Returns the new InstalledMap and all of the locally registered packages.
 getInstalled :: (M env m, PackageInstallInfo pii)
              => EnvOverride
              -> GetInstalledOpts
+             -- TODO: Clarify what the Map is used for and what properties it
+             -- should have.
              -> Map PackageName pii -- ^ does not contain any installed information
              -> m ( InstalledMap
                   , [DumpPackage () ()] -- globally installed
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -418,7 +418,7 @@ getToolMap mbp =
 
 loadResolver
     :: (MonadIO m, MonadLogger m, MonadReader env m, HasHttpManager env, HasConfig env, HasGHCVariant env, MonadBaseControl IO m, MonadMask m)
-    => Maybe (Path Abs File)
+    => Maybe (Path Abs File) -- ^ Optional path to project config file
     -> Resolver
     -> m (MiniBuildPlan, LoadedResolver)
 loadResolver mconfigPath resolver =
@@ -954,7 +954,7 @@ shadowMiniBuildPlan (MiniBuildPlan cv pkgs0) shadowed =
 
 parseCustomMiniBuildPlan
     :: (MonadIO m, MonadMask m, MonadLogger m, MonadReader env m, HasHttpManager env, HasConfig env, HasGHCVariant env, MonadBaseControl IO m)
-    => Maybe (Path Abs File) -- ^ Root directory for when url is a filepath
+    => Maybe (Path Abs File) -- ^ Optional path to project config file
     -> T.Text
     -> m (MiniBuildPlan, SnapshotHash)
 parseCustomMiniBuildPlan mconfigPath0 url0 = do
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -26,6 +26,7 @@ module Stack.Config
   ,loadConfig
   ,loadConfigMaybeProject
   ,loadMiniConfig
+  ,loadProjectConfig
   ,packagesParser
   ,resolvePackageEntry
   ,getImplicitGlobalProjectDir
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -31,7 +31,7 @@ import qualified Data.Traversable as T
 import           Network.HTTP.Client.Conduit (HasHttpManager)
 import           Prelude -- Fix redundant import warnings
 import           Stack.Build (withLoadPackage)
-import           Stack.Build.Installed (getInstalled, GetInstalledOpts(..))
+import           Stack.Build.Installed (getInstalled, defaultGetInstalledOpts)
 import           Stack.Build.Source
 import           Stack.Build.Target
 import           Stack.Constants
@@ -92,7 +92,7 @@ createDependencyGraph dotOpts = do
   let graph = Map.fromList (localDependencies dotOpts locals)
   menv <- getMinimalEnvOverride
   installedMap <- fmap snd . fst4 <$> getInstalled menv
-                                                   (GetInstalledOpts False False)
+                                                   defaultGetInstalledOpts
                                                    sourceMap
   withLoadPackage menv (\loader -> do
     let depLoader =
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -306,10 +306,7 @@ ghciSetup GhciOpts{..} = do
     menv <- getMinimalEnvOverride
     (installedMap, _, _, _) <- getInstalled
         menv
-        GetInstalledOpts
-            { getInstalledProfiling = False
-            , getInstalledHaddock   = False
-            }
+        defaultGetInstalledOpts
         sourceMap
     econfig <- asks getEnvConfig
     directlyWanted <-
--- a/src/Stack/Options.hs
+++ b/src/Stack/Options.hs
@@ -24,7 +24,9 @@ module Stack.Options
     ,testOptsParser
     ,haddockOptsParser
     ,hpcReportOptsParser
+    ,sdistOptsParser
     ,pvpBoundsOption
+    ,pvpBoundsOptsParser
     ,globalOptsFromMonoid
     ,splitObjsWarning
     ) where
@@ -60,6 +62,7 @@ import           Stack.Ghci                        (GhciOpts (..))
 import           Stack.Init
 import           Stack.New
 import           Stack.Nix
+import           Stack.SDist
 import           Stack.Types
 import           Stack.Types.TemplateName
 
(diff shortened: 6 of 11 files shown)


sjakobi:
### Remaining design issues:

* [ ] `loadDependencyConfigs` currently feels a bit flimsy and certainly misses some configuration options that influence that dependency versions. Possibly the internal machinery for `stack list-dependencies` could be used to get the actual dependency versions. Can someone comment whether that would work?

* [ ] The CLI usage info currently doesn't reflect that `--also-considering` only works in combination with `--pvp-bounds`.

    ```
    ~/s/servant (master) $ stack sdist --help
    Usage: stack sdist [DIR] [--pvp-bounds PVP-BOUNDS]
                       [--also-considering RESOLVER_OR_STACK_YAML] [--ignore-check]
                       [--sign] [--sig-server URL] [--help]
      Create source distribution tarballs

    Available options:
      --pvp-bounds PVP-BOUNDS  How PVP version bounds should be added to .cabal
                               file: none, lower, upper, both
      --also-considering RESOLVER_OR_STACK_YAML
                               Extra dependency configs to consider during bounds
                               generation
    <snip>

    ~/s/servant (master) $ stack sdist --also-considering stack-ghc-7.8.4.yaml
    Missing: --pvp-bounds PVP-BOUNDS
    <snip>
    ```

    It would be nice if the usage was formatted like
    ```
    [--pvp-bounds PVP-BOUNDS  [--also-considering RESOLVER_OR_STACK_YAML]]
    ```
    I'm not sure whether `optparse-applicative` currently supports this.

* [ ] Using `--pvp-bounds --also-considering` currently spews lots and lots of (badly worded) info on the bounds generation process:

    ```
    ~/s/servant (master) $ stack sdist --pvp-bounds both --also-considering stack-ghc-7.8.4.yaml --also-considering stack-ghc-8.0.1.yaml --also-considering lts-6.4 servant
    Getting file list for /home/simon/src/servant/servant/
    Building sdist tarball for /home/simon/src/servant/servant/
    Adjusting version range for dependency base-compat:
      Known versions:
        * 0.9.1 (current configuration, stack-ghc-7.8.4.yaml, stack-ghc-8.0.1.yaml, lts-6.4)
      Old range: >=0.9
      New range: ==0.9.*
      The new version range contains subranges compatibility with which cannot be inferred within the PVP:
        >=0.9 && <0.9.1
    Adjusting version range for dependency aeson:
      Known versions:
        * 0.8.0.2 (stack-ghc-7.8.4.yaml)
        * 0.11.2.0 (current configuration, stack-ghc-8.0.1.yaml, lts-6.4)
      Old range: >=0.7
      New range: >=0.7 && <0.12
      The new version range contains subranges compatibility with which cannot be inferred within the PVP:
        >=0.7 && <0.8.0.2 || >=0.9 && <0.11.2.0
    <snip>
    ```
    My reasoning behind this is that I want to be very explicit about these "PVP inference gaps". I'm worried that using the "PVP" name and not adhering to it very strictly might upset a lot of people.

    Still, the current output is way too much. Should I simply warn the users about those gaps once and ask them to use `-v` to get the full inference info? (Downside: The debug output is pretty hard to read.)

* [ ] Without context, it's pretty impossible to understand what `--also-considering` is supposed to mean.

  Would any of the following names be better?

   * `--extra-dependency-configs` ("dependency config" is AFAIK new terminology)
   * `--extra-package-configs`
   * Another name?

* [ ] Option syntax: I originally intended to have the following syntax:

  ```
  stack sdist --pvp-bounds both --also-considering stack-7.8.yaml stack-8.0.yaml
  ```

  My current understanding is that `optparse-applicative` doesn't readily support such syntax for options and that this syntax would clash with `[DIR]` arguments that can be passed to `stack sdist`. So I ended up having the following syntax instead:

  ```
  stack sdist --pvp-bounds both --also-considering stack-7.8.yaml --also-considering stack-8.0.yaml
  ```

  Would the syntax that we currently use e.g. for the ghc options be better?
  ```
  stack sdist --pvp-bounds both --also-considering "stack-7.8.yaml stack-8.0.yaml"
  ```
  (I think it's ok as it stands)

Comments on the design issues here would be highly appreciated! :)

### Other TODOs

* [ ] Add (integration) tests.
* [ ] Amend changelog

### Extensions to be added in another PR:

* I think it would be nice to be able to specify as many `sdist`/`upload` options as possible in the global and project configurations. Ideally, creating and uploading a tarball would then only require a single command `stack upload`.

sjakobi:
Regarding the new `Cabal >= 1.24` requirement:

This is necessary because I'm using a new function `Distribution.Version.invertVersionRange`. If the new bound is very problematic, I could simply duplicate that function instead.

mgsloan:
This is awesome work!  Sorry for letting it languish so long.  I merged it atop master and pushed that to a branch of the same name https://github.com/commercialhaskell/stack/tree/2262-multiple-config-pvp-bounds .  I should have probably rebased, but this works.

> loadDependencyConfigs currently feels a bit flimsy and certainly misses some configuration options that influence that dependency versions. Possibly the internal machinery for stack list-dependencies could be used to get the actual dependency versions. 

Makes sense!  RE `-- TODO: Do we need to look at extra package dbs etc, too?`, no the dep versions should be calculatable directly from the stack.yaml + resolver info.  It does need to handle extra-deps

>  The CLI usage info currently doesn't reflect that --also-considering only works in combination with --pvp-bounds.

What if it implied `--pvp-bounds both`?

> Using --pvp-bounds --also-considering currently spews lots and lots of (badly worded) info on the bounds generation process:

Seems fine.  I merged the pretty printer stuff today, so could use that too!

> EDIT2: Why not offer the "conservative"/holey bounds by default and have an additional flag, maybe --contiguous-bounds for the simpler bounds?! This way the user has the choice.

How about `--pvp-bounds holey`, in additional to `upper` / `lower` / `none`.

> Without context, it's pretty impossible to understand what --also-considering is supposed to mean.

> ```
> stack sdist --pvp-bounds both --also-considering "stack-7.8.yaml stack-8.0.yaml"
> ```

Maybe this is a crazy idea, but what about just jamming it into the `pvp-bounds` option: `stack sdist --pvp-bounds "both [stack-7.8.yaml, stack-8.0.yaml]"`

Agreed that `--also-considering` is weird out of context.

sjakobi:
> > The CLI usage info currently doesn't reflect that --also-considering only works in combination with --pvp-bounds.

> What if it implied --pvp-bounds both?

A line like `stack sdist --also-considering stack-x.yaml` doesn't make it clear that `stack` will change the bounds. That could be alleviated with a better option name but the implication of `both` might still not be  obvious.

I have opened https://github.com/pcapriotti/optparse-applicative/issues/220 asking for nested options in the usage info and it seems that we might get them in the future.

> > Using --pvp-bounds --also-considering currently spews lots and lots of (badly worded) info on the bounds generation process:

> Seems fine.

To be clear: You are ok with both the wording and the amount of output?

>  I merged the pretty printer stuff today, so could use that too!

I'll do that! The formatting was very ad-hoc so far.

> > EDIT2: Why not offer the "conservative"/holey bounds by default and have an additional flag, maybe --contiguous-bounds for the simpler bounds?! This way the user has the choice.

> How about --pvp-bounds holey, in additional to upper / lower / none.

Ah, that might actually be a good idea! I hadn't realized before that a "holey" mode would ignore any given lower or upper bounds from the `.cabal` file and would therefore behave the same for `upper`, `lower` and `both`. Maybe call it `skeptic` instead (with a `sceptic` as an alias)?

> Maybe this is a crazy idea, but what about just jamming it into the pvp-bounds option: stack sdist --pvp-bounds "both [stack-7.8.yaml, stack-8.0.yaml]"

That's short and elegant but means that users will have to remember this syntax for an option that they'd probably use only infrequently. I guess I'll stick with the existing, long but simple syntax.

> Agreed that --also-considering is weird out of context. Not sure what a better name is.

Right now I tend to go with `--extra-package-configs` which seems a bit clearer.

> Our flags do sometimes have rather context sensitive names. Like --open for build actually refers to opening haddocks, heh. Not sure if this is something to encourage.

Hopefully https://github.com/pcapriotti/optparse-applicative/issues/220 will alleviate this problem.

mgsloan:
> Ah, that might actually be a good idea! I hadn't realized before that a "holey" mode would ignore any given lower or upper bounds from the .cabal file and would therefore behave the same for upper, lower and both. Maybe call it skeptic instead (with a sceptic as an alias)?

Could call it `conservative` or `pessimal`, heh.

For naming, what about `--pvp-considering "stack-7.8.yaml stack-8.0.yaml"`?

borsboom:
In my way of using `--pvp-bounds=both`, I initially thought I'd prefer the conservative/holey bounds to be the default.  The motivating case here is something like aeson-0.10 (_shudder_).  Let's say I had `stack-lts-3.yaml` (uses aeson-0.8.0.2) and `stack-lts-6.yaml` (uses aeson-0.11.2.1).  I would absolutely _not_ want that to imply that aeson-0.10 should be allowed; it would probably break the build or break the application.  So in that case the generated bounds would be `(>= 0.8.0.2 && < 0.9) || (>= 0.11.2.1 && < 0.12)`.

But that would be ugly if we didn't have a major version gap (e.g something like `(>= 0.8.0.2 && < 0.9) || (>= 0.9.0.1 && < 0.10)`).  I think instead it would be safe to assume that lower _minor_ of the higher major version are OK and combine them if contiguous.  So for the example in this paragraph, since we already know 0.8.0.2 is allowed, that would imply that _any_ 0.9.* is OK, so then this would result: `>= 0.8.02 && < 0.10` (note it still keeps the lower minor version bound for 0.8.02 though).  __However__, for the case of a major version gap, this combining should not be performed (the resulting constraint should still be like the example in the first paragraph).

If there's a lower bound or upper bound already in the `.cabal` file, that should still be taken as "gospel" in all cases.

sjakobi:
> But that would be ugly if we didn't have a major version gap (e.g something like (>= 0.8.0.2 && < 0.9) || (>= 0.9.0.1 && < 0.10)). I think instead it would be safe to assume that lower minor of the higher major version are OK and combine them if contiguous. So for the example in this paragraph, since we already know 0.8.0.2 is allowed, that would imply that any 0.9.* is OK, so then this would result: >= 0.8.02 && < 0.10 (note it still keeps the lower minor version bound for 0.8.02 though).

I had a similar "cosmetic" smoothing feature in an earlier version of this PR. I currently think it's better not to do this kind of "minor version gap smoothing" because:

* It adds complexity to an already not-so-simple feature
* This mode is supposed to be somewhat "honest" about the PVP and the PVP doesn't allow us to assume compatibility in this case. In particular we might include buggy package versions in this way.
* It shouldn't have much impact on cabal's ability to find a matching package set (as upper version bounds are usually only specified like `< X.Y`, accepting any version with the same major version if at all)

I don't feel very strongly about this though, and am happy to be outvoted.

> If there's a lower bound or upper bound already in the .cabal file, that should still be taken as "gospel" in all cases.

I take it that "gospel" means that these bounds would be copied verbatim and not combined in any way with the inferred intervals, right? (Earlier, I believe I misunderstood you and thought that you'd want existing lower/upper bounds to be combined with the inferred intervals in such a way that the upper/lower bounds are preserved. I haven't found a principled way to do this though.)

borsboom:
I'm also fine with leaving the gaps, so if that's what you feel is right then go ahead.

I _think_ the current behaviour is that if there's only a lower bound in the .cabal file, then we leave the lower bound verbatim but can still infer the upper bound from the resolver(s).  And same idea for the upper bound.  Whatever the current behaviour, it should probably stay the same.

borsboom:
So, in a way, the algorithm could just be to apply the _existing_ single stack.yaml algorithm to each stack.yaml, and then combine all the results with `||`.  However, that would give you a repetitive constraint so finding a way to avoid some of the repetition would be nice.

mgsloan:
Pinging @sjakobi - I think it would be great to get this merged!  I'm not really sure what remains to be done, as there was quite a bit of discussion.

mgsloan:
Closing this due to age of the PR, and the conversation has gotten long.  A followup PR is certainly welcome!

PR #2326, opened 2016-07-04
Avoid re-running `getLocalPackageViews`

sjakobi:
`Stack.Build.Source.loadSourceMap` used to result in two calls to
`getLocalPackageViews`, calling it once directly and once via
`parseTargetsFromBuildOpts`.
This change adds an argument to `parseTargetsFromBuildOpts` through
which the local package views can optionally be passed in.


mgsloan:
Good idea!  This is a nitpick, but how about not having the `Maybe` and moving the `getLocalPackageViews` calls next to the `Nothing` usages?

sjakobi:
> Good idea! This is a nitpick, but how about not having the Maybe and moving the getLocalPackageViews calls next to the Nothing usages?

I agree that this was a bit ugly before.

What do you think of my second commit which solves the same issue slightly differently?
(`</bikeshed>` ;))

mgsloan:
LGTM!

PR #2327, opened 2016-07-04
Improve logging of sub-processes

sjakobi:
* Log the fully qualified command name (/usr/bin/ldd instead of ldd)
* Log how long the sub-process took to finish

`getTime Monotonic` takes ~150 ns on my machine while the processes all take > 10 ms. The overhead for the timings is therefore rather negligible.


sjakobi:
As one can see in [this example](https://github.com/commercialhaskell/stack/pull/2346#issuecomment-230849572) the repeated output of the full command can be a bit overwhelming.

Would it make sense to remove the `Run process: …` line and only have the `Process finished in X ms` line?

mgsloan:
I accidentally rebased this atop master instead of merging.  But it's on master now!  Thanks!

PR #2328, opened 2016-07-04
Move code for path, setup, hoogle commands and various withXConfig utils from Main into separate modules

sjakobi:
The main motivation for this is faster builds.

### DIFF
--- a/src/Stack/Hoogle.hs
+++ b/src/Stack/Hoogle.hs
@@ -1,0 +1,178 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+-- | A wrapper around hoogle.
+module Stack.Hoogle
+    ( hoogleCmd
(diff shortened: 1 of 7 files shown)


sjakobi:
I'll address `sdistCmd` and `uploadCmd` once #2322 has been merged.

`packagesCmd` and `targetsCmd` should probably be moved to `Stack.Ide` (or `Stack.IDE`???).

sjakobi:
Would anyone like to do some bikeshedding on whether to call the module `Stack.IDE` or `Stack.Ide`?

I think `IDE` looks nices but `Ide` is more easily typed.

We already have a few modules in the ALLCAPS style (`Stack.Sig.GPG`, `Network.HTTP.Download`) but also `Stack.Ghci`…

I think I'm leaning towards `IDE`…

sjakobi:
I have rebased this on `master`, renaming `Stack.Ide` to `Stack.IDE`.

Because I believe that this is a beneficial change and because it doesn't change any behavior I'll merge this once the CI jobs pass.

Issue #2332, opened 2016-07-04
Breakage related to http-client-0.5.0 / http-conduit-2.2.0


sjakobi:
I've had a Travis job fail because it picked a new version of http-conduit:

https://travis-ci.org/sjakobi/stack/jobs/142306535#L1131

I guess we should just adapt the code and add lower bounds for http-client and http-conduit.

PR #2338, opened 2016-07-05
Use http-client-0.5.0, http-conduit-2.2.0, http-client-tls-0.3.0

sjakobi:
Because I'm not really familiar with `http-client` I usually opted to replace the deprecated `parseUrl` with the synonymous `parseUrlThrow`.

Closes https://github.com/commercialhaskell/stack/issues/2332.

### DIFF
--- a/src/Network/HTTP/Download.hs
+++ b/src/Network/HTTP/Download.hs
@@ -16,6 +16,7 @@ module Network.HTTP.Download
     , download
     , redownload
     , downloadJSON
+    , parseRequest
     , parseUrlThrow
     , liftHTTP
     , ask
@@ -48,7 +49,7 @@ import           Data.Text.Encoding          (decodeUtf8With)
 import           Data.Typeable               (Typeable)
 import           Network.HTTP.Client         (path, checkResponse)
 import           Network.HTTP.Client.Conduit (HasHttpManager, Manager, Request,
-                                              Response,
+                                              Response, parseRequest,
                                               getHttpManager, parseUrlThrow,
                                               requestHeaders, responseBody,
                                               responseHeaders, responseStatus,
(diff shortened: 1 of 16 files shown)


borsboom:
I'd just gotten started with this process when I saw your PR come through, and looks like you took the same approach I'd begun, so great!

The only thing I was thinking of doing differently was that in cases where a `checkStatus = \_ _ _ -> Nothing` was set, I instead replaced `parseUrl` with `parseRequest` and removed the `checkStatus` (rather than changing to `parseUrlThrow` and `checkResponse`).  I _think_ that was safe, but I'm not 100% certain.

src/Stack/Upload.hs
```diff
@@ -195,10 +195,9 @@ mkUploader config us = do
     manager <- usGetManager us
     (creds, fromFile') <- loadCreds $ usCredsSource us config
     when (not fromFile' && usSaveCreds us) $ saveCreds config creds
-    req0 <- parseUrl $ usUploadUrl us
+    req0 <- parseRequest $ usUploadUrl us
     let req1 = req0
             { requestHeaders = [("Accept", "text/plain")]
-            , checkStatus = \_ _ _ -> Nothing
```
sjakobi: @borsboom:  I did the same in this case.

src/Network/HTTP/Download.hs
```diff
@@ -108,7 +108,7 @@ redownload req0 dest = do
                         requestHeaders req0 ++
                         [("If-None-Match", L.toStrict etag)]
                     }
-        req2 = req1 { checkStatus = \_ _ _ -> Nothing }
+        req2 = req1 { checkResponse = \_ _ -> return () }
```
sjakobi: @borsboom: In this case the we're handling the status codes manually afterwards, so I think it's still good to override the `checkResponse` method that's passed in.

sjakobi:
@borsboom: I've added a second commit that should be more in line with your comment. I previously hadn't checked which requests were handled by `redownload`.

borsboom:
LGTM!  Thanks!

PR #2340, opened 2016-07-06
Fix a Markdown error

sid-kap:


sjakobi:
Thanks!

@borsboom: How should we handle doc fixes that land in `master`? Can we just cherry-pick them into `stable`?

PR #2352, opened 2016-07-08
Parse local cabal files concurrently

sjakobi:
This should particularly benefit large multi-package projects.

### DIFF
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -22,6 +22,7 @@ module Stack.Build.Source
 
 import              Control.Applicative
 import              Control.Arrow ((&&&))
+import              Control.Concurrent.Async.Lifted (forConcurrently)
 import              Control.Exception (assert, catch)
 import              Control.Monad hiding (sequence)
 import              Control.Monad.Catch (MonadMask)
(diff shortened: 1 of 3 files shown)


sjakobi:
Oh, this does not perform as I expected:

With this commit:

```
~/s/servant (master) $ bench "stack-concurrent-getlpv clean"
benchmarking stack-concurrent-getlpv clean
time                 223.4 ms   (198.6 ms .. 254.8 ms)
                     0.982 R²   (0.939 R² .. 1.000 R²)
mean                 221.1 ms   (208.0 ms .. 231.7 ms)
std dev              14.63 ms   (10.76 ms .. 17.77 ms)
variance introduced by outliers: 15% (moderately inflated)
```

Before:

```
~/s/servant (master) $ bench "stack-master clean"
benchmarking stack-master clean
time                 186.7 ms   (181.6 ms .. 194.0 ms)
                     0.998 R²   (0.992 R² .. 1.000 R²)
mean                 187.8 ms   (185.1 ms .. 190.4 ms)
std dev              3.480 ms   (2.862 ms .. 4.316 ms)
variance introduced by outliers: 14% (moderately inflated)
```

My SSD is particularly slow though. I'll try to read the files in sequence and parse them concurrently.

sjakobi:
Argh, the above commands didn't even run `getLocalPackageViews`. I had totally forgotten about my optimization in `stack clean` which will now only run `getLocalPackageViews` if the user specifies a target manually.

This makes the performance difference even more surprising though. I'll get some sleep and take another look at this.

sjakobi:
I still can't see any performance gains from this, rather slight decreases. Reading all the cabal files (even when there are more than 100) concurrently might not be the best idea either.

I might look into this more once there has been a new release and we have better logs.

PR #2354, opened 2016-07-08
Nix: Adding LD_LIBRARY_PATH to shell (Template Haskell issues)

YPares:
GHCi and therefore template haskell need LD_LIBRARY_PATH when in a nix shell.
This makes stack (when using a list of nix packages instead of a shell file) behave more like `haskell.lib.buildStackProject` in nixpkgs.
See for instance https://github.com/tweag/HaskellR/issues/253


borsboom:
LGTM.  Thanks!

Issue #2040, opened 2016-04-17
Stack doesn't rebuild enough when using Custom build-type


mboes:
#### Steps to reproduce

Create a trivial stack project with the following Setup.hs (and setting `build-type: Custom`):

```Haskell
import Distribution.Simple

main = defaultMainWithHooks simpleUserHooks { postBuild = putStrLn "Hey!" }
```

Then `stack build` twice.

#### Expected

stack should run the preBuild/postBuild hooks each time `stack build` is called. In general these may be callouts to `make` or any other foreign build system, whose dependency graph is invisible to Cabal/Stack/GHC.

#### Actual

stack runs these hooks once only, unless Haskell source files change. Note that cabal-install does rerun these hooks every time. It's Stack's recompilation avoidance that seems to be the problem here.

mboes:
One possible solution here could be to treat `build-type: Custom` specially, i.e. be less aggressive about recompilation avoidance in that particular case.

mgsloan:
Yeah, special casing `built-type: Custom` for this makes sense to me.

mboes:
Problem is, if package D has `build-type: Custom`, and the dependency chain in a multi-package project looks like A -> B -> C -> D, then the hooks will be rerun each time for D and presumably A, B, C become out-of-date and rebuild too? Likely no object files will be rebuilt, but the null build may take quite a bit more time.

What about another solution: take into account the `extra-source-files` field, and posit that `build-type: Custom` only ever needs to rerun if any of the sources files listed there (+ those files already taken into account) are newer. This is perhaps not correct in general, but should cover nearly every sane use case for using build hooks.

mgsloan:
Yeah, though it seems to me like a strange behavior to toggle based on `build-type`.  Perhaps changes to  `extra-source-files` should always cause package dirtiness?  It's better to have false positives than false negatives for dirtiness checking.

borsboom:
I think this is resolved by the merge of #2362. Please re-open if not.

Issue #1891, opened 2016-03-09
Custom Preprocessed Files Don't Rebuild


deech:
I have a custom preprocessor that converts files in my `src` directory to Haskell files. However when any of those originals change `stack build` doesn't appear to reprocess. If I do  `stack clean` first that works.

mgsloan:
Are those files listed in `extra-source-files` in your cabal file?  If so, this could be a bug.  It seems prudent to treat a package as being dirty if these files are changed.  I don't see this being done anywhere [here](https://github.com/commercialhaskell/stack/blob/2bf17fbc2e84faadf13fe7a29d5954635e646b73/src/Stack/Package.hs#L223)

deech:
Yes they are. If it helps in my scenario I'm only building executables (no library). As I found out `extra-source-files` is only allowed once at the top of the Cabal file. If I put it in each executable's stanza, Cabal complains.

NicolasT:
I believe I'm having a similar issue: a source file is listed in `extra-source-files` in the global section of my package, and then converted into a `data-files` member (also listed in the global section) through ` Custom` build (`Setup.hs` calling out to some tool in the `buildHook`).

Unexpectedly, whenever I modify the source file, no rebuild is triggered: not when using `stack build --file-watch`, nor when running plain `stack build` multiple times.

borsboom:
I think this is resolved by the merge of #2362.  Please re-open if not.

PR #2362, opened 2016-07-10
Let extra-source-files mark a package as dirty (#1891, #2040).

judah:
This is necessary for packages that use a custom Setup.hs
(`build-type: Custom`).  Previously, stack wouldn't rebuild the package
if the only changes were to the `extra-source-files`.

Stack was already marking the package as dirty if `data-files` changed;
this just hooks `extra-source-files` into the same mechanism.


borsboom:
LGTM; thanks for your contribution!

PR #2363, opened 2016-07-10
stack unpack: Ignore pax headers (fix #2361)

### LINKED ISSUE #2361: stack install ghc-core fails because it mishandles special tar entries
`stack install ghc-core` fails, but `stack build` after `cabal unpack ghc-core` succeeds. Apparently it fails while trying to use `setFileMode` on file `pax_global_header`.
TL;DR. `ghc-core`'s tarball contains an entry `pax_global_header` which does not represent a file, and `stack` incorrectly assumes all entries are files, so `stack` fails unpacking. It turns out that `stack unpack ghc-core` fails while `cabal unpack ghc-core` works. Such entries are valid, and `git-archive` creates them.

Investigation shows that
- `pax_global_header` is a special tar entry contains extra attributes; newer `tar` versions special-case it, older ones don't (see e.g. https://github.com/backdrop-ops/contrib/issues/55#issuecomment-73814500)
- the tarball for ghc-core appears to contain a `pax_global_header` entry (confirmed e.g. through `gzcat ghc-core-0.5.6.tar.gz| strings |grep pax_global_header`, or with `zcat` on Linux).
- `stack` calls setFileMode on each entry in the tar file, without checking if they're real entries or not:
https://github.com/commercialhaskell/stack/blob/master/src/Stack/Fetch.hs#L524-L532
- The `tar` package itself seems to mark these entries as with [EntryContent = OtherEntryType ...](http://hackage.haskell.org/package/tar-0.5.0.3/docs/Codec-Archive-Tar-Entry.html#t:EntryContent), with specific type codes, see e.g.
https://github.com/haskell/tar/commit/6c0acf8edb9c5006a8fa1e1c555177df4cfa2d18.
- So `stack` should ignore the same entries — I'm just not sure how to do that without duplicating too much knowledge about the format from the Tar package. Looking at `unpack` (https://github.com/haskell/tar/blob/4557fdeab9363fcce16ed0be1f41eda022ba7c55/Codec/Archive/Tar/Unpack.hs#L77-L82), we might want to filter for entries of "type" (`EntryContent`) being `NormalFile`, `Directory`, `HardLink`, `SymbolicLink`.

Failing invocation:
```
$ stack install ghc-core
Using resolver: lts-6.6 from implicit global project's config file: /Users/pgiarrusso/.stack/global/stack.yaml
/private/var/folders/_7/hlxv4yv95x95vgnn416b4q4m0000gp/T/stack96236/pax_global_header: setFileMode: does not exist (No such file or directory)
```

Working:
```
$ cabal unpack ghc-core
Downloading ghc-core-0.5.6...
Unpacking to ghc-core-0.5.6/
# Standard setup (omitted)
$ stack build
ghc-core-0.5.6: configure
Configuring ghc-core-0.5.6...
ghc-core-0.5.6: build
Preprocessing executable 'ghc-core' for ghc-core-0.5.6...
[1 of 1] Compiling Main             ( ghc-core.hs, .stack-work/dist/x86_64-osx/Cabal-1.24.0.0/build/ghc-core/ghc-core-tmp/Main.o )
Linking .stack-work/dist/x86_64-osx/Cabal-1.24.0.0/build/ghc-core/ghc-core ...
ghc-core-0.5.6: copy/register
Installing executable(s) in
/Users/pgiarrusso/AeroFS/Mine/Sorgenti-Sync/Haskell/Stack-GHC-7.8.4/.stack-work/install/x86_64-osx/lts-0.7/7.8.4/bin
```

<s>(At this point, `stack install ghc-core` seems to work in cases where it failed before in the same snapshot, but I haven't investigated).</s> (EDIT: Given my setup that's not surprising).

Blaisorblade:
Stop trying to reset permissions on pax header entries. This fixed the bug as verified by `stack exec -- stack unpack ghc-core` and `stack exec -- stack install ghc-core`. Haven't tested on other packages or on the testsuite (I'm relying on CI).

XXX This is incomplete, see XXX case in the patch. I'm not sure how defensive to be for the case that some tar entry is not one of the known entry types. Also, I've not checked other tar formats than `pax`, so maybe there's more to expect.

Quite possibly, the code should just have a catchall case for all cases leading to False or at least for all `Tar.OtherEntryType` entries, and then I can dispense with all these complications.

Fix #2361.


borsboom:
This looks pretty good.  I'm honestly not sure what the best way to handle the other case; maybe a warning would be the safest just so there are no surprises.  Another idea would be to _always_ try to set the permissions, just like before this patch, but treat failure asa warning for the weird/unknown cases (i.e., still an error if it fails for a regular file, directory, etc. but just a warning for character devices, OtherEntryType, etc.).

Blaisorblade:
Thanks for the prompt response!

> treat failure as a warning for the weird/unknown cases (i.e., still an error if it fails for a regular file, directory, etc. but just a warning for character devices, OtherEntryType, etc.).

That could work. But FWIW, those don't even get extracted by `Tar.extract`, so we risk too many warnings.

Orthogonally: GNU archives can use other flags. Since none correspond to files or are handled by `Tar.extract`, I'd say I'm fine with ignoring entries of those types, or with just ignoring any `OtherEntryType`.
https://www.gnu.org/software/tar/manual/html_section/tar_91.html#SEC187

sjakobi:
Thanks for working on this, @Blaisorblade! :)

I think I'd like to get a warning if the tarball contains a named pipe, and maybe the author would like to know too. So I'm in favor of extracting what we can reasonably use  – just like what you have already implemented – and of warning about all the things that we can't.

And could you please add a regression test that checks that the warning is present and that the unpacking works?! Cheers!

Blaisorblade:
First down, the warnings (output yet untested). Using foldEntries is surprisingly ugly; I've tried converting to and from a list but it might not be better :-(. Next up, tests!

Blaisorblade:
So, since I like unit tests, I ended up with some refactorings and harmless changes to make this code unit-testable. I also haven't seen unit tests with data, so I had to get creative there.
I've run normal tests locally, for integration tests I'm relying on CI.

Blaisorblade:
I don't get the AppVeyor failure, and it doesn't seem about my testcase. If AppVeyor is flaky like Travis, could you restart the AppVeyor build? There *is* enough path manipulation that it matters that it passes.

There's no clear indication it's about my testcase, and the actual code in main should still do essentially the same thing (except for https://github.com/commercialhaskell/stack/pull/2363/commits/98e2ec0d4c090372e360658ec2281c25d20ca7c2).

Also master already broke in https://ci.appveyor.com/project/snoyberg/stack/build/1.0.1867, and 
https://ci.appveyor.com/project/snoyberg/stack/build/1.0.1868#L10 builds on that. That might be irrelevant though, since that's just a timeout.

sjakobi:
> could you restart the AppVeyor build?

No, I don't have the necessary permissions.

Pinging @snoyberg.

snoyberg:
Rebuilt. In the future, doing `git commit --amend && git push -f` also works well.

Blaisorblade:
It failed again — I started investigating the failure.

sjakobi:
Oh man, these Appveyor log files are absolutely unreadable.

```
[00:18:04] 016-07-18 14:46:38.490247: [debug] 
[00:18:04] Try
[00:18:04] ing 
[00:18:04] t
[00:18:04] o
[00:18:04]  d
[00:18:04] e
[00:18:04] c
[00:18:04] o
[00:18:04] de C:\
[00:18:04] s
[00:18:04] r
[00:18:04] \i
[00:18:04] nd
[00:18:04] i
```

WTF!

Blaisorblade:
@sjakobi Please worry not, I've figured it out, it was my fault and I've got green builds now:
https://ci.appveyor.com/project/Blaisorblade/stack/history
I'm almost done rebasing and squashing for review.

Blaisorblade:
I force-pushed based on this success (https://ci.appveyor.com/project/Blaisorblade/stack/build/1.0.11), got a hung build after rebasing on master but that's hopefully just flaky or not having to do with the new code (I'm trying to rebuild that).

Meanwhile, I think this is ready for review.

src/Stack/Fetch.hs
```diff
@@ -554,6 +533,69 @@ fetchPackages' mdistDir toFetchAll = do
...
 
                 atomically $ modifyTVar outputVar $ Map.insert ident destDir
+            $logWarn $ mconcat $ map (\(path, entryType) -> "Unexpected entry type " <> entryType <> " for entry " <> T.pack path) unexpectedEntries
+
+-- | Internal function used to unpack tarball.
+--
+-- Takes a path to a .tar.gz file, the name of the directory it should contain,
+-- and a destination folder to extract the tarball into. Returns unexpected
+-- entries, as pairs of paths and descriptions.
+untar :: FilePath -> FilePath -> FilePath -> IO [(FilePath, T.Text)]
```
sjakobi: I would prefer to see `Path` being used here as much as possible / reasonable. IMO the distinction between the different `Path` types makes code (especially type signatures) more readable and slightly more robust.

You can find utils for working with `Path` in `src/Path` (especially the amazing `Path.Extra.toFilePathNoTrailingSep` ;)) and in the `path-io` package.

sjakobi:
```
~/tmp $ stack-Blaisorblade unpack ghc-core
Populated index cache.    

Unpacked ghc-core-0.5.6 to /home/simon/tmp/ghc-core-0.5.6/
```

Nice!

Great code, @Blaisorblade, the tests are totally awesome!

This will be ready to merge once my comment above has been addressed!

src/Stack/Fetch.hs
Blaisorblade: I've seen path-io, but this is not trivial since this code has to talk to non-path-ified code.
But you have a point about signatures.

Also, the *existing* code uses FilePath quite a lot, see all the `FP.</>`. I guess you'd be fine with changes to the existing code, both as strictly needed to change `untar`'s signature and as it makes sense? I'll see what I can do later.

Blaisorblade:
> the tests are totally awesome!

Thanks! Happy to hear that, also because I got them very wrong the first time (so I should have asked for a close look there)—I was adding the tests only after all the actions in IO succeeded:
https://github.com/Blaisorblade/stack/commit/bd3436b92ad087b6bc305e15df6be1aa2031642b
I had never used `hspec`, but there's a surprising amount of ways to get those tests subtly wrong.

src/Stack/Fetch.hs
sjakobi: > but this is not trivial since this code has to talk to non-path-ified code.

Yes, I guess for the return value that is only going to be printed it doesn't matter as much. But for the input types it would be great, so we can encourage further usage of `Path` in any callers of `untar`.

> Also, the existing code uses FilePath quite a lot, see all the FP.</>.

Yes, there's a lot of room for improvement in this codebase! ;)

Blaisorblade:
@sjakobi Converting a bit more to `path` was easier than I expected. I ended up removing the import of `System.Directory`, replaced by `path-io`.

mgsloan:
Awesome, LGTM!  Merging

PR #2381, opened 2016-07-17
Various version bound/build-depends fixes

hesiod:
Various version bound fixes, mainly to make stack compile with `lts-2.22`. Please consider the individual commits for details.

### DIFF
--- a/stack.cabal
+++ b/stack.cabal
@@ -196,7 +196,7 @@ library
                    , persistent-sqlite (>= 2.1.4 && < 2.5.0.1) || (> 2.5.0.1 && < 2.6)
                    , persistent-template >= 2.1.1 && < 2.6
                    , pretty >= 1.1.1.1
-                   , process >= 1.2.0.0 && < 1.5
+                   , process >= 1.2.1.0 && < 1.5
                    , regex-applicative-text >=0.1.0.1 && <0.2
                    , resourcet >= 1.1.4.1
                    , retry >= 0.6 && < 0.8
--- a/stack.cabal
+++ b/stack.cabal
@@ -49,6 +49,10 @@ flag static
   -- Not intended for general use. Simply makes it easier to
   -- build a fully static binary on Linux platforms that enable it.
 
+flag use-old-locale
+  default:     False
+  description: Use old-locale instead of time >= 1.5
+
 library
   hs-source-dirs:    src/
   ghc-options:       -Wall -fwarn-tabs -fwarn-incomplete-uni-patterns -fwarn-incomplete-record-updates -fwarn-identities
@@ -206,7 +210,6 @@ library
                    , temporary >= 1.2.0.3
                    , text >= 1.2.0.4
                    , text-binary
-                   , time >= 1.4.2 && < 1.7
                    , tls >= 1.3.8
                    , transformers >= 0.3.0.0 && < 0.6
                    , transformers-base >= 0.4.4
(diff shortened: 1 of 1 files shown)


stack.cabal
```diff
@@ -49,6 +49,10 @@ flag static
   -- Not intended for general use. Simply makes it easier to
   -- build a fully static binary on Linux platforms that enable it.
 
+flag use-old-locale
+  default:     False
+  description: Use old-locale instead of time >= 1.5
```
sjakobi: Sorry again for the confusion in #2377.

There isn't much point in increasing the complexity of the cabal file only to make `stack` compatible with `lts-2` (of course `stack` then still needs _lots_ of extra-deps).

To build stack with ghc-7.8 you can already use `stack install --stack-yaml stack-7.8.yaml`.

stack.cabal
```diff
@@ -196,7 +196,7 @@ library
                    , persistent-sqlite (>= 2.1.4 && < 2.5.0.1) || (> 2.5.0.1 && < 2.6)
                    , persistent-template >= 2.1.1 && < 2.6
                    , pretty >= 1.1.1.1
-                   , process >= 1.2.0.0 && < 1.5
+                   , process >= 1.2.1.0 && < 1.5
```
sjakobi: Thanks for figuring this out, that's useful!

stack.cabal
```diff
@@ -219,7 +219,7 @@ library
                    , vector-binary-instances
                    , yaml >= 0.8.10.1
                    , zlib >= 0.5.4.2 && < 0.7
-                   , deepseq >= 1.4
+                   , deepseq >= 1.3 && < 1.5
```
sjakobi: Seems like there was a reason for the lower bound: 51813ffe (I found this with git blame)

Regarding the upper bound, I don't think we really need it, because `stack` uses `stack sdist --pvp-bounds both` (and hopefully soon the extension discussed in #2262) for publishing, so the releases already have upper bounds.

sjakobi:
I'd like to accept the `process` bounds fix but won't merge the other changes for the reasons I've lined out in my comments above.

stack.cabal
hesiod: I'm not quite sure how `stack-7.8.yaml` does manage to circumvent the dependency trouble regarding `deepseq-1.3`, but somehow it does, so I might as well remove this change.

stack.cabal
hesiod: As said in #2377, I just didn't realize there was a `stack-7.8.yaml`, so this addition is rather useless. I'll remove it in a moment.

hesiod:
Closed in favour of #2383.

stack.cabal
sjakobi: > I'm not quite sure how stack-7.8.yaml does manage to circumvent the dependency trouble regarding deepseq-1.3, but somehow it does

It has an [extra-dep on deepseq-1.4.1.2](https://github.com/commercialhaskell/stack/blob/master/stack-7.8.yaml#L25).

hesiod: Yes, I saw that extra-dep - however, there are packages in lts-2 that have an upper bound of 1.3 on deepseq, so it should be impossible to build stack with lts-2 and deepseq 1.4 at the same time. And stack indeed complains building itself with lts-2.22 and deepseq-1.4.1.2: I created a gist [here](https://gist.github.com/hesiod/4ada2c589b6f6a8613e7e4a14c24a161). In particular, take a look at [lines 465 to 664](https://gist.github.com/hesiod/4ada2c589b6f6a8613e7e4a14c24a161#file-stack-build-output-L465-L664).

sjakobi: I agree that this is a somewhat ugly situation but as long as it compiles it seems to be safe enough.

This [stackoverflow answer](http://stackoverflow.com/questions/9301122/haskell-cabal-package-indirectly-depends-on-multiple-versions-of-the-same-pack) explains it rather well.

PR #2383, opened 2016-07-17
cabal: Require process-1.2.1 or greater

hesiod:
Stack uses createProcess_ which has been added in `process-1.2.1`.
Therefore, the prior bound on process ( `>= 1.2.0 && < 1.5` ) was too lax.


sjakobi:
Cheers!

PR #2385, opened 2016-07-18
Add note on the different Cabal versions that stack uses

sjakobi:


sjakobi:
@hesiod: Do you think this is sufficiently clear?

Issue #2374, opened 2016-07-15
Improve error messages for yaml parse errors


sjakobi:
```
~/t/yaml-error $ cat stack.yaml 
resolver: lts-6.7
packages:
  - location: git: https://github.com/commercialhaskell/stack
~/t/yaml-error $ stack build
Could not parse '/home/simon/test/yaml-error/stack.yaml':
InvalidYaml (Just (YamlParseException {yamlProblem = "mapping values are not allowed in this context", yamlContext = "", yamlProblemMark = YamlMark {yamlIndex = 45, yamlLine = 2, yamlColumn = 17}}))
See http://docs.haskellstack.org/en/stable/yaml_configuration/.
```

At least we could use yaml's `prettyPrintParseException` to format the exception more nicely.

WillSewell:
This can be closed right due to #2388 right?

mgsloan:
I believe so, closing!

PR #2388, opened 2016-07-19
Pretty-print YAML parse errors

sid-kap:
Follow-up to #2374. Unfortunately, the Yaml library's `prettyPrintParseException` doesn't do a great job for the example in #2374:

```
Could not parse '/home/sidharth/sidharthkapur1@gmail.com/workspace/stack-testing/2374/stack.yaml':
YAML parse exception at line 2, column 17,
:
mapping values are not allowed in this context

See http://docs.haskellstack.org/en/stable/yaml_configuration/.
```

The second line of the error contains only a `:` because `yamlContext` is empty in this case (see [code](https://hackage.haskell.org/package/yaml-0.8.18/docs/src/Data-Yaml-Internal.html#prettyPrintParseException)). It would be nice to send them a PR to clean this up. Regardless, I think this is an improvement.

### DIFF
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -180,6 +180,7 @@ import qualified Data.Text as T
 import           Data.Text.Encoding (encodeUtf8, decodeUtf8)
 import           Data.Typeable
 import           Data.Yaml (ParseException)
+import qualified Data.Yaml as Yaml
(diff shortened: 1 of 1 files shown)


sid-kap:
Opened a related PR at http://github.com/snoyberg/yaml#89

sjakobi:
Thanks!

PR #2392, opened 2016-07-19
Travis: download stack for cabal builds

### LINKED ISSUE #2390: Travis build uses stack but doesn't download it on builds
The original build script build doesn't find `stack` when CI-ing my fork:
https://travis-ci.org/Blaisorblade/stack/jobs/145802415#L213-L217

That leads to a build failure because cabal uses the cabal file in master and selects the wrong versions of dependencies—in particular, it picks versions of `binary` and `Cabal` that both declare the same typeclass instance.

Working hypothesis: .travis.yml use stack in cabal builds, but `.travis-setup.sh` doesn't download it for cabal builds.

I conjecture that might work on your fork because there's stack in the build caches—since they cache the result of building stack.
But at some point this might lead to unexpected build configurations and hard-to-debug bugs.

My action plan:
1. I test my hypothesis by trying a fix to the build scripts
2. Ideally, somebody (I, if I manage) change the build to avoid storing the final binary.
3. If that works, you might want to verify my conjecture by clearing the Travis cache once and observing the build failure—or not, it's of course your call.

Blaisorblade:
Fix #2390:
- download stack before using it;
- error out if the uses of stack fails;
- pass enough options to stack to make it work without requiring `stack setup` (specifically, pick correct stack file and GHC version).

This still does not ensure a successful build from scratch on GHC-7.8.4, but it still fixes a clear bug.

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -15,15 +15,18 @@ fetch_stack_linux() {
   curl -sL https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack';
 }
 
+# We need stack to generate cabal files with precise bounds, even for cabal
+# builds.
+mkdir -p ~/.local/bin;
+if [ `uname` = "Darwin" ]; then
+  travis_retry fetch_stack_osx
+else
+  travis_retry fetch_stack_linux
+fi
+
 case "$BUILD" in
   stack)
-    mkdir -p ~/.local/bin;
-    if [ `uname` = "Darwin" ]; then
-      travis_retry fetch_stack_osx
-    else
-      travis_retry fetch_stack_linux
-    fi;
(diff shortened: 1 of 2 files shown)


mgsloan:
Thanks!  Looks like there's a CI error.

Blaisorblade:
@mgsloan I'm afraid most of the context is in #2390, hence I'm answering there—sorry this is awkward.

sjakobi:
These changes make a lot of sense regardless of whether https://github.com/commercialhaskell/stack/pull/2273 happens or not.

Please just rebase this if necessary once #2273 has been decided.

The `lifted-base` bound has been fixed on `master`.

borsboom:
These changes look good to me.  Go ahead and merge it after any necessary rebases and CI error fixes.

mgsloan:
LGTM!

PR #2394, opened 2016-07-19
stack sdist: Fix timestamp in tarball

Blaisorblade:
Here tar (the tool) [warns on Travis](https://travis-ci.org/commercialhaskell/stack/jobs/145921846#L254) that a time stamp in a tarball generated by stack sdist is very odd:

> tar: stack.cabal: implausibly old time stamp 1970-01-01 00:00:00

The warning *is* harmless, but it might still puzzle users (e.g. http://raspberrypi.stackexchange.com/a/13152/4395 in other scenarios), so here's a 2-line fix.


mgsloan:
LGTM, thanks!

PR #2403, opened 2016-07-22
Add FAQ entry regarding system library installation (#1384)

### LINKED ISSUE #1384: FAQ: automatic installation of extra-libraries
I scanned the FAQ and GUIDE but could not find an answer to this question: Does `stack` also install C libraries for me? There are several packages like `hmatrix` or `cuda` that depend on external C libraries. Could you please add an item to the FAQ, telling whether `stack` can install external C libraries or whether it is intended to support that in the future? Such support would mean that different Linux distributions, Mac and Windows would need individual information on how to install these libraries.

sjakobi:
Closes #1384.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -331,3 +331,7 @@ The workaround is to quote the `"--"`, e.g.:
     stack exec "--" cabal --version
 
 This is known to be a problem on Windows 7, but seems to be fixed on Windows 10.
+
+# Does stack also install the system/C libraries that some Cabal packages depend on?
+
+No, this is out of scope.


borsboom:
A few things this could mention:

* The Nix and Docker integration can be ways around this, where your `stack.yaml` can specify Nix packages containing system libraries or a Docker image that has them.

* Using `extra-include-dirs` and `extra-lib-dirs` in your `~/.stack/config.yaml` or project `stack.yaml` if you've installed system libraries in a non-standard location.

sjakobi:
> A couple of things this could mention:

I agree. I'll fix this when I'm back from my vacation.

mgsloan:
Thanks!  I added some more info in https://github.com/commercialhaskell/stack/commit/59c3db29764de5c7065f8c4332b45e5c02cca340

PR #2437, opened 2016-08-04
Delete Stack.Types to speedup builds.

### LINKED ISSUE #2405: Delete Stack.Types to improve build times
(Split off from #2334)

> Modules would then only depend on the Stack.Types.* modules that they actually need and e.g. Stack.Clean wouldn't have to be recompiled after a change in Stack.Types.Nix.

Blaisorblade:
Fix #2405.
The version bump might be controversial:
https://github.com/commercialhaskell/stack/issues/2405#issuecomment-237526090
> According to the PVP, this would forces a version bump to 1.2. Should I readd Stack.Types in or bump the version?


Blaisorblade:
Just confirmed stack officially ignores the PVP and documents that very loudly, so I'm happy to retract the version bump:

http://hackage.haskell.org/package/stack
> Also, note that the API for the library is not currently stable, and may change significantly, even between minor releases. It is currently only intended for use by the executable.

mgsloan:
LGTM, thanks!

Blaisorblade:
@mgsloan Whoops, I accidentally left in the version bump of `stack` to 1.2.0! I'm guessing that should be reverted, right? Sorry.

borsboom:
Yes, please revert it.  Also, for future reference, the master branch's version should always end in an odd number to make clear it's not a real release.

PR #2440, opened 2016-08-04
Travis: Propagate failures in Cabal builds

### LINKED ISSUE #2439: Travis doesn't *really* test cabal builds.
Or rather, it tests them but ignores the errors. See the failures here:
https://travis-ci.org/commercialhaskell/stack/jobs/149459127
https://travis-ci.org/commercialhaskell/stack/jobs/149736106#L435

Those ones are just harmless warnings turned into errors by `-Werror`.

That's because the travis script composes commands with `;` instead of `&&` in the `script` section. One could fix that also with `set -e`.

However, the stack build also gives similar warnings, but without `-Werror`:
https://travis-ci.org/commercialhaskell/stack/jobs/149459130

That's because of `cabal configure --enable-tests --enable-benchmarks -v2 --ghc-options="-O0 -Werror"` in .travis.yml picks -Werror for cabal, but not for Travis.
My plan:
- [x] add `-Werror` to stack builds
- [x] fix the build script to propagate failures
- [x] create an issue about those warnings and/or investigating them myself. In particular, the fix might not need conditional compilation: the warning [did not exist before 8.0](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html#ghc-flag--Wredundant-constraints), so hopefully the constraints are also redundant on previous versions.

Blaisorblade:
Fix #2439. As mentioned there:

-  fix warnings from -Wredundant-constraints on GHC 8
-  fix the build script to propagate failures
-  add -Werror also to stack builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ install:
    esac
 
 script:
- - case "$BUILD" in
+ - set -e; case "$BUILD" in
      stack)
        stack --no-terminal test --haddock --no-haddock-deps;;
      cabal)
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ install:
 script:
  - set -e; case "$BUILD" in
      stack)
-       stack --no-terminal test --haddock --no-haddock-deps;;
+       stack --no-terminal test --haddock --no-haddock-deps --ghc-options="-Werror";;
      cabal)
        cabal configure --enable-tests --enable-benchmarks -v2 --ghc-options="-O0 -Werror";
        cabal build;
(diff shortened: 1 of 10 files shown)


borsboom:
LGTM

Blaisorblade:
@borsboom Is https://github.com/commercialhaskell/stack/pull/2440/commits/6fdea5b61b8371ecb859c8fe70d3679a27ad2208 fine? That's less obvious to me.

borsboom:
I'm fine with requiring the newer version of hspec.  We used to really try to stay within the LTS snapshot but aren't so worried about that anymore.

mgsloan:
LGTM too!

Issue #2418, opened 2016-07-28
Strange leading newlines when logging build progress


afcowie:
For some reason, every time a configure step is logged, [13 spaces (?) and a] newline is preceding it. Was that on purpose?

```
...
              
quickcheck-io-0.1.3: configure
quickcheck-io-0.1.3: build
quickcheck-io-0.1.3: copy/register
               
hspec-core-2.2.3: configure
hspec-core-2.2.3: build
zlib-0.6.1.1: copy/register
               
zlib-bindings-0.1.1.5: configure
zlib-bindings-0.1.1.5: build
zlib-bindings-0.1.1.5: copy/register
hspec-core-2.2.3: copy/register
               
hspec-2.2.3: configure
hspec-2.2.3: build
hspec-2.2.3: copy/register
...
```

Not a problem, of course, just a surprise.

afcowie:
Employed `git bisect` and determined breaking change introduced in 6af46db4aa8ddb13d27bf2d78d90c6e41d7159db.

afcowie:
Referenced line has a 
```haskell
 $logWarn $ mconcat $ map (\(path, entryType) -> "Unexpected entry type " <> ...
```
thing going on. I'm guessing the result of there _not_ being an "unexpected" condition is `$logWarn` being called with an empty string, resulting in our spurious newlines.

PR #2445, opened 2016-08-05
Only call logWarn if there's anything to report

afcowie:
As reported in #2418, spurious newlines are being emitted. Doing a `git bisect` reduced it to a commit which introduced a `$logWarn` message which seemed responsible. This patch simply wraps the debug statements so that no call is made it the error report is the empty list.

(I leave aside the matter of whether this was a good way to do error reporting; I'm just trying to fix the output bug)

AfC

### DIFF
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -531,7 +531,9 @@ fetchPackages' mdistDir toFetchAll = do
                 S.writeFile cabalFP $ tfCabal toFetch
 
                 atomically $ modifyTVar outputVar $ Map.insert ident destDir
-            $logWarn $ mconcat $ map (\(path, entryType) -> "Unexpected entry type " <> entryType <> " for entry " <> T.pack path) unexpectedEntries
+
+            F.forM_ unexpectedEntries $ \(path, entryType) ->
+                logWarnN $ "Unexpected entry type " <> entryType <> " for entry " <> T.pack path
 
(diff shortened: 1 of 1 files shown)


src/Stack/Fetch.hs
```diff
@@ -531,7 +531,9 @@ fetchPackages' mdistDir toFetchAll = do
                 S.writeFile cabalFP $ tfCabal toFetch
 
                 atomically $ modifyTVar outputVar $ Map.insert ident destDir
-            $logWarn $ mconcat $ map (\(path, entryType) -> "Unexpected entry type " <> entryType <> " for entry " <> T.pack path) unexpectedEntries
+
+            F.forM_ unexpectedEntries $ \(path, entryType) ->
+                logWarnN $ "Unexpected entry type " <> entryType <> " for entry " <> T.pack path
```
Blaisorblade: FYI, stack never uses `logWarnN`, since it doesn't record source file and position. Otherwise LGTM.

afcowie: Oh. I didn't know if template haskell would work there. Fixed.

Blaisorblade:
LGTM, I'll merge as soon as tests pass again.

> (I leave aside the matter of whether this was a good way to do error reporting; I'm just trying to fix the output bug)

Good question. I used `$logWarn` just because it seems to be the standard here.
Sorry for the bug and thanks for fixing it so quickly!

PR #2451, opened 2016-08-06
[DOCUMENTATION] Request to emphasize `stack list-dependencies` in user guide

pellagic-puffbomb:
Hello,

I find that I'm often installing stuff and then later on I want to find out exactly which versions I installed (especially switching between projects created at different points in time). To do this, I rely on the command `stack list-dependencies`, but I always forget the command and have to look it up: is it `list-dependencies` or `show-dependencies`? Hyphen or no hyphen? Isn't it something simpler than `list-dependencies` (in the Python world, I'd use `pip freeze` or `conda env export`.)

To answer my question, there's the power-user section at the [bottom of the user guide](https://docs.haskellstack.org/en/stable/GUIDE/#power-user-commands) and it usually takes me a few minutes to find the command because it's buried there.

Anyway, it occurred to me that other people might also be wondering how to do this and highlighting this command might help them out and make it easier to remember. I also tend to think of this as a core function, not necessarily something only "power users" may want to use, but I may be generalizing broadly based only on my experience.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -457,9 +457,6 @@ library
                      , containers
 ```
 
-Often it's a good idea to specify exact versions of dependencies, but we didn't do that here: stack was able to
-find versions that are compatible with our project.
-
 After adding these two dependencies, we can again run `stack build` to have them installed:
 
 ```
@@ -486,9 +483,6 @@ helloworld 0.1.0.0
 integer-gmp 1.0.0.0
 text 1.2.2.1
 ```
-
-As our project grows in size, it may prove useful to have a comprehensive list of the packages and versions we are relying on.
-
 
 ### extra-deps
 
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -441,6 +441,55 @@ locally installed. Once that was done, we moved on to building our local package
 (helloworld). At no point did we need to ask stack to build dependencies — it
 does so automatically.
 
+### Listing Dependencies
+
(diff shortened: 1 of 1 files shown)


borsboom:
Just reading over the comment, it's possible you're misunderstanding Stack's approach a bit.  Stack basically uses in an "always frozen" approach and unless you explicitly using the `stack solver` or `stack init --solver` command it is never "figuring out"  package versions to use.  The `stack.yaml` always contains all the information to exactly determine which package versions will be used, so that makes a command like `list-dependencies` less crucial.  It's still a useful command, and I think making it more obvious in the guide is a good thing, but it may be good to adjust the wording to favour Stack's approach to versioning.  In particular, when using Stack we do not especially recommend setting version ranges in your `.cabal` file unless you're planning to distribute your package on Hackage (and, even then, using `stack upload --pvp-bounds` often makes it unnecessary to have them in your source .cabal file).

pellagic-puffbomb:
Those are fair points and it's likely I am misunderstanding stack's approach. This is why I wasn't sure if my perspective would be representative other users.

I can say that I tend to use this command when I'd like to look up documentation on a library and I'm not sure which one I'm have installed. 

I have edited my version to be much simpler, only highlighting that the command exists in the same location. If this doesn't seem like a useful contribution, I will withdraw the PR. 

On the other hand, if it does seem useful and worthy of inclusion, I'd probably also consider removing the mention of `list-dependencies` from the "power user" section down below.

borsboom:
Looks good, thanks!

PR #2457, opened 2016-08-08
Adding support for rendering GHCi scripts targeting different GHCi like applications.

AndrewRademacher:
This PR adds the capacity to StackGhci to render GHCi scripts for different targets like Intero. Or for differing versions of GHCi. Additionally, this takes advantage of setting the current working directory in Intero when available.

This is all aimed at supporting the loading of multiple packages simultaneously in a REPL.


mgsloan:
LGTM!

Blaisorblade:
Thanks for the PR, but src/test/Stack/Ghci/ScriptSpec.hs and src/test/Stack/GhciSpec.hs fail on Windows by introducing failures in Template Haskell code.

mgsloan:
From https://ci.appveyor.com/project/snoyberg/stack/build/1.0.2040, looks like

```
C:\stack\src\test\Stack\Ghci\ScriptSpec.hs:23:35:
    Exception when trying to run compile-time code:
      InvalidAbsDir "/src/package-a"
    Code: mkAbsDir "/src/package-a"
    In the splice: $(mkAbsDir "/src/package-a")
```

Surprising to me, considering that msys and such pretends to have a `/..` paths.

Blaisorblade:
@mgsloan I understand I have to use `</>` for work on WIndows. Some Haskell packages like to convert path separators to native ones, adding extra failures; I've seen that with `tar`, this might be another case.

@AndrewRademacher 
Reproducing it without Windows is hard (lest you use a Windows Virtual Machine); it might be easier to have a PR and wait for AppVeyor CI's results.

To fix it, I'd imagine one would to replace all "/" with </>, both in the paths (to fix compilation) and in the test results, make it compile again on your platform, and hope it still works on the CI. I understand this is annoying... maybe @mgsloan has a Windows machine? :-(

And, sorry for asking, but are you familiar with AppVeyor and the red cross in GitHub's UI next to your commits in this PR? Clicking on it leads to the logs from continuous integrations (that is, build servers); Appveyor tests on Windows, Travis on other OSs (and @mgsloan already fixed the Travis failures).

mgsloan:
Yeah, my mistake, should've gotten all CI results before merging.

AndrewRademacher:
@Blaisorblade, I actually didn't think that this error happened in the CI. @mgsloan and I had a conversation earlier about the errors that I had produced, so I falsely assumed that the error you were talking about was in addition to what happened in CI.

On the bright side, I have boot camped my Mac. So I can reproduce on my windows machine.

Blaisorblade:
@AndrewRademacher Ah that explains it. Please take no offense at the useless explanation, I should have just mentioned CI in my message!

Blaisorblade:
@AndrewRademacher Any progress? I'm working on this right now.

I've just learned that mkAbs* fails because "/a/b" is recognized as a relative path on Windows:
```
Prelude System.FilePath> isRelative "\a\b"
True
```
/cc @mgsloan.

AndrewRademacher:
@Blaisorblade, looks like I have a lot of work to do this weekend.

Blaisorblade:
@AndrewRademacher That's OK; I've now fixed the paths problem. (Actually I should have asked before starting, sorry about that; I inferred you lacked the time).
I'm dealing with Windows newlines now (working around https://github.com/nikita-volkov/neat-interpolation/issues/14).

Issue #2450, opened 2016-08-06
Have "stack --version" 's list of dependency versions include transitive deps


mgsloan:
The implementation in https://github.com/commercialhaskell/stack/pull/2223 only lists direct deps.  While not super likely, variations in the indirect deps chosen by cabal-install could also cause unexpected and unpredictable issues for users when it's used to build stack.

It would also be really nice to go back to a simple build type.

Blaisorblade:
I think I managed to avoid transitive dependencies! As an example, look for `zlib-bindings`. The code is even less hacky than I feared, but given the process I used, I won't remotely suggest the code is robust.
I'll take a look at using Template Haskell next.

```
$ stack exec -- stack --version
Version 1.1.3, Git revision 241df53b5c6ff9cc1d283710e69e740085d90482 (dirty) (3918 commits) x86_64
Compiled with:
- Cabal-1.22.8.0
- Glob-0.7.5
- HUnit-1.3.1.1
- MonadRandom-0.4.2.3
- QuickCheck-2.8.2
- SHA-1.6.4.2
- StateVar-1.1.0.4
- adjunctions-4.3
- aeson-0.11.2.0
- aeson-compat-0.3.3.0
- annotated-wl-pprint-0.7.0
- ansi-terminal-0.6.2.3
- ansi-wl-pprint-0.6.7.3
- array-0.5.1.0
- asn1-encoding-0.9.3
- asn1-parse-0.9.4
- asn1-types-0.3.2
- async-2.1.0
- attoparsec-0.13.0.2
- auto-update-0.1.4
- base-4.8.2.0
- base-compat-0.9.1
- base-orphans-0.5.4
- base16-bytestring-0.1.1.6
- base64-bytestring-1.0.0.1
- bifunctors-5.2
- binary-0.7.5.0
- binary-tagged-0.1.4.0
- blaze-builder-0.4.0.2
- blaze-html-0.8.1.1
- blaze-markup-0.7.0.3
- byteable-0.1.1
- bytestring-0.10.6.0
- bytestring-builder-0.10.6.0.0
- case-insensitive-1.2.0.6
- cereal-0.5.1.0
- clock-0.7.2
- comonad-4.2.7.2
- conduit-1.2.6.6
- conduit-extra-1.1.13.1
- connection-0.2.5
- constraints-0.8
- containers-0.5.6.2
- contravariant-1.4
- cookie-0.4.2
- cryptohash-0.11.9
[...]

PR #2458, opened 2016-08-09
List transitive deps in --version (#2450)

Blaisorblade:
As requested in #2450.
This PR does not switch to Template Haskell yet, that should be a separate PR.


Blaisorblade:
BTW, I can squash commits if you care. People have strong opinions on this ;-).

mgsloan:
LGTM

> People have strong opinions on this ;-).

As long as there aren't zillions of commits, fine with me!

PR #2471, opened 2016-08-10
Fix bulleted list presentation in the FAQ.

dkasak:
@mgsloan It seems `*` is rendered as expected after all, there just needs to be a blank line before the start of the list.


mgsloan:
Thanks!

PR #2477, opened 2016-08-11
Remove the `--os` flag.

### LINKED ISSUE #2227: Remove the "--os" flag
I was investigating how to build for production (I'm on a Mac, I want to build for Linux), and noticed that there's a flag:

```
stack - The Haskell Tool Stack

Available options:
  --os OS                  Operating system, e.g. linux, windows
```

I cannot find any documentation on this. If it's available I'd appreciate a nudge in the right direction, but otherwise, some documentation for the flag would be appreciated.

I understood the flag as being the target OS for builds, but when passing "--os linux" in a build on a Mac, I get a Mach-O binary, not a linux compatible binary.

0xmohit:
Fixes #2227

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -42,6 +42,8 @@ Bug fixes:
   [#2361](https://github.com/commercialhaskell/stack/issues/2361)
 * Fixes src directory pollution of `style.css` and `highlight.js` with GHC 8's
   haddock [#2429](https://github.com/commercialhaskell/stack/issues/2429)
+* Remove the `--os` flag
+  [#2227](https://github.com/commercialhaskell/stack/issues/2227)
 
 ## 1.1.2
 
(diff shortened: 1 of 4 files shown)


sjakobi:
Thanks for working on this!

Can you add an entry to the changelog and check if there's any documentation that needs to be changed or removed?

0xmohit:
Added a ChangeLog entry (wasn't familiar with it's existence).  Will look into the docs to determine if anything else needs to change.

sjakobi:
No worries about the AppVeyor failure BTW, that's caused by https://github.com/commercialhaskell/stack/pull/2457.

0xmohit:
Unable to find any reference to `--os` in the [docs](https://docs.haskellstack.org), so I suspect there's nothing else to change here.

PR #2480, opened 2016-08-12
Interpreter: Provide a way to hide implicit packages

### LINKED ISSUE #1208: interpreter: Don't make cached packages available that are not specified with `--package`

harendra-kumar:
Always pass all the explicitly specified packages in the script to ghc command
line. Also allow ghc options to be specified from the interpreter command. This
will give user the flixibility to, for example, use '-hide-all-packages' ghc
option. Like this:

{- stack runghc
    --package base
    --package exceptions
    --package transformers
    --package containers
    --package getopt-generics
    --package filepath
    --package path
    --package path-io
    --package process
    --package unix-compat
    --
    -hide-all-packages
-}

fixes #1208

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -545,7 +545,8 @@ interpreterHandler args f = do
       progName <- getProgName
       iargs <- getInterpreterArgs path
       let parseCmdLine = commandLineHandler progName True
-      let cmdArgs = stackArgs ++ iargs ++ "--" : path : fileArgs
+          separator = if "--" `elem` iargs then [] else ["--"]
+          cmdArgs = stackArgs ++ iargs ++ separator ++ path : fileArgs
        -- TODO show the command in verbose mode
        -- hPutStrLn stderr $ unwords $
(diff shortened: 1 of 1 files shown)


harendra-kumar:
Windows build failing due to #2457 fixed by #2482 . I will rebase and try again.

Blaisorblade:
LGTM!

Blaisorblade:
@harendra-kumar thanks for this. Maybe you can add a mention for this in the docs?

PR #2482, opened 2016-08-12
Fix Windows build

Blaisorblade:
Fix the regressions from #2457.


Blaisorblade:
Merged. I'd still welcome a code review, as there are quite a few non-obvious choices involved. But I'm uneasy with a broken build & rank fixing the build higher, lest other regressions slip in.

mgsloan:
LGTM, thanks for fixing!

PR #2484, opened 2016-08-13
simplify is_64_bit

0xmohit:
`test` (or `[`) would return 0 upon success and non-zero upon failure.
As such, explicit return statements are not required.


Blaisorblade:
Thanks! We indeed only care for success or failure, not exact exit codes, so LGTM!

PR #2485, opened 2016-08-13
Document config independent and reliable scripts

harendra-kumar:


doc/GUIDE.md
```diff
@@ -1712,8 +1717,43 @@ of a multi line block comment:
...
+
+Independent means that the script is independent of any prior deployment
+specific configuration. If required, the script will install everything it
+needs automatically on any machine that it runs on. To make a script always
+work irrespective of any specific environment configuration you can do the
+following:
+1. Use the `--install-ghc` option to install the compiler automatically
+2. Explicitly specify all packages required by the script using the
+`--package` option. Use `-hide-all-packages` ghc option to force
+explicit specification of all packages.
```
sjakobi: The Github Markdown renderer doesn't recognize this as a listing (and I believe `readthedocs` will behave the same).

I suspect a blank line is needed above the first point. Possibly the two lines belonging to `2.` need to be indented too, but I'm not sure.

sjakobi:
LGTM apart from my comment above.

Very nice solution of the original feature request, too! :+1:

doc/GUIDE.md
harendra-kumar: Ah, yes. Thanks! I faced this before but keep forgetting because my markdown renderer shows it fine.

Blaisorblade:
This PR makes sense to me too...

*However* I have one concern on the use case: *should* we have a good story for shebang versioning? New scripts won't run well with older stack versions.

Should we have some kind of `--require-stack 1.1.4` option (meaning, require a stack older than 1.1.4, or abort/upgrade)? This seems too much complexity, but it could be required for the strict guarantees @harendra-kumar seems to be looking for.

In most cases, one should just install the newest stack's release and be done. At some point you should switch to stack.yaml and a cabal config (or an hpack config, for that matter)—and stack.yaml does support `require-stack-version` (https://docs.haskellstack.org/en/latest/yaml_configuration/#require-stack-version).
(BTW, why are we porting stack.yaml options to the command line one by one? I guess there's an issue on it).

Blaisorblade:
Meanwhile, merging on account of @sjakobi's review—his comment appears addressed.

Issue #2489, opened 2016-08-14
Some execution functions lookup files incorrectly


Blaisorblade:
To reproduce:
1. Add to the PATH a non-executable `git` file
2. Watch stack try and fail to execute that file, if needed when fetching.

```
$ > ~/.local/bin/git
$ ls -l ~/.local/bin/git
-rw-r--r--+ 1 pgiarrusso staff 0 Aug 14 11:22 /Users/pgiarrusso/.local/bin/git
$ stack build --dry-run
/Users/pgiarrusso/.local/bin/git: streamingProcess: runInteractiveProcess: exec: permission denied (Permission denied)
$ stack upgrade
Fetching package index .../Users/pgiarrusso/.local/bin/git: createProcess: runInteractiveProcess: exec: permission denied (Permission denied)
```

For the failure to occur on build, the package needs git dependencies, for instance:
```
resolver: lts-6.6

packages:
- '.'
- location:
    git: https://github.com/agda/agda
    commit: e3472d7c1441c0d6db66be9e094e9dbcb7c33735
```

I noticed this by chance while debugging #2241. I can't remove Git from OS X 10.11, but I tried shadowing it with a failing one and made a mistake.

The bug is thankfully not in the `process` package. Nor is `stack` itself doing the lookup—the bug is somewhere between `System.Process.Read` and its supporting libraries. BTW, `System.Process.Read` appears confusing to me.

```
$ stack runghc ./runProcess.hs
git version 2.9.0
First git done
git version 2.9.0
```

`runProcess.hs`:
```haskell
import System.IO
import System.Process

args = ["--version"]
main = do
  (_, _, _, ph) <- createProcess (proc "git" args)
  ec <- waitForProcess ph
  hPutStrLn stderr "First git done"
  ph <- runProcess "git" args Nothing Nothing Nothing Nothing Nothing
  ec <- waitForProcess ph
  return ()
```

Blaisorblade:
BTW, all that was on OS X, I haven't rechecked on Linux yet.

Blaisorblade:
Ah, I had missed stack's `findExecutable`, the bug must be there!

Issue #2241, opened 2016-06-04
stack build should check whether git is available


jwaldmann:
I was running `stack build` on a (virtual) machine, with some github locations in stack.yaml. I got errors like
```
 .stack-work/downloaded/Lcy0nQgWISfb/: getDirectoryContents: does not exist (No such file or directory)
```
The reason is that my machine did not have `git` installed. 

Looks like `stack` called `git`, but did not check the exit (error) code.

sjakobi:
I'm under the impression that there were more problems here:

1. Distro packages of stack ought to have a package dependency on git. Did you install stack through a package manager, @jwaldmann, and what operating system were you running on that VM?

2. stack ought to have failed already when the `git clone` command failed.

jwaldmann:
This is on a gitlab builder, usind a docker image for haskell-7.10.3. Here is a typical build that works, but still produces a lot of noise: https://gitlab.imn.htwk-leipzig.de/waldmann/pure-matchbox/builds/931

Blaisorblade:
We should probably move `checkDependency` stuff from src/Stack/Setup.hs and use it for this.

Blaisorblade:
Alternatively, there's `isGitInstalled` in `Stack.PackageIndex`, some consolidation might be in order.

Blaisorblade:
So... at least on OS X, this seems to also relate to `System.Process.Read` and its different ways of launching processes and handling exceptions. Also, this is *not* about the exit code—that's handled correctly. The problem is specific to git not existing. There's some unclear overlap with #2489.

BTW, using `isGitInstalled` *would* probably hide the bug, but I'd rather learn where the failure is being ignored and fix that.

Experiment:
```
$ cat > ~/.local/bin/git << EOF
> #!/bin/sh
> exit 127
> EOF
$ chmod +x ~/.local/bin/git
$ stack build --dry-run
Running /Users/pgiarrusso/.local/bin/git clone --recursive https://github.com/agda/agda /Users/pgiarrusso/AeroFS/Repos/ilc-agda-bluevelvet/.stack-work/downloaded/kZ1ACjtmNMB9 in directory /Users/pgiarrusso/AeroFS/Repos/ilc-agda-bluevelvet/.stack-work/downloaded/ exited with ExitFailure 127


```

A problem is, in some functions different exceptions are thrown (unlike documented) for when the executable is missing and for when launching it fails.

Blaisorblade:
[...]

PR #2490, opened 2016-08-14
Cleanup Exception docs

Blaisorblade:
Branch born in response to #2241 and #2489.

I added docs to chase down a swallowed exception that I thought caused #2241, while the bug was simply fixed in the sources I looked at. Yet, the extra docs are hopefully useful.

Some of the inconsistencies I found should maybe be fixed by changing behaviors rather than docs, but I'm not 100% sure how to proceed. I'd guess we could want a single exception everywhere for simplicity, in particular Stack's `ReadProcessException` (since it can express more failures than the alternative `ProcessExitedUnsuccessfully`). But overall I'm not 100% sure these files have a coherent design, nor is it easy to find one (see also: `Stack/Exec.hs`).

### DIFF
--- a/src/System/Process/Read.hs
+++ b/src/System/Process/Read.hs
@@ -142,6 +142,8 @@ envHelper :: EnvOverride -> Maybe [(String, String)]
 envHelper = Just . eoStringList
 
 -- | Read from the process, ignoring any output.
+--
+-- Throws a 'ReadProcessException' exception if the process fails.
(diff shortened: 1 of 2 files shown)


Blaisorblade:
I've cherry-picked the fix to #2489 onto `master` (it's now 633ac50f03586326ec8f70a2a2c34b9e167f4bf0). This PR now contains only comments to review, which can wait more easily.

Blaisorblade:
For the record, the CI failure is spurious (master was broken at the time).

Issue #2491, opened 2016-08-14
Yaml file not found when Windows username contains non-ASCII characters


Forkest:
I'm encountering the same issue as described [here](https://stackoverflow.com/questions/34905113/how-to-change-the-default-location-of-stack-yaml-on-windows).

Almost any stack command (except of `stack --version`) results in an output like this (user René here):
```
Could not parse 'C:\Users\Ren├®\AppData\Roaming\stack\config.yaml':
InvalidYaml (Just (YamlException "Yaml file not found: C:\\Users\\Ren\233\\AppData\\Roaming\\stack\\config.yaml"))
See http://docs.haskellstack.org/en/stable/yaml_configuration.html.
```

The file is recreated successfully when I delete it, so in theory it should be possible to read it too.

Blaisorblade:
First, thanks for the report! Can you run a failing command with `-v` and post the complete log?

Then, have you tried the final tip on STACK_ROOT, as a workaround?
https://stackoverflow.com/questions/34905113/how-to-change-the-default-location-of-stack-yaml-on-windows#comment57558955_34905113

Blaisorblade:
Already minimized to a bug in the Yaml library.

`René/foo.yaml`
```
true
...
```

`testYaml.hs`:
```
import qualified Data.Yaml as Yaml

loadYaml :: FilePath -> IO (Either Yaml.ParseException Bool)
loadYaml = Yaml.decodeFileEither

main :: IO ()
main = do
  print =<< loadYaml "René\\foo.yaml"
  return ()
```

```
$ stack runhaskell testYaml.hs
Left (InvalidYaml (Just (YamlException "Yaml file not found: Ren\233\\foo.yaml")))
```

What I've shown is on a fresh Win10 VM, inside`Git Bash` coming from Git for Windows.

Blaisorblade:
Oh dear. `Data.Yaml` is a wrapper over a C library, but the issue is probably in its FFI code. I've filed an analysis over there. It doesn't appear fixed in later versions, but I haven't tried.

I think the easiest workaround doable inside stack is to read the file in Haskell into a bytestring and call — even if this changes the space complexity, `stack.yaml` files are probably usually small enough that this doesn't matter.

Blaisorblade:
Re workaround: since Snoyberg is in vacation, I guess we shouldn't expect a quickfix upstream. https://twitter.com/snoyberg/status/764514619822051328

Blaisorblade:
[...]

PR #2492, opened 2016-08-14
Workaround #2491 at all call sites

Blaisorblade:


Blaisorblade:
Investigating the failure.
Also, if merged, https://github.com/snoyberg/yaml/pull/91 would make this unnecessary.

Blaisorblade:
The AppVeyor failure appears spurious, so I've rebased and force-pushed to trigger a rebuild.

sjakobi:
Very clean workaround! :+1:

I defer merging this to see whether the AppVeyor failure reappears.

Blaisorblade:
Thanks!
But https://jenkins-public.fpcomplete.com/job/stack-integration-tests/177/console and https://jenkins-public.fpcomplete.com/job/stack-integration-tests/178/changes suggests something *is* wrong with these changes.
My 32bit Windows build (on 64bit Windows) keeps passing normal tests. I'm retesting with 64-bit Stack on the same Windows.

Blaisorblade:
It seems that these functions sometimes don't create YAML files when they should... and on Windows (64bit) that affects even the basic testsuite and not just the integration tests.

```
Failures:

  src/test\Stack\BuildPlanSpec.hs:55:
  1) Stack.BuildPlan finds missing transitive dependencies #159
       uncaught exception: IOException of type NoSuchThing (C:\sr\build-plan\lts-2.9.yaml: openFile: does not exist (No such file or directory))

  src/test\Stack\ConfigSpec.hs:127:
  2) Stack.Config.loadConfig finds the config file in a parent directory
       uncaught exception: IOException of type NoSuchThing (C:\sr\build-plan\lts-2.10.yaml: openFile: does not exist (No such file or directory))

  src/test\Stack\ConfigSpec.hs:138:
  3) Stack.Config.loadConfig respects the STACK_YAML env variable
       uncaught exception: IOException of type NoSuchThing (C:\sr\build-plan\lts-2.10.yaml: openFile: does not exist (No such file or directory))

Randomized with seed 690703526

Finished in 7.4692 seconds
684 examples, 3 failures
```

But the obvious basics work even on Windows—the snippet below creates a file with the expected content. While `C:\sr\build-plan\lts-2.10.yaml` indeed doesn't exist, though the directory does.

Frankly I'm out of ideas, other than closing this and waiting for the fix in `yaml`.

```
>stack exec ghci
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help
Prelude> import qualified Data.ByteString as B
Prelude B> import System.IO
Prelude B System.IO> withFile  "a.b" WriteMode $ \hnd -> B.hPut hnd (B.pack [1,2])

Prelude B System.IO>
Leaving GHCi.
```

Blaisorblade:
Ah, my decode wrapper reports file not found as the wrong exception type! Apparently those files are fetched when decode fails. Logs show stack trying to read those missing files, before they're written to. Testing this hypothesis.

src/Data/Yaml/Extra.hs
```diff
@@ -1,17 +1,21 @@
+{-# LANGUAGE ScopedTypeVariables #-}
 -- | Wrappers for Yaml functions to workaround
 -- https://github.com/commercialhaskell/stack/issues/2491.
 -- Import Data.Yaml.Extra in place of Data.Yaml to use this workaround.
 -- Beware these functions construct/deconstruct the entire file at once!
 module Data.Yaml.Extra (decodeFileEither, encodeFile, module Data.Yaml) where
 
+import           Control.Exception
```
Blaisorblade: Please double-check this — I think it's OK (this is a straightforward synchronous exception, handled in the IO monad) but I'm not yet fluent with Haskell exceptions.

Blaisorblade:
In local testing, this commit appears to help a bit... going to sleep now, we'll see tomorrow.

Blaisorblade:
All tests fixed (also integration: https://jenkins-public.fpcomplete.com/job/stack-integration-tests/179/changes). And the use of exceptions seems reasonable. @sjakobi, time to merge?

src/Data/Yaml/Extra.hs
sjakobi: I don't have much experience with Haskell exceptions either. Pinging @mgsloan.

Blaisorblade: Since he seems busy, I'll trust my (more educated) judgement, assume this is fine, and merge this. Details below but feel free to skip.

Based on https://github.com/commercialhaskell/haskelldocumentation/blob/master/content/exceptions-best-practices.md, having read some more docs (including for `MonadCatch`, which doesn't apply here), having read the papers on Haskell exceptions (including async ones, also not applying here*), I think this is fine.

*We risk catching any async `IOException`, and I understand that's bad. But `IOException` shouldn't be used for asynchronous exceptions, and even `safe-exceptions` won't help there.
https://www.fpcomplete.com/blog/2016/06/announce-safe-exceptions

PR #2497, opened 2016-08-16
Use package-id arg to stack ghc/runghc instead of package name

### LINKED ISSUE #1957: Running Haskell script with Stack uses multiple versions of one package
Hello, there

In Snowdrift, we have a database setup script called `sdb.hs`. It uses
`stack` along with the `turtle` package to run as a quasi-shell script.

When I tried to run it today to start the database server, I got this
error message:

```

sdb.hs:70:47:
    Couldn't match type ‘T.Text’
                   with ‘Text’
    NB: ‘T.Text’
          is defined in ‘Data.Text.Internal’ in package ‘text-1.2.2.1’
        ‘Text’ is defined in ‘Data.Text.Internal’ in package ‘text-1.2.2.0’
    Expected type: String -> Text
      Actual type: String -> T.Text
    In the first argument of ‘map’, namely ‘T.pack’
    In the second argument of ‘procs’, namely ‘(map T.pack as')’

sdb.hs:140:37:
    Couldn't match type ‘T.Text’
                   with ‘Text’
    NB: ‘T.Text’
          is defined in ‘Data.Text.Internal’ in package ‘text-1.2.2.1’
        ‘Text’ is defined in ‘Data.Text.Internal’ in package ‘text-1.2.2.0’
    Expected type: Char -> Text
      Actual type: Char -> T.Text
    In the first argument of ‘(<$>)’, namely ‘T.singleton’
    In the second argument of ‘(>>)’, namely
      ‘(T.singleton <$> (char '#' <|> newline))’

sdb.hs:221:41:
    Couldn't match type ‘T.Text’
                   with ‘Text’
    NB: ‘T.Text’
          is defined in ‘Data.Text.Internal’ in package ‘text-1.2.2.1’
        ‘Text’ is defined in ‘Data.Text.Internal’ in package ‘text-1.2.2.0’
    Expected type: Char -> Text
      Actual type: Char -> T.Text
    In the second argument of ‘(.)’, namely ‘T.singleton’
    In the first argument of ‘fmap’, namely ‘(Just . T.singleton)’
```

As you can see, stack is using multiple versions of `text`, and coming up with this ridiculous error.

## Bulleted information:

*   Script URL: <https://git.gnu.io/pharpend/snowdrift/blob/mech-sibling/sdb.hs>
*   `stack --version`:

        Version 1.0.5, Git revision 51492566184b4f98106de0e8199c8962dace567e x86_64

*   Steps to reproduce:

        git clone https://git.gnu.io/pharpend/snowdrift.git -b mech-sibling
        cd snowdrift
        ./sdb.hs init

harendra-kumar:
Current mechanism of using GHC_PACKAGE_PATH for runghc and ghc commands does
not seem to work well when we have multiple versions of the same package. GHC
does not always pick up the packages in the same order as GHC_PACKAGE_PATH.

This fix determines of the package-ids using ghc-pkg and then passes
package-ids on command line of ghc or runghc invocation. This works only when
the user explicitly passes --package to runghc or ghc commands. When --package
is not specified we have no easy way to determine what all packages will be
used by the file being compiled.

This will make sure that scripts which explicitly list all or multi-instance
packages will always run reliably.

fixes #1957 (Requires all packages to be listed explicitly)

### DIFF
--- a/src/Stack/GhcPkg.hs
+++ b/src/Stack/GhcPkg.hs
@@ -13,6 +13,7 @@ module Stack.GhcPkg
   (getGlobalDB
   ,EnvOverride
   ,envHelper
+  ,findGhcPkgField
   ,createDatabase
   ,unregisterGhcPkgId
   ,getCabalPkgVer
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -740,7 +740,14 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
                     Nothing -- Unlocked already above.
         ExecOptsEmbellished {..} ->
            withBuildConfigAndLock go $ \lk -> do
+               let targets = concatMap words eoPackages
+               unless (null targets) $
+                   Stack.Build.build (const $ return ()) lk defaultBuildOptsCLI
+                       { boptsCLITargets = map T.pack targets
+                       }
+
                config <- asks getConfig
+               menv <- liftIO $ configEnvOverride config eoEnvSettings
                (cmd, args) <- case (eoCmd, eoArgs) of
                    (ExecCmd cmd, args) -> return (cmd, args)
                    (ExecGhc, args) -> execCompiler "" args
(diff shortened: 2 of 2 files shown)


Blaisorblade:
This looks mostly good to me, and the fix idea and new code is probably good.

Only one concern: would you mind separating the behavior-preserving refactoring from the behavioral change, and document them as such? (That is, the refactoring should declare itself as such, since one can't tell from looking at the diff).
This is not critical; but it's hard to tell the different changes apart, and that can be a concern when looking at history, so I care a bit.

harendra-kumar:
The CI failure does not seem to have anything to do with the change.

Blaisorblade:
Failure seems indeed spurious (seen this before). LGTM & merging. Thanks!

PR #2514, opened 2016-08-18
Added option to add nix dependencies as nix GC roots

YPares:
This PR makes it easier to use stack nix with small disk spaces or low bandwidth internet.
Nix provides a command that removes all the packages that aren't registered as nix GC roots: nix-collect-garbage.
This PR gives you an option to tell stack to call nix-shell with the option to register the packages as GC roots. This way, you can run nix-collect-garbage to remove old packages you don't need anymore and nix will preserve your current depedencies, so you don't have to re-download them again.


mgsloan:
LGTM.  Want to review, @borsboom ?

mboes:
LGTM. Make sure to add a Changelog entry though. :)

PR #2516, opened 2016-08-18
Nix in docker

YPares:
This PR allows you to activate both docker & nix support at the same time.
This is really useful if you have a preconfigured imaged (based for instance on nixos/nix image)
already containing a nix store will all your nix dependencies, or else nix would just re-download all your dependencies everytime you build or run.

Note that if you use a custom shell.nix, it requires you to set both STACK_IN_NIXSHELL=1 (which lib.buildStackProject does) and STACK_IN_CONTAINER=1 (which it doesn't).

I'd like people to test it a bit, if the above description could fit someone's use case.


Blaisorblade:
Is stack outputting `>>> inContainer: False, inShell: False` because of this changes?
If that shows up in the output of `stack --compiler=ghc-$GHCVER path --dist-dir`, it would explain the Travis failure, and would make it genuine.

But I have no clue why that would only happen on the branch build and not on the merge with master.

YPares:
Ah yes it's a debug print I added and forgot to remove
Le 18 août 2016 21:18, "Paolo G. Giarrusso" <notifications@github.com> a
écrit :

mgsloan:
Wow, fancy!

Blaisorblade:
It seems the debug print is still there though and still failing some builds, so it should block the merge? Not that I really get the situation.

YPares:
@Blaisorblade Yes, I don't even understand why it's being still displayed. The last commit has removed it.

Blaisorblade:
Thanks for clarifying that! I suspect we're using a stack version from the cache, which was maybe infected by a previous build. Hopefully failed builds don't alter the cache; in that case, we probably lack a test detecting the failure.
I'll take a look and try fixing this later.

Blaisorblade:
Hypothesis confirmed; I filed #2527. Looks like somebody cleared the cache, which is the right thing to do now. @YPares Can you do a force-push to trigger a new build, or trigger a new one?

YPares:
@Blaisorblade I pushed a merge with master, but travis apparently is still using the old build.

Blaisorblade:
@YPares I'm confused as well, but since the merge with master works, let's just treat this as working and wait for @borsboom's review.

borsboom:
Is there an example Docker image that I can use to try this out?

borsboom:
Does `stack` get re-executed a second time in a nix-shell after being re-executed in the Docker container already?

YPares:
Yes, it does.
We were based on nixos/nix image and were using stack
`--no-docker-set-user --docker-stack-exe path/to/stack`.

2016-08-29 18:58 GMT+02:00 Emanuel Borsboom <notifications@github.com>:

YPares:
@mboes I've made a Dockerfile for HaskellR that does that: https://github.com/tweag/HaskellR/blob/nix-in-docker/etc/nix/Dockerfile (it requires that stack is built and executable is placed right next to Dockerfile)
But then I'm getting this error:

```
$ ~/tweag/stack/.stack-work/install/x86_64-linux/lts-6.0/7.10.3/bin/stack --nix --docker --docker-image 3732d34fbb91 --docker-stack-exe image --allow-different-user --no-docker-set-user build
no such file or directory
FATA[0000] Error response from daemon: Cannot start container 105c16ba4548b3c9bb1a0042657e5b858d1cd5edd5aedeeb8242b2b03ca038ee: [8] System error: no such file or directory
```

Which is weird, as stack is in the image and reachable by PATH. It also happens when I let stack on the host and do `--docker-stack-exe path/to/stack`.

YPares:
@borsboom Ok, you may use the dockerfile from 
https://github.com/tweag/HaskellR/blob/nix-in-docker/etc/nix/Dockerfile

and build HaskellR with `new_stack --docker --nix --docker-image e5aeb225f265 --docker-stack-exe $(which new_stack) --no-docker-set-user build`

borsboom:
Ok, this looks good to me.  Thanks for keeping the Nix support improving!

Issue #2229, opened 2016-06-02
Cannot generate internal documentation


osbugs:
Perhaps I did miss something, but it seems to me that generating internal documentation is not possible.

With cabal, I just use the following command line: `cabal haddock --internal`. Is there some similar switch for stack?

osbugs:
I think this bug/feature request is distinct from #2165. The difference is that I do not want to provide arguments to haddock but I do want haddock to be run on a different target than the one provided.

mgsloan:
I see!  Yes, currently I don't think we have `--internal`, but it'd be a great thing to add.  Not sure if we want to call it that, though.

taphu:
Hi guys. I really needed this for a project I'm working on, so I went ahead and took a swing at it. This is my first contribution to stack, so let me know if there is anything I missed in the way of completeness/style/policy. Thanks.

osbugs:
Updated PR in #2598

sjakobi:
Closed via #2598

PR #2529, opened 2016-08-23
Added `--haddock-internal` command line argument (#2229)

taphu:
This is a boolean argument which causes haddocks to be generated in the
same manner as `cabal haddock --internal`, implemented by simply passing
`--internal` to the `cabal haddock` invocation used to build the docs.

It defaults to False.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -1141,8 +1141,11 @@ singleBuild runInBase ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} in
                             ("Warning: haddock not generating hyperlinked sources because 'HsColour' not\n" <>
                              "found on PATH (use 'stack install hscolour' to install).")
                         return ["--hyperlink-source" | hscolourExists]
+            let internalDocs = if boptsHaddockInternal bopts
+                    then ["--internal"]
+                    else []
             cabal False (concat [["haddock", "--html", "--hoogle", "--html-location=../$pkg-$version/"]
-                                ,sourceFlag])
+                                ,sourceFlag, internalDocs])
 
         unless isFinalBuild $ withMVar eeInstallLock $ \() -> do
             announce "copy/register"
--- a/src/Stack/Types/Config/Build.hs
+++ b/src/Stack/Types/Config/Build.hs
@@ -52,6 +52,8 @@ data BuildOpts =
             -- ^ Open haddocks in the browser?
             ,boptsHaddockDeps :: !(Maybe Bool)
             -- ^ Build haddocks for dependencies?
+            ,boptsHaddockInternal :: !Bool
+            -- ^ Build haddocks for all symbols and packages, like @cabal haddock --internal@
             ,boptsInstallExes :: !Bool
             -- ^ Install executables to user path after building?
             ,boptsPreFetch :: !Bool
@@ -90,6 +92,7 @@ defaultBuildOpts = BuildOpts
     , boptsHaddockOpts = defaultHaddockOpts
     , boptsOpenHaddocks = False
     , boptsHaddockDeps = Nothing
+    , boptsHaddockInternal = False
     , boptsInstallExes = False
     , boptsPreFetch = False
     , boptsKeepGoing = Nothing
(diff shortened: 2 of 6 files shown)


taphu:
See https://github.com/commercialhaskell/stack/issues/2229

src/Stack/Build/Haddock.hs
```diff
@@ -114,6 +115,10 @@ shouldHaddockPackage bopts wanted name =
 shouldHaddockDeps :: BuildOpts -> Bool
 shouldHaddockDeps bopts = fromMaybe (boptsHaddock bopts) (boptsHaddockDeps bopts)
 
+-- | Determine whether to build haddocks for non-exported modules, like @cabal haddock --intenral@.
```
mgsloan: Typo here in the flag.  I'd prefer to just not have this function and instead use `boptsHaddockInternal`.

mgsloan:
Thanks for the PR! Looks good, just one comment to address

sjakobi:
I'm not familiar with `cabal haddock --internal`. This will generate haddocks for modules in a library's `other-modules` section but not for executables, right?

Please also document the option in `doc/yaml_configuration.md` and update the changelog!

:+1:

src/Stack/Build/Haddock.hs
taphu: Ok, I will fix that.

taphu:
> I'm not familiar with cabal haddock --internal. This will generate haddocks for modules in a library's other-modules section but not for executables, right?

Yes. More specifically, `cabal haddock --help` reports:
>     --internal              Run haddock for internal modules and include all symbols

.

> Please also document the option in doc/yaml_configuration.md

The first line of `doc/yaml-configuration.md` reports:

> This page is intended to fully document all configuration options available in the stack.yaml file.

This doesn't really seem applicable since this is a command line option instead of a config file option. Maybe I should update `doc/build_command.md` instead?

src/Stack/Types/Config/Build.hs
```diff
@@ -167,6 +171,7 @@ instance FromJSON (WithJSONWarnings BuildOptsMonoid) where
               buildMonoidHaddockOpts <- jsonSubWarnings (o ..:? buildMonoidHaddockOptsArgName ..!= mempty)
               buildMonoidOpenHaddocks <- First <$> o ..:? buildMonoidOpenHaddocksArgName
               buildMonoidHaddockDeps <- First <$> o ..:? buildMonoidHaddockDepsArgName
+              buildMonoidHaddockInternal <- First <$> o ..:? buildMonoidHaddockInternalArgName
```
sjakobi: Here is the line where the "haddock-internal" option is parsed from a `stack.yaml` file.

sjakobi:
> This doesn't really seem applicable since this is a command line option instead of a config file option. Maybe I should update doc/build_command.md instead?

As I pointed out with my inline comment above, you _did_ also add a `stack.yaml` option too! (And that's good IMO.)
So please do update `yaml_configuration.md` in the "Project configuration" section.

src/Stack/Types/Config/Build.hs
taphu: Oh, I see. Thanks.

src/Stack/Build/Haddock.hs
taphu: Fixed, using a rebase.

taphu:
> Please also document the option in doc/yaml_configuration.md and update the changelog!

updated, using a rebase.

taphu:
Rebased onto the current master because of whatever weirdness was causing the integration tests to fail.

src/Stack/Build/Execute.hs
```diff
@@ -1141,8 +1141,11 @@ singleBuild runInBase ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} in
                             ("Warning: haddock not generating hyperlinked sources because 'HsColour' not\n" <>
                              "found on PATH (use 'stack install hscolour' to install).")
                         return ["--hyperlink-source" | hscolourExists]
+            let internalDocs = if boptsHaddockInternal bopts
```
sjakobi: It's the `bopts` that aren't in scope and causing the Travis failures. From a superficial glance at the code above: are you possibly looking for the `eeBuildOpts`?

sjakobi:
Closed via https://github.com/commercialhaskell/stack/pull/2598.

Issue #1728, opened 2016-02-02
stack exec pacman fails


ndmitchell:
```
C:\Neil\hlint>stack exec pacman
[...]

PR #2537, opened 2016-08-28
Default to using a stack-installed, isolated GHC

### LINKED ISSUE #2221: Stop using system-ghc silently…or at all.
screen shot 2016-05-31 at 3 54 34 pm

This happens a fair bit, and not just in connection with `ghc-mod` although it's the most common culprit. While doing the training at LambdaConf we had a beginner get _very_ confused because their system installed GHC was pre-FTP and the Stack provided GHC was 7.10.3. From there, they didn't understand `ghci` and `stack ghci` would be from _substantially_ different versions of GHC.

I'd really like Stack to stop implicitly using system-installed versions of GHC without explicit direction from the user.

sjakobi:
Closes #2221.

### DIFF
--- a/src/Stack/SetupCmd.hs
+++ b/src/Stack/SetupCmd.hs
@@ -46,7 +46,7 @@ setupParser = SetupCmdOpts
                       "The default is to install the version implied by the resolver.")))
     <*> OA.boolFlags False
             "reinstall"
-            "reinstalling GHC, even if available (implies no-system-ghc)"
+            "reinstalling GHC, even if available (incompatible with --system-ghc)"
             OA.idm
     <*> OA.boolFlags False
             "upgrade-cabal"
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,7 +13,9 @@ Major changes:
   [#1364](https://github.com/commercialhaskell/stack/issues/1364)
 * Stack will now always use its own GHC installation, even when a suitable GHC
   installation is available on the PATH. To get the old behaviour, use
-  the `--system-ghc` flag or set `system-ghc: true` in your ~/.stack/config.yaml.
+  the `--system-ghc` flag or run `stack config set system-ghc --global true`.
+  Docker- and Nix-enabled projects continue to use the GHC installations
+  in their environment by default.
 
   NB: Scripts that previously used stack in combination with a system GHC
   installation should now include a `stack setup` line or use the `--install-ghc`
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,14 @@ Major changes:
   autogen files and run preprocessors. The `--no-build` flag is now deprecated
   because it should no longer be necessary. See
   [#1364](https://github.com/commercialhaskell/stack/issues/1364)
+* Stack will now always use its own GHC installation, even when a suitable GHC
+  installation is available on the PATH. To get the old behaviour, use
+  the `--system-ghc` flag or set `system-ghc: true` in your ~/.stack/config.yaml.
+
+  NB: Scripts that previously used stack in combination with a system GHC
+  installation should now include a `stack setup` line or use the `--install-ghc`
+  flag.
+  [#2221](https://github.com/commercialhaskell/stack/issues/2221)
 
 Behavior changes:
 
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -332,8 +332,7 @@ stack ghc, stack ghci, stack runghc, or stack exec
 ```
 
 Thankfully, the command is smart enough to know not to perform an installation
-twice. `setup` will either use the first GHC it finds on your PATH, or a sandboxed
-version after installing it. As the command output above indicates, you can use `stack
+twice. As the command output above indicates, you can use `stack
 path` for quite a bit of path information (which we'll play with more later).
 For now, we'll just look at where GHC is installed:
 
--- a/doc/docker_integration.md
+++ b/doc/docker_integration.md
@@ -78,6 +78,13 @@ The most basic configuration is to add this to your project's `stack.yaml`:
 See [configuration](#configuration) for additional options. You can enable it on
 the command-line using `stack --docker`.
 
+Please note that in a docker-enabled configuration, stack uses the GHC installed
+in the Docker container by default. To use a compiler installed by stack, add
+
+    system-ghc: false
+
+(see [`system-ghc`](yaml_configuration.md#system-ghc)).
+
 ### Use stack as normal
 
 With Docker enabled, most stack sub-commands will automatically launch
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -155,23 +155,39 @@ the following line to your .cabal file:
 
 ## I already have GHC installed, can I still use stack?
 
-Yes. stack will default to using whatever GHC is on your `PATH`. If that GHC is
-a compatible version with the snapshot you're using, it will simply use it.
-Otherwise, it will prompt you to run `stack setup`. Note that `stack setup`
-installs GHC into `~/.stack/programs/$platform/ghc-$version/` and not a global
-location.
+Yes. In its default configuration, stack will simply ignore any system GHC
+installation and use a sandboxed GHC that it has installed itself (typically
+via the `stack setup` command). You can find these sandboxed GHC installations
+in `~/.stack/programs/$platform/ghc-$version/`.
 
-Note that GHC installation doesn't work for all OSes, so in some cases the
-first option will need to install GHC yourself.
+If you would like stack to use your system GHC installation, use the
+[`--system-ghc` flag](yaml_configuration.md#system-ghc) or run
+`stack config set system-ghc --global true` to make stack check your
+`PATH` for a suitable GHC by default.
+
+Note that stack can only use a system GHC installation if its version is
+compatible with the configuration of the current project, particularly the
+[`resolver` setting](yaml_configuration.md#resolver).
+
+Note that GHC installation doesn't work for all OSes, so in some cases you
+will need to use `system-ghc` and install GHC yourself.
 
 ## How does stack determine what GHC to use?
 
-It uses the first GHC that it finds on the `PATH`. If that GHC does not comply
-with the various requirements (version, architecture) that your project needs,
-it will prompt you to run `stack setup` to get it. `stack` is fully aware of
-all GHCs that it has installed itself.
+In its default configuration, stack determines from the current project which
+GHC version, architecture etc. it needs. It then looks in
+`~/.stack/programs/$platform/ghc-$version/` for a compatible GHC, requesting
+to install one via `stack setup` if none is found.
 
-See [this issue](https://github.com/commercialhaskell/stack/issues/420) for a detailed discussion.
+If you are using the [`--system-ghc` flag](yaml_configuration.md/#system-ghc) or
+have configured `system-ghc: true` either in the project `stack.yaml`
+or the global `~/.stack/config.yaml`, stack will use the first GHC that it finds
+on your `PATH`, falling back on its sandboxed installations only if the found GHC
+doesn't comply with the various requirements (version, architecture) that your
+project needs.
+
+See [this issue](https://github.com/commercialhaskell/stack/issues/420) for a
+detailed discussion of stack's behavior when `system-ghc` is enabled.
 
 ## How do I upgrade to GHC 7.10.2 with stack?
 
--- a/doc/travis_ci.md
+++ b/doc/travis_ci.md
@@ -92,6 +92,10 @@ situation is simple:
 
 ```yaml
 before_install:
+  # Install stack as above
+  # ...
+  # Configure stack to use the system GHC installation
+  - stack config set system-ghc --global true
   - export PATH=/opt/ghc/7.10.2/bin:$PATH
 
 addons:
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -308,7 +308,8 @@ other projects by installing into your shared snapshot database.
 
 Enables or disables using the GHC available on the PATH.
 Useful to disable if you want to save the time, bandwidth or storage space needed to setup an isolated GHC.
-Default is `false`.
+Default is `false` unless the [Docker](docker_integration.md) or [Nix](nix_integration.md) integration is enabled.
+In a Nix-enabled configuration, stack is incompatible with `system-ghc: false`.
 
 ```yaml
 # Turn on system GHC
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -306,13 +306,13 @@ other projects by installing into your shared snapshot database.
 
 ### system-ghc
 
-Enables or disables using the GHC available on the PATH. Useful to disable if
-you want to force stack to use its own installed GHC (via `stack setup`), in
-cases where your system GHC my be incomplete for some reason. Default is `true`.
+Enables or disables using the GHC available on the PATH.
+Useful to disable if you want to save the time, bandwidth or storage space needed to setup an isolated GHC.
+Default is `false`.
 
 ```yaml
-# Turn off system GHC
-system-ghc: false
+# Turn on system GHC
+system-ghc: true
 ```
 
 ### install-ghc
@@ -451,6 +451,8 @@ Specify a variant binary distribution of GHC to use.  Known values:
 * any other value: Use a custom GHC bindist. You should specify
   [setup-info](#setup-info) so `stack setup` knows where to download it, or
   pass the `stack setup --ghc-bindist` argument on the command-line
+
+This option is incompatible with `system-ghc: true`.
 
 ### ghc-build
 
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -240,9 +240,12 @@ configFromConfigMonoid configStackRoot configUserConfigPath mresolver mproject C
 
          configGHCVariant0 = getFirst configMonoidGHCVariant
          configGHCBuild = getFirst configMonoidGHCBuild
+         configSystemGHC = fromFirst False configMonoidSystemGHC
 
-         configSystemGHC = fromFirst (isNothing configGHCVariant0) configMonoidSystemGHC
-         configInstallGHC = fromFirst False configMonoidInstallGHC
+     when (isJust configGHCVariant0 && configSystemGHC) $
+         throwM ManualGHCVariantSettingsAreIncompatibleWithSystemGHC
+
+     let configInstallGHC = fromFirst False configMonoidInstallGHC
          configSkipGHCCheck = fromFirst False configMonoidSkipGHCCheck
          configSkipMsys = fromFirst False configMonoidSkipMsys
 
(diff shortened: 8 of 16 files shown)


sjakobi:
I expect that this change will break a lot of CI scripts, e.g. those that use hvr's repo to install various GHC versions. Should I highlight this more in the Changelog?

src/Stack/Options.hs
```diff
@@ -861,7 +861,7 @@ ghcVariantParser hide =
         readGHCVariant
         (long "ghc-variant" <> metavar "VARIANT" <>
          help
-             "Specialized GHC variant, e.g. integersimple (implies --no-system-ghc)" <>
+             "Specialized GHC variant, e.g. integersimple" <>
```
Blaisorblade: That's still implied, no? Or is it now legal to use `--system-ghc=... --ghc-variant`?

src/Stack/SetupCmd.hs
```diff
@@ -46,7 +46,7 @@ setupParser = SetupCmdOpts
                       "The default is to install the version implied by the resolver.")))
     <*> OA.boolFlags False
             "reinstall"
-            "reinstalling GHC, even if available (implies no-system-ghc)"
+            "reinstalling GHC, even if available"
```
Blaisorblade: Still implied here too?

src/Stack/Options.hs
sjakobi: > That's still implied, no?

I guess that depends on how one understands "implies": I read it as "changes default settings" (in this case from `--system-ghc` to `--no-system-ghc`), while you seem to read it like "requires".

I haven't been able to test `--ghc-variant <sth-non-standard>` on Ubuntu-16.04 (I didn't try very hard) but according to #816 it's intended to be used only with `--no-system-ghc`.

I don't think `--ghc-variant` should silently override a user's `--system-ghc` setting, so how about I add  `(incompatible with --no-system-ghc)` to the CLI help line?
I'd also add a check that errors out if both `--ghc-variant X` and `--system-ghc` are passed in.

Blaisorblade: Both your proposals are good. FWIW I understand "implies" like you do, I had just misunderstood the code.

src/Stack/SetupCmd.hs
Blaisorblade: Then I guess `incompatible with no-system-ghc`.

src/Stack/Options.hs
sjakobi: Ok. I meant to write `(incompatible with --system-ghc)` above BTW.

src/Stack/SetupCmd.hs
sjakobi: I'll add `incompatible with --system-ghc` (my mistake).

src/Stack/Options.hs
Blaisorblade: I totally didn't notice and the discussion made sense anyway. But I need coffee...

borsboom:
> I expect that this change will break a lot of CI scripts, e.g. those that use hvr's repo to install various GHC versions. Should I highlight this more in the Changelog?

I think this probably should at least go in the Major Changes section to make it more likely to be noticed by someone skimming the changes.

borsboom:
> I don't think --ghc-variant should silently override a user's --system-ghc setting, so how about I add (incompatible with --no-system-ghc) to the CLI help line?
> I'd also add a check that errors out if both --ghc-variant X and --system-ghc are passed in.

I agree.

sjakobi:
@YPares: I'm not sure how this change affects Nix users. Could you take a look at that?

sjakobi:
> I expect that this change will break a lot of CI scripts, e.g. those that use hvr's repo to install various GHC versions.

LOL, here we go: https://travis-ci.org/commercialhaskell/stack/jobs/155970348

borsboom:
The `sanity` integration test also appears to have broken.  See https://jenkins-public.fpcomplete.com/job/stack-integration-tests/225/consoleFull

Blaisorblade:
Auto-installing GHC when needed would silently fix the issue with setup scripts. I assume that's a bad idea for some reason, but maybe that should be reconsidered? As long as Stack can be trusted to install a good GHC, that seems a fine default. The main downside, IIUC, is if GHC can't be trusted to install a good GHC, says on Linux platforms (HVR mentioned issues with Ubuntu 16.10 recently — no GHC bindist works there).

YPares:
@sjakobi Well, for it to work with nix I have to pass `--system-ghc`.
I think system ghc should be automatically activated if nix is.

borsboom:
> Auto-installing GHC when needed would silently fix the issue with setup scripts. I assume that's a bad idea for some reason, but maybe that should be reconsidered?

I remember we hashed this out a long time ago, and it boiled down to most people having a preference for not being "surprised" by Stack going ahead and installing GHC without being asked to.  There's no technical reason one way or the other.  The `--install-ghc` flag switches to the other behaviour.

snoyberg:
Just some ideas to throw out there. There are a number of people who really prefer to use system-installed GHCs. While it may be problematic as a default, we shouldn't make these people suffer through typing `--system-ghc` constantly, or have to edit all of their local `stack.yaml` files. Instead, what if we:

* Recognize a global configuration override to what the default for `system-ghc` is (true or false)
* Add a new command for changing that default (strawman: `stack set system-ghc` or `stack set no-system-ghc`, definitely up for debate)
* In the code which gives the user the message about running `stack setup`, also check if `ghc` is on the `PATH` and, if so, give a message about running `stack set system-ghc`

I think this will give us a situation where unexpecting users don't accidentally use a broken system install and determined users can easily use that install. CI scripts _will_ suffer for this (which is really unfortunate), but at least the fix will be as easy as adding `stack set system-ghc`.

borsboom:
> @sjakobi Well, for it to work with nix I have to pass --system-ghc.
> I think system ghc should be automatically activated if nix is.

The same probably applies to Docker.

borsboom:
> CI scripts will suffer for this (which is really unfortunate), but at least the fix will be as easy as adding `stack set system-ghc`.

While we're at it, perhaps also including `stack set install-ghc` (or something like it) to enable Stack automatically installing GHC without needing to run `stack setup` should be added.

Blaisorblade:
> stack set X, Y or Z

If we do that, should we hardcode an ad-hoc subset of config.yaml or support arbitrary keys? If we decide to hardcode some, let's pick an interface that is a subset of what we have later. (`git config` might be an inspiration here).

borsboom:
pinging @drwebb, who has done some thinking and work around supporting `stack set` or `stack config`-like commands in the past

sjakobi:
> Instead, what if we:

> * Recognize a global configuration override to what the default for system-ghc is (true or false)

Do you mean https://docs.haskellstack.org/en/stable/yaml_configuration/#system-ghc or are you asking for something different?

> * Add a new command for changing that default (strawman: stack set system-ghc or stack set no-system-ghc, definitely up for debate)

To modify project `stack.yaml`s we already have the `stack config set` command which currently only supports the `resolver` option. For CI scripts it might be enough to extend this command to support the `system-ghc` option so users could modify the `system-ghc` setting for the current _project_.

If we want to modify the _global_ configuration (`~/.stack/config.yaml`) we could maybe extend `stack config` with `--global` flag similarly to `git config`. Some users may think though that the `--global` flag is for the "implicit global project`…

> * In the code which gives the user the message about running stack setup, also check if ghc is on the PATH and, if so, give a message about running stack set system-ghc

:+1:. We should probably also check whether it's the right version of GHC too.

> While we're at it, perhaps also including stack set install-ghc (or something like it) to enable Stack automatically installing GHC without needing to run stack setup should be added.

I agree. To be clear, the [`install-ghc` config option](https://docs.haskellstack.org/en/stable/yaml_configuration/#install-ghc) already exists.

---

BTW, I believe that _all_ the ["non-project config" options](https://docs.haskellstack.org/en/stable/yaml_configuration/#non-project-config) can currently also be specified in  project-local `stack.yaml`s. If we agree that this is the way it should be, I'd fix the docs which currently claim that these options can only be specified in the global configuration. Oh, and the `/etc/stack/config.yaml` path is deprecated and should be removed, right?

snoyberg:
Sounds good, I'd simply forgotten that the global configuration would be included (shows how long since I've worked on the codebase...).

sjakobi:
> If we do that, should we hardcode an ad-hoc subset of config.yaml or support arbitrary keys?

I think ultimately we want to support arbitrary keys. https://github.com/commercialhaskell/stack/issues/115 somewhat tracks this idea, although currently only for `stack.yaml`s.

> If we decide to hardcode some, let's pick an interface that is a subset of what we have later. (git config might be an inspiration here).

:+1:

sjakobi:
@snoyberg 

> I'd simply forgotten that the global configuration would be included

I don't understand what you're trying to express here. Can you expand a bit?

sjakobi:
Also, to keep the `stack config set` syntax "regular", I'd propose to write `stack config set system-ghc false`. A shorter syntax for boolean options like `stack config set no-install-ghc` could be added later if deemed worth the hassle.

snoyberg:
Please don't put too much weight on my comments, the syntax was intended as strawmen, and you're in a better position to decide on the exact syntax.

snoyberg:
Just expressing that my mind is porous and I forgot many relevant details
of how Stack works. You really shouldn't pay much attention to me, I'm
causing more harm than good here.

Blaisorblade:
> If we want to modify the global configuration (~/.stack/config.yaml) we could maybe extend stack config with a --global flag similarly to git config. Some users may think though that the --global flag is for the "implicit global project`…

You could have `--config` and `--global` (or the longer `--global-config` and `--global-project` if needed).

sjakobi:
> Just expressing that my mind is porous and I forgot many relevant details of how Stack works. 

Gotcha! ;)

> You really shouldn't pay much attention to me, I'm causing more harm than good here.

Not at all! Your design input here is very valuable even if you aren't aware of each and every new stack feature! The `stack set system-ghc` idea is very good and should make the `system-ghc` transition much more tolerable for users! Please keep the ideas coming!

Blaisorblade:
BTW just realized that `stack setup` should also set `system-ghc` consistently. From https://github.com/commercialhaskell/stack/issues/1728#issuecomment-195138302:

> I think it is a funky default that the system ghc is preferred even if you've done stack setup --no-system-ghc.

sjakobi:
@YPares 
> @sjakobi Well, for it to work with nix I have to pass --system-ghc.
> I think system ghc should be automatically activated if nix is.

I'll update the code then so `nix` implies `system-ghc`.

Are `no-system-ghc` and `nix` actually compatible at all?

YPares:
For the time being, Stack with nix requires to use GHC from nixpkgs. So,
no, they're totally incompatible

2016-09-05 14:19 GMT+02:00 Simon Jakobi <notifications@github.com>:

borsboom:
On the other hand, in the Docker case it would be uncommon to use --system-ghc, but not totally incompatible.

mgsloan:
Seems like this PR is languishing due to the additional discussion.  I am confused why this isn't a simple matter of changing the default value of `system-ghc`?  We already have a `system-ghc` field in the configuration (including `config.yaml`).

So, all this would change is the base default.  If users put `system-ghc: true` in their config, then it will use the old semantics - it will be used if it matches, and otherwise it will use the version that stack installed.

So, as far as I can tell it will all "just work" with no extra work, and all that needs to happen is a change of the default.  Perhaps I'm missing something?

I would prefer to wait on doing `stack set` stuff until we have a good way of automatically modifying yaml.

mgsloan:
@Blaisorblade pointed out to me that the crux of the issue is the existing scripts, in particular Travis CI scripts, that this would break.  How about we encourage all users to make the decision on their own?

I propose that commands which depend on resolving a ghc version should end their output with something like

```
Warning: system-ghc is unspecified in .../config.yaml. It is recommended to set "system-ghc: false" in that file, to avoid a variety of potential problems with using a globally installed GHC.
See http://... for more information.
```

This message would also be suppressed via `system-ghc: true`, of course!

mgsloan:
EDIT: I wrote a couple comments without fully grokking the proposal above.

I think this is quite reasonable!  Here's my understanding of where this conversation has gone:

1) We are changing the default, and this will break some travis scripts and other scripts.
2) To save the travis script authors lots of boilerplate, we will implement `stack set system-ghc false --global`, perhaps along with other commands.  Initially, these will entirely re-serialize the yaml, which is acceptable for this usecase.

Do we need to open an issue or two for this?  I'm paranoid of this languishing and not happening for the next release.  I think it's important!

mgsloan:
@sjakobi Hey!  Thanks for writing this patch.  If you want to continue the iteration process, feel free.  This is a priority for next release, so I'm also willing to take implementation from here.  I don't want to step on toes though, does that sound like a good way to move forward with this?

sjakobi:
@mgsloan 

> This is a priority for next release, so I'm also willing to take implementation from here. I don't want to step on toes though, does that sound like a good way to move forward with this?

I wouldn't mind if someone else finished the implementation but I think I can do it myself this week.

Here's what I believe remains to be done:

* [ ] Make sure the default `system-ghc` setting is still `true` when either docker or nix are enabled.

* [ ] Implement `stack config set --global system-ghc [true/false]`.

  (If anyone has a strong preference for a different syntax here, please open an issue, so we can discuss it properly! At this point `stack config set resolver [snapshot]` has existed for quite awhile and I think we should stay consistent with that syntax.) 

* [ ] If `stack` can't find an isolated ghc installation, check whether a system-installed ghc can be used and suggest `stack config set --global system-ghc true` as an alternative to `stack setup`.

* [ ] Look over the documentation to see if `system-ghc: false` is still assumed anywhere.

sjakobi:
@YPares: Could you please test and review 5929f4c? Based on my local testing, the docker integration seems to work alrighty.

BTW would the new [Nix in Docker](https://github.com/commercialhaskell/stack/pull/2516) support make it easier for everyone to test their own PRs with Nix?! Could you, @YPares, maybe write a short guide on how to use that (for Nix & Docker noobs) and add it to the project documentation section?

The build failure should be fixed by a847c57.

YPares:
@sjakobi 
> BTW would the new Nix in Docker support make it easier for everyone to test their own PRs with Nix?!

No, I don't think so. Installing/removing Nix is really simple and everything Nix is already isolated (in /nix and in your home). After that, you just have to source the nix profile script and then you're good to run `stack --nix`.

I'm testing the PR.

YPares:
@sjakobi Ok, works for me! Thanks for the work :)

sjakobi:
> No, I don't think so. Installing/removing Nix is really simple and everything Nix is already isolated (in /nix and in your home). After that, you just have to source the nix profile script and then you're good to run stack --nix.

Ok, I'll give that a try a then.

Thanks for testing!

mgsloan:
@sjakobi Sounds like you have a handle on it :) Awesome!

Blaisorblade:
>> Could you, @YPares, maybe write a short guide on how to use that (for Nix & Docker noobs) and add it to the project documentation section?

> Installing/removing Nix is simpler and everything Nix is already isolated (in /nix and in your home).

While it's probably not your responsibility, *if* (big if) we're all supposed to be able to test Nix support, could you please *still* document a recommended installation method? As a Nix noob, googling for a few minutes doesn't tell me what I need, and I'm afraid I can't risk messing up my system.

Among the top 3 results for "installing nix", I assume you mean neither of https://nixos.org/wiki/Installing_Nix_on_Debian or https://nixos.org/wiki/How_to_install_nix_in_home_(on_another_distribution), and suspect that I should follow http://nixos.org/nix/download.html. That still mentions no "nix profile script" and neither does http://nixos.org/nix/manual/#chap-quick-start. And neither reassures me Nix will leave everything else alone—in fact, the example below makes no sense since none of the commands shown should be able to modify the PATH:

```
$ which hello
/home/eelco/.nix-profile/bin/hello
```

YPares:
@Blaisorblade The nix documentation https://github.com/commercialhaskell/stack/blob/master/doc/nix_integration.md indeed refers to http://nixos.org/nix/download.html .
I prefere not to duplicate any installation method in stack documentation (as it's nix maitainers responsibility to update http://nixos.org/nix/download.html if this installation method changes).
This page doesn't mention the nix profile script (the one in `~/.nix-profile/etc/profile.d/nix.sh`) because the installation process already adds its sourcing to your bashrc.

But I'll add this part to the doc anyway, yes.

Blaisorblade:
> This page doesn't mention the nix profile script (the one in ~/.nix-profile/etc/profile.d/nix.sh) because the installation process already adds its sourcing to your bashrc.

Thanks for pointing me to stack docs on nix integration, they seem really great, and installing Nix is not remotely invasive, unlike I feared.

sjakobi:
What's still missing here is mostly a fix for the integration tests. I'm not sure what to do about them yet, so if anyone has an idea, please chime in!

I'll probably get back to this tomorrow or on Tuesday.

Blaisorblade:
Sorry for the basic question, but for the integration tests, why isn't `stack config set --global system-ghc true` enough?

mgsloan:
Looks pretty good to me so far @sjakobi !  I think this can be merged soon.

I do not see the code changes for `config set`.  If it's not ready soon, feel free to make a P0 issue for implementing the config modification changes portion of this.

sjakobi:
> Sorry for the basic question, but for the integration tests, why isn't stack config set --global system-ghc true enough?

I guess you're right, for _fixing_ the integration tests that might just about be enough!

We could still add tests that use a sandboxed GHC later.

> I do not see the code changes for config set.

I did these in separate PRs / commits: https://github.com/commercialhaskell/stack/pull/2675 and https://github.com/commercialhaskell/stack/commit/0bdab6e3c44f8eef24b75a0a0e2fe8f6ca5596da.

sjakobi:
Also RE `stack config set`:

In my implementation, the `--global` flag is only available for some fields, e.g. not for `resolver`.

Sadly this means that `stack config set --global system-ghc true` is a syntax error:

```
~/t/hkl $ stack config set --global system-ghc true
Invalid option `--global'
```

The `--global` flag must come after the key: e.g. `stack config set system-ghc --global true` or `stack config set system-ghc true --global`.

Blaisorblade:
> Sadly this means that stack config set --global system-ghc true is a syntax error:

That's an implementation restriction right? Can you document it as such for now?
If there's some design reason why `--global` makes no sense for `resolver` maybe we should reconsider the API.

sjakobi:
@Blaisorblade 

> > Sadly this means that stack config set --global system-ghc true is a syntax error:

> That's an implementation restriction right? Can you document it as such for now?
> If there's some design reason why --global makes no sense for resolver maybe we should reconsider the (EDIT) command-line API.

I have opened https://github.com/commercialhaskell/stack/issues/2707 to discuss this. Is this what you meant by "Can you document it as such for now?" or would you like to see other documentation of this too?

sjakobi:
I think this issue is ready for a final review now. Sorry it took me so long!

Blaisorblade:
> Is this what you meant by "Can you document it as such for now?" or would you like to see other documentation of this too?

Read #2707, so I get it: using `--global` is not supported and can't be supported, because it doesn't make sense. It's not an implementation restriction, that is, something sensible we just don't handle (yet).

We need user docs for `stack config`—the more pitfalls, the more we need those docs.

mgsloan:
@sjakobi These changes look great!  I just realized that the `stack config set system-ghc` stuff is already on master.  So this is definitely good to go

PR #2541, opened 2016-08-29
Missing newline

whatsthepoint:
List doesn't render properly with Read the Docs without newline before start of list.


borsboom:
Thanks!

Issue #1714, opened 2016-01-28
`stack build network` on Windows via Cygwin SSH fails where `cabal install network` succeeds


acfoltzer:
So, right out of the gate, I know that Cygwin is not an officially supported build environment for `network`. However, what we do on some of our CI servers is use Cygwin only for the ssh support, and then make sure that the msys binaries come before Cygwin's in the PATH.

On a Windows 10 machine with MinGHC 7.10.2 installed, I can put MinGHC's msys binaries in the path:
    
    [...]/cygdrive/c/Users/hudson/AppData/Local/Programs/minghc-7.10.2-x86_64/ghc-7.10.2/bin:/cygdrive/c/Users/hudson/AppData/Local/Programs/minghc-7.10.2-x86_64/bin:/cygdrive/c/Users/hudson/AppData/Local/Programs/minghc-7.10.2-x86_64/ghc-7.10.2/mingw/bin[...]

With the path configured thusly, the following Just Works:

    cabal install network --configure-option --build=x86_64-pc-msys --configure-option --host=x86_64-pc-msys

However, I cannot seem to find an invocation of `stack` that works in this same environment. My first thought is that `stack build` needs a way to pass along `runhaskell Setup.hs configure` options, but maybe there's something else going on with paths?

acfoltzer:
It'd probably help to include _how_ `stack build network` fails:

    Configuring network-2.6.2.1...
    configure: WARNING: unrecognized options: --with-compiler, --with-gcc
    checking build system type... i686-pc-msys
    checking host system type... i686-pc-msys
    checking for gcc... gcc
    checking whether the C compiler works... no
    configure: error: in `/tmp/stack3424/network-2.6.2.1':
    configure: error: C compiler cannot create executables
    See `config.log' for more details

acfoltzer:
Found out a workaround for now by gleaning the output of `stack install -v network`. I copied the invocation of the `Setup.hs configure` step, manually added the `ghc-7.10.3/bin` and `ghc-7.10.3/mingw/bin` directories to my PATH, and then:

```
$ runhaskell Setup.hs configure --with-ghc='C:\Users\hudson\AppData\Local\Programs\stack\x86_64-windows\ghc-7.10.3\bin\ghc.exe' --with-ghc-pkg='C:\Users\hudson\AppData\Local\Programs\stack\x86_64-windows\ghc-7.10.3\bin\ghc-pkg.exe' --user
[...]

PR #2552, opened 2016-08-31
eager msys windows check

gbaz:
This should resolve https://github.com/commercialhaskell/stack/issues/1714

Prior, if you were using system ghc, no msys check was performed. Now, if you're using system ghc but `getInstalledTool installed $(mkPackageName "msys2") (const True) ` does not yield an msys, then you get an issue.

So a build of network with this PR now seems to bring up a message "Continuing despite missing tool: msys2" whereas before there was no warning. And now when one runs "stack setup" and it determines the system ghc is present, it doesn't bail early, but also goes and installs msys regardless.

This behavior could arguably be a bit better (with a warning instead of bailing on not finding msys) but at least teaching people to run "stack setup" is easy enough. Also it could try to be more clever about finding an msys, but that feels somewhat fragile, while this seems very straightforward.

I also made a few choices about how to hoist which code as I refactored. I tried to stick to a reasonable facsimile of house style, but this is my first Stack PR and I make no promises.

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -363,11 +363,38 @@ ensureCompiler sopts = do
                 arch /= expectedArch
         isWanted = isWantedCompiler (soptsCompilerCheck sopts) (soptsWantedCompiler sopts)
 
+    getSetupInfo' <- runOnce (getSetupInfo (soptsStackSetupYaml sopts) =<< asks getHttpManager)
+
+    let getMmsys2Tool = do
+            platform <- asks getPlatform
+            localPrograms <- asks $ configLocalPrograms . getConfig
+            installed <- listInstalled localPrograms
+
+            case platform of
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks @gbaz, this seems like an improvement for all cases, not just the HP use case. In 20-20 hindsight, I can't even figure why I would have coupled these together initially, unless it was something to do with MinGHC requirements.

I've reviewed, and it makes sense, but someone more recently familiar should probably have a look too.

@gbaz Can you confirm that this was tested with HP Minimal and works as expected?

gbaz:
I can confirm that I tested this fix on a windows machine with HP minimal installed, but with a stack with this patch substituted for the patch shipped by minimal, and it did allow network to install, with the results as described above.

borsboom:
Thanks @gbaz!

PR #2557, opened 2016-09-01
Force UTF8 encoding for tar paths in sdist (fixes #2549)

### LINKED ISSUE #2549: stack sdist does not handle filenames with non-ASCII characters
### Steps to reproduce

1. `git clone https://github.com/snoyberg/yaml`
2. `cd yaml`
3. `git checkout yaml/0.8.18.2`
4. `stack sdist`
5. Unpack the tarball printed to standard out, e.g. `tar xfv .stack-work/dist/x86_64-osx/Cabal-1.22.5.0/yaml-0.8.18.3.tar.gz`

### Expected

All files unpack correctly.

### Actual

I get the following error from tar due to an incorrectly generated tarball:

```
yaml-0.8.18.3/test/resources/accent?/foo.yaml: Can't create 'yaml-0.8.18.3/test/resources/accent?/foo.yaml'
tar: Error exit delayed from previous errors.
```

### Stack version

```
$ stack --version
Version 1.1.2, Git revision cebe10e845fed4420b6224d97dcabf20477bbd4b (3646 commits) x86_64 hpack-0.14.0
```

### Method of installation

* Executable distributed with the Haskell Platform Minimal (no joke)

Related to: https://github.com/snoyberg/yaml/issues/92

snoyberg:


snoyberg:
To be clear, I'm not 100% certain that this is the right way to fix the issue with the `tar` package, though it seems to be working for me.

snoyberg:
Hold up on merging, I need to fix the `unpack` side as well, the bug is deeper than expected.

snoyberg:
Never mind on the `unpack` side, it turns out that Hackage will just block the upload of non-ASCII characters. This patch is a good thing since it will make the errors more obvious for users.

PR #1290, opened 2015-11-05
Typo recommendations

mrkkrp:
This implements thing described in #158. I've chosen to stay conservative with threshold 1. Limiting number of matches can be confusing because desired package name can be eliminated from list of suggestions. With current implementation number of matches won't ever be too big to limit it.

Here is example of how it works:

```
[mark@arch ~]$ stack build stackage-metdata-0.3.0.0
Run from outside a project, using implicit global project config
Using resolver: lts-3.9 from implicit global project's config file: /home/mark/.stack/global-project/stack.yaml
Didn't see stackage-metdata-0.3.0.0 in your package indices.
Perhaps you meant "stackage-metadata"?
Updating and trying again.
Updating package index Hackage (mirrored at https://github.com/commercialhaskell                                                                                Fetched package index.
Populated index cache.
The following package identifiers were not found in your indices: stackage-metdata-0.3.0.0
[mark@arch ~]$ stack build parses-3.1.9
Run from outside a project, using implicit global project config
Using resolver: lts-3.9 from implicit global project's config file: /home/mark/.stack/global-project/stack.yaml
Didn't see parses-3.1.9 in your package indices.
Perhaps you meant "parsec", "parsek", or "parsers"?
Updating and trying again.
Updating package index Hackage (mirrored at https://github.com/commercialhaskell                                                                                Fetched package index.
Populated index cache.
The following package identifiers were not found in your indices: parses-3.1.9
[mark@arch ~]$
```

I also tested various edge cases and it behaves normally.

The refactoring should make some parts of code easier to read. I've also added comments describing what some functions do.

PR #2563, opened 2016-09-04
Switch from ‘edit-distance’ to ‘text-metrics’

mrkkrp:
The `text-metrics` package is ×2.5 faster and it works on strict `Text` values rather than on `String`s, which eliminates the need in unpacking of all package names (which should also speed-up things).

### DIFF
--- a/src/Stack/Fetch.hs
+++ b/src/Stack/Fetch.hs
@@ -67,6 +67,7 @@ import qualified Data.Set                       as Set
 import           Data.String                    (fromString)
 import qualified Data.Text                      as T
 import           Data.Text.Encoding             (decodeUtf8)
+import           Data.Text.Metrics
 import           Data.Typeable                  (Typeable)
 import           Data.Word                      (Word64)
 import           Network.HTTP.Download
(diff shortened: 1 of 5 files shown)


mrkkrp:
Please restart AppVeyor check, it should find it now.

sjakobi:
> Please restart AppVeyor check, it should find it now.

I don't have the proper permissions. Could you simply do a `git commit --amend` and `git push -f`?

Could you also do a quick demonstration, that our use case for the `damerauLevenshtein` function still works as before?

mrkkrp:
Nothnig could be broken by this commit, the new `function is well tested and is correct. I'll check if you wish, however.

sjakobi:
Sorry for my leeriness, but considering that `text-metrics` is brand new and that `stack` has rather bad test coverage, I'd prefer to see a short example usage.

mrkkrp:
@sjakobi Here

```
~/projects/programs/haskell/stack $ stack build stackage-metdata-0.3.0.0
Didn't see stackage-metdata-0.3.0.0 in your package indices.
Updating and trying again.
Fetched package index.
The following package identifiers were not found in your indices: stackage-metdata-0.3.0.0
Perhaps you meant stackage-metadata?
~/projects/programs/haskell/stack $ stack build parses-3.1.9
Didn't see parses-3.1.9 in your package indices.
Updating and trying again.
Fetched package index.
The following package identifiers were not found in your indices: parses-3.1.9
Perhaps you meant parsec, parsek, or parsers?
```

…that's what I get with `text-metrics`. Compare with my original implementation here: #1290. (There is no quotes, but that's OK, see #1298).

sjakobi:
Thanks!

PR #2564, opened 2016-09-05
Extend `stack config set` with `system-ghc` and `install-ghc` options

sjakobi:
This change adds the fields suggested in https://github.com/commercialhaskell/stack/pull/2537#issuecomment-243423410 and https://github.com/commercialhaskell/stack/pull/2537#issuecomment-243427665.


borsboom:
LGTM.  Thanks!

PR #2573, opened 2016-09-08
Wrap lines for easier reading of contrib docs

lwm:


Blaisorblade:
Thanks for caring, but how are you reading these docs? Markdown renders them as soft-wrapped HTML—in fact, the line breaks that you add will be preserved by GitHub Flavored Markdown (IIRC).

The preferred way to read that file is through [this link](https://github.com/commercialhaskell/stack/blob/master/CONTRIBUTING.md), especially since GitHub will add this link when creating an issue (see https://github.com/commercialhaskell/stack/issues/new).

lwm:
> but how are you reading these docs

In my text editor with no line wraps (old habit). 

>  the line breaks that you add will be preserved by GitHub Flavored Markdown (IIRC).

This shouldn't alter the representation at all?

Some of the `docs/*` files are nicely wrapped. It's a better read IMHO. 

Feel free to close/ignore ;)

borsboom:
I like my markdown to be wrapped so its readable as "plain text" too, so I'm inclined to merge this.

lwm:
@borsboom I will also take a look at wrapping the remaining unwrapped files in `docs/*` files if this gets accepted.

Blaisorblade:
Also fine with me. I'm sure Github markdown used to keep hard wraps 
but that's no more documented so maybe it has changed. Testing in this
message (admittedly with bad hard wraps).

The
content of doc sounds safer to hard wrap, since it's supposed to be processed by a different markdown parser.

Blaisorblade:
Merging as the output looks reasonable anyway.

lwm:
@Blaisorblade Ah, I see what you mean. Doesn't seem to affect markdown documents though (comparing [rendered][1] and [raw][2]). Thanks! :beers: 

[1]: https://github.com/commercialhaskell/stack/blob/master/CONTRIBUTING.md
[2]: https://raw.githubusercontent.com/commercialhaskell/stack/master/CONTRIBUTING.md

Blaisorblade:
OK, so GitHub switched to different Markdown dialects for issues and docs. I'm ambivalent on that, but good to know.

PR #2576, opened 2016-09-10
Add --license option to list-dependencies

ruuda:
Stack makes it a breeze to build a project with lots of (transitive) dependencies without worrying too much about them, but when it comes to distributing the final binary, it is useful to know what your dependencies actually are. In particular, I want to know the licenses of my dependencies, to make sure that they are compatible, and to make sure that I am not accidentally violating one.

For Cabal there is the [`cabal-dependency-licenses`](https://github.com/jaspervdj/cabal-dependency-licenses) package, but I haven’t been able to make it work with Stack. However, because Stack can already list all dependencies, I figured it shouldn’t be too hard to make Stack list the licenses too. So that is what this PR does.

* Add a `--license` flag to `stack list-dependencies`. When specified, `stack list-dependencies` will print the package license instead of the package version.
* As a drive-by change, unify the options for `stack dot` and `stack list-dependencies`. In order not to change the current behaviour,  `--external` defaults to false for `stack dot` and to true for `stack list-dependencies`.

My personal use case for this is that I want to validate the output of `stack list-dependencies --license --no-include-base` against a list of whitelisted licenses as part of CI, to ensure that I don’t accidentally depend on an incompatible library.

Also, I have a question about `--no-include-base`. Would it make sense to have this flag exclude the `template-haskell` and `integer-gmp` packages too? They appear to be “special” just like base.

This is my first time diving into the Stack codebase, so some things can likely be improved. Any feedback would be appreciated.

### DIFF
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -4,6 +4,7 @@
 module Stack.Dot (dot
                  ,listDependencies
                  ,DotOpts(..)
+                 ,DotPayload(..)
                  ,ListDepsOpts(..)
                  ,resolveDependencies
                  ,printGraph
@@ -83,8 +84,13 @@ dot dotOpts = do
   (localNames, prunedGraph) <- createPrunedDependencyGraph dotOpts
   printGraph dotOpts localNames prunedGraph
 
--- | Information about a package, when available.
-type Payload = (Maybe Version, Maybe License)
+-- | Information about a package in the dependency graph, when available.
+data DotPayload = DotPayload
+  { payloadVersion :: Maybe Version
+  -- ^ The package version.
+  , payloadLicense :: Maybe License
+  -- ^ The license the package was released under.
+  } deriving (Eq, Show)
 
 -- | Create the dependency graph and also prune it as specified in the dot
 -- options. Returns a set of local names and and a map from package names to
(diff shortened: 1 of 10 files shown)


Blaisorblade:
The feature makes sense. The build failures I see come from master (fix should be in #2580).

ruuda:
I rebased this on top of the fix.

Blaisorblade:
Thanks! Haven't had time to review the code yet though (not sure I'll find it myself, I'm afraid).

src/Stack/Dot.hs
```diff
@@ -70,16 +80,37 @@ dot :: (HasEnvConfig env
...
-    let pkgsToPrune = if dotIncludeBase dotOpts
-                         then dotPrune dotOpts
-                         else Set.insert "base" (dotPrune dotOpts)
-        prunedGraph = pruneGraph localNames pkgsToPrune resultGraph
-    printGraph dotOpts localNames prunedGraph
+  (localNames, prunedGraph) <- createPrunedDependencyGraph dotOpts
+  printGraph dotOpts localNames prunedGraph
+
+-- | Information about a package, when available.
+type Payload = (Maybe Version, Maybe License)
```
borsboom: Any reason not to make this a record type instead?

borsboom:
Overall, LGTM.

src/Stack/Dot.hs
ruuda: No reason in particular. I’ll change it to a record type as that makes the code below more readable.

ruuda:
I rebased this once more on top of master. All commits compile, but they don’t all pass the tests.

@lwm: You moved `dotOptsParser` into `Stack.Options.DockerParser` in #2587, was that by accident? I moved it into a new `Stack.Options.DotParser`.

lwm:
@ruuda Yes, that was a mistake. Thank you for catching that and adding the fix!

Blaisorblade:
Sorry for not reviewing that more closely (I wanted to merge it before it bitrotted).

ruuda:
Rebased once more to resolve changelog conflicts. Are there still any blockers for this?

Blaisorblade:
Merged based on the existing review.

ruuda:
Thanks!

PR #2578, opened 2016-09-11
Improved Travis instructions

snoyberg:
* Consolidate into a single page (with reference from guide)
* Put content into external files for easier copy/paste
* Include a simple configuration
* Give advice on how to choose the right configuration


doc/GUIDE.md
```diff
@@ -1939,213 +1939,8 @@ can use the `stack dot` command and Graphviz. More information is
 
 ### Travis with caching
 
-Many people use Travis CI to test out a project for every Git push. We have [a
```
Blaisorblade: There are still references to this section, e.g. in https://github.com/commercialhaskell/stack/pull/2578/files#diff-e96fbd8ff8894442360d3de9bdd33895R86, and at least that is now stale even with the redirection.

doc/travis_ci.md
```diff
@@ -1,11 +1,32 @@
...
+* [The complex Travis configuration](https://raw.githubusercontent.com/commercialhaskell/stack/master/doc/travis-complex.yml)
+  is intended for projects that need to support multiple GHC versions
+  and multiple OSes, such as open source libraries to be released to
+  Hackage. It tests against cabal-install, as well as Stack on Linux
+  and OS X. The configuration is significantly more involved to allow
+  for all of this branching behavior.
+
+Each of these configurations is ready to be used immediately, just
+copy-paste the content into the `.travis.yml` file in the root or your
+repo, enable Travis on the repo, and you're good to go.
```
Blaisorblade: I'd point out that the latter configuration might often need customization. It makes choices on which GHC versions to support, and assumes that the same `stack.yaml` (including extra-deps) works on all the chosen resolvers—that's likely if no extra-deps are used. Maybe add a pointer to the relevant sections below?

doc/GUIDE.md
snoyberg: Good point, I'll do a grep to find all such places in the repo.

Blaisorblade:
Looks like a big improvement! BTW there's progress on the Travis side, though they're not yet done it seems:
https://github.com/travis-ci/apt-package-whitelist/issues/379#issuecomment-246109122

doc/GUIDE.md
snoyberg: Commit pushed. I couldn't follow the link you provided, but I think I found the only two references in the repo.

snoyberg:
Yeah, I saw some of the Travis activity, but honestly gave up following the back-and-forth at some point. Glad it's moving forward.

And yes, I think this is a big improvement. Beyond including the simple config (which was definitely a good idea on your part), this scratches an itch of mine by putting the configs in standalone files for easier addition to project by just `curl`ing.

doc/travis_ci.md
snoyberg: Good call, added a paragraph.

Blaisorblade:
Oh, the Travis failure seems just "master stopped building pre-AMP". On it.

But the PR itself LGTM now.

PR #2584, opened 2016-09-12
stack upload uses HTTP digest authentication

snoyberg:


snoyberg:
I'll be pushing a new commit shortly, once http-client-tls-0.3.1 is available on the indices.

Blaisorblade:
stack-8.0.yaml needs an extra-dep—Travis gives `http-client-tls: needed (>=0.3.1), 0.3.0 found` (https://travis-ci.org/commercialhaskell/stack/jobs/159490050#L224).

snoyberg:
Ugh, I missed one! Hold on...

snoyberg:
I've just pushed another commit which bumps to http-client-0.5.3.1, due to: https://github.com/snoyberg/http-client/issues/225.

stack.cabal
```diff
@@ -187,7 +187,7 @@ library
                    , hit
                    , hpc
                    , http-client >= 0.5.0
```
Blaisorblade: Strictly speaking, shouldn't d52c9bcdd2bf717b8278d90a1cdc4ff68232545e also increase the cabal bound to `http-client >= 0.5.3.1`?

snoyberg: Yeah, it's a good idea.

src/Stack/Upload.hs
```diff
@@ -203,13 +205,14 @@ mkUploader config us = do
...
-            let req3 = applyBasicAuth
+            mreq3 <- applyDigestAuth
                     (encodeUtf8 $ hcUsername creds)
                     (encodeUtf8 $ hcPassword creds)
                     req2
+                    manager
             putStr $ "Uploading " ++ tarName ++ "... "
             hFlush stdout
-            withResponse req3 manager $ \res ->
+            withResponse (fromMaybe req2 mreq3) manager $ \res ->
```
Blaisorblade: As long as we handle `applyDigestAuth` failing, I'd prefer to have it logged, with sth. better than:
```haskell
when (isNull mreq3)
  $logWarn "Failure in applyDigestAuth"
```

So if a submission without password fails, there's a chance to realize that's because digest authentication failed, not because we didn't try authentication.

BTW sorry for thinking about this now, I realize multiple reviews are annoying 😅

Blaisorblade:
Other than this comment, LGTM. I'll leave merging to you 😉

src/Stack/Upload.hs
snoyberg: I don't think it's necessary. If `mreq3` is `Nothing`, then we're going to retry the same request again, and when it (almost) inevitably fails, we'll display the error message from the server to the user.

Blaisorblade: I see that; my point is, I'd expect the error would be misleading. If anything hinted at "no password provided", I'd wonder why stack isn't providing a password, not why Hackage is rejecting digest authentication. I'm not implying some hypothetical user would be confused, I'm implying *I* would be confused.

snoyberg: Then I'm not sure what message would be better than the proposed "Failure in applyDigestAuth."

Blaisorblade: That one's fine then!

PR #2587, opened 2016-09-12
Split `Stack.Options` module

### LINKED ISSUE #2406: Split Stack.Options up and move option parsers to option types
(Split off from #2334)

> Because Stack.Options has a lot of dependencies it is currently recompiled after most changes.
Splitting it up would also make changes on option types more localized.

lwm:
Closes #2406.

Had to rebase on https://github.com/commercialhaskell/stack/commit/e3aa2388cefc083605091c55e92d9af09e048182. This PR has the potential to rot pretty fast!

I created a [Stack.Options.Utils][2] module to store the common stuff.

Is this the desired approach? Do I need to add anything to the [CHANGELOG][1]?

[1]: https://github.com/commercialhaskell/stack/blob/master/ChangeLog.md
[2]: https://github.com/lwm/stack/blob/bcd1b84ce0d5542a839e42b49c1b88d0b15ff2fd/src/Stack/Options/Utils.hs


lwm:
44 minutes for the travis build! You folks are really serious about this eh?

Blaisorblade:
The structure looks reasonable to me (though I haven't reviewed this closely). I'll trust you (and the tests) that the change just moves the code without anything other refactoring with any potential for bugs (because if `git bisect` finds a regression here, there's no chance to know where it was introduced.

And finally, stack tries to avoid `Utils` modules since they're not very specific. But I think this is still a net improvement that can rot fast, so merging.

Blaisorblade:
BTW for the future: I squashed the commits because each commit is supposed to be a compiling state—exceptions creates problems when using `git bisect` to track regressions.

I think the only way to split this into modules would have been carving out one module at a time, and I don't think it's necessarily worth it on a simple refactoring (as long as you stick to copy-pasting code).

And overall: thanks a lot for this!

lwm:
> I'll trust you (and the tests) that the change just moves the code without anything other refactoring

:+1: 

> I squashed the commits because each commit is supposed to be a compiling state

OK. Makes sense.

> And overall: thanks a lot for this!

Happy to help! Thanks :sheep:

PR #2588, opened 2016-09-12
ghc-mod recommendation refactoring

lwm:
`ghc-mod` told me so.

Is there some way for the Travis build to complain about this stuff or it's already covered from the `script:` commands? In the case it is not, it could be [allowed to fail](https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail) in case it's too much overhead. I could submit a PR?

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -154,7 +154,7 @@ main = do
 
   eGlobalRun <- try $ commandLineHandler progName False
   case eGlobalRun of
-    Left (exitCode :: ExitCode) -> do
+    Left (exitCode :: ExitCode) ->
       throwIO exitCode
     Right (globalMonoid,run) -> do
       let global = globalOptsFromMonoid isTerminal globalMonoid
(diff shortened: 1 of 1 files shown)


Blaisorblade:
All refactorings look good, so merged.

Re the PR: sounds good, but how do I get the warnings locally to have a good workflow?

I keep hearing that installing ghc-mod is not necessarily easy. But maybe these warnings are in fact from the easier-to-install [hlint](https://github.com/ndmitchell/hlint#readme)?

lwm:
> I keep hearing that installing ghc-mod is not necessarily easy

Fair point! I had a hard time myself.

> but how do I get the warnings locally to have a good workflow?
> But maybe these warnings are in fact from the easier-to-install hlint?

Ah, you are right! It was in fact hlint :blush:.

It could be as simple as adding the following to a travis matrix:

```
$ stack install hlint
$ find src -name "*hs" -exec hlint {} +;
```

As it stands, there are 355 hints to be dealt with.

PR #2590, opened 2016-09-13
Wrap some docs to 80 chars.

lwm:
Follow on from #2573 and https://github.com/commercialhaskell/stack/pull/2573#issuecomment-246517282.

`[ci skip] [build skip]`


Blaisorblade:
@borsboom, you like this too?

PR #2591, opened 2016-09-13
add 'style' environment for travis

lwm:
Follow on from https://github.com/commercialhaskell/stack/pull/2588#issuecomment-246670280.

This could potentially take me a while to get right. My bash/travis-fu isn't great.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,11 +43,14 @@ matrix:
       compiler: ": #stack 8.0.1"
       addons: {apt: {packages: [ghc-8.0.1], sources: [hvr-ghc]}}
 
+    - env: BUILD=style GHCVER=8.0.1 STACK_YAML=stack-8.0.yaml
+      compiler: ": #stack 8.0.1"
+      addons: {apt: {packages: [ghc-8.0.1], sources: [hvr-ghc]}}
+
   allow_failures:
     - env: BUILD=stack STACK_YAML=stack.yaml
       compiler: ": #stack 7.10.3 osx"
       os: osx
-
 
 # Note: the distinction between `before_install` and `install` is not important.
 # - in before_install we install build tools
@@ -55,6 +58,8 @@ matrix:
 before_install:
  - unset CC
  - case "$BUILD" in
+     style)
+       export PATH=$HOME/.local/bin:/opt/ghc/$GHCVER/bin:$PATH;;
      stack)
        export PATH=$HOME/.local/bin:/opt/ghc/$GHCVER/bin:$PATH;;
      cabal)
@@ -65,6 +70,8 @@ before_install:
 install:
  - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
  - case "$BUILD" in
+     style)
+       stack --no-terminal install hlint;;
      stack)
        stack --no-terminal test --only-dependencies;;
      cabal)
@@ -81,6 +88,9 @@ install:
 
 script:
  - set -e; case "$BUILD" in
+     style)
+       hlint src/ test/ --cpp-simple --hint=HLint.hs;
+       stack --no-terminal build --pedantic;;
      stack)
        stack --no-terminal test --haddock --no-haddock-deps --ghc-options="-Werror";;
      cabal)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,7 +65,8 @@ Note that stack contributors need not dogmatically follow the suggested hints
 but are encouraged to debate their usefulness. If you find a hint is not useful
 and detracts from readability, consider marking it in the [configuration
 file](https://github.com/commercialhaskell/stack/blob/master/HLint.hs) to
-be ignored.
+be ignored. Please refer to the [HLint manual](https://github.com/ndmitchell/hlint#ignoring-hints)
+for configuration syntax.
 
 Quoting [@mgsloan](https://github.com/commercialhaskell/stack/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Amgsloan):
 
@@ -84,5 +85,8 @@ stack install hlint
 Once installed, you can check your changes with:
 
 ```
-hlint src/ test/ --cpp-simple
+hlint src/ test/ --cpp-simple --hint=HLint.hs
 ```
+
+Where `--cpp-simple` strips `#` lines and `--hint` explicitly specifies the
+configuration file.
--- a/HLint.hs
+++ b/HLint.hs
@@ -7,9 +7,9 @@ module HLint.HLint where
 import "hint" HLint.Default
 import "hint" HLint.Builtin.All
 
--- Related to 'Hide post-AMP warnings' comment
 ignore "Use fewer imports" =
-  System.Process.Read
+  System.Process.Read  -- Related to 'Hide post-AMP warnings' comment
+  Stack.Exec  -- ifdef for System.Process.Read
 
 -- Related to 'explicit pattern matching is clearer' comment
 ignore "Use fromMaybe" =


Blaisorblade:
Thanks for this. It's clever to do it with a matrix, so this doesn't slow down other builds 👍 .

But I suspect I'd rather keep this on hold ✋ , as I was writing on #2588, till the warnings are reduced.

If we add the warnings without fixing the existing hints, as a general (somewhat cynical) rule of thumb nobody will look at them—that's the same reason why one often keeps the build warning-free.

I'll keep accepting any PR like #2588. But to make any such tool part of the workflow, if somebody volunteers for it, the ideal path (or path of least resistance) is IMHO the following (note: this is mostly language and tool-agnostic, I'm aware of hlint but not sure I've ever used it so I might miss anything `hlint`-specific):

- Create an issue to get contributors consensus.
- If people care enough (which I'd guess), start fixing warnings and annotating false positives. Start from the ones are safe to fix and progress to the others.
- Document a good workflow so that dealing with `hlint` takes limited overhead (if this weren't Haskell I'd consider saying IDE support and one-click fixes).
- Once the last warnings are gone, consider making new warnings a build error so the codebase stays hlint-clean.

This ensures the codebase stays clean without annoyances from false positives, but it is a lot of work and may or may not be worth it—last I did it (unrelated project) (couple of days ago) I only volunteered because I was tired of reformatting the code every second PR ;-).

Blaisorblade:
After having looked at the hints, I agree with most. My favorite is "use if instead of casing on a boolean", though I hope the extra parens in the output aren't needed.

```
src/Stack/Build/Execute.hs:1459:5: Suggestion: Use if
Found:
  case toCoverage (boptsTestOpts bopts) of
      True -> do hpcIndexDir <- toFilePathNoTrailingSep <$>
                                  hpcRelativeDir
                 return ["--ghc-options", "-hpcdir " ++ hpcIndexDir ++ ddumpOpts]
      False -> return ["--ghc-options", ddumpOpts]
Why not:
  if toCoverage (boptsTestOpts bopts) then
    (do hpcIndexDir <- toFilePathNoTrailingSep <$> hpcRelativeDir
        return ["--ghc-options", "-hpcdir " ++ hpcIndexDir ++ ddumpOpts])
    else return ["--ghc-options", ddumpOpts]
```

lwm:
@Blaisorblade 

Thanks for your input!

> Create an issue to get contributors consensus.

So, I wait for some input either here or on #2592.

> This ensures the codebase stays clean without annoyances from false positives

OK. Agreed. I feel like doing it, so it should be fine. It's an easy one for learning.

> I hope the extra parens in the output aren't needed.

In my experience, they are not. But we can find out!

lwm:
Travis run seems to be stuck so I force pushed for a restart. I can't seem to cancel the last one (not enough permissions). Anyway, I will put this on hold as suggested and wait for input :beers:

Blaisorblade:
Canceled the last build. But I assume it's not stuck, there's just a longish queue. (Spurious failures on Travis are very common, but I've never seen stuck builds).

lwm:
Well, we got [some useful output](https://travis-ci.org/commercialhaskell/stack/jobs/159593400) but it didn't allow the failure. I wonder why ...

.travis.yml
```diff
@@ -17,6 +17,9 @@ cache:
         - $HOME/.stack
 
 matrix:
+  allow_failures:
+    - env: BUILD=qa
```
Blaisorblade: Spitballing: Shouldn't this say `BUILD=qa GHCVER=8.0.1 STACK_YAML=stack-8.0.yaml`? I strongly assume but am not certain you can omit `compiler` and `addons`

Blaisorblade: http://stackoverflow.com/a/35758162/53974 confirms you need `BUILD=qa GHCVER=8.0.1 STACK_YAML=stack-8.0.yaml?` (as value for `env:`).

lwm: Awesome. Pushing a fixup to test that one out (commenting out other envs to get quick feedback).

Blaisorblade:
I restarted the spurious failure on https://travis-ci.org/commercialhaskell/stack/jobs/159688176 — that was real Travis nonsense.

sjakobi:
In general I'm :+1: on adding a hlint pass to Travis, even making it a required job.

I think it will require some fine-tuning though:

* Some of the "replace `liftM` with `fmap`" suggestions won't work as long as `stack` is required to build with GHC-7.8. Not sure if hlint can run in "GHC-7.8-compat-mode"?!
* "Redundant `do`s" are super common IME and not a readability issue IMO. Maybe they should be excluded or only shown but not allowed to trigger a failure?! Some other hints should probably be treated similarly.

Some random suggestions:

* This job could also include a `stack build` with GHC-8 and `--pedantic`.
* Maybe call it more specifically "style" instead of "qa".

I guess the stack codebase has other, larger problems than those that `hlint` can catch, on stylistic and more fundamental levels. But I don't think it hurts to start addressing them like this.

Given that this PR will affect the workflows of other contributors as well, I think it would be good to get a few more `+1`s before this can be eventually merged.

mgsloan:
As mentioned on https://github.com/commercialhaskell/stack/pull/2615 , I am not so keen on following hlint religiously.  However, if folks feel strongly about it, go ahead!

lwm:
> I guess the stack codebase has other, larger problems than those that hlint can catch, on stylistic and more fundamental levels.

True. I realize these changes aren't worth much right now and only putting an extra effort on contributors to review my changes. But, as you said, it can be useful. And soon!

> This job could also include a stack build with GHC-8 and --pedantic.
> Maybe call it more specifically "style" instead of "qa".

Will follow those suggestions.

> Some of the "replace liftM with fmap" suggestions won't work as long as stack is required to build with GHC-7.8.

OK, I will keep that in mind.

> I am not so keen on following hlint religiously

As per [this post], I don't want to impose the hlint religion on anyone ;) Just a 'guard rail' against writing something silly. I suppose as the PR moves on, there will be a number of '--ignore' flags in use, depending on peoples preferences (https://github.com/commercialhaskell/stack/pull/2615#issuecomment-249517033).

[this post]: http://neilmitchell.blogspot.de/2009/09/how-i-use-hlint.html

mgsloan:
Yup, I think we're on the same page!  Feel free to proceed :D

lwm:
Rebased and added `--pedantic` build as per https://github.com/commercialhaskell/stack/pull/2591#issuecomment-248809797. Look like this one is ready to review once Travis goes green. Closes #2592.

mgsloan:
Cool, nice work @lwm !

So will this cause travis to fail if hlint isn't followed?  Should I expect to see hlint hints in https://travis-ci.org/commercialhaskell/stack/jobs/166878927 ?

lwm:
> So will this cause travis to fail if hlint isn't followed?

Yes. Currently the stack code base [is HLint hint free](https://travis-ci.org/commercialhaskell/stack/jobs/166878927#L227), so we're green :)

If HLint suggests any hints, then it will return 1 and the  build will fail.

Blaisorblade:
The documentation in #2699 doesn't mention how to use the joint config, could that be fixed?

lwm:
> The documentation in #2699 doesn't mention how to use the joint config

Not sure I get what 'joint config' means?

Blaisorblade:
Sorry, I meant hlint config, but autocorrect intervened.

lwm:
Ah, I got it ;) I added https://github.com/commercialhaskell/stack/pull/2591/commits/e20598b94e72451e9faa18cd503fed0987e6301e.

Blaisorblade:
New docs LGTM!

mgsloan:
Hmm, I'm not sure if I like that.  How about we have the style check, but have it in the list of allowed failures?

Then, periodically, or on release we can try to make the style check pass.

lwm:
> Hmm, I'm not sure if I like that. ... What do others think?

Huh? I thought I was clear about my intentions from the start with #2592. You saw +1s from https://github.com/commercialhaskell/stack/pull/2591#issuecomment-248809797 and https://github.com/commercialhaskell/stack/issues/2592#issuecomment-246682966?

> Then, periodically, or on release we can try to make the style check pass.

The warnings will stack up and no one will take care of them.

Blaisorblade:
> The warnings will stack up and no one will take care of them.

That's my assumption too, as I pointed out earlier: https://github.com/commercialhaskell/stack/pull/2591#issuecomment-246676805

My experience is probably colored by having to reformat code periodically in another project (different environment, I know).

> I don't want people (or me) to have to wrestle with their PRs to pass the style check.

For contributors where this is a barrier, there is now (worst case) the chance for committers to edit pull requests. For us... the question is to decide which warnings are worth fixing and which are better ignored, as I assume was discussed while fixing existing warnings.

lwm:
> the question is to decide which warnings are worth fixing and which are better ignored, as I assume was discussed while fixing existing warnings.

There are 175 ignored warnings in the code base based on decisions by contributors (See the [configuration file](https://github.com/commercialhaskell/stack/blob/master/HLint.hs)).  

In fact, now that I run HLint over HEAD, with the addition of [6e69e23](https://github.com/commercialhaskell/stack/commit/6e69e2306ddb56cef8887552a8a36715cf5a9b4d#diff-4968f87e674615592e572f2ec593e5b0R1076), there is a new hint which would fail the style build for all new PRs. That doesn't make sense unless everyone is on board.

It's your project folks, let me know :beers:

mgsloan:
> It's your project folks, let me know

It's your project too!  Thanks for contributing :)

I say we give this a try and see how it goes.

lwm:
:rocket:

borsboom:
There's an hlint failure that I'm not sure how to fix:

```
src/Stack/Docker.hs:83:1: Warning: Use fewer imports
Found:
  import Control.Monad.Trans.Control (MonadBaseControl)
  import Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)
  
Why not:
  import Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)
```

The code in question looks like this:

```
#ifdef WINDOWS
import           Control.Monad.Trans.Control (MonadBaseControl)
#else
import           Control.Concurrent (threadDelay)
import           Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)
import           System.Posix.Signals
import qualified System.Posix.User as PosixUser
#endif
```

That CPP is needed to avoid warnings from GHC about unused imports on Windows, but hlint seems to ignore the CPP.

Any thoughts on how to fix?

Blaisorblade:
Weird, hlint claims cpp support but might need configuration. Annoyingly
there's no --cpp-undefine documented (but maybe it's still there?): I'd
have suggested undefining WINDOWS since the windows variant has less code.
http://community.haskell.org/~ndm/darcs/hlint/hlint.htm
-- 
Paolo Giarrusso
From smartphone, sorry for typos or excessive brevity

Il 12 dic 2016 15:20, "Emanuel Borsboom" <notifications@github.com> ha
scritto:

There's an hlint failure that I'm not sure how to fix:

src/Stack/Docker.hs:83:1: Warning: Use fewer imports
Found:
  import Control.Monad.Trans.Control (MonadBaseControl)
  import Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)

Why not:
  import Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)

The code in question looks like this:

#ifdef WINDOWS
import           Control.Monad.Trans.Control (MonadBaseControl)
#else
import           Control.Concurrent (threadDelay)
import           Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)
import           System.Posix.Signals
import qualified System.Posix.User as PosixUser
#endif

That CPP is needed to avoid warnings from GHC about unused imports on
Windows, but hlint seems to ignore the CPP.

Any thoughts on how to fix?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/commercialhaskell/stack/pull/2591#issuecomment-266442262>,
or mute the thread

borsboom:
I see that it's actually already been fixed on master (by using a qualified import, which seems like more of a workaround for an hlint issue, but whatever).

lwm:
Ran into something similar and just used a HLint ignore:
  - https://github.com/commercialhaskell/stack/blob/master/HLint.hs#L12

PR #2594, opened 2016-09-15
Add dump-logs (fixes #426)

### LINKED ISSUE #426: `stack build --ghc-output` (or similar) to print GHC output during a build
Provide an option to pipe GHC's output to the terminal to better keep track of progress, instead of hanging on `Cabal-1.22.4.0: build` for dependencies that take some time to compile.

snoyberg:


src/Stack/Build/Execute.hs
```diff
@@ -801,6 +825,13 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
             logPath <- buildLogPath package msuffix
             ensureDir (parent logPath)
             let fp = toFilePath logPath
+
+            -- We only want to dump logs for local non-dependency packages
+            case taskType of
+                TTLocal lp | lpWanted lp ->
```
snoyberg: I'm not convinced that using `lpWanted` is the right thing here, we may instead want to check if this is listed as an `extra-dep` or not. Feedback welcome.

harendra-kumar:
How about adding a note at the end of the build suggesting `--dump-logs` to view the logs?

snoyberg:
That could work. Logic I'd follow would be:

    when (notDumpingLogs && wantedTargets > 1) displayDumpLogsSuggestion

Another idea: even when we don't dump the logs, we _could_ display the log file locations that were generated, though for a large project even that could be overwhelming output.

harendra-kumar:
Yes, it will be helpful when someone really needs it to diagnose problems. Otherwise you will have go figure the logs location first. Logs are also helpful to grep things when your console buffer has overflowed.

snoyberg:
That last comment implies to me that the full behavior should be:

```haskell
when dumpLogs doFullLogDumpOutput
when (not dumpLogs && wantedTargets > 1) $ logInfo "If you want to see build output on the console, please use --dump-logs"

-- and now, regardless of the above, print the list of a log files in case the buffer overruns
logInfo "Build logs have been generated in the following locations:"
mapM_ logInfo logFilePaths
```

Is that what you have in mind too?

snoyberg:
I've pushed a commit with this extra output. I'm on the fence as to whether it's a good idea or not.

harendra-kumar:
That sounds right to me. What are the multiple locations? Is that one for each package?

snoyberg:
It's one location per log file. There's a log file for each package's build, test, and bench output IIRC.

harendra-kumar:
That might be too much addendum to the build output. Why not have a `dump-logs` command as well which accepts the package name argument and dumps the logs for that package. If no argument is provided it will dump all. That way we need to suggest just one command at the end and you also do not have to copy/paste the path of the log file which is usually long.

snoyberg:
I think that's quickly reaching the point of diminishing returns. How about simply saying "log files can be found in *directory*"?

snoyberg:
I've pushed one more commit that just have the log directory on the output. I think this is a decent middle ground.

harendra-kumar:
That should work.

borsboom:
This looks good to me.  Looks like it needs a rebase/merge from master.

snoyberg:
Rebased against master, only conflict was a trivial one in the ChangeLog.md.

Issue #1295, opened 2015-11-06
Extra dep specified as git repo builds unnecessarily


chreekat:
On current HEAD of [Snowdrift's](https://github.com/snowdriftcoop/snowdrift) master  (ad236659f819eb220c282f35a4b04525a9fdc52b), the dependency ghci-runner is rebuilt after a 'stack clean':

```
$ stack clean
$ stack build
Snowdrift-0.1.4: unregistering (missing dependencies: ghci-runner)
ghci-runner-0.0.0.0: unregistering (local file changes: ghci-runner.cabal src/Dev/Runner.hs)
ghci-runner-0.0.0.0: configure
ghci-runner-0.0.0.0: build
ghci-runner-0.0.0.0: copy/register
Snowdrift-0.1.4: configure
Configuring Snowdrift-0.1.4...
```

Is this intended behavior?

If not, is this because both Snowdrift and ghci-runner have a source file 'src/Dev/Runner.hs'? But then why does it say that ghci-runner.cabal changes, too?

The [stack.yaml](https://github.com/snowdriftcoop/snowdrift/blob/ad236659f819eb220c282f35a4b04525a9fdc52b/stack.yaml):

```
packages:
- '.'
- location:
    git: https://github.com/chreekat/ghci-runner.git
    commit: bae5a558535087c8cc532da0fd192625361d8a25
  extra-dep: true
```

chreekat:
My mistake: "src/Dev/Runner.hs" *only* exists in ghci-runner. (Snowdrift has "dev/Runner.hs".)

So, there's still the question of why a extra-dep is rebuilt after 'stack clean', and why it things both ghci-runner.cabal and src/Dev/Runner.hs have changed when they are being pulled from a specific git revision.

mgsloan:
Hmm, yeah, there does seem to be something weird here.  Thanks for the report!  Extra-deps usually don't get rebuilt after `stack clean`.

So, this must be something to do with it being a git repo.  I figured maybe it's deleting the repo and recloning it, but that's not the case either.

chreekat:
Yeah. You may have already noticed this as well, but when stack *does*
build the dep, it seems to do nothing other than re-copy the object files.

Besides the 3-second delay, it's also annoying because "stack build
--pedantic" bails on the dep when I don't really care about it right now. :)

mgsloan:
Yup, that particular issue with `--pedantic` is tracked by https://github.com/commercialhaskell/stack/issues/1292

[...]

PR #2596, opened 2016-09-15
Don't treat local extra-dep packages as targets

snoyberg:
The immediate problem I noticed was that, when running `stack build`
with two local packages, one an extra-dep, the build output was not sent
to the console. This fixes the problem, but I believe there's still an
issue with the test suite being spuriously run. Still investigating
that.


chreekat:
Anecdotally (just tested with Snowdrift), this fixes #1295.

PR #2608, opened 2016-09-18
Display more useful digest auth failures

snoyberg:
Inspired by @blaisorblade's previous comments, I enriched the
http-client-tls API a bit.


Blaisorblade:
LGTM. The current failures look like "uploaded too early", so I restarted those builds.

PR #2612, opened 2016-09-20
Reclone all-cabal-hashes if --unshallow fails

### LINKED ISSUE #2602: Transitioning to a full clone of package index can fail
Example (from https://travis-ci.org/fpco/stackage/builds/160335230):

```
Shallow package index repo detected, transitioning to a full clone...
Updating package index Hackage (mirrored at https://github.com/commercialhaskell/all-cabal-hashes.git) ...Running /usr/bin/git fetch --unshallow in directory /home/travis/.stack/indices/Hackage/git-update/all-cabal-hashes/ exited with ExitFailure 128
fatal: error in object: unshallow 05a8ba81a30b775e3aab8269642317345708bcfa
```

The `--unshallow` option was added in git 1.8.2, released in early 2013.  It's entirely possible that Travis and other users could have an older version.

@snoyberg suggested a fallback incase of failure of deleting the directory and re-cloning.

I believe a workaround would be to delete `~/.stack/indices/Hackage/git-update/all-cabal-hashes/` and then try again.  Another workaround would be to upgrade `git`.

Blaisorblade:
Fix #2602. I only did some quick testing yet.

### DIFF
--- a/src/Stack/PackageIndex.hs
+++ b/src/Stack/PackageIndex.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE GADTs                      #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE KindSignatures             #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE PatternGuards              #-}
@@ -247,12 +248,19 @@ updateIndexGit menv indexName' index gitUrl = do
                   $(mkRelDir "git-update")
                 acfDir = suDir </> repoName
             repoExists <- doesDirExist acfDir
-            unless repoExists
-                   (readProcessNull (Just suDir) menv "git" cloneArgs)
+            let doClone = readProcessNull (Just suDir) menv "git" cloneArgs
+            unless repoExists doClone
             isShallow <- doesFileExist $ acfDir </> $(mkRelDir ".git") </> $(mkRelFile "shallow")
             when isShallow $ do
               $logWarn "Shallow package index repo detected, transitioning to a full clone..."
-              readProcessNull (Just acfDir) menv "git" ["fetch", "--unshallow"]
+              let handleUnshallowError =
+                    C.handle $ \case
+                      ReadProcessException{} -> do
+                        $logInfo $ "Failed to convert to full clone, deleting and re-cloning."
+                        ignoringAbsence (removeDirRecur acfDir)
+                        doClone
+                      err -> throwM err
+              handleUnshallowError (readProcessNull (Just acfDir) menv "git" ["fetch", "--unshallow"])
             $logSticky "Fetching package index ..."
             let runFetch = callProcessInheritStderrStdout
                     (Cmd (Just acfDir) "git" menv ["fetch","--tags"])


src/Stack/PackageIndex.hs
```diff
@@ -247,12 +248,19 @@ updateIndexGit menv indexName' index gitUrl = do
...
             when isShallow $ do
               $logWarn "Shallow package index repo detected, transitioning to a full clone..."
-              readProcessNull (Just acfDir) menv "git" ["fetch", "--unshallow"]
+              let handleUnshallowError =
+                    C.handle $ \case
+                      ReadProcessException{} -> do
+                        $logInfo $ "Failed to convert to full clone, deleting and re-cloning."
+                        ignoringAbsence (removeDirRecur acfDir)
+                        doClone
+                      err -> throwM err
```
sjakobi: I don't think this catch-all case changes anything. I think it can be dropped.

Blaisorblade: `ReadProcessException` is one data constructor of for in type `ReadProcessException`, so this catch all prevents a exhaustivity warning (which risks a match error at runtime). IIUC you assumed `ReadProcessException` did not have alternatives, is that because it violates some naming pattern?
```
Warning:
    Pattern match(es) are non-exhaustive
    In a case alternative:
        Patterns not matched:
            NoPathFound
            ExecutableNotFound _ _
            ExecutableNotFoundAt _
```

sjakobi: Oh ok, I simply forgot to consider the possibility that `ReadProcessException` could be only one of several constructors for the same type.

I guess a constructor with the same name as the the type does suggest that it is the only constructor but I'm not aware of any preexisting convention. But yeah, either the constructor name or the type name should be changed.

sjakobi: > But yeah, either the constructor name or the type name should be changed.

I can fix that myself.

sjakobi:
I've tested this locally with git-1.8.1.5 and Ubuntu-16.04:

```
~ $ git --version
git version 1.8.1.5
~ $ stack-1.1.2 --stack-root ~/.tmp-sr path --stack-root
Using latest snapshot resolver: lts-7.0
Writing implicit global project config file to: /home/simon/.tmp-sr/global-project/stack.yaml
Note: You can change the snapshot via the resolver field there.
Downloaded lts-7.0 build plan.    
Fetching package index ...remote: Counting objects: 227057, done.
remote: Compressing objects: 100% (183201/183201), done.
remote: Total 227057 (delta 60486), reused 179602 (delta 41523), pack-reused 0
Empfange Objekte: 100% (227057/227057), 53.44 MiB | 3.63 MiB/s, done.
Löse Unterschiede auf: 100% (60486/60486), vervollständigt mit 1 lokalen Objekten.
Von https://github.com/commercialhaskell/all-cabal-hashes
 * [neue Markierung] current-hackage -> current-hackage
Fetched package index.    
Populated index cache.    
Did not find .cabal file for th-desugar-1.6 with Git SHA of 61d07668c742d35133c1e3e786e8682d77cecb25
Right Nothing
...
Compiler version mismatched, found ghc-7.10.3 (x86_64), but expected minor version match with ghc-8.0.1 (x86_64) (based on resolver setting in /home/simon/.tmp-sr/global-project/stack.yaml).
Try running "stack setup" to install the correct GHC into /home/simon/.tmp-sr/programs/x86_64-linux/
~ $ stack-fallback-reclone --stack-root ~/.tmp-sr update
Shallow package index repo detected, transitioning to a full clone...                                     
Failed to convert to full clone, deleting and re-cloning.                                                 
Fetched package index.                                                                                    
```

Would this already be a candidate for `v1.2.2` with the purpose of preventing more people from running into the unshallow clone issue on Travis?

src/Stack/PackageIndex.hs
sjakobi: > > But yeah, either the constructor name or the type name should be changed.

> I can fix that myself.

Fixed in https://github.com/commercialhaskell/stack/commit/7a54085361260a469815b9aa408264bee4cd68e4.

PR #2636, opened 2016-09-26
Add PID-1 support for stack exec

snoyberg:
This improves behavior of stack exec in the Docker use case, ensuring
signals are properly handled and orphan processes are reaped.

### DIFF
--- a/src/Stack/Exec.hs
+++ b/src/Stack/Exec.hs
@@ -25,7 +25,7 @@ import           System.Process.Run (callProcess, callProcessObserveStdout, Cmd(
 #ifdef WINDOWS
 import           System.Process.Read (EnvOverride)
 #else
-import           System.Posix.Process (executeFile)
+import qualified System.Process.PID1 as PID1
 import           System.Process.Read (EnvOverride, envHelper, preProcess)
 #endif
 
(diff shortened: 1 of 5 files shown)


src/Stack/PID1.hs
```diff
@@ -0,0 +1,131 @@
...
+    (Nothing, Nothing, Nothing, ph) <- createProcess (proc cmd args)
+        { env = environment
+        }
+
+    -- Determine the child PID. We want to exit once this child
+    -- process is dead.
+    p_ <- modifyProcessHandle ph $ \p_ -> return (p_, p_)
+    child <-
+        case p_ of
+            ClosedHandle e -> assert False (exitWith e)
```
mgsloan: Is the `assert False` necessary?  exitWith already essentially returns `IO void`

mgsloan:
LGTM!  Perhaps @borsboom would also like to review?  This is interesting stuff!

src/Stack/PID1.hs
borsboom: I think the idea is that it should be impossible to reach that code, so if it does that should fail with prejudice.

borsboom:
@snoyberg is going to extract the Pid1 stuff to a separate package so that we can use it both within Stack and in a standalone executable.

mgsloan:
Fantastic, makes sense to me!

snoyberg:
Alright, I've moved over to the pid1 package for this, which significantly simplifies the PR. I may need to restart some builds if pid1 hasn't made it into the indices by the time tests run.

PR #2653, opened 2016-09-30
Download snapshots.json from S3

snoyberg:
S3 is more reliable than stackage.org. See discussion:
https://www.reddit.com/r/haskell/comments/556a43/stackageorg_down/d8818rh


sjakobi:
Noob question: Does it matter that S3 serves the file with `Content-Type: binary/octet-stream` while stackage used `Content-Type: application/json; charset=utf-8`?

A changelog entry would be good too, but I can add one myself.

sjakobi:
Oh, and the Travis failure can be ignored. The compile failure was introduced elsewhere and has since been fixed.

mgsloan:
LGTM!

snoyberg:
It won't break this, but it's worth fixing anyway, I'll add a note for
myself.

PR #2658, opened 2016-10-01
Parse .dump-hi files of library components only once

sjakobi:
This fixes a performance regression introduced in 5f3ffe5ea8b4d8d1f4d1e4d15bee8b481ec6a8d4.

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -734,7 +734,7 @@ libraryFiles lib = do
         resolveFilesAndDeps
             Nothing
             (dirs ++ [dir])
-            (names <> exposed)
+            names
             haskellModuleExts
     cfiles <- buildOtherSources build
     return (modules, files <> cfiles, warnings)
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -54,6 +54,9 @@ Bug fixes:
   [#2568](https://github.com/commercialhaskell/stack/issues/2568)
 * Fixed running `stack hpc report` on package targets.
   [#2664](https://github.com/commercialhaskell/stack/issues/2664)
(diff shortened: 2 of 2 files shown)


sjakobi:
With this change a null build of the [amazonka](https://github.com/brendanhay/amazonka) project (78 packages) takes ~14s on my machine vs. ~22s with v1.2.0.

src/Stack/Package.hs
```diff
@@ -734,14 +734,14 @@ libraryFiles lib = do
         resolveFilesAndDeps
             Nothing
             (dirs ++ [dir])
-            (names <> exposed)
+            names
```
sjakobi: This line alone would be a sufficient fix (see `names = bnames ++ exposed`!) but I felt that `Set` is the better datastructure than `[]` in this case. Order ~~isn't~~ doesn't seem to be important and we don't want any duplicates.

I actually think that we should use `Set` instead of `[]` much more pervasively to prevent ourselves from wondering whether duplicates and order actually matter in each usecase.

mgsloan: My reasoning for not having it be a Set is that there is cost to constructing the Set, and we always convert it from / to a list.  It is semantically more correct, though, indeed!

mgsloan:
I am fine with the use of Set, but are you sure it's worth it?  Seems to me like it's just extra work at runtime.  It does encode more invariants (no duplicates), but I imagine it will cost some performance.

sjakobi:
I have undone the switch to `Set` for now, will move the discussion to a separate issue.

mgsloan:
Thanks, glad this is fixed! :D

Issue #2592, opened 2016-09-13
Add 'style' env to Travis to add hlint-ing for stack code


lwm:
Following from https://github.com/commercialhaskell/stack/pull/2591#issuecomment-246676805, as recommended by @Blaisorblade.

I'd like to see if there is consensus amongst contributors to include hlint as a quality assurance tool for stack. There are currently 355 hints to deal with when running:

```
  $ find src -name "*hs" -exec hlint {} + ;
```

If people think this is a good change (IMHO, it is ;)), then I will do the following:

  - [x] Remove all warnings (let's see how it goes)
    - #2606, #2615, #2690
  - [x] Document a low overhead way to work with hlint and the stack codebase
    - https://github.com/commercialhaskell/stack/pull/2699
  - [x] Rebase #2591 and remove `allowed_failures` config

Blaisorblade:
Here's my 👍 , especially since you volunteer to fix this. If/when you get started on warnings, please make sure to separate potential behavioral changes, like ones that alter laziness:

```
src/Data/Attoparsec/Interpreter.hs:150:14: Suggestion: Use null
Found:
  length ctxs == 0
Why not:
  null ctxs
Note: increases laziness
```

Blaisorblade:
Pinging @sjakobi, @mgsloan, @harendra-kumar and @borsboom for opinions on addressing `hlint` warnings — @lwm is volunteering to do the work and already started in #2588.

In case there's no objection for a while, you might want to get started. As far as I can tell, we usually don't wait for explicit unanymous agreement for every little detail 😉 
https://github.com/commercialhaskell/stack/pull/2591#issuecomment-246679510

Also, googling suggests there's not so much controversy over hlint itself—people debate warning categories though. [Quoting hlint's author](http://neilmitchell.blogspot.de/2009/09/how-i-use-hlint.html):

> I do not intend people to slavishly apply the hints given by HLint - each hint should be carefully considered. For example, the darcs project uses HLint, but has decided that they are not interested in eta reduction hints, so have used HLint's ignoring facility.

lwm:
@Blaisorblade 

Thanks for the link. It's helpful. As with https://github.com/commercialhaskell/stack/pull/2606, I will move through each type of suggestion/warning and see what I can apply programmatically (`hlint --refactor --refactor-options -i`) and what I can reason about and manually do. Anything that seems questionable, we can start to use the ignore facility.

Blaisorblade:
Fixed by #2591 I guess.

PR #2659, opened 2016-10-01
Resolve HLint 'Use Foo' hints

lwm:
Some questionable, some not. Shout and I will ignore them. 

Refer to [HLint manual](http://community.haskell.org/~ndm/darcs/hlint/hlint.htm) for configuration/ignore syntax.

Nearly there. Just another PR to go and then #2591 can move forward.

Ref: https://github.com/commercialhaskell/stack/issues/2592

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -29,7 +29,7 @@ import           Control.Monad.IO.Class (MonadIO,liftIO)
 import           Control.Monad.Logger (MonadLogger,logError,logInfo,logWarn)
 import           Control.Monad.Reader (MonadReader,asks,runReaderT)
 import           Control.Monad.Writer (execWriter,runWriter,tell)
-import           Control.Monad.Trans.Control (MonadBaseControl)
+import           Control.Monad.Trans.Control (MonadBaseControl, liftBaseWith)
 import qualified "cryptohash" Crypto.Hash as Hash
 import           Data.Aeson.Extended (FromJSON(..),(.:),(.:?),(.!=),eitherDecode)
 import           Data.ByteString.Builder (stringUtf8,charUtf8,toLazyByteString)
@@ -82,7 +82,6 @@ import           Text.Printf (printf)
 
 #ifndef WINDOWS
 import           Control.Concurrent (threadDelay)
-import           Control.Monad.Trans.Control (liftBaseWith)
 import           System.Posix.Signals
 import qualified System.Posix.User as PosixUser
 #endif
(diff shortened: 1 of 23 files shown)


src/Stack/Types/FlagName.hs
```diff
@@ -54,7 +54,7 @@ newtype FlagName =
   FlagName Text
   deriving (Typeable,Data,Generic,Hashable,Store,NFData)
 instance Eq FlagName where
-    x == y = compare x y == EQ
+    x == y = x == y
```
sjakobi: It's funny that hlint would suggest this. :)

I'd feel safer with the old version

test/integration/lib/StackTest.hs
```diff
@@ -156,9 +148,7 @@ fileContentsMatch f1 f2 = do
     doesExist f2
     f1Contents <- readFile f1
     f2Contents <- readFile f2
-    if f1Contents == f2Contents
-          then return ()
-          else error
+    unless (f1Contents == f2Contents) $ error
                    ("contents do not match for " ++
                     show f1 ++
                     " " ++
```
sjakobi: The indentation should be reduced here.

src/Stack/Types/FlagName.hs
mgsloan: Yes, this will lead to an infinite loop

src/Stack/Docker.hs
```diff
@@ -720,7 +719,7 @@ checkDockerVersion envOverride docker =
        _ -> throwM InvalidVersionOutputException
   where minimumDockerVersion = $(mkVersion "1.6.0")
         prohibitedDockerVersions = []
-        stripVersion v = fst $ break (== '-') $ dropWhileEnd (not . isDigit) v
+        stripVersion v = takeWhile (not . (== '-')) (dropWhileEnd (not . isDigit) v)
```
lwm: Done!

test/integration/lib/StackTest.hs
lwm: Nice one, fixed up.

lwm:
Addressed the comments and squashed them. 

I can ignore `Use ++` hints if you rethink and feel strongly on that @sjakobi.

mgsloan:
LGTM!  I added a commit to fix warnings + build on 7.8

lwm:
Great! Thanks @mgsloan :rocket:

sjakobi:
I think it would be better to merge PRs that don't pass CI into a different branch than master and fix them there. Developing based on a master branch that won't compile is no fun.

lwm:
> Developing based on a master branch that won't compile is no fun.

Agreed!!!

mgsloan:
True, but in this case it compiled, just didn't pass travis.  I am fine with having master states that don't pass travis, but I realize that this varies from some other people's workflow.

lwm:
Oh, you don't use protected branches? Is that even less fun?

mgsloan:
I just like to minimize overall development friction, and sometimes that means breaking travis occasionally.  We can certainly move to a more PR-centric development workflow if it becomes a problem.

Issue #2657, opened 2016-10-01
Issue with corruption in downloaded executable


rdnetto:
### General summary/comments (optional)

I just spent half a day trying to figure out why Stack was failing with exit code 139 and no output when Docker support was enabled, and eventually tracked it down to `~/.stack/programs/x86_64-linux/stack-1.1.2/stack` being corrupted (`file` and `ldd` couldn't parse the header). Deleting the folder and tarball fixed this, but the failure was pretty non-obvious, and I can easily see someone not as familiar with strace as me getting stuck on it.

I haven't read the source, but I suspect that computing a hash of the tarball before extracting it would prevent this from happening. Ideally there'd be a more helpful error message, but I don't 

### Steps to reproduce
* Add `docker: enable: true` to a project
* Corrupt `~/.stack/programs/x86_64-linux/stack-1.1.2/stack` in some way

### Expected
An error message telling the user to re-run `stack setup`, or Stack automatically re-downloading the file.

### Actual
Stack silently terminates with exit code 139.

### Stack version

```
$ stack --version
Version 1.1.2 x86_64 hpack-0.14.1
```

### Method of installation
* Ebuild from the [Haskell overlay](git://github.com/gentoo-haskell/gentoo-haskell.git)

mgsloan:
Thanks for the report!  It does indeed check the hash of the file and length - see this file https://github.com/fpco/stackage-content/blob/master/stack/stack-setup-2.yaml

Perhaps the issue is something like stack getting killed while writing out this file.

rdnetto:
So, this is interesting: for some reason the bad copy of the tarball is *larger* than the good copy.

Good copy of stack-1.1.2.tar.gz:
```
SHA1: 7147c2c912e7c12ebe81c6f3d3cf1e128f7b40fb
File size: 9.1M
```

Bad copy:
```
SHA1: 07f86433c3a4d4d42b05646327ec75e407d4f190
File size: 16M
```

The bad copy of the tarball can't be read by tar, and its extracted directory only contains the corrupted stack executable (there are other files in the good copy of the tarball).

The start of both files is the same, but they differ from 0x0002623e, where the corrupted version seems to 'reset' and goes back to the start of the file. My guess is it tried to resume downloading after the connection dropped out / it was killed, but the server didn't resume from the point the client expected it to. This also happens again at 0x006176bb.

[...]

PR #2661, opened 2016-10-01
Fix retry of file download to not append #2657

mgsloan:
@snoyberg @borsboom Please review, I believe this has been a bad longstanding bug in the download code.  When retry occurs, it continues appending to the same file.  The fix is simple - have `withBinaryFile` be within the retry.  Since it uses `WriteMode`, writing will restart from the beginning.


snoyberg:
Good catch

PR #2668, opened 2016-10-03
'custom snapshot' link to use docs instead of wiki

samdoshi:


sjakobi:
Thanks! I have cherry-picked your commit onto the `stable` branch where we keep the docs of the last released version of stack! We should document this better!

mgsloan:
Thanks good catch @samdoshi !  Thanks for cherry-picking to stable, @sjakobi !

Issue #2650, opened 2016-09-28
Systematic cleanup and colorization of stack output


mgsloan:
This issue is tracking doing a systematic sweep of the stack code and converting it to use the pretty-printer where appropriate.

I added some pretty print machinery to stack.  In particular, we now have the following TH functions:
- `$prettyDebug` is just like `$logDebug`, except it expects an `AnsiDoc` value.
- `$prettyInfo` is just like `$logInfo`, except it expects an `AnsiDoc` value.
- `$prettyWarn` is similar to `$logWarn`, but it prepends a newline and "Warning: " colored yellow.
- `$prettyError` is similar to `$logError`, but it prepends a newline and "Error: " colored red.

One thing to note is that I would prefer colorization to be consistent across all of the stack messages.  So, for the most part, colors should not be directly used in the stack source.  Instead, add type and context specific utilities to `Stack.PrettyPrint.  For example, I have these ones for plan construction:

``` haskell
displayTargetPkgId :: PackageIdentifier -> AnsiDoc
displayTargetPkgId = cyan . display

displayCurrentPkgId :: PackageIdentifier -> AnsiDoc
displayCurrentPkgId = yellow . display

displayErrorPkgId :: PackageIdentifier -> AnsiDoc
displayErrorPkgId = errorRed . display
```

Related issues that this work would ideally cover:
- [ ] Unclear error message for invalid extra-package-dbs #1738"
- [ ] Add a proper system for stack warnings suppression / collection / display #1745"
- [ ] Pretty print long commands #1836
- [ ] Better formatting and parameter documentation when initializing a project #1911
- [ ] Improve error messages for yaml parse errors #2374
- [ ] Improve error message for bad build targets #2384
- [ ] Prepend warnings with "Warning:" outside verbose mode so they're not taken as errors #2470
- [ ] Improve error/warning reporting in `stack sdist` and `stack upload` #2507

mgsloan:
Assigning myself, but if someone is interested in learning this pretty print stuff and doing the work, feel free to take it over / pitch in!

Are you interested @luigy ?  I know you were interested in improving some aspects of stack's CLI display.

kadoban:
I think I'd be interested enough to take on at least some of this. Is there a branch or any context I should look at, or just start in `s/S/PrettyPrint.hs` and look for places it should be used and etc.?

mgsloan:
[...]

PR #2683, opened 2016-10-06
Try to limit line lengths in exception messages to 100 chars

sjakobi:

This is just an experiment to improve the formatting of exception messages which can sometimes get quite long.

Further refinements or other suggestions are welcome.


sjakobi:
On second thought, I think this would break any kind of indentation.

Leaving it open for now in case someone wants to discuss.

mgsloan:
I think it makes sense to discuss this in the context of https://github.com/commercialhaskell/stack/issues/2650 .  Closing for now, but it's a good idea!

I would prefer that we switch over to pretty printing for everything, including exceptions.

PR #2695, opened 2016-10-10
Pass -threaded when building Setup.hs

enolan:
Cabal-install builds `Setup.hs` files with `-threaded`, this patch changes Stack so it does too. This makes it possible to have concurrent build systems inside `Setup.hs`. Found in the context of ndmitchell/shake#490.

This was real quick and I don't know Stack internals, so it might be wrong. Tests pass, and my test in the linked issue works passes now.


enolan:
The OS X Travis build actually succeeded "Done. Your build exited with 0." but something else went wrong. Says "running `casher push` took longer than 180 seconds and has been aborted" and "/Users/travis/build.sh: line 109: shell_session_update: command not found".

mgsloan:
Makes sense to me, thanks!

Issue #1364, opened 2015-11-18
Avoid building the target packages with "stack ghci"


mgsloan:
In order to make `stack ghci` reliable / consistent, the `cabal_macros.h` and `Paths_*` files need to be generated.  Currently, this means that we do a full build of the targets (unless `--no-build` is passed).

Really all that needs to happen is to run [`initialBuildSteps`](https://github.com/haskell/cabal/blob/4e8733632a52d386ab4893af31d87e350a0c657c/Cabal/Distribution/Simple/Build.hs#L556) before running ghci.  Running this function requires `LocalBuildInfo`, which gets stored in `dist/setup-config`.  Unfortunately, we can't call it directly from stack - it uses Binary / Generic, so we can't rely on it being compatible between cabal versions (and stack may be linked with a different version).

I think a reasonable solution to this is to have a separate program which runs `initialBuildSteps` on the specified packages.  This separate program could either be a script that gets `runhaskell`-ed, or something that gets build and installed to the snapshot (like the default Setup.hs).

This solution also has relevance to other issues:

* It's also the information necessary to allow Setup.hs to modify the cabal info (see https://github.com/commercialhaskell/stack/issues/970).  Doing this would be more complicated.  I think the best way to do this would be to serialize `localPkgDescr` as a .cabal file (I really hope this is possible).  Then, we'd would parse the resulting cabal file, and use that description to generate the arguments passed to ghci.

* This may also allow us to avoid doing full rebuilds for haddock - I noted that in cabal this function [is also called by the haddock stuff](https://github.com/haskell/cabal/blob/945dbc4ccb46a60a3854be9b19d00109b093900a/Cabal/Distribution/Simple/Haddock.hs#L211).

hesselink:
Another reason this would be very useful (IIUC) is an issue I just ran into. If you have two packages, A and B, and A depends on B, then running `ghci A B` will now build B. This means that is B has type errors, you won't get a ghci prompt! Often the reason I start ghci is because I'm developing package B, so getting type errors is expected.

Let me know if this is not related, and I'll file a new issue.

mgsloan:
Hmm, that is surprising.  Is this with stack >= 0.1.8?  Currently it should carry on running ghci despite build failures.

hesselink:
That was with stack 0.1.6. I'll try with 0.1.8 and HEAD tomorrow.

hesselink:
[...]

PR #2697, opened 2016-10-10
Use a shim to invoke initialBuildSteps for ghci #1364

mgsloan:
Here's how this works:

* Outputs a `StackSetupShim.hs` file to the build's temporary directory.

* Builds both simple setup and custom setup files using this file as an input, along with `-main-is: StackSetupShim.mainOverride`.

* When we want to run the `initialBuildSteps` cabal function, we run the setup exe with the arguments `repl stack-initial-build-steps`.  This condition is special-cased by the shim, and causes the function we want to be run.  Any other invocations of the executable get passed through to the standard simple / default setup.

In summary, this is a bit of a hack, but I think it works perfectly in all cases we care about.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -51,7 +51,6 @@ Other enhancements:
 * `stack dot` and `stack list-dependencies` now take targets and flags.
   [#1919](https://github.com/commercialhaskell/stack/issues/1919)
 
-
 Bug fixes:
 
 * Fixed a gnarly bug where programs and package tarballs sometimes have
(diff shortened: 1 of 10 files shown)


mgsloan:
Pinging @chrisdone - this should be a major improvement for intero as well!  What do you think of this approach?

chrisdone:
Sounds good. I looked through the patch. I'm not too fresh on Stack's codebase at the moment for it to be a good review, but the approach is okay (with the one problem that anyone's custom build repl hooks won't be launched, it seems to me) and yet will yield a smoother user experience.

mgsloan:
Cool!  I think this approach is solid, so merging.

> with the one problem that anyone's custom build repl hooks won't be launched, it seems to me

Yeah, if somehow the user uses the repl hook to affect the initialBuildSteps, then that won't be applied. I don't anyone does that, though.  Ordinary repl commands / hook will still work, just not integrated with `stack ghci`.

PR #2703, opened 2016-10-11
Deprecate `stack-setup-yaml`

### LINKED ISSUE #2647: Improve UI for `stack setup --stack-setup-yaml` option
Is the `--stack-setup-yaml` flag ignored by `stack setup`? If I pass something with that flag, it doesn't seem to affect the operation. The only place I see `--stack-setup-yaml` mentioned is in the [Changlog entry for v0.1.5.0](https://docs.haskellstack.org/en/stable/ChangeLog/#0150).

The reason I'm bringing this up is that I was confused for a bit trying to specify a different `stack.yaml` file other than the default. I eventually realized it should be `stack --stack-yaml <file> setup`. So having both `stack setup --stack-setup-yaml <file>` and `stack --stack-yaml <file> setup` is confusing. If `--stack-setup-yaml` is not being used, I think it should be removed.

lwm:
Closes #2647. Hope that covers it.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -60,6 +60,8 @@ Other enhancements:
   [#2653](https://github.com/commercialhaskell/stack/pull/2653)
 * `stack dot` and `stack list-dependencies` now take targets and flags.
   [#1919](https://github.com/commercialhaskell/stack/issues/1919)
+* Deprecate `stack setup --stack-setup-yaml` for `--setup-info-yaml` based
+  on discussion in [#2647](https://github.com/commercialhaskell/stack/issues/2647).
 
 Bug fixes:
 
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -479,6 +479,12 @@ setup-info:
         url: "https://example.com/ghc-7.10.2-i386-unknown-mingw32-foo.tar.xz"
 ```
 
+Or without using `ghc-variant`:
+
+```yaml
+setup-info: "https://raw.githubusercontent.com/fpco/stackage-content/master/stack/stack-setup-2.yaml"
+```
+
 `url` may be either URL or (since 1.2.0) absolute file path.
 
 ### pvp-bounds
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -17,7 +17,7 @@ module Stack.Setup
   , ensureDockerStackExe
   , getSystemCompiler
   , SetupOpts (..)
-  , defaultStackSetupYaml
+  , defaultSetupInfoYaml
   , removeHaskellEnvVars
   ) where
 
@@ -102,8 +102,8 @@ import              System.Process.Run (runCmd, Cmd(..))
 import              Text.Printf (printf)
 
 -- | Default location of the stack-setup.yaml file
-defaultStackSetupYaml :: String
-defaultStackSetupYaml =
+defaultSetupInfoYaml :: String
+defaultSetupInfoYaml =
     "https://raw.githubusercontent.com/fpco/stackage-content/master/stack/stack-setup-2.yaml"
 
 data SetupOpts = SetupOpts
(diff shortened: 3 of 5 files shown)


lwm:
Oh wait, it doesn't compile .... :face-palm:

Blaisorblade:
Deprecating `--stack-setup-yaml` should mean still supporting it (with some sort of warning in the help text and at runtime), but the code doesn't seem to do that... In particular I doubt you want to have a separate field—instead of `SetupCmdOpts <$>` I'd use a smart constructor for `SetupCmdOpts`.

lwm:
Thanks for tips @Blaisorblade but I am a little confused.

Regarding what the `stack setup --help` should look like, would it be:

```
λ  ~  stack setup --help
Usage: stack setup [GHC_VERSION] [--[no-]reinstall] [--[no-]upgrade-cabal]
                   [--stack-setup-yaml ARG] [--ghc-bindist URL] [--help]
  Get the appropriate GHC for your project
Available options:
  --stack-setup-yaml ARG   <deprecation notice>
  --setup-info-yaml URL   <new explanation>
...
```

Not sure how I can achieve this with a smart constructor given I need a new field to represent both? Or should the `--help` just show `--setup-info-yaml` and the code will still accept `--stack-setup-yaml` and output a runtime warning.

Blaisorblade:
I'd have said expected `URL`, not `ARG` in  `--stack-setup-yaml URL`. But otherwise yes, that's the message that makes sense to me.

Disclaimer: I haven't tried this out, but at least here's what I imagined—with a bit more detail added. Feel free to ignore tips that don't actually make sense (I'm also learning optparse-applicative 😉 ).

> Or should the --help just show --setup-info-yaml and the code will still accept --stack-setup-yaml and output a runtime warning.

No, deprecations are usually announced, just like in API—how do you study existing scripts otherwise? I also wouldn't know how to achieve that.

> Not sure how I can achieve this with a smart constructor given I need a new field to represent both? 

Well, why a new field? Again, that sounds bad. What I had in mind was (ignoring other fields):

```
setupCmdOpts :: Maybe String -> Maybe String -> SetupCmdOpts
setupCmdOpts stackSetupYaml setupInfoYaml = SetupCmdOpts (stackSetupYaml <|> setupInfoYaml <|> Just defaultStackSetupYaml).

setupCmdOpts <$> ... <*> ...
```

I wondered how to build parsers for Option String though, but that seems doable through [`optional`](https://www.stackage.org/haddock/lts-6.14/base-4.8.2.0/Control-Applicative.html#v:optional)—it has the right type and it seems to have the right behavior.

And to get the right output, it's probably best to leave the default in `setupInfoYaml`:

```
setupCmdOpts :: Maybe String -> String -> SetupCmdOpts
setupCmdOpts stackSetupYaml setupInfoYaml = SetupCmdOpts (stackSetupYaml <|> setupInfoYaml).

setupParser = setupCmdOpts
    <$> OA.optional (OA.strOption
            ( OA.long "stack-setup-yaml"
           <> OA.metavar "URL"
           <> OA.help "DEPRECATED in favour of 'setup-info-yaml'"))
    <*> OA.strOption
            ( OA.long "setup-info-yaml"
           <> OA.metavar "URL"
           <> OA.help "Alternate URL or absolute file path for various tools stack relies on (GHC, msys2)"
           <> OA.value defaultStackSetupYaml
           <> OA.showDefault )
```

Blaisorblade:
Forgot: the above doesn't allow per se a runtime warning. If needed we might need to store a boolean flag in `SetupCmdOpts`—but that could be computed as `isJust stackSetupYaml`.

Blaisorblade:
BTW, what I proposed is more complicated than having separate fields for the options. The problem with having two fields is making sure they're *not treated differently* by the command—so at some point `stackSetupYaml <|> setupInfoYaml` must happen, and preferably early enough that most of the code can't possibly forget it—that's a facet of designing types to prevent bugs.

Generally speaking, the option parser is IMHO still the best place, but having two fields in a new type `SetupCmdOpts'` from which one produces a `SetupCmdOpts` would not IMHO be absurd.

Finally: `stackSetupYaml <|> setupInfoYaml` gives priority to `stackSetupYaml`, not sure we want that O_O. Boy is this tricky.

lwm:
Thanks for this brain dump!!! I will work towards something useful :)

lwm:
@Blaisorblade Is skipping the run time warning acceptable? Given that `OA.value` gurantees a value is present, it's hard to detect when the user passed `--stack-setup-yaml`?

Blaisorblade:
Not 100% sure on that, that's why I asked in #2647. I just realized that *you* answered me...
I'd wait a bit in case others want to answer, but I guess I'm fine with it, since we're not in a hurry to drop the deprecated options. (I'm not sure there's a pattern).

lwm:
> I just realized that you answered me...

:grinning:

mgsloan:
LGTM, thanks!

PR #2723, opened 2016-10-20
Add extra variables in the hastache context

gxmas:
When a cabal file contains a `data-files` attribute, `cabal` generates a path module.  The module name includes the package name with hyphens replaced by underscores.  Currently `stack new` only adds the variables `name` and `year` in the hastache context available in stack templates.  That breaks templates that include a `data-files` attribute when a package name contains hyphens is passed to `stack new`.

Adding a variable (name-as-varid) in the hastache context to be able to reliably be able to support such cases.  Also add another variable (name-as-module) to use the package name as a module name.


mgsloan:
Seems sensible to me!

PR #2731, opened 2016-10-24
Replace script note with built in.

lwm:


mgsloan:
Sensible, though I think this will wait less time than travis long.  If people encounter issues with it, then they can read the doc and increase the wait time.

PR #2736, opened 2016-10-24
Remove confusing default. 

### LINKED ISSUE #2700: `{}` in config.yaml is confusing
I've often seen `{}` as a default in config.yaml. That's confusing—assuming that comes from Stack-generated config, is there something else we can use going forward by default for an empty config? I guess that's allowed because you could use JSON-like map literals, but that's not an intentional part of Stack's interface...

I've filed this issue because I was confused by this, and others might be too—see https://github.com/commercialhaskell/stack/issues/2575#issuecomment-252951103 (not arguing "that user is confused", just that he might be, and I'd like to reduce the possibility).

lwm:
Closes #2700.

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -905,9 +905,7 @@ getDefaultUserConfigPath stackRoot = do
         liftIO $ S.writeFile (toFilePath path) $ S.concat
             [ "# This file contains default non-project-specific settings for 'stack', used\n"
             , "# in all projects.  For more information about stack's configuration, see\n"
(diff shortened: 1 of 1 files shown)


mgsloan:
Hmm, my comment on #2700 is actually wrong.  It's a syntactic error to have a file that just has comments.  Thanks for having a swing at this, though!

I've changed the default to be a non-empty document, but still equivalent to an empty configuration.  It now encourages you to provide values for various template parameters.  https://github.com/commercialhaskell/stack/commit/e4092597ca6238022437192057512f384d33e2dc

lwm:
Oh, that's much better. Nice one.

PR #2739, opened 2016-10-24
Redirect stdout of docker run to stderr

### LINKED ISSUE #2733: stack path --docker-auto-pull prints docker output to stdout, making the result hard to parse
When running `stack path --docker-auto-pull` if the docker image hasn't been pulled the output of `docker pull` is printed to `stdout`, with the actual data of `stack path` printed at the end.

I think it would be more appropriate to print this kind of output to `stderr`, to keep the `stdout` output reliable and amenable to parsing.

bitonic:
Fixes #2733 .


mgsloan:
LGTM!

PR #2740, opened 2016-10-25
Bump http-client version to fix broken https proxy functionality

robbinch:
Not sure if I'm doing this correctly, but http-client version needs to be bumped, else https proxy does not work. See https://github.com/snoyberg/http-client/pull/235.


mgsloan:
Yep, that's the correct way to do it.  Thanks!

Issue #2678, opened 2016-10-05
`stack config set resolver` doesn't accept `ghc-X` type resolvers


sjakobi:
```
~ $ stack config set resolver ghc-7.10.3
Invalid snapshot name: ghc-7.10.3
~ $ stack config set resolver ghc-7.10
Invalid snapshot name: ghc-7.10
~ $ stack config set resolver ghc-7
Invalid snapshot name: ghc-7
```

The error probably comes from [this line](https://github.com/commercialhaskell/stack/blob/a11c2c2c43f779c6fe7086a2e4c8d974305609a0/src/Stack/ConfigCmd.hs#L87).

Once this is fixed, the help text for the `RESOLVER` argument should be enhanced to suggest e.g. `ghc-7.8` as a valid argument.

mgsloan:
Agreed!  Marking this as a P2 bug!

mgsloan:
Fixed!

PR #2747, opened 2016-10-28
Allow `stack config set resolver` to accept ghc-X resolvers.

lwm:
~~What would `defaultGhcBuildPlans` be? Doing a lot of guessing in this PR ;)~~

~~Also, I don't see any tests for `parseSnapName`. Should I add some?~~

Closes https://github.com/commercialhaskell/stack/issues/2678.

### DIFF
--- a/src/Stack/Types/BuildPlan.hs
+++ b/src/Stack/Types/BuildPlan.hs
@@ -10,6 +10,7 @@
 module Stack.Types.BuildPlan
     ( -- * Types
       BuildPlan (..)
+    , BuildPlanTypesException
     , PackagePlan (..)
     , PackageConstraints (..)
     , TestState (..)
--- a/HLint.hs
+++ b/HLint.hs
@@ -34,3 +34,4 @@ ignore "Use fmap"  -- specific to GHC 7.8 compat
 ignore "Parse error"  -- we trust the compiler over HLint
 ignore "Use ==" -- Creates infinite loops in `EQ` using expressions
 ignore "Evaluate"
+ignore "Use unless"
--- a/src/Stack/ConfigCmd.hs
+++ b/src/Stack/ConfigCmd.hs
@@ -14,7 +14,7 @@ module Stack.ConfigCmd
 
 import           Control.Applicative
 import           Control.Monad
-import           Control.Monad.Catch (throwM)
+import           Control.Monad.Catch (throwM, try)
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
 import           Control.Monad.Reader (asks)
@@ -77,14 +77,15 @@ cfgCmdSet cmd = do
             $logInfo (T.pack configFilePath <> " has been updated.")
 
 cfgCmdSetValue
-    :: (StackMiniM env m, HasConfig env, HasGHCVariant env)
+    :: forall m env . (StackMiniM env m, HasConfig env, HasGHCVariant env)
     => ConfigCmdSet -> m Yaml.Value
 cfgCmdSetValue (ConfigCmdSetResolver newResolver) = do
     -- TODO: custom snapshot support?
     newResolverText <- fmap resolverName (makeConcreteResolver newResolver)
-    -- We checking here that the snapshot actually exists
-    snap <- parseSnapName newResolverText
-    _ <- loadMiniBuildPlan snap
+    _ <- try $ parseSnapName newResolverText >>= loadMiniBuildPlan
+               :: m (Either BuildPlanTypesException MiniBuildPlan)
+    _ <- try $ parseResolverText newResolverText
+               :: m (Either ConfigException Resolver)
     return (Yaml.String newResolverText)
 cfgCmdSetValue (ConfigCmdSetSystemGhc _ bool) =
     return (Yaml.Bool bool)
@@ -113,7 +114,7 @@ configCmdSetParser =
                     OA.argument
                         readAbstractResolver
                         (OA.metavar "RESOLVER" <>
-                         OA.help "E.g. \"nightly\" or \"lts-7.2\""))
+                         OA.help "E.g. \"nightly\" or \"lts-7.2\" or \"ghc-7.8\""))
                    (OA.progDesc
                         "Change the resolver of the current project. See https://docs.haskellstack.org/en/stable/yaml_configuration/#resolver for more info."))
         , OA.command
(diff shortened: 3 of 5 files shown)


lwm:
CI failure because of [2eaece](https://github.com/commercialhaskell/stack/pull/2747/files#diff-2eaece127623942969ceecfc37bd459bR23).

src/Stack/Types/BuildPlan.hs
```diff
@@ -74,6 +74,7 @@ import           Stack.Types.Version
 -- | The name of an LTS Haskell or Stackage Nightly snapshot.
 data SnapName
     = LTS !Int !Int
+    | GHC !Int !Int !Int
```
sjakobi: I'm not sure if this is the right way to go about this. Ideally, I think, the parser for `stack config set resolver` should get closer to what we have for the `resolver` field in `stack.yaml`s. So, in `Stack.ConfigCmd`, instead of using `parseSnapName` we should probably parse a `Resolver` or `AbstractResolver`, not sure about the details.

mgsloan:
Yeah, I don't think this is the right way to do it.  We've already got a `Resolver` type that fits the bill.  `ghc` resolvers aren't snapshots, and so this is why there are parts of the code where it's hard to say what belongs.

Take a look at `Stack.Types.parseResolverText`.

lwm:
OK, thanks for the review folks. I'll come up with something based on your comments!

lwm:
I think I have the right patch but I can't figure out the compiler error. Anyone? If not, I think I'll just close this one off and leave it to the pro's ;)

With diff:

```diff
diff --git a/src/Stack/ConfigCmd.hs b/src/Stack/ConfigCmd.hs
index 0218ec1..99fc675 100644
--- a/src/Stack/ConfigCmd.hs
+++ b/src/Stack/ConfigCmd.hs
@@ -14,7 +14,7 @@ module Stack.ConfigCmd
 
 import           Control.Applicative
 import           Control.Monad
-import           Control.Monad.Catch (throwM)
+import           Control.Monad.Catch (throwM, try)
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
 import           Control.Monad.Reader (asks)
@@ -82,9 +82,11 @@ cfgCmdSetValue
 cfgCmdSetValue (ConfigCmdSetResolver newResolver) = do
     -- TODO: custom snapshot support?
     newResolverText <- fmap resolverName (makeConcreteResolver newResolver)
-    -- We checking here that the snapshot actually exists
-    snap <- parseSnapName newResolverText
-    _ <- loadMiniBuildPlan snap
+
+    -- resolvers /= snapshots, so we skip parse failure
+    -- we check that snapshots exist, if parse is successful
+    _ <- try (parseSnapName newResolverText >>= loadMiniBuildPlan) :: (Either BuildPlanTypesException SnapName)
+
     return (Yaml.String newResolverText)
 cfgCmdSetValue (ConfigCmdSetSystemGhc _ bool) =
     return (Yaml.Bool bool)
diff --git a/src/Stack/Types/BuildPlan.hs b/src/Stack/Types/BuildPlan.hs
index ae568a6..25b4a6f 100644
--- a/src/Stack/Types/BuildPlan.hs
+++ b/src/Stack/Types/BuildPlan.hs
@@ -10,6 +10,7 @@
 module Stack.Types.BuildPlan
     ( -- * Types
       BuildPlan (..)
+    , BuildPlanTypesException
     , PackagePlan (..)
     , PackageConstraints (..)
     , TestState (..)
```

I get the nasty:

```
/home/lwm/hobby/stack/src/Stack/ConfigCmd.hs:79:1: error:
    Couldn't match type ‘Either BuildPlanTypesException’ with ‘IO’
      arising from a functional dependency between:
        constraint ‘Control.Monad.Base.MonadBase
                      IO (Either BuildPlanTypesException)’
          arising from the type signature for:
                         cfgCmdSetValue :: (StackMiniM env m, HasConfig env,
                                            HasGHCVariant env) =>
                                           ConfigCmdSet -> m Yaml.Value
        instance ‘Control.Monad.Base.MonadBase (Either e) (Either e)’
          at <no location info>
    Inaccessible code in
      the type signature for:
        cfgCmdSetValue :: (StackMiniM env m, HasConfig env,
                           HasGHCVariant env) =>
                          ConfigCmdSet -> m Yaml.Value
```

lwm:
#haskell IRC saves the day. https://github.com/commercialhaskell/stack/pull/2747/commits/b9b5f45a2089bc741417ada69467a7ac3cbcfaf4 is worth another look now.

sjakobi:
Doesn't this swallow errors from non-existent snapshot versions now, e.g. `stack config set resolver lts-8`?

lwm:
> Doesn't this swallow errors from non-existent snapshot versions now, e.g. stack config set resolver lts-8?

Yep, something else on the way!

mgsloan:
I get the following output when I run the test

```
Failures:

  test/integration/IntegrationSpec.hs:73: 
  1) 2747-config-set-resolver
       uncaught exception: TestFailure (Exited with ExitFailure 1
       
       stdout:
       
       
       stderr:
       Running: /home/mgsloan/fpco/stack/.stack-work/install/x86_64-linux/lts-6.14/7.10.3/bin/stack config set resolver ghc-8.0.1
       Could not parse '/tmp/stack-integration-2747-config-set-resolver28943/stack.yaml':
       Aeson exception:
       Error in $.resolver: failed to parse field 'resolver': failed to parse field resolver: Invalid resolver value: nightly. Possible valid values include lts-2.12, nightly-YYYY-MM-DD, ghc-7.10.2, and ghcjs-0.1.0_ghc-7.10.2. See https://www.stackage.org/snapshots for a complete list.
       See http://docs.haskellstack.org/en/stable/yaml_configuration/
       Main.hs: Exited with exit code: ExitFailure 1
       )

Randomized with seed 1606489183

Finished in 8.3794 seconds
1 example, 1 failure                                                                                                                                                                                                                               
Completed 2 action(s).
Test suite failure for package stack-1.2.1
    stack-integration-test:  exited with: ExitFailure 1
Logs printed to console
```

I also don't see how this can work, because the parsing functions are wrapped in `try`, and the results don't get used.  So it seems like no checking is being done.

lwm:
Hmm, couldn't really figure that one out. Sure someone else can grok it.

Issue #2407, opened 2016-07-22
Split up Stack.Types.Config


sjakobi:
(Split off from #2334)

This is very large module that takes a long time to compile for minor changes in itself or in its dependencies.

I said in #2334:

> And why are types like PackageIndex not in Stack.Types.PackageIndex?
> 
> The structure that I added in #1326 should probably be largely disregarded. I didn't really understand the various types at that time.

@chrisdone said in #2334:

> > This module takes a really long time to compile. And why are types like PackageIndex not in Stack.Types.PackageIndex?
> 
> There might be a cyclic dependency problem. Feel free to normalize it. Long compile times are obstacles to feeling motivation to contribute.

I think that mainly `Config`, `ConfigMonoid` and `HasConfig` should stay where they are but most other types should be moved into different / their own modules.

PR #2764, opened 2016-11-07
A stab at slimming Stack.Types.Config down.

lwm:
See https://github.com/commercialhaskell/stack/issues/2407.


mgsloan:
Seems reasonable!

Issue #996, opened 2015-09-16
Avoid writing large amounts of data to system temp directory


mgsloan:
Reported by [joseph07 in #haskell](http://ircbrowse.net/day/haskell/today/recent?id=21440325&timestamp=1442372927#t1442372927).  `stack setup` failed due to not having enough space on the device, despite there being 38GB.  According to pdxleif, archlinux defaults to using tmpfs, which stores it in ram, and so is limited in size.

Here's where a system temp directory is used as a place to unpack the GHC tarball: https://github.com/commercialhaskell/stack/blob/c0525a2431b7e383e06473efacee1e2084abf2d0/src/Stack/Setup.hs#L760

It'd also likely be worthwhile to take a look at other usages of the system temp directory.  On one hand, it seems odd to have a system configuration that has such limited space in tmp.  On the other hand, I see few downsides to instead storing such temp directories somewhere in `~/.stack` (and the big upside of resolving this issue)

LeifW:
Their article says tmpfs on Arch defaults to having its max size be half total RAM. https://wiki.archlinux.org/index.php/Tmpfs#Examples

snoyberg:
The downside to not using a proper temp directory is that if the process fails halfway through, the disk space is never reclaimed.

LeifW:
If that's a concern, what about using say https://hackage.haskell.org/package/temporary-1.2.0.3/docs/System-IO-Temp.html#v:withSystemTempDirectory ? Or are you referring to the Haskell process crashing?

snoyberg:
I'm referring to the Haskell process crashing (or system simply shutting down). And that's the function we're using already; the problem is Arch's setup not providing enough disk space in the temp directory.

LeifW:
I think other [distros using systemd](https://en.wikipedia.org/wiki/Systemd#Adoption_and_reception) will have /tmp be tmpfs, too. E.g. Fedora since version 18. http://fedoraproject.org/wiki/Features/tmp-on-tmpfs

mgsloan:
> The downside to not using a proper temp directory is that if the process fails halfway through, the disk space is never reclaimed.

I figured we'd have stack check for that and free it up.  This way, the only case where disk space is never reclaimed is if stack isn't run again.  Could get a bit tricky with concurrent stack executions, though..

LeifW:
[...]

PR #2766, opened 2016-11-08
Allow configuring programs directory

### LINKED ISSUE #1644: ghc-paths and other installation directories are hard-coded on Windows
This honestly wouldn't be a problem except that the sheer nesting depth is causing node.js/ghcjs to fail at compiling ghcjs-dom. It'd be nice if ghc-paths and others depended on STACK-ROOT, or were at least somehow variable.

ruuda:
This removes the hard-coded `$LOCALAPPDATA\Programs\stack` directory, and uses `$STACK_ROOT\programs` instead, unifying this path for all programs. This resolves #1644 and also [my comment](https://github.com/commercialhaskell/stack/issues/996#issuecomment-258642367) on #996.

@borsboom: You noted in #1644 that it would be good to make this configurable in `stack.yaml`. That would certainly be possible, but currently I do not see how that would be useful. I think it would introduce more knobs than necessary. Is there a compelling use case for downloading and extracting GHC to a different location than the snapshots?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -32,6 +32,10 @@ Behavior changes:
 * If a remote package is specified (such as a Git repo) without an explicit
   `extra-dep` setting, a warning is given to the user to provide one
   explicitly.
+* On Windows, `stack setup` will now download and extract additional programs
+  such as GHC to `$STACK_ROOT\programs`, like on other platforms. You may need
+  to remove programs in `$LOCALAPPDATA\Programs\stack` manually.
+  [#1644](https://github.com/commercialhaskell/stack/issues/1644).
 
 Other enhancements:
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -32,10 +32,6 @@ Behavior changes:
 * If a remote package is specified (such as a Git repo) without an explicit
   `extra-dep` setting, a warning is given to the user to provide one
   explicitly.
-* On Windows, `stack setup` will now download and extract additional programs
-  such as GHC to `$STACK_ROOT\programs`, like on other platforms. You may need
-  to remove programs in `$LOCALAPPDATA\Programs\stack` manually.
-  [#1644](https://github.com/commercialhaskell/stack/issues/1644).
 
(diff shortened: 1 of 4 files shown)


mgsloan:
Thanks for working on this!  The problem with this approach is that it means people will need to re-install their GHC versions, and will also leave the old GHC versions still there.  This is why we'd prefer to make this an explicit option.

The reason to allow it to be specified in the config subset of `stack.yaml` is that then users can specify it in their `config.yaml`

ruuda:
I see. In that case, I think the proper thing to do (from a user perspective) would be to automatically move the data to the new location. But that would mean adding even more special cases for a while. Making this a configurable option with different defaults on Windows than on other platforms looks like an easier solution from an implementation point of view. What do you think?

mgsloan:
> Making this a configurable option with different defaults on Windows than on other platforms looks like an easier solution from an implementation point of view.

Sounds good to me!

ruuda:
I rebased this and changed it to keep the `$LOCALAPPDATA` default. There now is a `local-programs-path`, which defaults to `$STACK_ROOT/programs` except on Windows, where it defaults to `$LOCALAPPDATA\Programs\stack`. When getting `$LOCALAPPDATA` fails (I don’t think that should ever happen), it now defaults to `$STACK_ROOT\programs` on Windows too.

I have not tested this on Windows yet, I’ll try that tomorrow. I did test this on Linux.

mgsloan:
LGTM, thanks!

wowofbob:
> I rebased this and changed it to keep the `$LOCALAPPDATA` default. There now is a `local-programs-path`, which defaults to `$STACK_ROOT/programs` except on Windows, where it defaults to `$LOCALAPPDATA\Programs\stack`. When getting `$LOCALAPPDATA` fails (I don’t think that should ever happen), it now defaults to `$STACK_ROOT\programs` on Windows too.
> 
> I have not tested this on Windows yet, I’ll try that tomorrow. I did test this on Linux.

This pretty much seems to work on windows!

Issue #1238, opened 2015-10-26
Binary upgrade code path


snoyberg:
- [ ] Add a new option for the `stack upgrade` command, `--binary`, that will download the latest binary version from the Github release (`--no-binary` will also be available)
- [ ] When compiling official releases, have some flag that indicates that this is an official release
- [ ] When using an official binary release, `--binary` is on by default. Otherwise, it's off by default.

Blaisorblade:
- [ ] Show the "official release" flag in `stack --version`, as suggested by https://github.com/commercialhaskell/stack/issues/2222#issuecomment-234030699.

borsboom:
Bumped priority of this in light of planning to stop maintaining Linux APT/YUM repos.

PR #2769, opened 2016-11-09
Allow binary upgrades #1238

snoyberg:
This PR is not yet complete, please see comments.

### DIFF
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -173,7 +173,7 @@ binaryUpgrade (BinaryOpts mplatform force' mver morg mrepo) = do
             ec <- rawSystem (toFilePath tmpFile) ["--version"]
 
             unless (ec == ExitSuccess)
-                    $ error $ "Non-success exit code from running newly downloaded executable"
+                    $ error "Non-success exit code from running newly downloaded executable"
 
 sourceUpgrade
   :: (StackM env m, HasConfig env)
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -11,30 +11,16 @@ module Stack.Upgrade
     , upgradeOpts
     ) where
 
-import qualified Codec.Archive.Tar           as Tar
-import           Control.Exception.Safe      (catchAny, throwM)
-import           Control.Monad               (guard, unless, when)
+import           Control.Exception.Safe      (catchAny)
+import           Control.Monad               (unless, when)
 import           Control.Monad.IO.Class
 import           Control.Monad.Logger
-import           Control.Monad.Reader        (MonadReader, asks)
-import           Data.Aeson                  (Value (Array, Object, String))
-import qualified Data.ByteString.Lazy        as L
-import           Data.Conduit                ((.|))
-import           Data.Conduit.Lazy           (lazyConsume)
-import           Data.Conduit.Zlib           (ungzip)
-import           Data.Foldable               (fold, forM_)
-import qualified Data.HashMap.Strict         as HashMap
+import           Data.Foldable               (forM_)
 import qualified Data.Map                    as Map
-import           Data.Maybe                  (isNothing, fromMaybe)
+import           Data.Maybe                  (isNothing)
 import           Data.Monoid.Extra
 import qualified Data.Text as T
-import qualified Data.Version
-import           Distribution.System         (Platform (Platform), Arch (..), OS (..))
 import           Lens.Micro                  (set)
-import           Network.HTTP.Client         (parseUrlThrow)
-import           Network.HTTP.Simple         (Request, httpJSON, withResponse,
-                                              setRequestHeader, getResponseBody,
-                                              getResponseStatusCode, parseRequest)
 import           Options.Applicative
 import           Path
 import           Path.IO
@@ -51,16 +37,9 @@ import           Stack.Types.Config
 import           Stack.Types.Internal
 import           Stack.Types.Resolver
 import           Stack.Types.StackT
-import qualified System.Directory            as IO
 import           System.Exit                 (ExitCode (ExitSuccess))
-import qualified System.FilePath             as FP
 import           System.Process              (rawSystem, readProcess)
 import           System.Process.Run
-import           Text.ParserCombinators.ReadP (readP_to_S)
-
-#if !WINDOWS
-import           System.Posix.Files (setFileMode)
-#endif
 
 upgradeOpts :: Parser UpgradeOpts
 upgradeOpts = UpgradeOpts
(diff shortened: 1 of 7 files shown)


stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-6.14
+resolver: lts-6.23
```
snoyberg: This bump is needed to get a newer version of conduit-combinators for tar-conduit. If this is a problem, two alternatives:

* Switch back to the tar package
* Modify tar-conduit to have a more lenient lower bound

src/main/Main.hs
```diff
@@ -283,16 +283,9 @@ commandLineHandler progName isInterpreter = complicatedOptions
...
                     upgradeCmd
-                    ((,) <$> switch
-                              ( long "git"
-                             <> help "Clone from Git instead of downloading from Hackage (more dangerous)" )
-                         <*> strOption
-                              ( long "git-repo"
-                             <> help "Clone from specified git repository"
-                             <> value "https://github.com/commercialhaskell/stack"
-                             <> showDefault ))
+                    (upgradeOpts Nothing) -- FIXME insert the default platform to use
```
snoyberg: We need to figure out the right way to pass in the platform string (e.g., `osx-x86_64`). I see two options:

1. Pass in some CPP variable during official builds, and have only official builds use binary upgrade by default (probably the right approach).
2. Use the build system info from the Cabal package or similar to construct the string. Overall I'm -1 on this due to likely breaking cross-compilation and possibly having an incorrect configuration.

src/Stack/Upgrade.hs
```diff
@@ -29,16 +48,182 @@ import           Stack.Types.Config
...
+#if WINDOWS
+                        FP.<.> "exe"
+#endif
+        tmpFile = destFile FP.<.> "tmp"
+
+    liftIO $ do
+      case () of
+        ()
+          | ".tar.gz" `T.isSuffixOf` archiveURL -> handleTarball archiveURL tmpFile
+          | ".zip" `T.isSuffixOf` archiveURL -> error "FIXME: Handle zip files"
```
snoyberg: While we _could_ add zip file support, that's IMO a very bad idea: we likely don't want to add a zip library dependency (I think the ones available on Hackage are fairly unstable), and shelling out to an external command will be painful. Worse yet, this will introduce a lot of code that is conditional for Windows only.

What do you think about adding a tarball release option for Windows in addition to the zip files?

snoyberg:
Also worth discussing:

* I'm not at all convinced I did a good job of the options parsing, there may be far better approaches to it.
* I used the download functions from `Network.HTTP.Simple`, but perhaps I should have used a Stack-internal HTTP function?

src/Stack/Upgrade.hs
mgsloan: Makes sense to me!

src/main/Main.hs
mgsloan: I think it makes sense to use the same logic as is currently used to figure out the platform.  Since we have a configuration with this command, perhaps `asks getPlatform` is a sufficient default?

This info does indeed correspond to the build platform, but is overidable in the case of cross compilation - https://github.com/commercialhaskell/stack/blob/master/src/Stack/Config.hs#L254 .

stack.yaml
borsboom: No problem bumping the LTS release.

src/main/Main.hs
borsboom: We already have logic to determine it for downloading GHC, so that could be re-used.

An advantage of passing it in at compile time would be that we could ensure that _exactly_ the same platform is used for the downloaded version as is currently running.  On the other hand, it may sometimes make sense to update the logic for determining the platform (e.g. when new distro versions are released that change shared library versions such as GMP), in which case having it be "static" would be a disadvantage.

src/Stack/Upgrade.hs
```diff
@@ -29,16 +48,182 @@ import           Stack.Types.Config
...
+        val <- getResponseBody <$> httpJSON req
+        let ret = do
+                Object top <- return val
+                Array assets <- HashMap.lookup "assets" top
+                getFirst $ fold $ fmap (First . findMatch pattern') assets
+        case ret of
+            Just url -> return url
+            Nothing -> error $ "Could not find Stack archive for platform: " ++ T.unpack pattern
+      where
+        -- FIXME make the Github repo configurable?
```
borsboom: This could also use https://www.stackage.org/stack instead of the Github releases directly.

src/Stack/Upgrade.hs
borsboom: I don't feel particularly strongly about it, but we do already use an external zip command for extracting GHC on Windows so it may be easier than you think.  Adding a tarball for Windows is no problem though.

borsboom:
> I used the download functions from Network.HTTP.Simple, but perhaps I should have used a Stack-internal HTTP function?

There is also already logic for downloading and extracting stack on Linux in [ensureDockerStackExe](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Setup.hs#L594), that could be re-used.  Or at least, there probably shouldn't be two different ways of doing the same thing.

src/Stack/Upgrade.hs
snoyberg: It could, but why add the dependency on another external service that could fail? The logic I've put in here is essentially copy-pasted from stackage-server. (This _might_ argue for making a shared library for easier maintainability going forward, but that's a general Stack maintenance question we should be discussing.)

src/main/Main.hs
snoyberg: Assuming we have no reason to believe that the stack executable will be more OS-specific than the GHC bundle (which seems like a fair assumption), I agree with reusing that logic.

borsboom:
A couple of things we've discussed in the past that I don't see here (we can decide now if they matter):

- Should we default to a binary upgrade only if the currently running binary is an official one?  We'd either have to add something to the build process to add a compiled-in flag that indicates the current binary is official (that could include the platform, as I mentioned in another comment), or we could put checksums of the `stack` executable itself in https://github.com/fpco/stackage-content/blob/master/stack/stack-setup-2.yaml and compare.  On the other hand, if people are using third-party packages, they may still want to use a binary `stack upgrade`.

- It would be nice to have a new version released notification, since people's package manager won't do that for them anymore if we don't release official packages.  Check max once per day, and also offer a way to disable the check.

snoyberg:
> Should we default to a binary upgrade only if the currently running binary is an official one?

That's what I thought we'd achieve with the platform option I mentioned above.

Auto-check for updates is a good idea.

src/main/Main.hs
borsboom: Stack will be less OS-specific than GHC, because it doesn't use libtinfo/libncurses.

snoyberg: OK, it now determines automatically which package to download based on `getPlatform`. It also:

* Automatically checks for and prefers a `-static` version
* Will append `-gmp4` if needed

This commit also adds in support for comparing version numbers so we don't needlessly upgrade

snoyberg:
I've given auto-check a bit more thought; I think I'm overall 👎 on it. This could seriously hamper the usage of Stack in an environment without internet, by (from a user perspective) randomly adding in delays of unknown duration. With the addition of version number checking, `stack upgrade` is now a safe thing for users to run regularly.

What I _would_ consider is giving a message after the completion of a `stack upgrade` command to let the user know that a new version is available upstream. That way, we're just piggy-backing on existing information.

snoyberg:
Reviewing the `ensureDockerStackExe` (EDSE) logic and comparing with `binaryUpgrade` (BU), I'm not convinced it's a good idea to merge these into the same code path, as they're doing something quite different:

* EDSE needs to peg a specific executable version, BU does not
* EDSE wants to hard-code a specific tarball per version, BU wants to discover the appropriate tarball for the target platform
* EDSE seems to keep multiple copies of the tool in place at once, and marks each of them as installed. BU just overwrites the current Stack executable with a newer one
* BU will test the executable on the host OS before completing the install, EDSE needs to do it inside the Docker container (if it tests it at all)

I'm still open to making the change, but I'd like to make sure we're synched up on what we're trying to unify here.

borsboom:
In general I agree about EDSE/BU.  One exception is:

> EDSE wants to hard-code a specific tarball per version, BU wants to discover the appropriate tarball for the target platform

In fact, EDSE doesn't need to do this.  It could just as well use the Github release thing like BU does to find the right binary for a given version.

borsboom:
> What I would consider is giving a message after the completion of a stack upgrade command to let the user know that a new version is available upstream. That way, we're just piggy-backing on existing information.

Not sure I understand.  If you've just run `stack upgrade`, wouldn't you always be on the latest version already?

borsboom:
> EDSE seems to keep multiple copies of the tool in place at once, and marks each of them as installed. BU just overwrites the current Stack executable with a newer one

This is a side-effect of the fact that the Docker-compatible stack executables are potentially platform-specific and so installed next to GHC.  In practice, this is never actually needed at the moment, although it potentially could be in the future.

snoyberg:
> In fact, EDSE doesn't need to do this. It could just as well use the Github release thing like BU does to find the right binary for a given version.

And then we'd just do an exact version match for the current version of Stack? That sounds OK to me, and would mean less maintenance on that YAML config file going forward, which sounds good.

* [ ] EDSU gets file locations from releases API

> Not sure I understand. If you've just run stack upgrade, wouldn't you always be on the latest version already?

Stupid fingers. I meant after a `stack update`.

> This is a side-effect of the fact that the Docker-compatible stack executables are potentially platform-specific and so installed next to GHC. In practice, this is never actually needed at the moment, although it potentially could be in the future.

Let me poke at this code a bit more, maybe I can share some of the logic of "download the tarball containing the Stack executable and write it out to a certain location."

* [ ] Try to share tarball unpacking logic with EDSU

snoyberg:
Alright, logic is now unified. The only thing left I think would be adding some kind of warning about out-of-date Stack on the `stack update` command, but I think that would be best handled separately.

borsboom:
Awesome, I'll plan to merge today and then cut an RC branch.

snoyberg:
Woohoo!

PR #2779, opened 2016-11-13
Update YAML configuration docs to include `scm-init` option

devlinzed:
I was about to patch this `git init` feature in before noticing it was already in the source code.  I was unable to find documentation for it anywhere so I thought I would add some.  The documentation is based on my preliminary understanding of the code and my own experimentation.

This is also my first pull request, so I apologize if I neglected any protocol.


mgsloan:
Thanks!

PR #2782, opened 2016-11-15
Remove verbatim linebreak

kini:
Text inside `` is rendered into \<code> tags in HTML.  The CSS style
for \<code> tags in the resulting page appears to understand line
breaks in the inner HTML of the tag as being verbatim.  Hence it is
probably wise not to break literal expressions in the markdown source
over multiple lines.


mgsloan:
Thanks!

kini:
Ever since this was merged, I've been getting build failure notification emails from "Jenkins fpco public", as has @devlinzed, apparently. What do I need to do to get myself un-CC'd from these emails?

PR #2793, opened 2016-11-20
Drop Data.Yaml.Extra #2491

snoyberg:
This has been fixed upstream, see:
https://github.com/snoyberg/yaml/pull/91


stack-7.8.yaml
```diff
@@ -76,6 +76,7 @@ extra-deps:
 - text-metrics-0.1.0
 - pid1-0.1.0.0
 - file-embed-0.0.10
+- yaml-0.8.20
```
Blaisorblade: I expect we'll need this `- yaml-0.8.20` also in `stack.yaml` and `stack-8.0.yaml` since the snapshots use older versions, but CI will tell.

Blaisorblade:
LGTM—it reverts everything in #2492, matches the plan in https://github.com/commercialhaskell/stack/issues/2491#issuecomment-239697902, and updates the dependency in the cabal file.

stack-7.8.yaml
snoyberg: Doh, I'm confusing myself with these separate PRs, I'd already bumped the LTS versions in a separate one. I'll do that here too.

Blaisorblade: OK, should work now!

mgsloan:
LGTM!

PR #2801, opened 2016-11-22
DWARF debugging support

### LINKED ISSUE #877: Add a `--dwarf` flag for DWARF debugging
Say I wanted to build my project with all the debugging options, including stuff like DWARF data. The [page about DWARF support in GHC](https://ghc.haskell.org/trac/ghc/wiki/DWARF) says that you have to pass `-g` to ghc to get it to generate these symbols, but what about all the packages I depend on, and the RTS itself? ("you will need to compile both libraries and the runtime system with debugging support (GhcRtsHcOpts += -g and GhcLibHcOpts += -g).") I guess to make it easy to have a GHC RTS with debugging we'd need bindists that have symbols in them -- I'm not sure if the default 7.10.2 ones have them.

Likewise all this stuff for GHC's profiling options. Is there a way to easily just turn on all profiling options for a project and rebuild everything so I can profile everything?

Maybe this is already pretty easy and this should just be resolved with some documentation in the guide.

SASinestro:

It passes all tests that are in the repo, and in my own tests on OS X it does indeed produce output that contains debugging symbols. I don't have access to any other operating systems to test on at the moment, though.

This would close #877.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -79,6 +79,9 @@ Other enhancements:
   [#2235](https://github.com/commercialhaskell/stack/issues/2235)
 * Replace enclosed-exceptions with safe-exceptions.
   [#2768](https://github.com/commercialhaskell/stack/issues/2768)
+* `stack build` and related commands now allow the user to disable debug symbol stripping
+  with new `--no-strip`, `--no-library-stripping`, and `--no-executable-shipping` flags, 
+  closing [#877](https://github.com/commercialhaskell/stack/issues/877).
 
 Bug fixes:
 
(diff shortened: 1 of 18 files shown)


SASinestro:
It won't pass the CI because of `-Werror` and my lack of knowledge of how to test for debugging on most of the platforms in Distribution.System.OS.

SASinestro:
It does build correctly on Linux and OS X, Travis just seems to have had a bad hair day when my commit was going through.

mgsloan:
While I'm a bit wary of the approach used to detect debugging symbols, this seems like a good change.  Let's give it a try - thanks for implementing!

Issue #2805, opened 2016-11-24
RFC: A new script command


snoyberg:
I've discussed this in some part with various people, so I thought I'd get the thoughts down in writing. Currently, script interpreter usage of Stack has (IMO) three limitations:

* It's slower than just using `runghc` directly. Some of that is inherent: it needs to check for the presence of GHC and libraries, for example, which `runghc` does not. Nonetheless, there is more overhead than necessary. From my understanding, settings file checking is a significant component of this (and also likely something we don't want to impact the behavior of script usage).
* Scripts violate some reproducibility rules, because:
  * They don't require that a `resolver` be set
  * They don't require that all packages be listed
  * They can be impacted by other packages installed

  These trade-offs all make sense when considering the `runghc` feature in general, but are less appetizing when applied to reproducible scripts.
* There are multiple options that should almost always be set but aren't enforced: `--install-ghc`, a `--resolver` as implied above.

Proposal: let's add a new Stack command: `stack script`. It would have the following semantics:

* Require that a `--resolver` be set
* Hide all packages from the package database except those specified with `--package`. (Perhaps make an exception for wired in packages, or at the very least `base`.)
* Automatically imply `--install-ghc`
* Ignore all settings files

As an alternative (or addition) to the above, we could consider:

* Making these behaviors options available to `runghc` (e.g., `--no-parse-settings-file`)
* Don't actually add a `script` command, but add support to the script interpreter bits of Stack to recognize a `script` command and do the right thing.

silky:
i prefer the first proposal; a new command. it will be cleaner indication that it is doing something special.

chrisdone:
I think the script command is a good proposal, better than changing runghc.

harendra-kumar:
Sounds good to have a packaged command since this is a common enough use case. Users will not have to scratch their head and consult documentation on what all options to specify. Making this part of runghc will not solve the complexity of too many options (we already have tricky ways to specify options, for example see #2486).

I also do not like the second alternative of a magical command only visible to interpreter, it will cause confusion.

How about using `stack run` or `stack runsure` instead of `stack script`? Just  a quick thought.

snoyberg:
I think `stack run` would be confused with `cabal run`, which (IIRC) builds an executable in a project and runs it.

I agree with the overall consensus that an explicit new command is better than a bunch of new options or a secret command.

snoyberg:
[...]

PR #2806, opened 2016-11-24
Script command #2805

snoyberg:
This adds a basic script command. Most of the work involved has to do
with preventing config files from being loaded. There's at least one
ugly hack involved (adding a dummy bcStackYaml value), and some usages
of error that should probably be converted to proper exception types.

While this addresses reproducibility, it doesn't help with the speed
concerns: script is now about 100ms faster than runghc on my system for
the case where --package is provided, but it's still over a second for
Hello World. The slowdown is inherent right now in checking if the
relevant packages are installed. It would be nice to figure out a way to
optimize the package check.

Also, this should include some integration tests. It should be a simple
matter of a test that includes a bogus stack.yaml and proving that stack
script ignores it.

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -23,11 +23,15 @@ import           Control.Monad.Trans.Either (EitherT)
 import           Control.Monad.Writer.Lazy (Writer)
 import           Data.Attoparsec.Args (parseArgs, EscapingMode (Escaping))
 import           Data.Attoparsec.Interpreter (getInterpreterArgs)
+import qualified Data.ByteString.Char8 as S8
 import qualified Data.ByteString.Lazy as L
+import qualified Data.Conduit.List as CL
 import           Data.List
+import           Data.List.Split (splitWhen)
 import qualified Data.Map as Map
 import           Data.Maybe
 import           Data.Monoid
+import qualified Data.Set as Set
 import           Data.Text (Text)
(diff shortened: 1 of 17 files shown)


src/main/Main.hs
```diff
@@ -779,6 +788,68 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+        mapM_ parsePackageNameFromString targets
+
+        unless (null targets) $ do
+            -- Optimization: use the relatively cheap ghc-pkg list
+            -- --simple-output to check which packages are installed
+            -- already. If all needed packages are available, we can
+            -- skip the (rather expensive) build call below.
+            bss <- sinkProcessStdout
+                Nothing menv (ghcPkgExeName wc)
+                ["list", "--simple-output"] CL.consume -- FIXME use the package info from envConfigPackages, or is that crazy?
```
mgsloan: RE the FIXME, I don't think that's necessary - envConfigPackages is for local packages.  If there is no settings file, then there can't be local packages.

mgsloan:
Seems reasonable to me!  Accidentally introduced merge conflicts with https://github.com/commercialhaskell/stack/commit/fd6d4509ff0d8c35645813d497c44f78aec78407

snoyberg:
I'm busy refactoring this still in my spare time, closing until it's ready.

PR #2823, opened 2016-12-02
Fix HLint/CPP failure in Travis

enolan:

HLint was failing because it was using --cpp-simple, which just strips
every directive, which in turn created duplicate import statements. This
commit turns off --cpp-simple and lints with the WINDOWS macro on and off.
--cpp-simple is kept on for the tests, which don't have the duplicate import
problem and some of which use CPP in more complex ways than just turning a
single variable on and off.

Tested via Travis. No need to do anything else since `.travis.yml` is the only file changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,9 @@ install:
 script:
  - set -e; case "$BUILD" in
      style)
-       hlint src/ test/ --cpp-simple --hint=HLint.hs;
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks! I wish I'd seen this earlier, I just modified the imports a bit to avoid the same failure.

Issue #2825, opened 2016-12-03
`stack --no-nix-pure` should not imply `--nix`


fmthoma:
I have a Haskell script I want to run with stack as a script interpreter. This script should run both on NixOS and on non-Nix systems. The script depends on the locale, which is unset by the pure environment, therefore I want to run the script in a non-pure environment if Nix is enabled, but adding `--no-nix-pure` also enables `--nix` if `stack` isn't configured to use nix, breaking the script on non-Nix systems.

### Steps to reproduce

`~/.stack/config.yaml`:
```
…
nix:
  enable: false
```

`echo.hs`:
```
#!/usr/bin/env stack
-- stack --no-nix-pure runhaskell

main :: IO ()
main = getLine >>= putStrLn
```

Running this on a non-Nix system produces:
```
> echo äöü | ./echo.hs
Executable named nix-shell not found on path: [ … ]
```

Removing the `--no-nix-pure` on a Nix system (`nix: enable: true`) breaks the script because the default locale in a pure environment does not support UTF-8 (see also #2358 for the locale problem):
```
> echo äöü | ./echo.hs
echo.hs: <stdin>: hGetLine: invalid argument (invalid byte sequence)
```

### Expected

```
> echo äöü | ./echo.hs
äöü
```
on both Nix and non-Nix systems.

I would expect `--no-nix-pure` to switch to non-pure nix build if `--nix` is enabled, not enabling `--nix` in the same turn.
The current behavior makes it impossible to use `stack` as a script interpreter in a way compatible with both Nix and non-Nix systems.

### Stack version

```
$ stack --version
Version 1.2.0, Git revision 241cd07d576d9c0c0e712e83d947e3dd64541c42 (4054 commits) x86_64 hpack-0.14.0
```

fmthoma:
Possible fix:

```diff
diff --git a/src/Stack/Options/NixParser.hs b/src/Stack/Options/NixParser.hs
index f1b0fcf..5db1570 100644
--- a/src/Stack/Options/NixParser.hs
+++ b/src/Stack/Options/NixParser.hs
@@ -10,7 +10,7 @@ import           Stack.Options.Utils
 import           Stack.Types.Nix
 
 nixOptsParser :: Bool -> Parser NixOptsMonoid
-nixOptsParser hide0 = overrideActivation <$>
+nixOptsParser hide0 =
   (NixOptsMonoid
   <$> pure (Any False)
   <*> firstBoolFlags nixCmdName
[...]

Issue #2358, opened 2016-07-09
Enabling nix causes LANG to be lost.


seanparsons:
This took me a while to figure out, but the pre-requisites of it appear to be:
- Nix is enabled in stack.yaml.
- Code is run from within stack itself (for me it manifests with `ghcid` and by using `stack exec <binaryname>`).

An easy way to see what's going on here is this:

``` bash
# diff <(env) <(stack exec env) | grep LANG
Run from outside a project, using implicit global project config
> GDM_LANG=en_GB.utf8
> LANG=en_GB.UTF-8
< LANG=en_GB.UTF-8
< GDM_LANG=en_GB.utf8

# diff <(env) <(stack --nix exec env) | grep LANG
Run from outside a project, using implicit global project config
< LANG=en_GB.UTF-8
< GDM_LANG=en_GB.utf8
```

In my case this manifested as a failure in my code where I tried to create files with non-ASCII filenames using `Data.ByteString.writeFile`.

seagreen:
I think this is an issue on the Nixpkgs side, see [this issue](https://github.com/NixOS/nix/issues/318) or just a [search for "lang" in their issues](https://github.com/NixOS/nix/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20lang).

chris-martin:
It seems fair to me for a build not to inherit `LANG` from the environment. Not all projects use the same language, right?

seagreen:
@chris-martin: Agreed. Happily it looks like nix isn't inheriting LANG:

```
[~/code/mine/hjsonschema]traveller@nivian $ export LANG=POSIX
[~/code/mine/hjsonschema]traveller@nivian $ echo $LANG
POSIX
[~/code/mine/hjsonschema]traveller@nivian $ nix-shell 

[nix-shell:~/code/mine/hjsonschema]$ echo $LANG 
en_US.UTF-8

[nix-shell:~/code/mine/hjsonschema]$
```

seagreen:
Hmm, not I'm confused. The above uses this `stack.yml`:

```
resolver: lts-7.2
nix: 
  enable: true                                                                                              
  shell-file: shell.nix
```

Note that I haven't set `pure` to `false`. The `shell.nix` file it mentions doesn't mess with `LANG` at all.

[...]

PR #2826, opened 2016-12-03
`--nix-*` should not imply `--nix` by default

fmthoma:
Only enabling `--nix-pure` explicitly on the command line implies `--nix`.

### Rationale:

Users who prefer to use Nix (e.g. NixOS users) will have `--nix` enabled
globally in their `~/.stack/config.yaml`. On the other hand, users who
do not have nix installed will not have a `nix` section in their global
stack config.

When distributing scripts runnable by stack (`#!/usr/bin/env stack`),
they should be runnable independent from the user's Nix preference. In
order to run in a Nix environment, however, sometimes certain tweakings
have to be made to the Nix settings:
* Disabling a pure build because of missing locale (see also #2358)
* Adding packages like `ncurses` or `zlib` that are usually present
  in the user environment, but not in a pure nix-shell.

This was not possible before, since any of the `--nix-*` options, even
`--no-nix-pure`, would enable `--nix` as well, therefore breaking
portability for scripts on systems without Nix installed.

This change preserves the implication for `--nix-pure`, but removes it
for all other `--nix-*` options.

### ChangeLog & Documentation

    * Added documentation for the `--nix-pure` ⇒ `--nix` implication to the parser. The other implications were undocumented anyways.

### See also
Corresponding issue: #2825

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,9 @@ Behavior changes:
   as well. If you manually specify a package index with only a Git
   URL, Git will still be used. See
   [#2780](https://github.com/commercialhaskell/stack/issues/2780)
+* Passing `--(no-)nix-*` options now no longer implies `--nix`, except for
+  `--nix-pure`, so that the user preference whether or not to use Nix is
+  honored even in the presence of options that change the Nix behavior.
 
 Other enhancements:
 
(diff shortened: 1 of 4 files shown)


fmthoma:
Rebased onto master.

Ping @lwm, @sjakobi and @YPares based on `git blame`ing the relevant lines.

mgsloan:
Sorry for the delay on reviewing this!  Seems like a sensible usecase.  Hopefully the change doesn't break many scripts.

fmthoma:
Thank you! I'll close #2825.

PR #2837, opened 2016-12-12
Update travis-complex.yml

mitchellwrosen:
For GHC 7.8.4, bump cabal from 1.18 to 1.22, because PatternSynonyms is only recognized by 1.22+

### DIFF
--- a/doc/travis-complex.yml
+++ b/doc/travis-complex.yml
@@ -47,9 +47,9 @@ matrix:
   - env: BUILD=cabal GHCVER=7.6.3 CABALVER=1.16 HAPPYVER=1.19.5 ALEXVER=3.1.7
     compiler: ": #GHC 7.6.3"
     addons: {apt: {packages: [cabal-install-1.16,ghc-7.6.3,happy-1.19.5,alex-3.1.7], sources: [hvr-ghc]}}
-  - env: BUILD=cabal GHCVER=7.8.4 CABALVER=1.18 HAPPYVER=1.19.5 ALEXVER=3.1.7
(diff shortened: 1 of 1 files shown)


snoyberg:
I'm not sure this is a good idea:

1. It doesn't fix the GHC 7.6.3 entry
2. It's good to include testing of the cabal version that is tied to a GHC release for users who are stuck with them at the same version (e.g. someone using Ubuntu packages)

I'd rather go with documentation of this case.

mitchellwrosen:
Sounds good. But, what's wrong with the GHC 7.6.3 entry? Pattern synonyms were introduced in 7.8.

snoyberg:
I don't follow, can you clarify?

mitchellwrosen:
When building on 7.8.4 with the given travis.yml, with PatternSynonyms, the build breaks because PatternSynonyms are only recognized by Cabal-1.22+.

To your second point, although I'm not sure _why_ Cabal-1.18 was bundled with GHC-7.8.4 (since it lacks PatternSynonyms support), but it makes sense to keep them in sync with each other.

However, I'm not sure what you meant by your first point, since the GHC 7.6.3 entry wasn't modified.

snoyberg:
Regarding 7.6.3: it seems arbitrary to update 7.8.4 without updating 7.6.3. Either there was a bug with older Cabals, or there wasn't. I understand the incompatibility you're getting at though.

As to why it was bundled this way: I think it was just a bug in Cabal that the appropriate language extension wasn't added.

Given this approach to keeping them in sync, should we close out this PR?

Issue #2829, opened 2016-12-05
Stack hoogle --rebuild is broken in 1.3.1


rikvdkleij:
### General summary/comments (optional)
After upgrading to version 1.3.1 Hoogle rebuild option gives error.

### Steps to reproduce
Run `stack hoogle --rebuild`

### Expected
Rebuild

### Actual
Error:`
Rebuilding database ...
readelf: Error: '/home/rik/.stack/programs/x86_64-linux/ghc-7.10.3/lib/ghc-7.10.3/ghc_0AG9TOjDEtx4Ji3wSwHOBe/libHSghc-7.10.3-0AG9TOjDEtx4Ji3wSwHOBe-ghc7.10.3.a': No such file
readCreateProcess: readelf "--debug-dump=info" "--dwarf-depth=1" "/home/rik/.stack/programs/x86_64-linux/ghc-7.10.3/lib/ghc-7.10.3/ghc_0AG9TOjDEtx4Ji3wSwHOBe/libHSghc-7.10.3-0AG9TOjDEtx4Ji3wSwHOBe-ghc7.10.3.a" (exit 1): failed
`

What actually happened.
Nothing, only error. Same problem for ghc 8.0.1

### Stack version

```
$ stack --version
Version 1.3.1, Git revision 737ee751654cba4b2547b50628c9a5a78ddc6f8b x86_64 hpack-0.15.0
```

### Method of installation
stack upgrade --git

sjakobi:
I see a very similar issue with `stack haddock`:

```
~/tmp $ stack new haddock-issue
~/tmp $ cd haddock-issue/
~/t/haddock-issue $ stack haddock
Warning: File listed in haddock-issue.cabal file does not exist: README.md
readelf: Fehler: '/home/simon/.stack/programs/x86_64-linux/ghc-8.0.1/lib/ghc-8.0.1/ghc-8.0.1/libHSghc-8.0.1-ghc8.0.1.a': Keine solche Datei
readCreateProcess: readelf "--debug-dump=info" "--dwarf-depth=1" "/home/simon/.stack/programs/x86_64-linux/ghc-8.0.1/lib/ghc-8.0.1/ghc-8.0.1/libHSghc-8.0.1-ghc8.0.1.a" (exit 1): failed
~/t/haddock-issue $ stack --version
[...]

PR #2854, opened 2016-12-16
Fix DWARF heuristics (closes #2851, #2829)

### LINKED ISSUE #2851: `readelf` error when building with profiling enabled
Encountered by @dysinger.  Additional info from him:

> what may have triggered this bug above is that I had profiling enabled for libraries & exes
> yep. If I comment out the profiling options in my stack.yaml then it works fine.
screen shot 2016-12-16 at 7 24 42 am

Possible culprit: https://github.com/commercialhaskell/stack/pull/2801/files#diff-eda6dc79c51b46a15020a1b28b794f50R97

Likely related: #2829

SASinestro:
Now runs correctly on Linux and OS X. Tested on Ubuntu 16.04 and OS X 10.11.4.


borsboom:
Thanks for the quick action!

PR #2860, opened 2016-12-19
Fix version bound for conduit-extra

thoferon:
Due to a use of `sinkFileCautious` introduced in 1.1.14. See https://github.com/snoyberg/conduit/commit/73b5fc28f1bb3ae52095c404ff64f07ba0da92ac


snoyberg:
Thanks!

PR #2866, opened 2016-12-21
Cabal custom setup

snoyberg:

I'm still testing this in a more complex use case. However, the simple test case I devised was:

* `stack new foo`
* Modify the `Setup.hs` file to add `import Acme.Missiles`
* Add `acme-missiles-0.3` as an extra-dep
*   Add the following `custom-setup` block to `foo.cabal`:

    ```
    custom-setup
      setup-depends: base >= 4.6,
                     Cabal >= 1.24 && < 1.25,
                     acme-missiles
    ```
* Set `build-type` to `Custom`
* Confirm that without the patch, the package refuses to compile. With the patch, it compiles just fine.

@borsboom Maybe we can sync up later on adding this as an integration test case?

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -946,7 +946,7 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
                 getPackageArgs =
                     case (packageSetupDeps package, mdeps) of
                         -- The package is using the Cabal custom-setup
-                        -- configuraiton introduced in Cabal 1.24. In
+                        -- configuration introduced in Cabal 1.24. In
                         -- this case, the package is providing an
                         -- explicit list of dependencies, and we
                         -- should simply use all of them.
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -933,29 +933,58 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
                         ["-package=" ++ packageIdentifierString
                                             (PackageIdentifier cabalPackageName
                                                               eeCabalPkgVer)]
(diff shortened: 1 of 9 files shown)


src/Stack/Build/Execute.hs
```diff
@@ -933,29 +933,58 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
...
+                    ) ++
+                    ( ("-package-db=" ++ toFilePathNoTrailingSep (bcoSnapDB eeBaseConfigOpts))
+                    : ("-package-db=" ++ toFilePathNoTrailingSep (bcoLocalDB eeBaseConfigOpts))
+                    : ["-hide-all-packages"]
+                    )
+
+                getPackageArgs =
+                    case (packageSetupDeps package, mdeps) of
+                        -- The package is using the Cabal custom-setup
+                        -- configuraiton introduced in Cabal 1.24. In
```
lwm: s/configuraiton/configuration/g

metaleap:
Just what I need! This isn't out yet right? I have stack v1.3.2 and get:

    Cabal file warning in path/testdep.cabal: Ignoring unknown section type: custom-setup

Curious, will this be in 1.3.3?

lwm:
@metaleap, try `stack upgrade --git` and `stack setup --ugrade-cabal`.

The functionality is in HEAD.

metaleap:
I would, had I not just found out about stack's `explicit-setup-deps` which let's me proceed with business luckily  =)  (btw probly cu at one of the upcoming Berlin Haskell meetups as soon as the snow&ice melt away!)

lwm:
> btw probly cu at one of the upcoming Berlin Haskell meetups

Ah! It would be cool but I already moved city. But send my regards! Good people. :+1:

PR #2867, opened 2016-12-21
stack new always uses UTF-8 encoding

snoyberg:

Basic sanity checking of using `stack new`, which worked. I'd like to have the user who reported the original problem confirm that this fixes it.


snoyberg:
User confirmed it worked in the Yesod mailing list.

PR #2894, opened 2017-01-04
Remove trailing slash for package list in `stack.yaml`.

### LINKED ISSUE #2887: Backslash should NOT be appended in stack.yaml created on Windows
### Steps to reproduce

1. Run `stack init` for some multi-package-project on Windows.
    For example:
    ```
    $ git clone git clone https://github.com/haskell/haddock.git
    $ cd haddock
    $ stack init
    ```

### Expected

A stack.yaml like below should be created:

```yaml
# ... snip ...
packages:
- '.'
- haddock-api/
- haddock-library/
- haddock-test/
# ... snip ...
```

### Actual

```yaml
# ... snip ...
packages:
- '.'
- haddock-api\
- haddock-library\
- haddock-test\
# ... snip ...
```

#### Why this is bad:

Using backslash as the path separator causes a cryptic error like below, when **porting the stack.yaml file to non-Windows environment such as Linux.**

```
/home/yu/prg/foreign/haddock/haddock-api\: canonicalizePath: does not exist (No such file or directory)
```

### Stack version

```
$ stack --version
Version 1.3.2, Git revision 3f675146590da4f3edf768b89355f798229da2a5 x86_64 hpack-0.15.0
```

### Method of installation

lwm:
Closes #2887.

### DIFF
--- a/src/Stack/Init.hs
+++ b/src/Stack/Init.hs
@@ -8,36 +8,36 @@ module Stack.Init
     , InitOpts (..)
     ) where
 
-import           Control.Exception (assert)
-import           Control.Exception.Safe (catchAny)
+import           Control.Exception               (assert)
+import           Control.Exception.Safe          (catchAny)
 import           Control.Monad
-import           Control.Monad.Catch (throwM)
+import           Control.Monad.Catch             (throwM)
 import           Control.Monad.IO.Class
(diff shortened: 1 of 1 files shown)


mgsloan:
Makes sense, thanks!

Issue #2888, opened 2017-01-02
`stack solver` is broken


sjakobi:
### General summary/comments (optional)

`stack solver --update-config` doesn't seem to work with either `cabal-install-1.22.9.0` or `cabal-install-1.24.0.1`. My hypothesis is that the problem is caused by `cabal` not seeing the package index.

stack v1.2.1 @ [0a96df](https://github.com/commercialhaskell/stack/commit/0a96df) works without issue.

### Steps to reproduce

For example:

```
$ stack --resolver lts-7.14 new solver-repro
$ cd solver-repro
# Add "Cartesian" to the library build-depends in the cabal file
$ stack solver --update-config
Using configuration file: stack.yaml
Using cabal packages:
- solver-repro.cabal

Using resolver: lts-7.14
Using compiler: ghc-8.0.1
Asking cabal to calculate a build plan...
Trying with packages from lts-7.14 as hard constraints...
Attempt failed.

>>>> Cabal errors begin
Warning: The package list for 'Hackage' does not exist. Run 'cabal update' to
download it.
Warning: Constraint refers to an unknown package: AC-Vector ==2.3.2 (main
config /tmp/cabal-solver13270/cabal.config)
cabal: Could not resolve dependencies:
trying: solver-repro-0.1.0.0 (user goal)
next goal: Cartesian (dependency of solver-repro-0.1.0.0)
Dependency tree exhaustively searched.
<<<< Cabal errors end

Retrying with packages from lts-7.14 as preferences...
Attempt failed.

>>>> Cabal errors begin
Warning: The package list for 'Hackage' does not exist. Run 'cabal update' to
download it.
Warning: Preference refers to an unknown package: AC-Vector ==2.3.2
cabal: Could not resolve dependencies:
trying: solver-repro-0.1.0.0 (user goal)
next goal: Cartesian (dependency of solver-repro-0.1.0.0)
Dependency tree exhaustively searched.
<<<< Cabal errors end

*** Failed to arrive at a workable build plan.

Solver could not resolve package dependencies.
You can try the following:
    - Update external packages with 'stack update' and try again.
    - Tweak stack.yaml and try again
        - Remove any unnecessary packages.
        - Add any missing remote packages.
        - Add extra dependencies to guide solver.
        - Adjust resolver.
```

### Expected

`stack solver --update-config` adds a version of  `Cartesian` to the extra-deps and the following `stack build` succeeds.

### Actual

[...]

PR #2901, opened 2017-01-06
For stack solver, use old index filename #2888

mgsloan:

Tested by getting expected results from the repro for #2888


snoyberg:
For future-proofing, should we just copying over the file to _both_ locations?

mgsloan:
Hmm, I suppose that makes sense!  A bit inefficient, but I think that's ok for `stack solver`.

When/if 02-index.tar happens, we can do something a little smarter and check the cabal version.

PR #2906, opened 2017-01-08
Copy 01-index.tar to 00-index.tar

snoyberg:
This is done for backwards compatibility with any external tools relying
on the presence of a 00-index.tar file.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,6 +17,13 @@ Behavior changes:
   URL, Git will still be used. See
   [#2780](https://github.com/commercialhaskell/stack/issues/2780)
 
+  For backwards compatibility with tools relying on the presence of a
+  `00-index.tar`, Stack will copy the `01-index.tar` file to
+  `00-index.tar`. Note, however, that these files are different; most
+  importantly, 00-index contains only the newest revisions of cabal
+  files, while 01-index contains all versions. You may still need to
+  update your tooling.
+
 Other enhancements:
 
 * Internal cleanup: configuration types are now based much more on lenses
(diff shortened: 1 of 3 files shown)


phadej:
I'm not sure whether it's worth adding a note, but e.g. `00-index.tar` had only at most one instance of each file, and had files in alphabetical order. E.g. `packdeps` couldn't cope with `01-index.tar` without changes (have to look for the latest instance of cabal file in the index). And I have "lazily" used the fact 00-index.tar is sorted in some internal tools.

snoyberg:
Oh, that's interesting. Can you explain how packdeps breaks? All the other tools I've worked on seem to work reliably with swapping in 01-index.

phadej:
packdeps *broke*, before https://github.com/snoyberg/packdeps/pull/30/files 

Before it used the first `foo.cabal` (`>`), but for `01-index.tar` it has to use `>=`, so the latter occurrences (new revisions) override.

snoyberg:
Ah, so essentially it's an issue of left- vs right-biased merging, right? That makes a lot of sense. I'd agree with putting a note _somewhere_, I'm just not sure where a good place to put it would be (unless you just mean changelog).

phadej:
I think changelog would be enough, if it's close to index-backend change note.

snoyberg:
Cool, changelog updated.

borsboom:
LGTM.  Does this fix the `cabal-solver` integration test?

borsboom:
Does this obviate #2888?

snoyberg:
No, #2901 is the proper fix for #2888. This solves a separate but related issue. What's going on:

* `stack update` today will generate a ~/.stack/indices/Hackage/00-index.tar file
* Some external tools may refer to that file (I know that stackage-curator does, for instance, though I already updated it to use 01-index.tar when available)
* `stack solver`, by contrast, always creates a temporary cabal configuration directory with its own 00-index.tar file. My previous update broke `stack solver` by renaming to 01-index.tar; #2901 fixes that temporary directory filename.

dabrowski-adam:
Couldn't 00-index.tar be just a symlink to 01-index.tar? That would significantly reduce the space used.

Issue #2911, opened 2017-01-10
stack setup --nix --upgrade-cabal fails


mboes:
### Steps to reproduce

```
$ stack new foo
$ cd foo
$ stack --nix setup --upgrade-cabal
```

### Expected

For subsequent `stack build` in the local project to be built using the latest Cabal library. Or to display an error saying that `--upgrade-cabal` is not supported in conjunction with `--nix` (arguably the Cabal library should be set in a `shell.nix` file).

### Actual

```
[...]
/nix/store/cff78wmha1b363jkhxfsxp4vg1vyslz2-ghc-8.0.1/new-cabal:
createDirectory: permission denied (Read-only file system)
Exit code ExitFailure 1 while running ["/run/user/1000/stack-cabal-upgrade6275/Cabal-1.24.2.0/Setup","install"] in /run/user/1000/stack-cabal-upgrade6275/Cabal-1.24.2.0/
```

### Stack version

```
$ stack --version
Version 1.3.2 x86_64
```

cc @robinbb @YPares @facundominguez

mboes:
Related: #2910.

YPares:
Opened a PR for that: https://github.com/commercialhaskell/stack/pull/2912

PR #2912, opened 2017-01-10
Failing when using setup --upgrade-cabal with --nix

YPares:
Aims to resolve https://github.com/commercialhaskell/stack/issues/2911

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -97,6 +97,7 @@ import           Stack.Types.Version
 import           Stack.Types.Config
 import           Stack.Types.Compiler
 import           Stack.Types.Resolver
+import           Stack.Types.Nix
 import           Stack.Types.StackT
 import           Stack.Upgrade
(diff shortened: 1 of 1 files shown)


YPares:
Is this the right point to trigger a failure?

src/main/Main.hs
```diff
@@ -563,6 +564,8 @@ pathCmd keys go = withBuildConfig go (Stack.Path.path keys)
 setupCmd :: SetupCmdOpts -> GlobalOpts -> IO ()
 setupCmd sco@SetupCmdOpts{..} go@GlobalOpts{..} = do
   lc <- loadConfigWithOpts go
+  when (scoUpgradeCabal && (nixEnable (configNix (lcConfig lc)))) $ do
+    error "--upgrade-cabal cannot be used when nix is activated"
```
borsboom: We avoid using `error` and instead throw specific Exceptions.

PR #2916, opened 2017-01-11
Update on missing SHA

snoyberg:

OK, this is a bit involved. Ignoring cabal file revisions for a moment, consider you run `stack build` in a project using lts-7.15, and the project uses `foldl-1.2.2`. Suppose you had last updated your package index when foldl-1.2.2 wasn't available. Stack will automatically update your index before trying to build so that it can get the .cabal file for foldl-1.2.2.

Now consider a similar situation: your project uses yesod-core, which for nightly-2017-01-10 is version 1.4.30, with cabal file revision #1. If your index is missing that revision, previously you would get an error message about a missing revision. With this patch, you'll instead get the behavior that the index will automatically update.

In addition, there are commits here to improve the error messages delivered (to indicate whether the lookup occurred in a Git repo or in the tarball), and to standardize logic around choosing Git vs HTTP index methods.

## Testing

In order to test, I first created a `00-index.tar` file with the most recent content (by running `stack update`), and then pruned all recent .cabal files (including the first revision of yesod-core-1.4.30) with the following program run from `~/.stack/indices/Hackage`:

```haskell
#!/usr/bin/env stack
{- stack --resolver lts-7.14 --install-ghc runghc
    --package tar
    --package classy-prelude-conduit
-}
{-# LANGUAGE NoImplicitPrelude, OverloadedStrings #-}
import ClassyPrelude.Conduit
import qualified Codec.Archive.Tar as Tar
import qualified Codec.Archive.Tar.Entry as Tar
import Codec.Compression.GZip (decompress)
import System.Directory

main :: IO ()
main = do
    void $ tryIO $ removeFile "01-index.cache"
    void $ tryIO $ removeFile "01-index.tar"
    lbs <- readFile "01-index.tar.gz"
    let entries = Tar.read $ decompress lbs
    writeFile "01-index.tar" $ Tar.write $ go entries
  where
    go (Tar.Next e es)
        | toKeep e = e : go es
        | otherwise = go es
    go Tar.Done = []
    go (Tar.Fail e) = impureThrow e

    toKeep e = Tar.entryTime e < 1483747200
```

Then, I ran the following:

```
$ rm -rf yesod-core-1.4.30/ && stack --resolver nightly-2017-01-10 unpack yesod-core
```

Before this change: I would get an error message about a missing Git SHA. After this change, I get the following:

```
$ rm -rf yesod-core-1.4.30/ && stack --resolver nightly-2017-01-10 unpack yesod-core
Missing some cabal revision files, updating indices
Downloading package index from https://s3.amazonaws.com/hackage.fpcomplete.com/01-index.tar.gz                  
Populated index cache.                                                                                          
Unpacked yesod-core-1.4.30 to /Users/michael/Desktop/yesod-core-1.4.30/
```

I also confirmed that the output yesod-core-1.4.30/yesod-core.cabal file contains:

```
x-revision: 1
```

### DIFF
--- a/src/Stack/Hoogle.hs
+++ b/src/Stack/Hoogle.hs
@@ -100,6 +100,17 @@ hoogleCmd (args,setup,rebuild) go = withBuildConfig go pathToHaddocks
         hooglePackageIdentifier <-
             do (_,_,resolved) <-
                    resolvePackagesAllowMissing
+
+                       -- FIXME this Nothing means "do not follow any
+                       -- specific snapshot", which matches old
+                       -- behavior. However, since introducing the
+                       -- logic to pin a name to a package in a
+                       -- snapshot, we may arguably want to ensure
+                       -- that we're grabbing the version of Hoogle
+                       -- present in the snapshot currently being
+                       -- used.
+                       Nothing
+
                        mempty
                        (Set.fromList [hooglePackageName])
                return
--- a/src/Stack/Hoogle.hs
+++ b/src/Stack/Hoogle.hs
@@ -98,8 +98,10 @@ hoogleCmd (args,setup,rebuild) go = withBuildConfig go pathToHaddocks
             hoogleMinIdent =
                 PackageIdentifier hooglePackageName hoogleMinVersion
         hooglePackageIdentifier <-
-            do (_,_,resolved) <-
+            do menv <- getMinimalEnvOverride
+               (_,_,resolved) <-
                    resolvePackagesAllowMissing
+                       menv
 
                        -- FIXME this Nothing means "do not follow any
                        -- specific snapshot", which matches old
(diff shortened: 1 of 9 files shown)


src/Stack/Hoogle.hs
```diff
@@ -100,6 +100,17 @@ hoogleCmd (args,setup,rebuild) go = withBuildConfig go pathToHaddocks
...
                    resolvePackagesAllowMissing
+
+                       -- FIXME this Nothing means "do not follow any
+                       -- specific snapshot", which matches old
+                       -- behavior. However, since introducing the
+                       -- logic to pin a name to a package in a
+                       -- snapshot, we may arguably want to ensure
+                       -- that we're grabbing the version of Hoogle
+                       -- present in the snapshot currently being
+                       -- used.
```
borsboom: The behaviour described in the FIXME does seem more expected to me.  I wonder how many people rely on the current behaviour.

snoyberg: I'll open up a new Github issue about this then, but changing the behavior of this doesn't seem like it should block the current PR.

Issue #2895, opened 2017-01-04
hs-source-dirs starting with `../` are not included in the -i<dir> arguments when using `stack ghci file.hs`


tallen-imvu:
### Summary
`stack ghci <target file>` will correctly determine include directories using the cabal file's hs-source-dirs property unless the source directory starts with a `../`
`stack build` however deals with this correctly.

### Steps to reproduce

1. add a relative path to hs-source-dirs:
    ex:
          `hs-source-dirs:      . ../extra`

2. Add a module to the ../extra directory and import it in your main file
3. run command `stack build`
    everything should work
4. run command `stack ghci <path to main>`

### Expected

ghci launched with package loaded

### Actual

Error in compiling due to failing to load the extra module. Because it cannot find it.

running the `stack ghci` command with `--version` shows that the `../extra` directory is not being added to the list of `-i<directory>` flags passed to ghci.
If the module is moved to a directory that does not require a `../` prefix, then that directory is included properly.

### Stack version

```
$ stack --version
Version 1.3.2, Git revision 3f675146590da4f3edf768b89355f798229da2a5 (4395 commits) x86_64 hpack-0.15.0
```

### Method of installation

* static binary from https://www.stackage.org/stack/linux-x86_64-static

mgsloan:
Hmm, I cannot reproduce the issue.  For me if I have

```
library
  hs-source-dirs:      src ../issue2895-extra
  exposed-modules:     Lib Huh
```

It works (`Huh.hs` is in `../issue2895-extra`).  Perhaps your module is not listed in `exposed-modules` / `other-modules`?

tallen-imvu:
I put up a reproducing repo [here](https://github.com/tallen-imvu/stack-ghci-i)

tallen-imvu:
Oh, there it is.
It is the `parseRelDir` on line 423 of src/Stack/Package.hs
Pull request incomming

mgsloan:
Thanks for finding and fixing!!

PR #2920, opened 2017-01-13
for 'stack ghci', don't omit hsSourceDirs that have '..' in them. #2895

tallen-imvu:
* [NA] The documentation has been updated, if necessary.

Tested using [this test repo](https://github.com/tallen-imvu/stack-ghci-i)
With these code changes, all three examples can correctly load in `stack ghci`

### DIFF
--- a/src/Path/Extra.hs
+++ b/src/Path/Extra.hs
@@ -44,7 +44,7 @@ parseCollapsedAbsDir = parseAbsDir . collapseFilePath
 parseCollapsedAbsFile :: MonadThrow m => FilePath -> m (Path Abs File)
 parseCollapsedAbsFile = parseAbsFile . collapseFilePath
 
--- | Combine a 'Path Abs Dir' and a relative 'FilePath' into a 'Path Abs Dir'
+-- | Add a relative FilePath to the end of a Path
 -- We can't parse the FilePath first because we need to account for ".."
 -- in the FilePath (#2895)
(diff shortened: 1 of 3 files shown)


src/Path/Extra.hs
```diff
@@ -43,6 +44,12 @@ parseCollapsedAbsDir = parseAbsDir . collapseFilePath
 parseCollapsedAbsFile :: MonadThrow m => FilePath -> m (Path Abs File)
 parseCollapsedAbsFile = parseAbsFile . collapseFilePath
 
+-- | Combine a 'Path Abs Dir' and a relative 'FilePath' into a 'Path Abs Dir'
```
lwm: I'm not sure about this description. It's a `m (Path Abs Dir)`, not `Path Abs Dir`? I applaud the effort to write a nice description but I don't think we should describe the type. Ignore if pedantic ;)

tallen-imvu: Heh. It makes me a bit uncomfortable as well. Perhaps it is better to be more colloquial and terse?
`+-- | Add a relative FilePath to the end of a Path`

lwm: Oh, that reads nicely.

tallen-imvu:
ping ?

mgsloan:
LGTM.  Thanks!  Sorry for the delay

PR #2940, opened 2017-01-20
Use hackage security

snoyberg:

Normal usage of Stack will test this out plenty. I made this change because, once we switched from the Git index to HTTPS, updates became very slow, which I'm not happy with. As long as we're doing partial updates, may as well use hackage-security I suppose.

Since the hackage-security-http-client package isn't yet on Hackage, I inlined that module. Also, manually specified indices will be able to opt in or out of Hackage Security for downloads, so compatibility with existing non-HS indices will be retained.

I know the initial build will fail due to missing extra-deps. After that, I'm planning on using this myself a bit before merging. Review certainly welcome, this was far more complicated to get right than I would have ever expected.

### DIFF
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -321,6 +321,11 @@ package-indices:
   # optional fields, both default to false
   gpg-verify: false
   require-hashes: false
+
+  # Starting with stack 1.4, we default to using Hackage Security
+  hackage-security:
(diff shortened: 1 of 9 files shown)


snoyberg:
I've been using this regularly for the past few days and haven't experienced any issues. Since I've been doing Stackage curation in that time, I've been very regularly running `stack update`, and my results have always matched up with the Travis job using the current version of Stack. I think this is safe to merge.

PR #2949, opened 2017-01-26
Updated archlinux installation docs

MichielDerhaeg:
The maintainers of the Archlinux community repository have been doing a very good job of keeping a lot of haskell packages very up to date. So this is in my opinion the recommended way of installing stack on the platform and this should be mentioned in the documentation.


MichielDerhaeg:
tried merging the wrong branch, redoing it..

PR #2950, opened 2017-01-26
update archlinux installation docs

MichielDerhaeg:
The maintainers of the Archlinux community repository have been doing a very good job of keeping a lot of haskell packages very up to date. So this is in my opinion the recommended way of installing stack on the platform and this should be mentioned in the documentation.

### DIFF
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -182,7 +182,17 @@ it.  To install it:
 
 ## Arch Linux
 
-Use the [generic Linux option](#linux).
+There is an official package in the Arch community repository. So you can
+install it by simply doing:
+
+    sudo pacman -S stack
+
+Note that this version may slightly lag behind, but it should be updated within
+the day. The package is also always rebuilt and updated when one of it's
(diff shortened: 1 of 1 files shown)


doc/install_and_upgrade.md
```diff
@@ -194,13 +204,6 @@ If you use the
 [ArchHaskell repository](https://wiki.archlinux.org/index.php/ArchHaskell), you
 can also get the `haskell-stack-tool` package from there.
 
-There is also an unofficial package in the Arch community repository.  Note that this Stack
-version may lag behind, so we recommend running `stack upgrade` after installing
```
lwm: The `stack upgrade` advice seems like it shouldn't be lost?

MichielDerhaeg: It doesn't seem necessary to me, the version in the repository is always kept up to date. And in the small chance this isn't' the case, and you run it once and forget about it, your local version will lag behind if you don't keep doing it.

mgsloan:
Seems reasonable, thanks!

Issue #2386, opened 2016-07-18
Add `stack setup --install-cabal VERSION` in addition to `stack setup --upgrade-cabal`.


sjakobi:
This should make it easier to test `stack` with different versions of `Cabal`, as required for e.g. #2350.

sjakobi:
A few pointers:

* Most (all?) of the related code lives in either `Stack.SetupCmd` or `Stack.Setup`.
* I'd suggest generalizing `Stack.Setup.ugradeCabal` to `installCabal` so it can accept a concrete version or default to the newest available version of `Cabal`

Blaisorblade:
What about picking a fixed `cabal-install` version for each stack version, and installing it in isolation? Ideally, the newest one for the supported cabal library.
(Why is stack not yet on Cabal 1.24 BTW?)

borsboom:
Stack doesn't use cabal-install for much (just the solver AFAIK) so not really any need to have cabal-install match the Cabal library version it was built with.  For building Stack, we stick with the versions of packages that are in the current LTS snapshot unless there's a really compelling reason to add an extra-dep, so once LTS 7 is released Stack will upgrade to a newer Cabal library version (along with building with GHC 8.0.1). And we haven't been in a rush to have Stack use Cabal-1.24 because we don't need any of its new features.

lwm:
Can be closed off now that #2995 went in.

mgsloan:
Indeed! Closing

PR #2995, opened 2017-02-10
Add `--cabal-install`

lwm:

Manually tested.

Based on https://github.com/commercialhaskell/stack/issues/2386#issuecomment-233304228. Thanks @sjakobi for the pointers!

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -138,7 +138,7 @@ data SetupOpts = SetupOpts
     -- ^ Don't check for a compatible GHC version/architecture
     , soptsSkipMsys :: !Bool
     -- ^ Do not use a custom msys installation on Windows
-    , soptsUpgradeCabal :: !Bool
+    , soptsUpgradeCabal :: !(Maybe UpgradeTo)
     -- ^ Upgrade the global Cabal library in the database to the newest
     -- version. Only works reliably with a stack-managed installation.
     , soptsResolveMissingGHC :: !(Maybe Text)
@@ -234,7 +234,7 @@ setupEnv mResolveMissingGHC = do
             , soptsSanityCheck = False
             , soptsSkipGhcCheck = configSkipGHCCheck config
             , soptsSkipMsys = configSkipMsys config
-            , soptsUpgradeCabal = False
+            , soptsUpgradeCabal = Nothing
             , soptsResolveMissingGHC = mResolveMissingGHC
             , soptsSetupInfoYaml = defaultSetupInfoYaml
             , soptsGHCBindistURL = Nothing
@@ -493,11 +493,11 @@ ensureCompiler sopts = do
                 m <- augmentPathMap (edBins ed) (unEnvOverride menv0)
                 mkEnvOverride (configPlatform config) (removeHaskellEnvVars m)
 
-    when (soptsUpgradeCabal sopts) $ do
+    forM_ (soptsUpgradeCabal sopts) $ \version -> do
         unless needLocal $ do
-            $logWarn "Trying to upgrade Cabal library on a GHC not installed by stack."
+            $logWarn "Trying to change a Cabal library on a GHC not installed by stack."
             $logWarn "This may fail, caveat emptor!"
-        upgradeCabal menv wc
+        upgradeCabal menv wc version
 
     case mtools of
         Just (Just (ToolGhcjs cv), _) -> ensureGhcjsBooted menv cv (soptsInstallIfMissing sopts) (soptsGHCJSBootOpts sopts)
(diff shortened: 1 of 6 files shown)


src/Stack/Setup.hs
```diff
@@ -140,6 +140,8 @@ data SetupOpts = SetupOpts
     , soptsUpgradeCabal :: !Bool
     -- ^ Upgrade the global Cabal library in the database to the newest
     -- version. Only works reliably with a stack-managed installation.
+    , soptsInstallCabal :: !(Maybe Version)
+    -- ^ Install a specific version of Cabal.
```
sjakobi: Can you fuse the `soptsUpgradeCabal` and `soptsInstallCabal` fields into one? This should make it impossible to have nonsensical states where `soptsUpgradeCabal == True` and `soptsInstallCabal == Just <x>`.

Maybe it would even be good to avoid the additional `--install-cabal` flag and only add an optional argument to `--upgrade-cabal`?! That would help combating CLI inflation :). Not sure if it's possible with optparse-applicative though…

sjakobi:
I haven't done a thorough review yet, only added an inline comment on the design so far.

Could you in any case please fix your commit description to use the correct flag name? Cheers! :)

lwm:
> Could you in any case please fix your commit description to use the correct flag name? Cheers! :)

Oh yes, fixed that one :smile:

src/Stack/Setup.hs
lwm: This sounds like a good idea. I haven't seen an 'optional argument' but will have a look and see what is possible.

lwm: Oh, I see that [options with optional arguments are not supported](https://github.com/pcapriotti/optparse-applicative#faq).

lwm:
I've done another pass on this and reduced the diff quite a bit. It should be much easier to review. If there are any other ideas on the merge of `--upgrade-cabal/--install-cabal`, I'm all ears (see comments above).

src/Stack/Setup.hs
sjakobi: Fusing the two fields and parsing the two options as alternatives would still make sense IMO. As a result the usage info should look roughly like this:

```
stack setup … [--upgrade-cabal | --install-cabal VERSION] …
``` 

The combined field should have a type like `Maybe UpgradeTo` with

```
data UpgradeTo = Latest | Specific Version
```

or so.

The default value (which applies when neither option is used) is `Nothing`.

sjakobi: BTW, as an example `stack clean` uses the `Alternative` instance of the option parser:

```
Usage: stack clean ([PACKAGE] | [--full]) [--help]
```

lwm: Oh! That was really helpful, thanks for even more pointers :)

lwm:
Latest stab at it is https://github.com/commercialhaskell/stack/pull/2995/commits/c0942c65c64703801d47347cb3b2c880a0b5b0e6.

src/Stack/Setup.hs
```diff
@@ -489,11 +489,12 @@ ensureCompiler sopts = do
...
 
-    when (soptsUpgradeCabal sopts) $ do
+    let cabalVersion = soptsUpgradeCabal sopts
+    when (isJust $ cabalVersion) $ do
         unless needLocal $ do
-            $logWarn "Trying to upgrade Cabal library on a GHC not installed by stack."
+            $logWarn "Trying to change a Cabal library on a GHC not installed by stack."
             $logWarn "This may fail, caveat emptor!"
-        upgradeCabal menv wc
+        upgradeCabal menv wc (fromJust cabalVersion)
```
sjakobi: It would be nicer to use `forM_` instead of `when isJust …` and `fromJust`.

src/Stack/Setup.hs
```diff
@@ -622,46 +623,47 @@ ensureDockerStackExe containerPlatform = do
...
     rmap <- resolvePackages menv Nothing Map.empty (Set.singleton name)
-    newest <-
-        case map rpIdent rmap of
+    version <- case cabalVersion of
+        Specific v -> return v
+        Latest     -> case map rpIdent rmap of
             [] -> error "No Cabal library found in index, cannot upgrade"
-            [PackageIdentifier name' version]
-                | name == name' -> return version
+            [PackageIdentifier name' version] | name == name' -> return version
```
sjakobi: This block seems unnecessarily monadic, could be pure.

src/Stack/Setup.hs
```diff
@@ -622,46 +623,47 @@ ensureDockerStackExe containerPlatform = do
...
         then $logInfo $ T.concat
             [ "Currently installed Cabal is "
             , T.pack $ versionString installed
-            , ", newest is "
-            , T.pack $ versionString newest
-            , ". I'm not upgrading Cabal."
+            , case cabalVersion of
+                  Specific _ -> " (specified)"
+                  Latest     -> " (latest)"
+            , ". No work to be done."
```
sjakobi: The string that is being formed here doesn't look right. Anyway, I think we should support installing an older version here. Not sure if that requires adding logic to unregister or hide the old version.

src/Stack/Setup.hs
```diff
@@ -622,46 +623,47 @@ ensureDockerStackExe containerPlatform = do
...
                 , T.pack $ versionString installed
                 ]
-            let ident = PackageIdentifier name newest
+            let ident = PackageIdentifier name version
             -- Nothing below: use the newest .cabal file revision
             m <- unpackPackageIdents menv tmpdir Nothing (Map.singleton ident Nothing)
 
             compilerPath <- join $ findExecutable menv (compilerExeName wc)
-            newestDir <- parseRelDir $ versionString newest
+            newestDir <- parseRelDir $ versionString version
```
sjakobi: For consistency, that should probably be "versionDir".

src/Stack/Types/Version.hs
```diff
@@ -64,6 +65,9 @@ instance Exception VersionParseFail
 instance Show VersionParseFail where
     show (VersionParseFail bs) = "Invalid version: " ++ show bs
 
+-- | A package version or the latest.
+data UpgradeTo = Specific Version | Latest deriving (Show)
+
```
sjakobi: The description is way more general than the name of the type. Can you come up with a more general name maybe?

src/Stack/Setup.hs
lwm: I don't know what this means :smile:

src/Stack/Setup.hs
lwm: This formatting corresponds to stopping the _same_ version being installed. If you want to get an old version, you just fire off `--install-cabal FOO` instead of `--upgrade-cabal`. Good question regarding the unregister/hide stuff. Will investigate.

src/Stack/Types/Version.hs
lwm: Hmmm ... tried the description again. I like the name :)

src/Stack/Setup.hs
lwm: Fixed :+1:

src/Stack/Setup.hs
lwm: Not sure how to get the `when isJust` out of it?. See https://github.com/commercialhaskell/stack/pull/2995/commits/734dae44615208d8404229cf1fc64e7ec989bbf7.

src/Stack/Setup.hs
lwm: Examples are:

```
λ :main ["setup", "--upgrade-cabal"]
Currently installed Cabal is 1.24.2.0 (latest). No work to be done.
```
And:
```
λ :main ["setup", "--install-cabal", "1.24.2.0"]
Currently installed Cabal is 1.24.2.0 (specified). No work to be done.
```

src/Stack/Setup.hs
```diff
@@ -494,7 +494,7 @@ ensureCompiler sopts = do
         unless needLocal $ do
             $logWarn "Trying to change a Cabal library on a GHC not installed by stack."
             $logWarn "This may fail, caveat emptor!"
-        upgradeCabal menv wc (fromJust cabalVersion)
+        forM_ cabalVersion (upgradeCabal menv wc)
```
sjakobi: No, the `forM_` should wrap the entire block:

```
forM_ cabalVersion $ \v -> do
  unless …
```

src/Stack/Setup.hs
sjakobi: The only monadic bit about this case expression is the `return`s – this should be a let-expression instead of using `<-`.

src/Stack/Setup.hs
sjakobi: Ok, I think I misread the diff a bit. Still, even when `version` < `installed`, we want to setup that old `version` of Cabal.

mgsloan:
Looks like good stuff!  I certainly recommend making sure the downgrade case works.  I'd also have expected it to require unregistering the old one.

snoyberg:
Moving over the discussion from #3019:

Upgrading the global Cabal package is an ugly hack that I added in towards the beginning of Stack. It was necessary for cases where a custom setup script required some new functionality in Cabal that was not present in the shipped-with-GHC version. However, since then, we've had a significant change: `custom-setup` stanzas. With these, we're no longer trapped by random whims of the package database, and trying to guess what set of packages should be available to Setup.hs. In particular:

* Stackage snapshots will now build packages reliably against the snapshot version of Cabal
* If a specific Cabal version is needed, it can be specified in a .cabal file
* Instead of being pinned to building against the global version of Cabal, we can build against the snapshot version

So I'd like to go in almost the opposite direction of this PR (apologies to @lwm and @sjakobi for only commenting now):

* Get rid of the hack of upgrading global Cabal versions at all
* Give massive warnings if a custom setup script does not specify its dependencies via custom-setup

I'd go so far as considering upstream—in Stackage itself—beginning to block packages that don't define a custom-setup stanza.

sjakobi:
@snoyberg Interesting!

So to check that stack is compatible with Setup.hs files built with several versions of `Cabal` I'd simply use a different resolver (see e.g. https://github.com/commercialhaskell/stack/issues/2350 for a usecase)?
Will overriding the resolver's version of `Cabal` via `extra-deps` or `packages` be supported?

snoyberg:
I'm pretty sure that overriding the Cabal version will work with `master`, though to be honest I haven't tested it explicitly.

sjakobi:
@snoyberg 

> I'm pretty sure that overriding the Cabal version will work with master

Not sure if I'm missing something but it doesn't seem to work:

```
~/tmp $ stack --resolver lts-8.2 new cabal-setup
~/tmp $ cd cabal-setup/
~/t/cabal-setup $ vim stack.yaml # Add Cabal-1.24.0.0 to extra-deps
~/t/cabal-setup $ stack build
Warning: File listed in cabal-setup.cabal file does not exist: README.md
cabal-setup-0.1.0.0: configure (lib + exe)
Configuring cabal-setup-0.1.0.0...
cabal-setup-0.1.0.0: build (lib + exe)
Preprocessing library cabal-setup-0.1.0.0...
[1 of 1] Compiling Lib              ( src/Lib.hs, .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/Lib.o )
Preprocessing executable 'cabal-setup-exe' for cabal-setup-0.1.0.0...
[1 of 1] Compiling Main             ( app/Main.hs, .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/cabal-setup-exe/cabal-setup-exe-tmp/Main.o )
Linking .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/cabal-setup-exe/cabal-setup-exe ...
Warning: File listed in cabal-setup.cabal file does not exist: README.md
cabal-setup-0.1.0.0: copy/register
Installing library in
/home/simon/tmp/cabal-setup/.stack-work/install/x86_64-linux/lts-8.2/8.0.2/lib/x86_64-linux-ghc-8.0.2/cabal-setup-0.1.0.0-Bd2jGUn5HBKJ5lzwApT2Rh
Installing executable(s) in
/home/simon/tmp/cabal-setup/.stack-work/install/x86_64-linux/lts-8.2/8.0.2/bin
Registering cabal-setup-0.1.0.0...
```

Judging from the build paths stack used `Cabal-1.24.2.0` instead of `1.24.0.0` to build the Setup.hs.

snoyberg:
Is this using custom setup at all?

sjakobi:
> Is this using custom setup at all?

No, but should that be a requirement for testing compatibility between stack and Cabal?

snoyberg:
Ah, I hadn't understood that this is the case you're trying to address. I
guess this feature makes sense in that context. Ignore my objections.

sjakobi:
Ah, good! Maybe the usage info for `--upgrade-cabal` and `install-cabal` should note that these flags are only for debugging now.

snoyberg:
SGTM 👍

lwm:
Kind of fell off on this one. Is this PR still relevant? I see lots of movement around Cabal stuff.

mgsloan:
@lwm I think the main unresolved issue is handling the downgrade case.  Currently with this patch, if I have Cabal-1.24.2.0 and run `stack setup --install-cabal 1.24.0.0`, I get `Currently installed Cabal is 1.24.2.0 (specified). No work to be done.`

I think this is still relevant.  I don't see how custom-setup obviates the need for this, since the global cabal is still used for everything else.

lwm:
Thanks for feedback @mgsloan and @sjakobi.

I've given this another stab. Apologies that the diff on this latest patch diverged so far but I tried a new approach. It appears the downgrade case is covered now but it takes so long to get a cabal install on my crappy machine, I didn't spend long testing ;)

lwm:
Failure unrelated ... tests seem to be borked:

```haskell
src/test/Stack/ConfigSpec.hs:154:22:
    No instance for (FromJSON (WithJSONWarnings ConfigMonoid))
      arising from a use of ‘decodeEither’
    In the expression: decodeEither defaultConfigYaml
    In an equation for ‘parsed’:
        parsed = decodeEither defaultConfigYaml
    In the expression:
      do { let parsed :: Either String (WithJSONWarnings ConfigMonoid)
               parsed = decodeEither defaultConfigYaml;
           isRight parsed `shouldBe` True }
```

mgsloan:
@lwm My mistake, sorry!  Should be fixed after rebasing.

lwm:
No worries @mgsloan! Travis said OK but now Appveyor said:

```
C:\stack\src\test\Stack\ConfigSpec.hs:154:53: error:
    * Exception when trying to run compile-time code:
        InvalidAbsDir "/"
  error, called at src/Path.hs:242:17 in path-0.5.12-DPixZY6jeMPCLEenb9rPGQ:Path
      Code: mkAbsDir "/"
    * In the untyped splice: $(mkAbsDir "/")
--  While building package stack-1.4.1 using:
```

The never ending pull request lives on!

mgsloan:
Oops, fixed that too.  I'm going to go ahead and merge this.  Thanks!!

PR #3002, opened 2017-02-12
Guard fixOutput with setOutput instead of setInput

RyanGlScott:
I adapted the `fixCodePage` function for use in my standalone [`code-page`](http://hackage.haskell.org/package/code-page) library, and while doing so I noticed a small mix-up in its implementation. `fixCodePage` uses the internal functions `fixInput` and `fixOutput` to actually change the input and output codepages respectively, but while `fixInput` is guarded by the value `setInput`, `fixOutput` isn't guarded by `setOutput`—it's guarded by `setInput` too! This seems like a mistake, unless I'm mistaken.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -351,7 +351,7 @@ fixCodePage inner = do
                     (liftIO $ setConsoleCP origCPI)
                 | otherwise = id
             fixOutput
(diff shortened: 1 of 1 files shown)


snoyberg:
Thank you!

PR #3003, opened 2017-02-12
Bump to LTS-8.

lwm:
Not sure what policy is here but taking a stab. I suppose there should still be a `compiler: ": #stack 8.0.1"` in the build matrix?


borsboom:
Oops, looks like we were working on this concurrently.  I've pushed nearly identical changes to master.

lwm:
:+1:

PR #3004, opened 2017-02-14
Use cryptonite directly instead of cryptohash

vincenthz:
As a side effect:

* Add a direct dependency to memory instead of byteable
* No new dependencies cryptohash depends on cryptonite, and byteable on memory
* Remove 3 dependencies (byteable, bytestring-base16, cryptohash

ran `stack tests` succesfully, but not sure how much that cover the code that I changed.


snoyberg:
This is great, removing dependencies is one of the best activities we can ever undertake! :)

vincenthz:
thanks for merging @snoyberg . wait until you see the PR to move stack to foundation :joy:

Issue #2022, opened 2016-04-11
.prof file is not created when running 'stack exec' with +RTS -p flag on Windows.


ahgibbons:
When building a program with the `--executable-profiling --library-profiling --ghc-options="-fprof-auto -rtsopts"` flags, a .prof file is not produced when using `stack exec` with the `+RTS -p` flag. This problems occurs on Windows 10 with stack version 1.0.4

**Steps to Reproduce:**
1. Create new stack project with `stack new A`
2. Build project with profiling flags with `stack build --executable-profiling --library-profiling --ghc-options="-fprof-auto -rtsopts"`
3. Run the executable with profiling flags with `stack exec A-exe.exe -- +RTS -p`

**Expected**
I expect a profiling file A-exe.prof to be created somewhere within the project directory.

**Actual**
No profile summary file can be found. A profile file can be made by executing the file directly without using stack.

Here is the stack version output:

```
> stack --version
Version 1.0.4, Git revision cf18703b1392a96a5a4896a560309e501af63260 (3220 commits) x86_64
```

This is the stack command I ran with the output:

```
> stack exec A-exe.exe --verbose -- +RTS -p
Version 1.0.4, Git revision cf18703b1392a96a5a4896a560309e501af63260 (3220 commits) x86_64
2016-04-11 19:12:53.846001: [debug] Checking for project config at: C:\Users\andrew\Dropbox\Haskell\A\stack.yaml @(stack_7LDm9brrZDRGPali8XJ7ve:Stack.Config src/St
ack\Config.hs:761:9)
2016-04-11 19:12:53.892878: [debug] Loading project config file stack.yaml @(stack_7LDm9brrZDRGPali8XJ7ve:Stack.Config src/Stack\Config.hs:779:13)
2016-04-11 19:12:53.908504: [debug] Trying to decode C:\stack_root\build-plan-cache\x86_64-windows\lts-5.12.cache @(stack_7LDm9brrZDRGPali8XJ7ve:Data.Binary.Versio
nTagged src/Data\Binary\VersionTagged.hs:55:5)
2016-04-11 19:12:53.955378: [debug] Success decoding C:\stack_root\build-plan-cache\x86_64-windows\lts-5.12.cache @(stack_7LDm9brrZDRGPali8XJ7ve:Data.Binary.Versio
nTagged src/Data\Binary\VersionTagged.hs:64:13)
[...]

PR #3012, opened 2017-02-16
Rts opts

AndreasPK:
* [X ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [ X?] The documentation has been updated, if necessary. 
 I'm not sure if the change should also be documented in the user guide. For not it's only in the Changelog

I've tested it by running various variants of stack exec with --rts-options to make sure the arguments are passed properly. Mostly with echo but also on other executables.

This provides a workaround for https://github.com/commercialhaskell/stack/issues/2022 and allows the same invocation to work on both Linux and Windows.

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -729,7 +729,8 @@ sdistCmd (dirs, mpvpBounds, ignoreCheck, sign, sigServerUrl) go =
 
 -- | Execute a command.
 execCmd :: ExecOpts -> GlobalOpts -> IO ()
-execCmd ExecOpts {..} go@GlobalOpts{..} =
+execCmd e@ExecOpts {..} go@GlobalOpts{..} =
+    print e >>
     case eoExtra of
         ExecOptsPlain -> do
             (cmd, args) <- case (eoCmd, eoArgs) of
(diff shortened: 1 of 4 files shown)


snoyberg:
Generally speaking, I would expect a plural argument like this to parse the input and allow for multiple options at once, whereas a singular-named argument (i.e., `--rts-option`) would have the current behavior.

Also, generally, it would be better if PRs did not contain unrelated whitespace changes. For others reviewing, you can add `?w=1` to ignore those.

AndreasPK:
Thank you for the feedback.
I missed the issue with multiply arguments since echo doesn't differentiate between two arguments or one including spaces. It's fixed now.

src/main/Main.hs
```diff
@@ -773,7 +773,7 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
                 -- Add RTS options to arguments
                 let argsWithRts args = if null eoRtsOptions 
                             then args :: [String]
-                            else args ++ ["+RTS"] ++ eoRtsOptions ++ ["-RTS"]
+                            else args ++ ["+RTS"] ++ concatMap words eoRtsOptions ++ ["-RTS"]
```
snoyberg: Sorry, I should have clarified that there's support in Stack already for parsing these kinds of arguments, please see: https://github.com/commercialhaskell/stack/blob/master/src/Options/Applicative/Args.hs. I _think_ `argsOption` will handle this correctly, including issues like escape characters and quoted whitespace.

AndreasPK: the --package and --ghc-options flags still use text/stringOption and take multiple arguments I assume these are just that way because of age?

I assume string/textOptions is just legacy method to get a list of these?

snoyberg: It makes sense for --package, since (until the script command) it only took one package at a time, and even in the script command case, it doesn't follow normal escaping rules.

The --ghc-options though concerns me. Perhaps the String value is being parsed afterwards. There are a few different --ghc-options at play. Which one were you looking at?

snoyberg:
Thank you!

src/main/Main.hs
AndreasPK: The build one here:
https://github.com/commercialhaskell/stack/blob/master/src/Stack/Options/BuildParser.hs#L39

snoyberg: After staring at this for a while, I think it's OK. That usage gets passed to Cabal inside a `--ghc-options` flag itself, so therefore it becomes Cabal's responsibility to parse it into individual flags to hand off to GHC. Though to be fair I'm not 100% certain. I've asked some others to take a look.

Issue #2795, opened 2016-11-21
Building project dependent on gtk2hs fails on lts-7.8 and later


fujiyan:
### General summary
Building project dependent on glib fails on lts-7.8 and later after Cabal becomes 1.24.1.0.

### Steps to reproduce

1. Add glib to a project's .cabal file
('gtktest' is the project name.)
```
    build-depends:       base
                       , gtktest
                       , glib

```
2. Run following commands

```
    stack install gtk2hs-buildtools
    stack build

```

### Expected

I expected that the build will succeed.

### Actual

The following error occurred
verbose : [failure.txt](https://gist.github.com/fujiyan/973a506b0e30aef429fed24261134b65)
```
...

2016-11-17 22:07:54.899952: [debug] Getting Cabal package version
@(Stack\GhcPkg.hs:171:5)
2016-11-17 22:07:54.900952: [debug] Run process: C:\Users\foo\AppData\Local\Programs\stack\x86_64-windows\ghc-8.0.1\bin\ghc-pkg.EXE --no-user-package-db field --simple-output Cabal version
@(System\Process\Read.hs:277:3)
2016-11-17 22:07:54.900952: [debug] Getting global package database location
@(Stack\GhcPkg.hs:54:5)

...

[debug] Ignoring package Cabal due to wanting version 1.24.1.0 instead of 1.24.0.0
@(Stack\Build\Installed.hs:196:5)

...

--  While building package glib-0.13.4.1 using:
  C:\Users\foo\AppData\Local\Programs\stack\x86_64-windows\ghc-8.0.1\bin\ghc.EXE --make -odir C:\Users\foo\AppData\Local\Temp\stack5512\glib-0.13.4.1\.stack-work\dist\b7fec021\setup -hidir C:\Users\foo\AppData\Local\Temp\stack5512\glib-0.13.4.1\.stack-work\dist\b7fec021\setup -i -i. -package=Cabal-1.24.0.0 -clear-package-db -global-package-db -package-db=C:\sr\snapshots\a78c6a89\pkgdb C:\Users\foo\AppData\Local\Temp\stack5512\glib-0.13.4.1\Setup.hs -o C:\Users\foo\AppData\Local\Temp\stack5512\glib-0.13.4.1\.stack-work\dist\b7fec021\setup\setup
Process exited with code: ExitFailure 1
Logs have been written to: C:\Users\foo\Documents\Files\Haskell\gtktest\.stack-work\logs\glib-0.13.4.1.log

[...]

PR #3019, opened 2017-02-22
Remove --upgrade-cabal recommendation

snoyberg:
As we started discussing on Slack, and related to #2795. The original issue is no longer present in Stack due to the usage of custom setup deps now being fully supported. If I manually remove the custom-setup stanza from the glib cabal file, the problem reappears. In theory, the right solution is to notice that there's a Cabal library present in the snapshot database and use that instead. I don't want to do that though, because:

1. It will add extra code (and possibly a performance hit) to do all of the checking for Cabal library versions. I _think_ this information is already loaded up by Stack, but if not, it's an extra `ghc-pkg` call, which isn't cheap.
2. It adds some level of non-reproducibility to build plans, as the results will depend on whether we first ran `stack build Cabal` or not.

I've instead added a warning in the case of a missing `custom-setup` stanza. I believe this is the right approach: all of these build issues will disappear (in theory) with properly provided custom-setup, so we should be encouraging them.

I'd go so far as to say that, in the not-too-distant future, removing the ability to upgrade the global Cabal library entirely may be the right approach.

Pinging @lwm @borsboom

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -982,14 +982,21 @@ withSingleContext runInBase ActionContext {..} ExecuteEnv {..} task@Task {..} md
                         -- This branch is taken when
                         -- 'explicit-setup-deps' is requested in your
                         -- stack.yaml file.
-                        (Nothing, Just deps) | explicitSetupDeps (packageName package) config ->
+                        (Nothing, Just deps) | explicitSetupDeps (packageName package) config -> do
+                            $logWarn $ T.pack $ concat
+                                [ "Package "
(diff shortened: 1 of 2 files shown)


lwm:
> all of these build issues will disappear (in theory) with properly provided custom-setup, so we should be encouraging them.

Ah right. I wasn't aware.

Thanks for the explanation.

> I'd go so far as to say that, in the not-too-distant future, removing the ability to upgrade the global Cabal library entirely may be the right approach.

Have you seen #2995? Please do weigh in there if you think it's not something useful.

borsboom:
Sounds good

Issue #3035, opened 2017-03-01
Unnecessary rebuilding of dependencies


sjakobi:
I'm seeing this strange behavior at https://github.com/sjakobi/store/commit/1a4c632ebe156320035a122443e56224a2f9047a. Every time I run a benchmark, some dependencies are unregistered and rebuilt, even when there is _no_ change on any file that stack should care about.

The most interesting line from the verbose logs appears to be this one:

```
2017-03-01 15:59:36.313604: [debug] Ignoring package distributive, from (InstalledTo Local,"/home/simon/src/store/.stack-work/install/x86_64-linux/nightly-2017-02-28/8.0.2/pkgdb/"), due to wrong location: (Just (InstalledTo Local),Snap)
```

(It would be nice if the log said what the expected location was.)

`distributive` and a few (related?) packages are rebuilt again and again and apparently registered in the wrong package db. I think this might be related to containers being a local `extra-dep` in [this configuration](https://github.com/sjakobi/store/blob/1a4c632ebe156320035a122443e56224a2f9047a/stack.yaml#L8).

I haven't properly minimized this yet, just wanted to report it in case someone else could reproduce it.

Full logs – with and without `-v` – [here](https://github.com/commercialhaskell/stack/files/811138/bench.txt).

```
$ stack --version
Version 1.3.3, Git revision 8b0ab0bf687a225b2d724f743db50a20ddcc303c (dirty) (4613 commits) x86_64 hpack-0.17.0
```

mgsloan:
Whew, this was tricky to track down.

Minimal repro is to create a new project with the default template, add distributive to build-depends and have the following in your stack.yaml:

```yaml
packages:
- '.'
- location:
    git: https://github.com/sjakobi/containers
    commit: 9612b20
  extra-dep: true
# - location:
#     git: https://github.com/ekmett/tagged
#     commit: 298215b
#   extra-dep: true
```

Uncommenting the `tagged` repo causes things to work properly.

I figured out that the issue is that `loadSourceMapFull`, for `distributive`, is yielding `PSUpstream` with `Snap`.  Later during the build process it figures out that it needs to be `Local`, due to it depending on things in the local db, causing reinstallation.

Here's why this is happening:

* `stack dot --external` shows that `distributive` has a direct dependency on Cabal, despite it not being in build-depends.  `Cabal` depends on `containers`, and so when a local version of `containers` is used, the `Cabal` version must also be local.

[...]

PR #3039, opened 2017-03-03
Fix a case of unnecessary dependency rebuilding #3035

mgsloan:
See https://github.com/commercialhaskell/stack/issues/3035

I think it makes sense to include this in the release candidate, since it's the first version with custom-setup.

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -515,11 +515,11 @@ getBuildComponentDir (Just name) = parseRelDir (name FilePath.</> (name ++ "-tmp
 
 -- | Get all dependencies of the package (buildable targets only).
 packageDependencies :: PackageDescription -> Map PackageName VersionRange
-packageDependencies =
-  M.fromListWith intersectVersionRanges .
-  concatMap (fmap (depName &&& depRange) .
-             targetBuildDepends) .
-  allBuildInfo'
+packageDependencies pkg =
+  M.fromListWith intersectVersionRanges $
+  map (depName &&& depRange) $
+  concatMap targetBuildDepends (allBuildInfo' pkg) ++
+  maybe [] setupDepends (setupBuildInfo pkg)
 
 -- | Get all build tool dependencies of the package (buildable targets only).
 packageToolDependencies :: PackageDescription -> Map Text VersionRange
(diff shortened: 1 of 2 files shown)


sjakobi:
The change makes sense to me but it doesn't appear to fix original issue. Even when I delete `~/.stack/snapshots` and `stack clean --full` I still see the behavior described in #3035.

borsboom:
Including in rc/v1.4.0 makes sense to me, so feel free to merge when ready.

mgsloan:
@sjakobi Ah, good catch!  Issue was that this computed info gets stored in a cache.  Normally, we'd just bump the version tags.

However, when investigating this I noticed that the errors for peeking when both the tag and format have changed are rather poor.  Decode errors in the contents took priority over tag matching.  I've released store-0.4.1, which handles this.

There was also an encoding change with Map for this store release.  If we delayed on updating store version, then this would cause further cache invalidation down the road.  Best to avoid this when feasible!

@borsboom I think it's good to go, seems to work just fine.  Since I've just made this change quite a bit larger, it might be worth further review.  Feel free to merge if it looks alright

sjakobi:
OOM while building store with GHC-7.10.3: https://travis-ci.org/commercialhaskell/stack/jobs/207616671#L327

Not sure if this needs further investigation in `store`.

Here, we should probably switch to using the [sudo-enabled Travis environment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) that has 7.5 GB memory instead of 4.

sjakobi:
Looks like the other build failures were caused because they weren't aware of store-0.4.1 yet.

sjakobi:
Works on my machine though! Nice work, @mgsloan! :+1:

borsboom:
We should make sure the CI works before merging.

borsboom:
@mgsloan: is the `store` upgrade actually necessary to fix the bug?  I'm hesitant to do that for v1.4.0.

mgsloan:
@borsboom Not strictly necessary.  But if we don't do it now, then the cache for MiniBuildPlan will be invalidated for both this release and next release.  It's straightforward to do it that way instead.

mgsloan:
I've gone with the more conservative change.  Merging!

borsboom:
Tests passed, so I merged it.

PR #3042, opened 2017-03-04
Fix location of global config file

massysett:
Apparently the location of the global `stack.yaml` has changed.


borsboom:
Thanks!

Issue #2984, opened 2017-02-05
stack build/test/bench --profile should recompile if previous compile was without --profile


janpath:
If you build without `--profile` and then with `stack` doesn't recompile the second time to create a binary with profiling options if no files have changed. This is in opposition to `ghc` behaviour. If you build without `-prof` and then with, `ghc` will recompile the second time even if no files changed.

### Steps to reproduce

For example:

1. Run command `stack clean && stack build`.
2. Run command `stack build --profile`.

### Expected

Stack builds two times, first without profiling options and then another time with.

### Actual

Stack builds only once without profiling. It ignores the second build instruction because no files changed.

### Stack version

```
$ stack --version
Version 1.3.2 x86_64
```

### Method of installation

* ArchLinux package `community/stack`.

mgsloan:
Hmm, I can't reproduce the issue:

```
mgsloan@computer:~/fpco/test-stack$ stack new issue2984
...
mgsloan@computer:~/fpco/test-stack$ cd issue2984
mgsloan@computer:~/fpco/test-stack/issue2984$ stack build --system-ghc
Warning: File listed in issue2984.cabal file does not exist: README.md
[1 of 2] Compiling Main             ( /home/mgsloan/.stack/setup-exe-src/setup-mPHDZzAJ.hs, /home/mgsloan/.stack/setup-exe-src/setup-mPHDZzAJ.o )
[2 of 2] Compiling StackSetupShim   ( /home/mgsloan/.stack/setup-exe-src/setup-shim-mPHDZzAJ.hs, /home/mgsloan/.stack/setup-exe-src/setup-shim-mPHDZzAJ.o )
Linking /home/mgsloan/.stack/setup-exe-cache/x86_64-linux/tmp-Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2 ...
issue2984-0.1.0.0: configure (lib + exe)
Configuring issue2984-0.1.0.0...
issue2984-0.1.0.0: build (lib + exe)
Preprocessing library issue2984-0.1.0.0...
[1 of 1] Compiling Lib              ( src/Lib.hs, .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/Lib.o )
Preprocessing executable 'issue2984-exe' for issue2984-0.1.0.0...
[...]

PR #3046, opened 2017-03-07
Have switching on/off profiling cause package dirtiness #2984

mgsloan:
This seems like the simplest way to fix https://github.com/commercialhaskell/stack/issues/2984 .

I suspect it may cause unnecessary rebuilds, as there's likely a reason for these to be present in the list.  Thoughts, @snoyberg ?  The alternative seems complicated / error prone - https://github.com/commercialhaskell/stack/issues/2984#issuecomment-284611410


mgsloan:
In my limited testing, it seems like this doesn't cause excessive rebuilding.  Merging!

PR #3054, opened 2017-03-10
Add optional shorter form for `--test-arguments`.

lwm:

I could just be lazy but it is handy when fooling around with tasty framework args.

Looks like:

```
  --ta,--test-arguments TEST_ARGS
                           Arguments passed in to the test suite program
```


mgsloan:
I can certainly see the convenience of such aliases.  How about also having `--ba` for `--benchmark-arguments`?

lwm:
> How about also having --ba for --benchmark-arguments?

Rebased and done!

mgsloan:
Cool, thanks!

Issue #2884, opened 2016-12-30
Stack exec incorrectly setting no buffering


houli:
### General summary/comments (optional)
A [recent commit](https://github.com/commercialhaskell/stack/commit/6c876293a4ca215c2645de162aa0314f415fb754) to stack has made it so that line buffering is turned off. This results in programs that are run by stack exec printing out control characters such as backspace instead of having the appropriate action on stdin. This can be seen clearly with a simple example such as `cat` vs `stack exec cat`. I first noticed this when using a program that made use of the standard `getLine` Haskell function.
### Steps to reproduce

1. Run `stack exec cat`
2. Type some input
3. Attempt to use backspace to delete it
4. Compare results to just running plain `cat`

### Expected

Backspaces and other control characters have their desired effect on the input.

### Actual

Backspaces and other control characters are printed to the terminal.

### Stack version

```
$ stack --version
Version 1.3.2 x86_64 hpack-0.15.0
```

### Method of installation
Homebrew

lwm:
Just triaging here, the relevant issue for the commit was https://github.com/commercialhaskell/stack/issues/2702.

geekosaur:
Note that this is an unintended side effect; some ghc versions back, ghc's Unix runtime started conflating NoBuffering with `stty -icanon`, which is convenient for users who don't know about termios but has this  side effect if you want to use NoBuffering for some other reason.

koterpillar:
Another consequence of this is `Ctrl-D` to signal EOF doesn't work either.

mgsloan:
Interesting!  Is there a way to resolve this while addressing the concern in #2702?

> ghc's Unix runtime started conflating NoBuffering with stty -icanon

@geekosaur Do you know of a way to undo this conflation?

geekosaur:
I don't know of one. You probably need to raise this with the ghc devs.

andersk:
This affects not only programs run inside `stack exec`, but also programs run _after_ `stack exec`!

```console
$ stty
speed 38400 baud; line = 0;
-brkint -imaxbel iutf8
$ stack exec /bin/true
$ stty
speed 38400 baud; line = 0;
min = 1; time = 0;
-brkint -imaxbel iutf8
-icanon
```

geekosaur:
I think that's a missed edge case in ghc: termios gets reset if you switch from `NoBuffering` to `BlockBuffering`, but not if you switch to `LineBuffering`. (Found in `#haskell` the other day)

[...]

Issue #2702, opened 2016-10-11
Respect executable NoBuffering


crockeea:
I explicitly adjust `hSetBuffering` in my benchmarks so that I can see incremental output. `stack bench` seems to ignore this and only produces output when a newline occurs.

foo.cabal

```
name:                foo
version:             0.0.0.1
build-type:          Simple
cabal-version:       >= 1.10

library
  default-language:   Haskell2010
  build-depends:
      base

  exposed-modules: Lib

Benchmark bench-foo
  type:               exitcode-stdio-1.0
  default-language:   Haskell2010
  main-is:            Main.hs
  hs-source-dirs:     benchmarks

  build-depends:
    base
```

Lib.hs

```
module Lib where
```

Main.hs

```
import Control.Concurrent
import System.IO

main :: IO ()
main = do
  hSetBuffering stdout NoBuffering
  putStrLn "Hello world"
  putStr "Going to sleep..."
  threadDelay 2000000
  putStrLn "awake again!"
```

I'm using the 7.0 resolver (i.e. GHC 8.0.1).

To reproduce,

```
> stack bench foo
```

This prints "Hello world", then waits 2 seconds, then prints "Going to sleep...awake again!" and exits.
By comparison,

```
> ./.stack-work/dist/*/Cabal-*/build/bench-foo/bench-foo
```

prints "Hello world" and immediately prints "Going to sleep..." on a new line. _Then_ it waits 2 seconds and prints "awake again" on the same line, and then exits.

```
> stack --version
Version 1.2.0 x86_64 hpack-0.14.0
> cabal --version
cabal-install version 1.24.0.0
compiled using version 1.24.0.0 of the Cabal library
```

I explicitly set `hSetBuffering stdout NoBuffering` so that I could see output without a new line. I've also tried setting `stderr` to `NoBuffering`, but that didn't help either.

mgsloan:
I've added some code to disable buffering before invoking the executable.  Does using the latest git version of stack resolve the problem for you?  Please comment and re-open if this doesn't work.

andersk:
This fix is wrong. Experimentally, it does not fix the problem reported here (I reproduced it in stack 1.4.0 e714f1dd3fade19496d91bd6a017e435a96a6bcd using the reporter’s instructions). Worse, it also created the regression in #2884, where the termios state is corrupted for processes running inside `stack exec` and after `stack exec`. Please revert 6c876293a4ca215c2645de162aa0314f415fb754.

[...]

PR #3086, opened 2017-03-23
Revert "Set NoBuffering mode io, before exec #2702"

andersk:
This reverts commit 6c876293a4ca215c2645de162aa0314f415fb754.

Running `hSetBuffering` just before exec had no effect on any actual buffering, but it screwed up the persistent terminal state in such a way that the runtime fails to reset it.  The effect was that keys like Backspace, Ctrl-C, Ctrl-D, etc. were not working correctly in programs launched inside `stack exec` and after `stack exec`.

Closes: #2884.

Reopens: #2702 (actually, that was never fixed in the first place).

> Please include the following checklist in your PR:

Tested using the example in #2702, and by running `stty sane; stty; stack exec stty; stty` (which should show the same state three times).

### DIFF
--- a/src/Stack/Exec.hs
+++ b/src/Stack/Exec.hs
@@ -21,7 +21,6 @@ import           System.Process.Log
 import           Control.Exception.Lifted
 import           Data.Streaming.Process (ProcessExitedUnsuccessfully(..))
 import           System.Exit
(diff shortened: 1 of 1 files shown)


mgsloan:
Thanks for looking into this!

PR #3102, opened 2017-03-31
Actually use haddock-arguments

Fuuzetsu:
Until now, the only place I can tell that these arguments are being used
is when generating index.

* Any changes that could be relevant to users have been recorded in the ChangeLog.md
Done
* The documentation has been updated, if necessary.
Do we need it?

`stack --nix test --flag stack:integration-tests stack:stack-integration-test --test-arguments "-m haddock-options"`


mgsloan:
Thanks!

Issue #2369, opened 2016-07-12
Add options to override locations of programs called from stack


sjakobi:
(This issue is split off from https://github.com/commercialhaskell/stack/issues/593.)

It should be possible to specify the locations of the programs that stack calls, like `ldd` or `strip`.

The location of `gcc` can be specified via `--with-gcc` since https://github.com/commercialhaskell/stack/pull/2264.

mgsloan:
Good idea!  Should also share code with https://github.com/commercialhaskell/stack/issues/652

nh2:
This would allow stack users to link with the `lld` linker, see https://www.reddit.com/r/haskell/comments/63y43y/liked_linking_3x_faster_with_gold_link_10x_faster/

kgadek:
I published #3118 as a quick solution to overriding `ld` and `strip`.

Also, my cabal (`1.24.0.2`) reports:
```
The flags --with-PROG and --PROG-option(s) can be used with the following programs:
  alex ar c2hs cpphs gcc ghc ghc-pkg ghcjs ghcjs-pkg greencard haddock happy
  haskell-suite haskell-suite-pkg hmake hpc hsc2hs hscolour jhc ld lhc lhc-pkg
  pkg-config strip tar uhc
```
Quite a few. The complete solution should cover them all.

I've started playing around more comprehensive code.

dukedave:
Hi, sorry if this is unrelated:
Could this (and #593) be and underlying reason for the issue mentioned in [this comment](https://github.com/LumiGuide/haskell-opencv/issues/104#issuecomment-358023847)?

mgsloan:
I've commented on that issue, I don't think it's related

snoyberg:
Would #1438 (adding a `configure-options`) fully subsume this? At least for the examples given, I don't believe Stack calls these executables directly.

snoyberg:
I believe the new `configure-options` feature on master will address this, closing. If someone has a use case which isn't addressed, please add some clarification.

Issue #652, opened 2015-07-21
Consistent '--*-options' options


manny-fp:
Currently there are a few different conventions in use for stack options that specify options to pass to external programs.  These should be made consistent, preferably sharing an optparse-applicative between them.  I'm thinking they should all allow two different ways:

`--*-option`: pass a single argument, which may contain spaces.
`--*-options`: pass multiple space-separated arguments (parsed with [argsParser](https://github.com/commercialhaskell/stack/blob/e328d63042d5472a25741c6ad988f461d826e036/src/Data/Attoparsec/Args.hs#L38)

It should be possible to pass both of these multiple times, in which case the options are concatenated in the order given.

In some cases, we also specify options in YAML configuration files.  The configuration field should be named `*-options`, and should contain a YAML sequence.  Sequences are concatenated when cascading configuration files (and any additional options from the command-line are concatenated after those from the configuration file).

(motivated by #601)

DanBurton:
+1 I generally agree with everything proposed here.

On Tuesday, July 21, 2015, Emanuel Borsboom <notifications@github.com>
wrote:

> Currently there are a few different conventions in use for stack options
> that specify options to pass to external programs. These should be made
> consistent, preferably sharing an optparse-applicative between them. I'm
> thinking they should all allow two different ways:
>
> --*-option: pass a single argument, which may contain spaces.
> --*-options: pass multiple space-separated arguments (parsed with
> argsParser
> <https://github.com/commercialhaskell/stack/blob/e328d63042d5472a25741c6ad988f461d826e036/src/Data/Attoparsec/Args.hs#L38>
>
> It should be possible to pass both of these multiple times, in which case
> the options are concatenated in the order given.
>
> In some cases, we also specify options in YAML configuration files. The
> configuration field should be named *-options, and should contain a YAML
> sequence. Sequences are concatenated when cascading configuration files
> (and any additional options from the command-line are concatenated after
> those from the configuration file).
>
> (motivated by #601 <https://github.com/commercialhaskell/stack/issues/601>
> )
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/commercialhaskell/stack/issues/652>.
>


-- 
-- Dan Burton

mgsloan:
RyanGIScott just mentioned in #haskell-stack that he'd like a way to pass options to `ghc-pkg`, so we should be sure this covers that use case. It's to work around this issue https://github.com/gtk2hs/gtk2hs/issues/139

[...]

PR #3118, opened 2017-04-10
Add a config option to override path and add options to each PROG

kgadek:
This allows one to specify with-ld: and with-strip: in stack.yaml. Ref #2369

Ripoff #2264 from @khanage, includes subsequent changes in 113c6375.

While it ignores the existence of #652 and only covers `ld` and `strip`, I would propose to include this anyway, as it's consistent with existing code.

Main motivation is the https://www.reddit.com/r/haskell/comments/63y43y/liked_linking_3x_faster_with_gold_link_10x_faster/ post — this patch should allow people to compare the compilation times using clang.

Checklist:

* [*] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [*] The documentation has been updated, if necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/commercialhaskell/stack/3118)


kgadek:
The following commit is a more comprehensive solution. A lot of similar code there, but does the job.

mgsloan:
Wow, thanks for working on this!

1) Beyond supplying these arguments to configure, we will also want to use these in stack's invocations of ghc / ghcjs (building Setup.hs, ghci), ghc-pkg, ghcjs-pkg, and haddock.  I think that covers it, but there may be more.

2) Might also consider leaving out the uhc, lhc and jhc stuff as stack doesn't support them.

3) These are going to really clutter up the help documentation.  I think it'd be best to hide them by default, and have something similar to cabal's doc

```
    --with-PROG=PATH                 give the path to PROG
    --PROG-option=OPT                give an extra option to PROG (no need to
                                     quote options containing spaces)
    --PROG-options=OPTS              give extra options to PROG
```

Unfortunately this is not so straightforward with optparse-applicative, but once https://github.com/pcapriotti/optparse-applicative/pull/253 is merged, something like this should be possible.  I don't think it makes sense to delay this PR on that point.

It'd also be cool to avoid adding so many fields to the configuration type, possibly by having a datatype for programs and having `Map Program [Text]` and similar?  Not really required, can be deferred to a later change.

ChangeLog.md
```diff
@@ -37,6 +37,8 @@ Other enhancements:
 * `--[no-]haddock-hyperlink-source` flag added which allows toggling
   of sources being included in Haddock output.
   See [#3099](https://github.com/commercialhaskell/stack/issues/3099)
+* Add the ability to explictly specify a ld and strip executable.
+  [#2369](https://github.com/commercialhaskell/stack/issues/2369)
```
kgadek: self: update this to match 2ᴺᴰ commit.

doc/yaml_configuration.md
```diff
@@ -392,12 +392,14 @@ extra-lib-dirs:
 - /opt/foo/lib
 ```
 
-### with-gcc
+### with-gcc/with-ld/with-strip
```
kgadek: self: update this to match 2ᴺᴰ commit.

src/Stack/Config.hs
```diff
@@ -298,7 +298,58 @@ configFromConfigMonoid
...
+         configProgHsc2hsPath          = getFirst configMonoidProgHsc2hsPath
+         configProgHscolourPath        = getFirst configMonoidProgHscolourPath
+         configProgJhcPath             = getFirst configMonoidProgJhcPath
+         configProgLdPath              = getFirst configMonoidProgLdPath
+         configProgLhcPath             = getFirst configMonoidProgLhcPath
+         configProgLhcPkgPath          = getFirst configMonoidProgLhcPkgPath
+         configProgPkgConfigPath       = getFirst configMonoidProgPkgConfigPath
+         configProgStripPath           = getFirst configMonoidProgStripPath
+         configProgTarPath             = getFirst configMonoidProgTarPath
+         configProgUhcPath             = getFirst configMonoidProgUhcPath
```
kgadek: self: drop this.

kgadek:
1. oh right. I'll ripgrep for usages
1. makes sense. I'll also make a note about it in the docs
1. 100% agree, I didn't know what to propose as the proper solution. I can very well add dummies `--with-PROG` & `--PROG-option`, and hide all of the new options… though obviously it's a hack. For the record, while the pcapriotti/optparse-applicative#253 is merged, it's not a part of the release (at very least `lts-8.5 --> optparse-applicative 0.13.2.0` doesn't have that). I think we could go the hacky-way and after merge make a ticket to _fix_ that ugliness once pcapriotti/optparse-applicative#253 is available.

The `Map Program [Text]` approach is okay. I was thinking about making a separate data type for this, but `Map` works fine.

As a related note, `src/Stack/Options/ConfigParser.hs`:`configOptsParser` is on the edge of getting obscure bugs (e.g. flip any of the `*Path` params) and being just unmaintainable. I don't know `optparse-applicative` enough to propose any solution to that. Ideas?

kgadek:
1. I don't know how to properly trace all places where I'd need to put the config options. Could you help me with this?
1. Hmm, is there any other place in documentation that I shall update?
1. Done (the hacky way). Shall I made some additional behaviour to exit when someone calls this dummy option (`stack --with-PROG foo`) by accident?
1. Used maps. I'm not sure if I like `Map ProgName (First (Path Abs File))` though. WDYT?
1. Rebased & cleaned up commit history a bit.

lwm:
Oh damn, it's a real pity to see this rotting :bomb: 

> Hmm, is there any other place in documentation that I shall update?

How about I take a stab at updating that, while you cover the rest.

> Shall I made some additional behaviour to exit when someone calls this dummy option (stack --with-PROG foo) by accident?

Probably a good idea to include some nice message for this case.


> 1. I don't know how to properly trace all places where I'd need to put the config options. Could you help me with this?
> 4. Used maps. I'm not sure if I like Map ProgName (First (Path Abs File)) though. WDYT?

@mgsloan, any quick ones to get @kgadek on the right direction?

mgsloan:
Sorry!!  I don't have any good explanation of why I hadn't answered the earlier inquiries, other than misplacing browser tabs and having lots of other stuff to contend with.

One reason for this is that I am not so keen on having so much boilerplate.  Maybe this is what Cabal does, but I would really prefer something that either:

1. Has a list of programs at runtime, generates all the options etc from that, doesn't have separate fields.

2. Alternatively, uses TH to generate the code.

Maybe that's overkill, but it'd be nice if adding a program to the list could be a modification in one or two spots.

> I don't know how to properly trace all places where I'd need to put the config options. Could you help me with this?

Anything that invokes processes.  Search for `Process`.  In particular, the stuff in `System.Process.*` modules.

This is part of the idea with making this more of a runtime thing rather than having separate code paths for each overidable command.  The code in System.Process.* would check if the location of a particular command name has been overridden, and instead use that.  To ensure that this matches, up, can have a constant like `ghcProgramName = "ghc"` that is used in both spots.

> Done (the hacky way). Shall I made some additional behaviour to exit when someone calls this dummy option (stack --with-PROG foo) by accident?

Seems reasonable to check that.

> Used maps. I'm not sure if I like Map ProgName (First (Path Abs File)) though. WDYT?

Seems alright.  I hope there is a good way to do (String -> Maybe ProgName).  If there isn't, then there won't be a convenient way to do the Process code.  Other than perhaps listing what is accepted for the with options, I feel like cabal's ProgName sum type is overkill.  Might be nice to use it so that exhaustiveness checks tell us when it's been updated.  However, I think `newtype ProgName = ProgName String` would be just as good and a lot less boilerplate.

Anyway, apologies again @kgadek your good intentions here are appreciated.  I hope this hasn't put you off doing other contributions.  Could still rebase and implement the stuff described above.

mgsloan:
Ah, perhaps something like `ProgName` is good.  Could have it actually be directly provided to the System.Read.* functions and other process functions.  Or, perhaps a layer atop.

The idea is that if TH is used to generate the options, then adding a constructor to a `ProgName` datatype will automatically cause there to be a `--with-PROG` option for that new program.

snoyberg:
It looks like this has stalled, and has begun accumulating merge conflicts. I'm going to close out for now. @kgadek if you're still interested in pushing forward with this, please drop a comment and we can figure out how to get this to continue.

Issue #3006, opened 2017-02-15
Speed up Store instances for (Byte)Strings of known size


sjakobi:
`Stack.Types.BuildPlan.GitSHA1` currently derives its `Store` instance from the underlying ByteString, so store must assume a variable size. A handwritten instance could use the fact that a SHA1 digest has 20 bytes and would probably be a bit faster, and – depending on architecture – 4 or 8 bytes shorter.

The same could be done for `Stack.Types.PackageIndex.PackageDownload.pdSHA256` and probably other types.

If there isn't already a set of helpers for ByteStrings of known size in `store`, maybe it should be added! :)

sjakobi:
> If there isn't already a set of helpers for ByteStrings of known size in store, maybe it should be added! :)

See [`StaticSize`](https://hackage.haskell.org/package/store-0.3.1/docs/Data-Store-Internal.html#t:StaticSize).

mgsloan:
Yep, I think `StaticSize` is our answer to this, closing

sjakobi:
Reopening, as stack's Store instances still need to make use of `StaticSize`, right?!

mgsloan:
Oh, true!  I thought this was on the store tracker.

alexeyzab:
Hi there!

I'd like to try implementing this. I'll be sending a PR shortly, let me know if I am on the right track. Thanks!

alexeyzab:
I believe this can be closed.

snoyberg:
Sounds good, thanks for catching that!

PR #3120, opened 2017-04-12
Start implementing #3006

alexeyzab:

This changes `GitSHA1` from using a `ByteString` to a `StaticSize 40 ByteString`. Other edits make sure the code compiles and the tests pass.

It was not possible to derive a `Hashable` instance after the change, so I've implemented one by hand.

I have commented out the implementation of `peek` and `poke` for `GitSHA1`'s `Store` instance as I am not sure if those are needed or not.

If these changes are satisfactory, I'll start working on the `Stack.Types.PackageIndex.PackageDownload.pdSHA256`'s instances as well.


alexeyzab:
I went ahead and made the changes to `Stack.Types.PackageIndex.PackageDownload.pdSHA256`. If this needs more work, let me know.

mgsloan:
Hey!  Sorry for the delay on feedback.  Looks good to me!  One thing is that the version tags, like (`storeVersionConfig "pkg-v2" ...)` should have its name updated (so, `pkg-v2`), since this will actually cause a difference in serialization.  With the StaticSize version, the size won't get stored in the file at all.

alexeyzab:
@mgsloan To clarify, should it be changed `pkg-v3`? I've adjusted the hash, but didn't touch this part.

On a separate note, should I remove the commented out definitions of [`poke` and `peek` for `GitSHA1`](https://github.com/commercialhaskell/stack/pull/3120/files#diff-c915b6fa64991afe9ae33739482900ffR470)? I put them in just in case they might be necessary here.

Thanks!

snoyberg:
I totally broke this PR with extensible-snapshots, which switched from GitSHA1 to a SHA256. I'm going to this it up and open a new PR, I'll add a comment when that's done.

snoyberg:
PR #3260 is now open. When it's merged, it will merge this PR too.

snoyberg:
Continuing the discussion from the other PR: here's a really hacky implementation of the `Word64` idea I mentioned: https://github.com/snoyberg/static-bytes

alexeyzab:
That's pretty interesting. If we were to use that library, we'd change `StaticSHA256` to something like 

```haskell
data StaticSHA256 = StaticSHA256 !Bytes32
```

Is that right? And then get rid of the `StaticSize` altogether.

snoyberg:
Yes, though I'd prefer a `newtype` wrapper in this case. But I'd recommend considering that separate from this PR to avoid slowing it down.

alexeyzab:
Makes sense. In that case it seems all that's left here is to edit the `ChangeLog.md`, I'll do that now.

alexeyzab:
@snoyberg Thanks for walking me through what needed to be done here!

snoyberg:
No problem, thank you! It looks like the PR is showing all of my changes on extensible-snapshots as differences, which is a bit annoying... anyway, I think this is good to go. @mgsloan, did you have any objections to merge?

alexeyzab:
I wonder if that's because I merged something the wrong way.

I believe these are the relevant commits for this PR:
https://github.com/commercialhaskell/stack/pull/3120/commits/14533459f7cef7e9aa43a7e30f87870383a731ed
https://github.com/commercialhaskell/stack/pull/3260/commits/8c87d12b963956aeeb8b9f8802cc9eeefcf0212b
https://github.com/commercialhaskell/stack/pull/3260/commits/696d9b4e61617c59e017e5cc922c79bef08df0b1
https://github.com/commercialhaskell/stack/pull/3120/commits/54ff6326c40718395da8bd0deb72d68bdee7eac9

Where the last two overwrite some parts of the first two.

snoyberg:
Thanks!

PR #3127, opened 2017-04-14
Improve filtering of ghc messages

sergv:
I added stripping of linker warnings on ghc 7.8 on windows (x64 IIRC). I'm not sure how useful it would be in general going forwards, but it's really annoying when it occurs.

I also took liberty to improve `Stack.Build.Execute` module a bit by adding type signatures no non-obvious parts and by introducing (hopefully) descriptive enums instead of booleans.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -453,10 +453,11 @@ withExecuteEnv menv bopts boptsCli baseConfigOpts locals globalPackages snapshot
     dumpLog :: String -> (Path Abs Dir, Path Abs File) -> m ()
     dumpLog msgSuffix (pkgDir, filepath) = do
         $logInfo $ T.pack $ concat ["\n--  Dumping log file", msgSuffix, ": ", toFilePath filepath, "\n"]
+        compilerVer <- view actualCompilerVersionL
         runResourceT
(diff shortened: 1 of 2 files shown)


mgsloan:
@sergv Thanks!  Are these warnings sure to be spurious?  Perhaps the filtering should only be done when we know we're using ghc 7.8?

sergv:
@mgsloan The warnings are only visible on windows x64 and are definitely uninteresting as they don't affect the executables produced. I agree that it's good idea to only filter them for ghc 7.8. I'll prepare a followup patch in this PR soon.

sergv:
@mgsloan I've rebased the branch and ensured that filtering happens only for ghcs below 7.10.

mgsloan:
@sergv Sorry for taking so long to get this merged!  Will try to be more on it in the future.

sergv:
@mgsloan Thanks for merging it anyway! It's reassuring to know than PR will be merged at some point, even if not so quickly.

lwm:
> I also took liberty to improve Stack.Build.Execute module a bit by adding type signatures no non-obvious parts and by introducing (hopefully) descriptive enums instead of booleans.

:+1: :+1: :+1:

PR #3135, opened 2017-04-24
stack uninstall: Update the static help message

Tarrasch:
It seems like --local-bin-path is deprecated in favour of --local-bin.

----

I've not compiled this change, but I did run the two commands and indeed there's no deprecation warning printed for the latter.

```
❯ stack path --local-bin-path

'--local-bin-path' will be removed in a future release.
Please use '--local-bin' instead.


/home/arash/.local/bin
```
```
❯ stack path --local-bin     
/home/arash/.local/bin
```


Blaisorblade:
Thanks!

Issue #1659, opened 2016-01-15
Allow stack test to skip certain test suites


dmcclean:
It would be great to have a flag like `stack test --skip long-tests` that would skip the test suite named `long-tests` in the cabal file.

Duration isn't the only reason for wanting something like this. Some tests are also difficult to get running on older LTS versions (because they are on old versions of various libraries that are used by the tests but not the library under test). Some tests are difficult to get running on the CI server because they require external resources like a database.

For symmetry it would be good if it was also accepted by `stack bench`. I'm not sure if cabal allows a benchmark and a test suite from the same project to have overlapping names, but if it does then there might need to be two flags instead, `--skip-tests` and `--skip-benchmarks`.

borsboom:
Agreed, that would be handy.  A pull request adding this feature would be welcome.

Note: It's possible to achieve something like this using flags that enable or disable tests in the .cabal file (see Stack's own stack.cabal for an example with the `integration-tests` flag).

mgsloan:
Adding `--skip COMPONENT` to `build` options makes sense to me, and probably everywhere else that takes per-component targets.  In other words, the feature can be part of determining targets, not specific to tests / benchmarks.

snoyberg:
This is implemented via #3148, closing.

PR #3148, opened 2017-04-30
Allow skipping components when building

martin-kolinek:
This is a naive attempt at implementing #1659. It adds `--skip-component` option which can be specified multiple times. Any component specified using such option is then skipped.

The filtering happens in `loadLocalPackage`, the same place which decides whether test and bench components will be built if the user hasn't specified targets.

The issue mentions making this part of the target resolution. I tried to look at that, but if I understood the code correctly, that would probably require either making `StLocalAll` `SimpleTarget` also contain components to skip, or removing `StLocalAll` and finding out which components to build earlier. I'm not sure if something like that is wanted.

Let me know what you think.


src/Stack/Build/Source.hs
```diff
@@ -405,6 +405,12 @@ loadLocalPackage boptsCli targets (name, (lpv, gpkg)) = do
                 Just STUnknown -> assert False mempty
                 Nothing -> mempty
 
+        filterSkippedComponents = Set.filter (not . (`elem` boptsSkipComponents bopts))
+
+        (exes, tests, benches) = (filterSkippedComponents exeCandidates,
```
lwm: You said in the documentation that executables cannot be skipped?

src/Stack/Options/BuildMonoidParser.hs
```diff
@@ -148,3 +149,10 @@ buildOptsMonoidParser hide0 =
             "split-objs"
             ("Enable split-objs, to reduce output size (at the cost of build time). " ++ splitObjsWarning)
             hide
+    skipComponents = many
+        (fmap
+            T.pack
+            (strOption
+                (long "skip-component" <>
```
lwm: `--skip` is nice and short right? Do we really need `component`?

src/Stack/Build/Source.hs
martin-kolinek: Yes, but that is because of the code in `Stack.Build.Execute.exesToBuild`. I added a comment with details. I think it's better if this part behaves consistently for all types of components.

martin-kolinek:
@lwm Thank you for the feedback. I tried to fix the issues.

src/Stack/Build/Source.hs
```diff
@@ -405,6 +405,14 @@ loadLocalPackage boptsCli targets (name, (lpv, gpkg)) = do
                 Just STUnknown -> assert False mempty
                 Nothing -> mempty
 
+        filterSkippedComponents = Set.filter (not . (`elem` boptsSkipComponents bopts))
+
+        -- the part for executables currently does nothing as Stack.Build.Execute.exesToBuild
+        -- ignores lpComponents, but once that is changed it should also make this work with executables
+        (exes, tests, benches) = (filterSkippedComponents exeCandidates,
```
lwm: You could  point this to the relevant issue (is there one? Should we make one?) for the future? 

Otherwise, looks good!

lwm:
> @lwm Thank you for the feedback. I tried to fix the issues.

Happy to help and thanks for coming back to this. Hopefully one of the core maintainers will pass by here soon and this will get into master :beers:

src/Stack/Build/Source.hs
martin-kolinek: I added a link to https://github.com/haskell/cabal/issues/2780 as that's the issue which needs to be fixed to make skipping of executables work.

snoyberg:
Thank you for the contribution.

With the merge of #3237, which works around the cabal limitation around executable component specification, is this still relevant?

martin-kolinek:
@snoyberg I updated the documentation, tests and comments in the code to make it accurate after the latest changes. Skipping now works for executables if they were built at least once already.

snoyberg:
Thank you!

PR #3159, opened 2017-05-04
Fixed pid1 ownership inside fpco/stack-build docker image

lehins:
Here is the issue with the file ownership that this PR fixes.
```bash
$ docker run --rm fpco/stack-build ls -laF /usr/local/sbin
total 372
drwxr-xr-x  2  501 staff   4096 Nov 23 06:15 ./
drwxr-xr-x 16 root root    4096 Feb 12 23:38 ../
-rwxr-xr-x  1  501 staff 371956 Nov 23 06:15 pid1*
```

### DIFF
--- a/etc/dockerfiles/stack-build/lts-8.0/Dockerfile
+++ b/etc/dockerfiles/stack-build/lts-8.0/Dockerfile
@@ -43,6 +43,7 @@ RUN stack --system-ghc --resolver=$LTS_SLUG --local-bin-path=/usr/local/bin inst
 #
 
 RUN wget -O- "https://github.com/fpco/pid1/releases/download/pid1%2F$PID1_VERSION/pid1-$PID1_VERSION-linux-x86_64.tar.gz" |tar xzf - -C /usr/local
+RUN chown -R root:root /usr/local/sbin
 
 #
 # Set up pid1 entrypoint and default command


etc/dockerfiles/stack-build/lts-8.0/Dockerfile
```diff
@@ -43,6 +43,7 @@ RUN stack --system-ghc --resolver=$LTS_SLUG --local-bin-path=/usr/local/bin inst
 #
 
 RUN wget -O- "https://github.com/fpco/pid1/releases/download/pid1%2F$PID1_VERSION/pid1-$PID1_VERSION-linux-x86_64.tar.gz" |tar xzf - -C /usr/local
+RUN chown -R root:root /usr/local/sbin
```
borsboom: Perhaps better to include this in the previous RUN command rather than create a new layer.  Also specifically change the ownership of `pid1` rather than all files in /usr/local/sbin.

```
RUN wget -O- "https://github.com/fpco/pid1/releases/download/pid1%2F$PID1_VERSION/pid1-$PID1_VERSION-linux-x86_64.tar.gz" |tar xzf - -C /usr/local && \
     chown root:root /usr/local/sbin/pid1
```

lehins: @borsboom, yeah, adding `chown` to the same layer makes total sense. Changing the ownership of `pid1` only, wouldn't be sufficient though, since `/usr/local/sbin` ownership is also set to `501:staff` (it seems to be created during the untar process). Considering that `sbin` contents is usually owned by root and especially since the folder is initially empty or nonexistent in that image, doing `chown -R` seems to be reasonable, but I can see how it can have an undesired effect in case of future customizations. In that case I can change it to:

```
RUN wget -O- "https://github.com/fpco/pid1/releases/download/pid1%2F$PID1_VERSION/pid1-$PID1_VERSION-linux-x86_64.tar.gz" | tar xzf - -C /usr/local && \
     chown root:root /usr/local/sbin && \
     chown root:root /usr/local/sbin/pid1
```
Would that be better?

borsboom: Looks good.

Issue #3147, opened 2017-04-28
New option for stack build


psibi:
To understand the context of this issue, reading this will be helpful:
https://github.com/yesodweb/yesod/issues/1384

Proposal: A new option named `--external` for `stack build` which will
be used along with the `--file-watch` option.

This feature is needed for the yesod development server specifically
to fix the above bug. When the option `--external` is given, the
`stack build --file-watch --external` won't recognize external
commands like `watching` or `help` in that mode. It will only
recognize `quit` and `rebuild` and the Return key for rebuild. On
giving some dummy text, it will just ignore the input and won't emit
anything. This feature is necessary for the `yesod-bin` package.

I would like to know what other's think of this.

kadoban:
From reading #3165 it sounds like the decision was to go a different way than this proposal (fixing the problem on the yesod side). Let me know if I misunderstood.

psibi:
Yeah, youre right. The plan is to handle it in devel code itself.

PR #3165, opened 2017-05-06
Add --external flag (Needed for yesod dev server)

psibi:

> Please also shortly describe how you tested your change. 

Tested it using the `--external` flag that has been implemented in this patch.

Fixes https://github.com/commercialhaskell/stack/issues/3147

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -49,6 +49,10 @@ Other enhancements:
   a method to ensure PVP compliance without having to proactively fix
   bounds issues for Stackage maintenance.
 * Expose a `save-hackage-creds` configuration option
+* Add a build option named `--external` which when coupled with
+  the `file-watch` flags will change the interactivity behavior of
+  the tool. This option is specifically being developed for yesod
+  development server.
 
 Bug fixes:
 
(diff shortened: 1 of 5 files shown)


psibi:
@snoyberg Can this be reviewed and merged ?

snoyberg:
I know I should have thought about this earlier in the discussion process, but I'm only doing so now, apologies. Two concerns:

1. I'm not a big fan of the name `--external`. What about `--file-watch-silent` or something like that?
2. Understanding the use case again, it now sounds like this is to work around a user error of typing in commands besides `build`. Is that accurate? If so, what about having `yesod devel` intercept the `stdin` and only passing on valid commands to Stack?

Apologies for not noting this at design phase, as my previous comments probably make clear, I wasn't fully grokking the problem initially.

psibi:
> Understanding the use case again, it now sounds like this is to work around a user error of typing in commands besides build. Is that accurate? If so, what about having yesod devel intercept the stdin and only passing on valid commands to Stack?

Yeah, that's right. I think intercepting at the `yesod devel` itself is a better idea now. :) I will see if I can patch it up in the coming days (unless you want to fix this up in the current flow).

> Apologies for not noting this at design phase, as my previous comments probably make clear, I wasn't fully grokking the problem initially.

No issues!

mgsloan:
I think it makes sense to have yesod devel do the interception, since I don't see this flag being useful for other cases.

PR #3187, opened 2017-05-24
Support building the sdist and upload tarball

martin-kolinek:
This adds the option `--test-tarball` to `stack sdist` and `stack upload`. When this option is specified, the tarball is extracted, and tests are executed as part of the process. This should implement part of #717.

The implementation works by building the extracted tarball using the default build options. The package directory in the `BuildConfig` used for building and testing is replaced by the location of the extracted tarball.

There is currently no way to specify build options.

The added tests work with packages which fail to build/test because the build depends on a file which is not packaged into the resulting tarball.


lwm:
Holy moly! I've been wanting this feature since a year now :smile: Don't have time to look at this now but hopefully someone will stop by soon. Nice one for this work!

mgsloan:
Awesome, thanks!  Implementation looks good, refactors appreciated too.

Issue #2287, opened 2016-06-19
Nix: warn when pkg-config or git not present in nix-shell


mboes:
As a followup from https://github.com/commercialhaskell/stack/issues/2089, we should be helpful to the user and suggest fixes whenever we can. In particular, it's confusing that a `stack exec` command might fail because `git`, `pkg-config` or some other tool is not present, when it's most likely installed in the user's host environment.

In particular, if `--nix` is passed an the `stack` subprocess fails to find the git command, Stack should suggest adding it to the buildInputs. Likewise for `pkg-config`. There's really not much else that Stack would need for building that isn't app specific.

cc @YPares @rrnewton

PR #3205, opened 2017-06-09
When using Nix, depend on git.

domenkozar:
Partially addresses #2287, as git is used in stack internally
to fetch package index and LTS.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,6 +19,8 @@ Behavior changes:
 
 Other enhancements:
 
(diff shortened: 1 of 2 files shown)


mgsloan:
Seems reasonable, thanks!

mgsloan:
@domenkozar @mboes Does it make sense to do the same for pkg-config?

domenkozar:
I think so, it has a very small size footprint.

mboes:
Yup.

PR #3219, opened 2017-06-17
doc: Link docker integration to docker section in guide

nh2:
This PR creates a missing link on the docker page to the page that explains how to build docker images with stack.

### DIFF
--- a/doc/docker_integration.md
+++ b/doc/docker_integration.md
@@ -1,6 +1,10 @@
 Docker integration
 ===============================================================================
 
+**Note:** This page is mainly about building Haskell packages inside docker containers.
+If you want to deploy your built Haskell programs into a docker container,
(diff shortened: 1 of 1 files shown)


doc/docker_integration.md
```diff
@@ -1,6 +1,10 @@
 Docker integration
 ===============================================================================
 
+**Note:** This page is mainly about building Haskell packages inside docker containers.
+If you want to deploy your built Haskell programs into a docker container,
+look [here](../GUIDE.md) instead.
```
snoyberg: I think this should just be `GUIDE.md`, the file is in the same directory. And what about providing an anchor to the appropriate part of the page?

nh2: Hmm you're right, not sure where I got the `../` from. I updated the PR.

PR #3226, opened 2017-06-21
Fix #2886: include package names in the haddock index.

### LINKED ISSUE #2886: No package names in haddock index
The package index generated by `stack haddock` does not contain the package names. The package names seem to be present for the packages distributed with ghc but not for the packages installed with stack.

Here's a screenshot of my `~/.stack/snapshots/x86_64-linux/lts-7.0/8.0.1/doc/index.html`. Notice the difference between, say, `BasicTypes` (distributed with ghc) and `Bio.Core` (from the `biocore` package, installed with stack from hackage).

stack-haddock

It's been a while since I looked at haddock, but my guess is that it has a command line option to specify the package name that stack does not pass.

The practical significance of this is that it makes it harder to search the index page, especially for packages that are named differently from their modules.

### Stack version

```
$ stack --version
Version 1.3.3, Git revision 87068d8edb64e81f4062ae889472ef74452a2b97 (4441 commits) x86_64 hpack-0.15.0
```

### Method of installation

from git master

judah:
It turned out that Haddock expects to see packages with those names
in the GHC package DB, and omits the names otherwise.  Pointing
Haddock to the local and snapshot DBs resolved the issue.

Tested by running "stack haddock" for a local package and verifying that all the generated `index.html` files (local packages, local packages and deps, snapshot packages) list the package name for each module.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -84,6 +84,8 @@ Bug fixes:
   See [#3142](https://github.com/commercialhaskell/stack/issues/3142)
 * `stack script`'s import parser will now properly parse files that
   have Windows-style line endings (CRLF)
+* `stack haddock` now includes package names for all modules in the
+   Haddock index page.
 
 
 ## 1.4.0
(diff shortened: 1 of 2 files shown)


snoyberg:
Thank you!

PR #3231, opened 2017-06-27
Turn on --install-ghc by default

snoyberg:
This is in response to feedback from the Haskell website working group,
who requested this change to ease the new user onboarding process.

__NOTE__ Please feel free to click on the 👍 or 👎 buttons on this description to indicate your support or opposition to this change.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -12,6 +12,11 @@ Behavior changes:
   benchmarks and tests to the beginning of the args, instead of the end.
   See [#2399](https://github.com/commercialhaskell/stack/issues/2399)
 * Support for Git-based indices has been removed.
+* The `--install-ghc` flag is now on by default. For example, if you
+  run `stack build` in a directory requiring a GHC that you do not
+  currently have, Stack will automatically download and install that
+  GHC. You can explicitly set `install-ghc: false` or pass the flag
+  `--no-install-ghc` to regain the previous behavior.
 
 Other enhancements:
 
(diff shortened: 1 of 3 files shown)


bitemyapp:
Strongly in favor. Seems unlikely anyone would be using Stack and not want this. I think a no-install argument so that a GHC not being installed is still a good idea for things like CI or build scripts that should fail if the environment isn't configured correctly.

snoyberg:
> I think a no-install argument... is still a good idea.

Agreed, `--no-install-ghc` will be available after this change for this purpose. And scripts can be written for both old and new Stack by explicitly putting in one of these arguments at all times.

vaibhavsagar:
This doesn't currently work with NixOS, you have to do `stack --nix --install-ghc`.

snoyberg:
1. I was under the impression that `nix` is supposed to maintain its own installation of GHC, since it can't install from standard tarballs
2. I believe (and could be very mistaken) that `stack --nix` calls out to a separately downloaded `stack` executable which would not have this modified behavior. If this is true: this modification would only work seamlessly once officially released

Perhaps some Nix experts can weigh in.

iand675:
I'm in favor. When I pull a new project and run stack build, I tend to leave it for a bit and do something else in anticipation of some build process. I'll come back later and sometimes find that nothing has happened and then I have to context switch again.

eborden:
:+1:

snoyberg:
Thanks for the feedback everyone.

Issue #3232, opened 2017-06-27
Improved PATH-aware behavior for `stack upgrade`


snoyberg:
Problem: users will often run the install script and end up with `stack` in `/usr/local/bin`, and when they run `stack upgrade`, are surprised that the next `stack` call still uses the old version. To mitigate this, @borsboom and I propose:

1. `stack upgrade` continues to copy the executable to `~/.local/bin` as it has until now
2. If the current Stack executable is running from there: we're done. Otherwise: continue.
3. Attempt to copy the `stack` executable to the location of the current executables.
4. If that fails with a permission denied error:
    1. Explain the situation to the user, and recommend copying from `source file path` to `dest file path`
    2. Ask if the user wants us to try to use `sudo` to perform the copy (and explain that it may fail due to OS specific things)
    3. If the user says yes, try running `sudo cp ...`, and potentially similar commands with `su`.

Presumably we won't even bother trying this on Windows.

PR #3234, opened 2017-06-27
Perform path check on stack upgrade #3232

snoyberg:

I used the following script to ensure things worked correctly:

```shell
#!/usr/bin/env bash

set -exu

sudo rm -rf /owned-by-root
sudo mkdir -p /owned-by-root
sudo chmod 755 /owned-by-root
sudo chown root /owned-by-root

sudo cp $(stack exec which stack) /owned-by-root/tmp-stack
sudo chown root /owned-by-root/tmp-stack
/owned-by-root/tmp-stack upgrade --binary-version 1.4.0
```


src/Stack/Setup.hs
```diff
@@ -1821,6 +1827,69 @@ downloadStackExe platforms0 archiveInfo destDir testExe = do
...
+      liftIO $ setFileMode (toFilePath tmpFile) 0o755
+#endif
+      liftIO $ renameFile tmpFile executablePath'
+      $logInfo "Stack executable copied successfully!"
+    case eres of
+      Right () -> return ()
+      Left e
+        | isPermissionError e -> do
+            $logWarn $ T.pack $ "Permission error when trying to copy: " ++ show e
+            $logWarn "Should I try to perform the file copy using sudo? This may fail"
```
borsboom: It might be good to show the users exactly which commands will be run (that way they have the option to ctrl-C and do it themselves if they don't trust stack to run commands as root itself)

Issue #3229, opened 2017-06-26
stack build [target] no longer works, builds all targets


nh2:
There is a regression in stack that `stack build myexe` no longer builds just the `executable myexe`, but all executables from the package.

Especially projects with many executables, and projects where linking takes a long time suffer from this (both are the case for IOHK's [`cardano-sl`](https://github.com/input-output-hk/cardano-sl)).

The reason according to @mgsloan:

> I think this was the result of switching from using `runhaskell Setup.hs install` to copy + register - https://github.com/commercialhaskell/stack/commit/f71a515ef02aac4ea864bc4a5979c56b884b4ae2 .
> `copy` fails if it can't find every executable, and there's no way in released cabal to tell it to only copy particular targets.  There was a patch well before the most recent cabal release, but for some reason it is only available on master.  So, with future cabal this may be fixable.

snoyberg:
Let's finally set to rest the theory that it's the switch from `install` to `copy`/`register` (it's not):

```
stack new foo
cd foo
stack exec --no-ghc-package-path -- runghc Setup.hs configure --user
stack exec --no-ghc-package-path -- runghc Setup.hs build lib:foo
stack exec --no-ghc-package-path -- runghc Setup.hs install
```

Results in:

```
Installing library in
/Users/michael/.cabal/lib/x86_64-osx-ghc-8.0.2/foo-0.1.0.0-K7FRl2SdaSPJF1E2Gf1Eav
Installing executable(s) in /Users/michael/.cabal/bin
Warning: The directory /Users/michael/.cabal/bin is not in the system search
path.
dist/build/foo-exe/foo-exe: copyFile: does not exist (No such file or
directory)
```

snoyberg:
The problem: Cabal (the library) will fail and die if you call the `copy` command when some of the executables are unavailable.

Original solution: oh well, user will get an error message, realize he/she has to call `build` once without arguments, and then can give components.

Current solution: never tell `build` to only build a subset of components, since that can lead to `copy` failing.

Proposed solution: before calling `Cabal build`, Stack should check if all executables are built. If they are, pass in components. Otherwise, give a `logInfo` that we're building all components once, and then future calls will build individual components.

chrisdone:
Found the related commit: dc6313f16bf9eb09f0607718cef1dab608b5705c

snoyberg:
Merged to master, closing.

PR #3237, opened 2017-06-28
Build all exes once; then only specified (#3229)

chrisdone:
Works like this:

* We simply look at the expected executable binary files e.g. `.stack-work/dist/x86_64-osx/Cabal-1.22.4.0/build/alpha/alpha` and collect a `Map Text ExecutableBuildStatus` to determine which executable targets have or have not been built. 
* Then in `exesToBuild` we can build the user-supplied targets if all executables have been built once, otherwise we build all executables in the cabal file.

This steps around the issue described [here](https://github.com/commercialhaskell/stack/issues/3229#issuecomment-311236016), while letting us avoid build more executables than necessary.

Example output: https://gist.github.com/chrisdone/98a8dfbc7a793f4c6307e5aecddeb3ea

I added an integration test.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -1296,7 +1296,10 @@ singleBuild runInBase ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} in
         $ \package cabalfp pkgDir cabal announce _console _mlogFile -> do
             executableBuildStatuses <- getExecutableBuildStatuses package pkgDir
             when (not (cabalIsSatisfied executableBuildStatuses) && taskIsTarget task)
-                 ($logInfo "Building all executables once. After a successful build of all of them, only specified executables will be rebuilt.")
+                 ($logInfo
+                      ("Building all executables for `" <> packageNameText (packageName package) <>
+                       "' once. After a successful build of all of them, only specified executables will be rebuilt."))
+
             _neededConfig <- ensureConfig cache pkgDir ee (announce ("configure" <> annSuffix executableBuildStatuses)) cabal cabalfp
 
             let installedMapHasThisPkg :: Bool
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -25,8 +25,8 @@ import           Control.Arrow ((&&&), second)
 import           Control.Concurrent.Execute
 import           Control.Concurrent.MVar.Lifted
 import           Control.Concurrent.STM
-import           Control.Exception.Safe (catchIO)
 import           Control.Exception.Lifted
+import           Control.Exception.Safe (catchIO)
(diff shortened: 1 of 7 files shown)


src/Stack/Build/Execute.hs
```diff
@@ -1287,7 +1294,10 @@ singleBuild runInBase ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} in
 
     realConfigAndBuild cache allDepsMap = withSingleContext runInBase ac ee task (Just allDepsMap) Nothing
         $ \package cabalfp pkgDir cabal announce _console _mlogFile -> do
-            _neededConfig <- ensureConfig cache pkgDir ee (announce ("configure" <> annSuffix)) cabal cabalfp
+            executableBuildStatuses <- getExecutableBuildStatuses package pkgDir
+            when (not (cabalIsSatisfied executableBuildStatuses) && taskIsTarget task)
+                 ($logInfo "Building all executables once. After a successful build of all of them, only specified executables will be rebuilt.")
```
snoyberg: Perhaps include the name of the package in the log message? It's possible this will apply to multiple packages in a single build.

src/Stack/Build/Execute.hs
```diff
@@ -1754,27 +1815,36 @@ extraBuildOptions wc bopts = do
...
-    --
-    -- if lpWanted lp
-    --     then exeComponents (lpComponents lp)
-    --     -- Build all executables in the event that no
-    --     -- specific list is provided (as happens with
-    --     -- extra-deps).
-    --     else packageExes (lpPackage lp)
+      map (T.unpack . T.append "exe:") (Set.toList $ exesToBuild executableBuildStatuses lp)
+
+-- | History of this function:
```
snoyberg: Nice addition

src/Stack/Build/Execute.hs
```diff
@@ -1754,27 +1815,36 @@ extraBuildOptions wc bopts = do
...
+--   (modulo success), and thereafter pay attention to user-wanted
+--   components.
+--
+exesToBuild :: Map Text ExecutableBuildStatus -> LocalPackage -> Set Text
+exesToBuild executableBuildStatuses lp =
+    if cabalIsSatisfied executableBuildStatuses && lpWanted lp
+        then exeComponents (lpComponents lp)
+        else packageExes (lpPackage lp)
+
+-- | Do the current executables satisfy Cabal's bugged out requirements?
```
snoyberg: 😆 :rofl:

src/Stack/Build/Execute.hs
chrisdone: Good idea.

chrisdone: Added a commit for that.

PR #3260, opened 2017-07-13
Alexeyzab 3006 speed up store instances

snoyberg:
This updates #3120 to work with the new extensible snapshots code. Related to #3006. Pinging @alexeyzab.

### DIFF
--- a/src/Network/HTTP/Download/Verified.hs
+++ b/src/Network/HTTP/Download/Verified.hs
@@ -55,7 +55,7 @@ import              Network.HTTP.Types.Header (hContentLength, hContentMD5)
 import              Path
 import              Prelude -- Fix AMP warning
 import              System.Directory
-import              System.FilePath ((<.>))
+import qualified    System.FilePath as FP ((<.>))
 import              System.IO
 
 -- | A request together with some checks to perform.
@@ -245,7 +245,7 @@ verifiedDownload DownloadRequest{..} destpath progressSink = do
         if p then m >> return True else return False
 
     fp = toFilePath destpath
-    fptmp = fp <.> "tmp"
+    fptmp = fp FP.<.> "tmp"
     dir = toFilePath $ parent destpath
 
     getShouldDownload = do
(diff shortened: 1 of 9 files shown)


snoyberg:
Possible enhancement: instead of storing a `StaticSize 64 ByteString` with the base16-encoded value, we could store 4 `Int64` values in the unencoded values, which will avoid the `ByteString` overhead and related heap fragmentation.

alexeyzab:
To make sure I understand this correctly, a checklist of what needs to be done here:

* [ ] Possibly replace `StaticSize 64 ByteString` with 4 `Int64` values. Would that look something like `StaticSize 64 (Int64, Int64, Int64, Int64)`?

Should I also remove the commented-out implementations of `poke` and `peek`?

snoyberg:
@alexeyzab I'd leave off the `Int64` change unless you're particularly enthusiastic about it. _If_ you want to take a crack at it, the best way to define it is:

```haskell
data StaticSHA256 = StaticSHA256 !Word64 !Word64 !Word64 !Word64
```

We don't want to use a signed representation since we're just storing the raw bits. And we need the strictness annotations so that GHC can unpack the values directly into the constructor, making the representation much smaller.

snoyberg:
Also, if you're planning on continuing updates on this, can I recommend that you merge my branch into the one you created, and I'll close out this PR?

alexeyzab:
@snoyberg Ah, I understand, thanks for explaining that! Sure, is that done by using the `Resolve conflicts` button on the other PR or is there anything else I can do?

snoyberg:
No, you'll have to do it on your machine. Check out the branch you were working on and run:

    git fetch origin
    git merge origin/alexeyzab-3006-speed-up-store-instances

Then push to your repo and you should have all the changes.

Depending on how adventurous I'm feeling, I may write a helper library right now for the `Word64` trick, unless you tell me you're taking a crack at it.

alexeyzab:
Ah, that's what I've done already, wasn't sure if there was something else I missed. Thanks again! Hm, it'd probably take me longer to implement this and I wouldn't want to slow things down. I am curious what that trick would end up looking like though.

snoyberg:
FYI, I don't see the commits I pushed here on your PR, so I don't think it's been pushed to the right branch yet.

alexeyzab:
Whoops, my bad. Should be done now.

snoyberg:
Cool, I'll close this up

Issue #2750, opened 2016-10-30
Documentation should mention hpack


chris-martin:
Stack's support for hpack `package.yaml` files as an alternative to `.cabal` files seems to be [missing from the documentation](https://readthedocs.org/search/?q=hpack&check_keywords=yes&area=default&project=stack&version=stable&type=file). Having never heard of hpack before, this caused me [much confusion](https://stackoverflow.com/questions/40332040/what-goes-in-a-stack-package-yaml-file). The docs ought to discuss this feature a bit.

lwm:
Fair enough. I also didn't know what a `package.yaml` was either.

Blaisorblade:
Agreed, PRs welcome (traveling & busy right now, sorry).

chris-martin:
My point in raising this issue was to make it possible to figure out what a "package.yaml" file does in a Stack project by reading the Stack documentation, so IMO it's not resolved since the Stack docs still don't mention package.yaml.

lwm:
@chris-martin, can you amend the documentation entry then?

PR #3286, opened 2017-07-23
Mention hpack a bit more in the docs

chris-martin:
This addresses #2750 - When I first started using Stack I was puzzled by projects that had `package.yaml` files, because I'd never heard of Hpack and the Stack documentation barely mentions it. The aim of this PR is to work the word "package.yaml" into the docs in a few places, so that someone wondering what a `package.yaml` file does can have a better chance of figuring it out.


mgsloan:
Thanks!

PR #3289, opened 2017-07-24
Upgrade stack to the latest version of hpack

mgattozzi:
hpack recently updated to version 0.18 and one of the big changes from
that is the ability to handle Happy and Alex files. Previously this made
using hpack with projects that contained these files impossible. As such
they were stuck with using cabal files.

This commit updates stack to this newer version of hpack so that stack
can be used with hpack or cabal based projects interchangeably if Alex
or Happy files are involved.

We have a large Haskell code base at work that we were trying to migrate over to hpack that failed because of the lack of Alex and Happy support. This was added to hpack but stack still didn't have it. I tested this version of stack on my machine and it worked but I'd like another set of eyes if possible.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -40,6 +40,8 @@ Other enhancements:
   assist with the user experience when the `PATH` environment variable
   has not been properly configured, see
   [#3232](https://github.com/commercialhaskell/stack/issues/3232).
+* The `hpack` dependency was updated allowing projects with Alex and Happy files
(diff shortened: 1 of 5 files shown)


borsboom:
I think `stack-7.10.yaml` needs to be updated as well.  `stack-nightly.yaml` should be OK since nightly snapshots already include hpack-0.18.

mgattozzi:
Apparently this already happened when updating the LTS so closing.

Issue #3309, opened 2017-08-03
GHCI script shouldn't load packages separately


mgsloan:
https://github.com/commercialhaskell/stack/pull/2457 added a change to load packages individually.  This has caused some non-ideal behavior when loading multiple packages into ghci / intero.  Also, I've observed that the approach doesn't work in general, its success is non-deterministic.  In particular:

1) This causes it to output "Ok, modules loaded: " messages N types, where N = number of packages.

2) If an error occurs, it can be repeated upto N times, filling up your console with duplicated errors.

3) It doesn't order the packages based on dependency order, so if package A is before package B in the list, and A depends on B, by the time it asks to load B's modules, they've already been loaded.  So sometimes the source path setting doesn't work.  So, the primary motivation for that change doesn't always work.

4) Intero re-collects type info for every load statement.  This is probably an issue with intero.  But since it is 

5) The setting of cwd does not replay on reload.

So generally, that change was a major downgrade in how `stack ghci` behaves, and it does not really properly support switching cwd.

I'm working on a change to go back to pretty much how scripts were generated before, this is documenting the motivation for the change.  It will still benefit from `:cd-ghc` in the case that intero is just loading one package, which I believe is the default in emacs

lwm:
@mgsloan, while dogfooding with the latest stack, I discovered that the default `stack ghci` doesn't load up any modules any more. I tried `stack ghci --load` and nothing was loaded either. I found my way here - do I understand correctly that this change shouldn't affect this default behavior? Maybe I'm on the wrong track too though.

mgsloan:
@lwm Ah, good catch, silly mistake on my part.  Fix pushed to master.

lwm:
Nice one! :beers:

PR #3311, opened 2017-08-03
Have "stack ghci" load modules all at once instead of grouped by pkg #3309

mgsloan:


mgsloan:
Went ahead and added another change to this, since I didn't want to deal with merge conflicts

chrisdone:
Cool, I'll test it out locally.
---
As an aside regarding the test suite, perhaps we could copy/paste the process launching helper I use in intero's own test suite to launch itself and check outputs: https://github.com/commercialhaskell/intero/blob/master/src/test/Main.hs#L500-L554 If such a thing would be handy.

Issue #3033, opened 2017-03-01
"The following package identifiers were not found in your indices"


sjakobi:
```
~ $ stack --resolver nightly build tasty-stats
Selected resolver: nightly-2017-03-01
Selected mirror https://s3.amazonaws.com/hackage.fpcomplete.com/                                 
Downloading timestamp                                                                            
No updates to your package list were found                                                       
Updating package index Hackage (mirrored at https://s3.amazonaws.com/hackage.fpcomplete.com/) ...The following package identifiers were not found in your indices: tasty-stats-0.2.0.2
```

This might be related to the recent Hackage maintenance or the S3 downtime. I have notified the Hackage maintainers [here](https://www.reddit.com/r/haskell/comments/5wpkhh/heads_up_short_hackage_downtime_today_30_minutes/decpxvo/).

It would be nice if `stack` could handle this better or at least make a suggestion like "report a bug".

mgsloan:
I could indeed reproduce this when you opened it, however, it seems to have resolved itself!  Indeed likely due to something like hackage maintenance or s3 downtime.

sjakobi:
I'm seeing similar exceptions once more:

```
~/s/hackage-server (master|✔) $ stack init --ignore-subdirs --force
Looking for .cabal or package.yaml files to use to init the project.
Using cabal packages:
- hackage-server.cabal

Selecting the best among 10 snapshots...

Missing some cabal revision files, updating indices
Selected mirror https://s3.amazonaws.com/hackage.fpcomplete.com/                                 
Downloading timestamp                                                                            
No updates to your package list were found                                                       
Did not find .cabal file for safe-exceptions-checked-0.1.0 with SHA of 2d07e15162937261823e35786206c8a218db8ae1 in tarball-based cache
Selected mirror https://s3.amazonaws.com/hackage.fpcomplete.com/                                 
Downloading timestamp                                                                            
No updates to your package list were found                                                       
Missing some cabal revision files, updating indices                                              
Selected mirror https://s3.amazonaws.com/hackage.fpcomplete.com/                                 
Downloading timestamp                                                                            
No updates to your package list were found                                                       
Did not find .cabal file for safe-exceptions-checked-0.1.0 with SHA of 2d07e15162937261823e35786206c8a218db8ae1 in tarball-based cache
[...]

PR #3324, opened 2017-08-07
Clear index cache if it is old #3033

mgsloan:

Not sure if this fixes the problem described in #3033, but I believe this will help in cases where stack is terminated after downloading the index but before clearing the cache.

### DIFF
--- a/src/Stack/PackageIndex.hs
+++ b/src/Stack/PackageIndex.hs
@@ -50,6 +50,7 @@ import           Network.HTTP.Client.TLS (getGlobalManager)
 import           Network.HTTP.Download
 import           Network.URI (parseURI)
 import           Path (toFilePath, parseAbsFile)
+import           Path.Extra (tryGetModificationTime)
 import           Path.IO
 import           Stack.Types.Config
 import           Stack.Types.PackageIdentifier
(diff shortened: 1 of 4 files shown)


src/Stack/PackageIndex.hs
```diff
@@ -337,7 +338,23 @@ updateIndexHackageSecurity indexName' url (HackageSecurity keyIds threshold) = d
...
+            -- could happen if a prior run of stack updated the index,
+            -- but exited before deleting the cache.
+            --
+            -- See https://github.com/commercialhaskell/stack/issues/3033
+            mindexModTime <- tryGetModificationTime =<< configPackageIndex indexName'
+            mcacheModTime <- tryGetModificationTime =<< configPackageIndexCache indexName'
+            case (mindexModTime, mcacheModTime) of
+                (Right indexModTime, Right cacheModTime) | cacheModTime < indexModTime -> do
+                    deleteCache indexName'
+                    $logInfo "No updates to your pakcage list were found, but clearing the index cache as it is older than the index."
```
lwm: s/pakcage/package/g

mgsloan: Thanks!  Fixed.

Issue #3154, opened 2017-05-02
Can't build gi-gtk on win10: Did not find a GI repository for Dependency-name in ["C:\\Program Files\\gir-1.0"]


Anrock:
### Steps to reproduce
0. Be on windows 10.
1. `stack new foobar simple`
2. Add `gi-gtk` to `build-depends` section in `foobar.cabal`
3. Run `stack setup`
4. Run `stack build`

### Expected
Successful build

### Actual
```
gi-gdkpixbuf-2.0.12: configure
gi-cairo-1.0.12: configure
gi-pango-1.0.13: configure

--  While building package gi-pango-1.0.13 using:
[...]

PR #3325, opened 2017-08-07
Put msys on the PATH before ghc, potential fix for #3154

mgsloan:
As mentioned [here](https://github.com/commercialhaskell/stack/issues/3154#issuecomment-320775528), I'm really not sure if this is a good change.  It may fix that particular problem, but it is hard to predict all the impacts of this behavior change.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,6 +19,8 @@ Behavior changes:
   currently have, Stack will automatically download and install that
   GHC. You can explicitly set `install-ghc: false` or pass the flag
   `--no-install-ghc` to regain the previous behavior.
+* On windows, the msys bin directory now comes before the ghc bin
+  directory. See [#3154](https://github.com/commercialhaskell/stack/issues/3154)
 
 Other enhancements:
 
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -487,7 +487,7 @@ ensureCompiler sopts = do
         Nothing -> return Nothing
         Just (compilerTool, mmsys2Tool) -> do
             -- Add GHC's and MSYS's paths to the config.
-            let idents = catMaybes [compilerTool, mmsys2Tool]
+            let idents = catMaybes [mmsys2Tool, compilerTool]
             paths <- mapM extraDirs idents
             return $ Just $ mconcat paths
 


Anrock:
@mgsloan how do i install it? I can't find info about building stack on windows and there is no artifacts in appveyour build.

mgsloan:
@Anrock After doing a `git clone https://github.com/commercialhaskell/stack` and `git checkout msys-before-ghc-on-path-3154`, run `stack install`

Anrock:
@mgsloan looks ok.
1. I've done `stack install` of this branch.
2. I've changed to STACK_ROOT to another dir to be sure packages are built using stack from this branch
3. I've removed XDG_DATA_DIR from Windows env vars.
Then i've tried to reproduce, using steps from original issue
`stack new foobar simple` produced error:
```
Downloaded nightly-2017-08-13 build plan.
AesonException "Error in $.flags['bytestring--lt-0_10_4']: Invalid flag name: \"bytestring--lt-0_10_4\""
```
So i've created project with lts-9.0 resolver.
Put gi-gtk dependency in cabal file and run `stack build`. Build succeded, though i still had to define XDG_DATA_DIRS so gi-* builds could locate gir directory.

atwupack:
@mgsloan Today I also ran into the issue with the wrong zlib dll being used during the execution of my program. Maybe a solution with less impact would be to add a new option `--no-ghc` to the `stack exec` command which prevents the path to the ghc bin from being added to the environments path.

lwm:
> Today I also ran into the issue

@atwupack, did the existing fix on this branch solve your issue?

atwupack:
@lwm I have tested the following:

1. Changed back the zlib1.dll to the original one in ...\AppData\Local\Programs\stack\x86_64-windows\ghc-8.0.2\mingw\bin. I had previously copied the dll from ...\AppData\Local\Programs\stack\x86_64-windows\msys2-20150512\mingw64\bin there in order to fix the issue temporarily.
2. Ran my program with `stack exec`. This caused the error about the missing "inflateValidate" entry point.
3. Built stack based on this branch and replaced the version in ...\AppData\Roaming\Local\bin with it.
4. Ran my program again with `stack exec`. This then worked fine w/o any errors.

This fix seems to solve the issue and I did not experience any problems caused by changing the PATH order.

mgsloan:
Sounds like this branch is not a direct fix for the issue.  I don't grok this windows stuff very well, and don't feel comfortable merging a somewhat arbitrary change to semantics.  So, closing this.

Issue #849, opened 2015-08-25
Unify behaviour of flags and ghc-options


borsboom:
In particular, ghc-options in stack.yaml should also promote snapshot dependencies to extra-deps (to avoid breaking the snapshot sandbox).  This also will open the door to more easily adding other options that effect what is passed to `runhaskell Setup.hs configure`.

borsboom:
This may be subsumed by #1265 (Unify extra-deps and extensible snapshots)

mgsloan:
We have a couple cases where a ghc-option *must* be applied to all packages.  I'm not even sure how feasible that is considering that we have wired-in packages (that come with GHC / GHCJS).

Nonetheless, it makes a lot of sense for ghc-options to actually be declarative rather than having the meaning be sensitive to what's already installed.  I think this is one of the ugliest parts of stack right now.

I agree that this is likely to be subsumed by #1265

mgsloan:
I've spent some time looking into this, and figured I'd note down the current state of things:

The `flags` specified in the stack.yaml must refer to packages declared in the config.  So in this case promotion to an extra-dep is the responsibility of the user.  See `checkFlagsUsed` in `Stack.Build.Source`.  The automatic promotion to extra-dep only happens for CLI flags (`convertSnapshotToExtra`).

Flags specified in the config and cli *replace* those specified in the snapshot.  Conversely, ghc-options get added to those specified in the snapshot.  See `extraDeps3` in `Stack.Build.Source`.

snoyberg:
I specifically _avoided_ making this change when doing my refactoring to match the previous behavior, together with FIXMEs in the code to reconsider. I'm happy to see this issue. After the PR is merged (#3249), I'll attack this with a separate PR.

snoyberg:
Alright, here's a bit of a wrinkle from a design standpoint. We've got two kinds of ghc-options overrides:

* By package
* For all packages

For by package options (specified in the `stack.yaml` file), I think it's pretty straightforward that we should promote packages to the local database when they are set. It's less obvious was good behavior is for options applied to all packages (either from the `stack.yaml` file or from the command line). Here are the choices I see:

1. If any such options are specified, promote _all_ packages to the local database. This will result in a lot of rebuilding for confusing reasons to the user
2. Modify behavior to only apply the ghc-options to already-local packages, which is a perhaps significant behavior change
3. Continue the current behavior: options are applied and silently affect the snapshot database

[...]

PR #3327, opened 2017-08-08
Promote packages to local database by ghc-options #849

snoyberg:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -26,6 +26,10 @@ Behavior changes:
   paths.  TH relative paths will still work when loading a single
   package into intero. See
   [#3309](https://github.com/commercialhaskell/stack/issues/3309)
+* Setting GHC options for a package via `ghc-options:` in your
+  `stack.yaml` will promote it to a local package, providing for more
+  consistency with flags and better reproducibility. See:
+  [#849](https://github.com/commercialhaskell/stack/issues/849)
 
 Other enhancements:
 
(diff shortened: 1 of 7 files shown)


snoyberg:
@mgsloan @borsboom Will you review the comments I've put in the changelog in this PR before I merge? I think this is the right call, but would appreciate a second (and third) opinion.

borsboom:
So if I read this correctly, this doesn't implement the `local-only` option mentioned in https://github.com/commercialhaskell/stack/issues/849#issuecomment-320892095.

In a way, this seems similar to setting flags on a package, but IIRC that doesn't automatically promote a package to local (instead, it requires you to make the package an extra-dep).  I'm wondering if there's a justification for that difference in behaviour or if it's just by accident.

snoyberg:
Just to unify the discussion from here and the issue: I _think_ we're in agreement that this is a fair first step, and are considering some additional changes. Let me document them here:

* Some way to apply flags in a stack.yaml file to all local packages. @mgsloan mentioned a `$locals` key. I like that, and would move to replace `*` with `$everything` to be more explicit (with a warning printed for the old syntax).
* A broader question about implicit snapshots. We could add them in under @mgsloan's idea of only having project packages and things depending on them end up in the local database. I'm open to revisiting that. And once that's in place, yes, it would make sense to promote all packages to a custom snapshot.

My recommendation is, in the interest of keeping the changesets manageable, we consider if this change is an improvement. If so, let's merge and open new issues for these enhancements.

mgsloan:
@snoyberg Yep, that seems reasonable to me!

snoyberg:
OK, two new issues opened. I'm going to merge this so I can get started on the other two.

Issue #3254, opened 2017-07-09
ghc output is not coloured when run through stack


duog:
ghc 8.2 will output error and warning diagnostics using ANSI terminal codes by default. I just tried this with stack installed by stack upgrade --git, and it no colours appear. The messages are not garbled.

tfausak:
This would be really nice to have now that GHC 8.2.1 produced colored output.

PR #3331, opened 2017-08-09
GHC color output

kadoban:

Tested `stack build` with `ghc-8.2.1` resolver and got colored error output and tested in older GHC that `stack build` does not error out (would if the new flag were being passed to old GHC).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -55,6 +55,11 @@ Other enhancements:
 * Added `stack ghci --only-main` flag, to skip loading / importing
   all but main modules. See the ghci documentation page
   for further info.
+* Allow GHC's colored output to show through. GHC colors output
+  starting with version 8.2.1, for older GHC this does nothing.
+  Sometimes GHC's heuristics would work fine even before this change,
+  for example in `stack ghci`, but this override's GHC's heuristics
+  when they're broken by our collecting and processing GHC's output.
 
 Bug fixes:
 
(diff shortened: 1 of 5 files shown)


snoyberg:
This is great, thanks!

One request, one question:

* Can you add a note to the changelog explaining the change?
* Does this cause the contents of the log files themselves to be unreadable? Do we need to filter out control characters when save to the log files?

kadoban:
Anytime.

Sure, I'll add a note. I didn't initially because I wasn't sure if one was appropriate since it's more of a GHC change than a stack change, that we're then just passing through. But it makes sense to mention it, since yeah older stack versions will behave differently obviously.

Ah, that's a good question, I'll test that, the log stuff.

I'm also not sure if this covers every place that GHC output would be coming from that could use coloring. I looked around a little bit, but could definitely do some more.

duog:
This would solve #3254

snoyberg:
I've tested locally, and this looks good. The only (potential) downside is that the color codes are making it into the logs, e.g.

screen shot 2017-08-10 at 12 09 43 pm

I put up a little poll on Twitter to get feedback on correct behavior. There is definitely an advantage to just keeping the color codes in the log, as `cat logfile` is nice and colorful too. However, it does make it harder to use a text editor/less on the file.

snoyberg:
I've pushed an extra commit that, when color is turned on, creates a second version of the log files with the ANSI codes stripped. Once that commit passes Travis, I'll merge this PR to master. Thanks again!

borsboom:
> However, it does make it harder to use a text editor/less on the file.

`less -R` FTW.

kadoban:
Thanks! Yep I couldn't find anything wrong either on my end.

PR #3337, opened 2017-08-10
Kadoban ghc color output

snoyberg:
Just adds in one extra commit to the original PR, which strips ANSI codes from the log files.


mgsloan:
Maybe it's pretty unlikely, and this would add complication, but to me it makes sense to have a bound on the `CB.dropWhile (/= 109)`, as color codes are bounded by length.  Otherwise it could omit arbitrarily large portions of the output, particularly if it ever uses any non-color codes, or they somehow make it into the output (TH could `putStrLn` an ESC char for example).

PR #3359, opened 2017-08-17
Use path-0.6

felixonmars:


mgsloan:
Thanks!  I actually think it'd be best to use a more restricted range, since 0.6 has different behavior and will cause some warnings.  Adding

```
- path-0.6.1
- path-io-1.3.3
```

to the extra-deps of each yaml file should do the trick.  Then, just a matter of fixing warnings / compile errors.

felixonmars:
@mgsloan Updated accordingly :)

mgsloan:
LGTM, thanks!

PR #3314, opened 2017-08-06
Bump to latest LTS and use new HLint.

lwm:
* [ ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [x] The documentation has been updated, if necessary.

PR #2812, opened 2016-11-28
Implementation of "--copy-compiler-tool" #2643

mgsloan:
* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [x] The documentation has been updated, if necessary.

As described in #2643, this implements a mechanism for having a bin directory for tools that match a particular ghc version.  This compiler specific dir looks like is `/home/mgsloan/.stack/compiler-tools/x86_64-linux/ghc-8.0.1/bin`.  It is included on the PATH when stack is building projects and when running exec. It adds:

1) `stack path --compiler-tool-bin`, yielding a path to the dir compiler-specific tools are installed to
2) `stack build --copy-compiler-tool`, which causes executables for target packages to get copied to the compiler-specific dir.

Tested this change by using it to install a local intero.  Querying `stack exec -- which intero` yielded the correct path.  `stack path --local

PR #3368, opened 2017-08-20
Explicitly install happy for style builds

lwm:
Attempt to solve build failure (seen [on master](https://travis-ci.org/commercialhaskell/stack/jobs/266365181#L670) and also reported [over here](https://github.com/commercialhaskell/stack/pull/2812#issuecomment-323574408)).

This failure is a result of #3314 and was seen also in build failures in #2812.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,10 @@ install:
    set -ex
    case "$BUILD" in
      style)
+       # This should be resolved (and removed) by Stack 1.6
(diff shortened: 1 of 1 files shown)


lwm:
Seems to work: https://travis-ci.org/commercialhaskell/stack/jobs/266488094#L549

mgsloan:
Seems like a build is still timing out, but this is moving closer to CI being green.  Thanks!

Issue #3376, opened 2017-08-22
readthedocs.org search is broken


mgsloan:
Thinking this is likely a readthedocs bug, but perhaps it is some configuration issue.  Seems that the search on https://docs.haskellstack.org/en/stable/README/ doesn't work.  If I input keywords that are definitely used, like `stack`, I get no results - https://readthedocs.org/search/?q=stack&check_keywords=yes&area=default&project=stack&version=stable&type=file

Pinging @borsboom

lwm:
https://github.com/commercialhaskell/stack/issues/3298 :disappointed:

mgsloan:
This guy's workaround - https://github.com/rtfd/readthedocs.org/issues/1487#issuecomment-323290547 - seems ok: https://github.com/nodemcu/nodemcu-firmware/commit/7dd89dd15ef993c9740ba4d4361e6be8edd1284b .  Stack's docs are not huge, so doing clientside search seems fine.

borsboom:
That workaround certainly seems like a reasonable thing to try.  We'd probably need to add an `extra_javascript` clause to mkdocs.yml (like [this](https://github.com/nodemcu/nodemcu-firmware/blob/7dd89dd15ef993c9740ba4d4361e6be8edd1284b/mkdocs.yml#L18))

lwm:
Nice detective work :shipit:

borsboom:
The docs for the branch, for whatever reason, decided to build successfully (even though most builds are still failing).  In the JS console, I get a 404 when it tries to load `https://docs.haskellstack.org/en/fixup-search/etc/mkdocsjs/searchhack.js`.  That's because the `extra_javascript` setting in `mkdocs.yml` should be within the `doc` directory.  I've pushed a new branch `3376-fix-search` that fixes that... hopefully mkdocs will build it.

borsboom:
Mkdocs does seem to have rendered the docs: https://docs.haskellstack.org/en/3376-fix-search/README/.

However, search still doesn't work.

borsboom:
I suspected that the `if (window.location.origin.indexOf('readthedocs') > -1)` check was the reason, so I tried using http://stack.readthedocs.io/en/stable/README/ instead but also didn't work at that URL.

borsboom:
[...]

PR #3379, opened 2017-08-23
Try the mkdocs search hack

lwm:
Something for https://github.com/commercialhaskell/stack/issues/3376.

Unfortunately, I couldn't test this (because mkdocs search actually works locally). For someone who has access to the RTD builder, I think you can build off a branch and view the documentation. Then just try to search something and see if you get the mkdocs search bar and not the broken RTD one.

\cc @mgsloan @borsboom


etc/mkdocsjs/searchhack.js
```diff
@@ -0,0 +1,29 @@
+// Please see https://github.com/commercialhaskell/stack/issues/3376
+
+(function () {
```
mgsloan: Haven't tested it, but it looks to me like this function won't be run, needs another () on the last line.

borsboom:
@lwm I've copied your `fixup-search` branch to the main repo, and told readthedocs to build it.  Unfortunately, all documentation builds seem to be failing for the last day with "An unexpected error occurred", so we can't actually test it yet.  You should be able to push changes to the `fixup-search` branch yourself too.

etc/mkdocsjs/searchhack.js
lwm: Ah yes, you're right. I should have listened to my JavaScript linter ;)

lwm:
> all documentation builds seem to be failing for the last day with "An unexpected error occurred"

Hmm, seems like the RTD build needs some attention then. Any stack trace? 

Feel like adding me to the RTD account so I can play around with it or?

borsboom:
RTD seems to be failing most builds, but occasionally passing.  I've made `lwm` a maintainer there.

lwm:
Ugh, seems to be a bunch of these 'unexpected error' messages plaguing people, as far as I can see in https://github.com/rtfd/readthedocs.org/issues/2985. I'll keep an eye there but can't seem to get any good of it now.

lwm:
I'll close this off for now and try to fix when things are clear on the RTD side.

PR #3384, opened 2017-08-24
pretty up `src/Stack/Package.hs`

kadoban:
Use `$pretty*` functions where appropriate in `src/Stack/Package.hs`.

Related to #2650

b85364d is probably the commit most in need of review, had to do some things I barely understood to get that working.

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -70,6 +70,7 @@ import qualified Distribution.Types.LegacyExeDependency as Cabal
 import qualified Distribution.Types.UnqualComponentName as Cabal
 import qualified Distribution.Verbosity as D
 import           Distribution.Version (showVersion)
+import           Lens.Micro (lens)
 import qualified Hpack
 import qualified Hpack.Config as Hpack
 import           Path as FL
@@ -96,6 +97,22 @@ import qualified System.Directory as D
 import           System.FilePath (splitExtensions, replaceExtension)
 import qualified System.FilePath as FilePath
 import           System.IO.Error
+
+data Ctx = Ctx { ctxFile :: !(Path Abs File)
+               , ctxDir :: !(Path Abs Dir)
+               , ctxEnvConfig :: !EnvConfig
+               }
+
+instance HasPlatform Ctx
+instance HasGHCVariant Ctx
+instance HasLogFunc Ctx where
+    logFuncL = configL.logFuncL
+instance HasRunner Ctx where
+    runnerL = configL.runnerL
+instance HasConfig Ctx
+instance HasBuildConfig Ctx
+instance HasEnvConfig Ctx where
+    envConfigL = lens ctxEnvConfig (\x y -> x { ctxEnvConfig = y })
 
 -- | Read the raw, unresolved package information.
 readPackageUnresolved :: (MonadIO m, MonadThrow m)
@@ -269,10 +286,11 @@ packageFromPackageDescription packageConfig pkgFlags pkg =
         \cabalfp -> debugBracket ("getPackageFiles" <+> display cabalfp) $ do
              let pkgDir = parent cabalfp
              distDir <- distDirFromDir pkgDir
+             env <- view envConfigL
              (componentModules,componentFiles,dataFiles',warnings) <-
                  runReaderT
                      (packageDescModulesAndFiles pkg)
-                     (cabalfp, buildDir distDir)
+                     (Ctx cabalfp (buildDir distDir) env)
              setupFiles <-
                  if buildType pkg `elem` [Nothing, Just Custom]
                  then do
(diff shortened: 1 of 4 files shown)


kadoban:
Hmm, somehow I though there was more to do in that module. This might be less WIP than I thought. I still haven't really reviewed my work though, and there's a few TODOs I added that I can maybe resolve if I look around a bit. Will try to do that tomorrow.

src/Stack/Package.hs
```diff
@@ -1108,43 +1141,45 @@ findCandidate dirs exts name = do
...
+        [ flow "Unable to find a known candidate for the Cabal entry"
+        , (styleModule . fromString $ D.display mn) <> ","
+        , flow "but did find:"
+        , line <> bulletedList (map dispPoss possibilities)
+        , flow "If you are using a custom preprocessor for this module"
+        , flow "with its own file extension, consider adding the file(s)"
+        , flow "to your .cabal under extra-source-files."
+        ]
   where
+    dispPoss = styleFile . fromString . toFilePath
```
mgsloan: I think `display` can be used instead because there is an instance for `Path t File`.

mgsloan:
LGTM!  Just one comment to address.  I like the `Ctx` refactoring.

src/Stack/Package.hs
kadoban: Nice catch, thanks!

kadoban:
Did that one change, nice catch.

Cool, thanks. Was hoping that `Ctx` part made sense. I might actually be getting a handle on what those `Has*` typeclasses mean, heh. They seem quite nice in practice. Need to learn more `lens` sometime still though.

src/Stack/Package.hs
```diff
@@ -1108,42 +1141,45 @@ findCandidate dirs exts name = do
...
+        , fromString . showName $ name
+        , line <> bulletedList (map dispOne rest)
+        , line <> flow "picking:"
+        , dispOne candidate
+        ]
+  where showName (DotCabalModule name') = D.display name'
+        showName (DotCabalMain fp) = fp
+        showName (DotCabalFile fp) = fp
+        showName (DotCabalCFile fp) = fp
+        dispOne = fromString . toFilePath
```
kadoban: This is unfortunate, I can't figure out exactly why this can't be `display` instead of `dispOne` in this function. I get the error:

```
    • Couldn't match type ‘Ann (Path b t)’ with ‘AnsiAnn’
      Expected type: Path b t -> AnsiDoc
        Actual type: Path b t -> Doc (Ann (Path b t))
    • In the first argument of ‘map’, namely ‘display’
      In the first argument of ‘bulletedList’, namely
        ‘(map display rest)’
      In the second argument of ‘(<>)’, namely
        ‘bulletedList (map display rest)’
    • Relevant bindings include
        rest :: [Path b t] (bound at src/Stack/Package.hs:1146:29)
        candidate :: Path b t (bound at src/Stack/Package.hs:1146:19)
        warnMultiple :: DotCabalDescriptor
                        -> Path b t -> [Path b t] -> m ()
          (bound at src/Stack/Package.hs:1146:1)
```

when I try.

kadoban: Must be something about `Path b t` being polymorphic? A little confused.

mgsloan: Yeah, it's because it's polymorphic.  I think adding `Display (Path b t)` to the function's constraints should solve the problem.  Alternatively, just make it monomorpic, I think it's just used for `(Path Abs File)` or something.

mgsloan:
LGTM.  Nice idea to use hlint for this!  Though, I think there could be cases where `fromString . toFilePath` is used to create a type other than `AnsiDoc`.  Not a big deal though, lets see if those cases ever crop up.

src/Stack/Package.hs
kadoban: Hmm, thanks for looking at it. I thought I tried that constraint and got the same error. Very possible I screwed it up though. I'll investigate more.

kadoban:
> LGTM. Nice idea to use hlint for this! Though, I think there could be cases where fromString . toFilePath is used to create a type other than AnsiDoc. Not a big deal though, lets see if those cases ever crop up.

Thanks! Yeah, hopefully not many. If there's only a few we can just add exceptions for them. Can definitely re-evaluate if they pop up all over.

> Will need to be updated due to changes merged in #3389

Hopefully just did that, crossing fingers for the CI builds.

mgsloan:
Generalizing to something like `(HasCallStack, HasRunner env, MonadReader env m, MonadLogger m)` should do the trick

kadoban:
Yep, that seemed to do the trick, thanks. Now to investigate that `Display (Path b t)` thing while waiting for CI.

mgsloan:
Cool, looks like CI is passing, feel free to merge this.

kadoban:
That works, I'll just come back to the `Display` thing for `Path b t` later, there's a TODO in there at least. can't seem to figure that one out though, heh.

On to finishing up terminal detection I hope :)

PR #3392, opened 2017-08-29
Unbuildable test stanzas don't cause problems

### LINKED ISSUE #3390: On master: buildable: False test suites break stack build
Repro steps:

1. Start a new project (e.g. `stack new foo`)
2. Set the test suite to `buildable: False`
3. Run `stack test`

Expected: test suite doesn't build or run

Actual: Get the error message:

```
2017-08-28 19:03:16.833002: [warn] Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2: Cannot build the test suite
@(Stack/Build/Execute.hs:1099:67)
2017-08-28 19:03:16.833218: [warn] 'foo-test' because the component is marked as disabled in the .cabal file.
```

The reason is obvious, the command run is `Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2 --builddir=.stack-work/dist/x86_64-osx/Cabal-1.24.2.0 build lib:foo test:foo-test --ghc-options " -ddump-hi -ddump-to-file"`. This is almost certainly caused by a bug in my Cabal 2.0 upgrade. There were some hairy issues around the buildable stuff due to changes in how Cabal works.

snoyberg:
Included a new test case for this. It would be really nice to clean up the package types at some point. Fixes #3390.

### DIFF
--- a/src/Stack/BuildPlan.hs
+++ b/src/Stack/BuildPlan.hs
@@ -208,7 +208,9 @@ gpdPackageDeps gpd cv platform flags =
     Map.filterWithKey (const . (/= name)) (packageDependencies pkgDesc)
     where
         name = gpdPackageName gpd
-        pkgDesc = resolvePackageDescription pkgConfig gpd
+        -- Since tests and benchmarks are both enabled, doesn't matter
+        -- if we choose modified or unmodified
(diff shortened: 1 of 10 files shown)


mgsloan:
LGTM, merging

PR #3395, opened 2017-08-30
detect terminal width and use it for pretty-printed output

kadoban:
Related to #2650, detect the width of the terminal and use it for proper output.

* [x] The documentation has been updated
* [x] Test terminal detection more, ideally on several OSes
* [x] Use detected width when pretty-printing
* [x] Allow user configuration of widths (hardcode a width, set min and max?, disable detection?)
* [x] Make sure this doesn't screw up log files too badly
* [x] Figure out why hsc2hs isn't available on appveyor, or if that's just a CI problem

Note: this current implementation is only ever going to work on POSIX. I don't currently have ability to dev/test on windows, though it appears that there *are* APIs that could give this info if anyone else wants to sub that in.

**Testing:** It seems to work fine on all of the OSes I have. Unfortunately, that's a pretty unrepresentative set of OSes. I mostly have linux variants and one BSD I tried on and it worked fine. I don't have access to OSX or Windows really.

If anyone can, testing on other OSes would really be good. Windows won't do any terminal detection, but the command-line option should work.

Log output doesn't seem to be messed up by this. If anyone can verify that'd be nice though. I don't use logs really ever, so I'm hoping I looked the right places.

**Docs:** I couldn't find any places the docs actually need to be updated. The ```stack --help``` output is correct already obviously.

### DIFF
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -28,6 +28,8 @@ module Stack.Constants
     ,platformVariantEnvVar
     ,compilerOptionsCabalFlag
     ,ghcColorForceFlag
+    ,minTerminalWidth
+    ,defaultTerminalWidth
     )
     where
 
@@ -216,3 +218,9 @@ compilerOptionsCabalFlag Ghcjs = "--ghcjs-options"
 
 ghcColorForceFlag :: String
 ghcColorForceFlag = "-fdiagnostics-color=always"
+
+minTerminalWidth :: Int
+minTerminalWidth = 20
+
+defaultTerminalWidth :: Int
+defaultTerminalWidth = 100
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -221,7 +221,7 @@ ghcColorForceFlag :: String
 ghcColorForceFlag = "-fdiagnostics-color=always"
 
 minTerminalWidth :: Int
-minTerminalWidth = 20
+minTerminalWidth = 40
 
 maxTerminalWidth :: Int
 maxTerminalWidth = 200
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -29,6 +29,7 @@ module Stack.Constants
     ,compilerOptionsCabalFlag
     ,ghcColorForceFlag
     ,minTerminalWidth
+    ,maxTerminalWidth
     ,defaultTerminalWidth
     )
     where
@@ -221,6 +222,9 @@ ghcColorForceFlag = "-fdiagnostics-color=always"
 
 minTerminalWidth :: Int
 minTerminalWidth = 20
+
+maxTerminalWidth :: Int
+maxTerminalWidth = 200
 
 defaultTerminalWidth :: Int
 defaultTerminalWidth = 100
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -217,14 +217,20 @@ compilerOptionsCabalFlag :: WhichCompiler -> String
 compilerOptionsCabalFlag Ghc = "--ghc-options"
 compilerOptionsCabalFlag Ghcjs = "--ghcjs-options"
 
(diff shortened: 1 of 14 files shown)


kadoban:
So far I just have the raw terminal width detection in. It seems to work fine but needs more testing, and obviously I need to put in the parts that actually use it, and some configurability as well.

kadoban:
Hmm, is that hsc2hs error on appveyor a CI problem or an actual problem, anyone know offhand?

mgsloan:
LGTM sofar, though I am not really all that experienced with FFI

kadoban:
Thanks! Yeah, this is my first time doing C FFI (done a tad of JS FFI), so there's a lot of finger crossing going on. But it seems pretty painless so far surprisingly.

kadoban:
This basically seems to be working, yay. Need to clean up and test a decent amount still though.

src/Stack/Constants.hs
```diff
@@ -216,3 +218,9 @@ compilerOptionsCabalFlag Ghcjs = "--ghcjs-options"
 
 ghcColorForceFlag :: String
 ghcColorForceFlag = "-fdiagnostics-color=always"
+
+minTerminalWidth :: Int
+minTerminalWidth = 20
```
mgsloan: 20 is **really** small.  I think 40 might be a better minimum.  Reasoning is that the user might resize their terminal.

I think having a max makes sense too, it's unpleasant to read a whole paragraph of text stretched on a line.  Maybe 200 chars?  Of course, an answer to that would be why doesn't the user resize their terminal.  I use XMonad, though, so if I open up a terminal on a 4k display, I probably don't really want paragraphs to be single lines, though that does happen.

stack.cabal
```diff
@@ -268,6 +268,7 @@ library
                    , store-core >= 0.4 && < 0.5
                    , annotated-wl-pprint
                    , file-embed >= 0.0.10
+  build-tools:       hsc2hs >= 0.68
```
kadoban: Is this even correct? It doesn't seem to help the appveyor problem, but maybe it's supposed to be there anyway?

I've never really used `build-tools` before, and I also can't tell if `hsc2hs` is part of GHC and thus doesn't need to be listed here or what.

src/Stack/Constants.hs
kadoban: Yeah I wasn't sure what to do for a min. 40 does look more reasonable. I'm indifferent to a maximum personally, but that reasoning seems to make sense, I'll throw it in.

Hm, I just noticed that I'm doing the checking all wrong too, it's only checking the user-supplied one, and it's just ignoring it completely if it's too small. Should probably just clip to the allowed range if it's outside it. I'll redo that part, heh.

stack.cabal
mgsloan: Hmm, could be that hsc2hs needs to be added to stackage.  For now, adding `stack install hsc2hs` to the travis script should fix the issue.  Adding hsc2hs to extra-deps might help, as it would specify a version to install, not 100% sure if that works, it used to not, but that may have been fixed.

kadoban: Oh, nice catch, it isn't in stackage. Appveyor seems to be the only one that doesn't like it. Huh, I'm getting other travis CI errors that look like actual code problems. Wacky that I'm not getting it locally though, I'll have to look into that.

kadoban:
Once this goes in, I'll make an issue in the tracker for Windows not doing terminal detection, just as a reminder. I figure someone on windows will care enough eventually.

lwm:
Is this ready to merge then? Serious effort @kadoban!

mgsloan:
This looks good to me, is it ready for merge @kadoban ?

kadoban:
Thanks! Yep, as far as I know it's good to go.

I'm a bit uncertain about other OSes still, like OSX (no way to test that), but I'm as close to sure it should work as I can be.

mgsloan:
I bet it will work fine on osx.  Merging!

PR #3402, opened 2017-08-31
Fix docker comment

TheWizardTower:
So, previously, if you ran

```shell
stack --docker-help
```

you'd see this:

```
  --[no-]docker            Enable/disable using a Docker container. Implies
                           'system-ghc: true'
```

Well, that's not strictly true! --no-docker does not imply system-ghc: true. Only --docker does. 

Looking at the code, it's super-easy to see how this happened, but I'm inclined towards pedantry and I want my documentation to be super-clear. So, PR!

* [*] Any changes that could be relevant to users have been recorded in the ChangeLog.md
Done, in this PR
* [*] The documentation has been updated, if necessary.
This is a documentation update, of a sort. I don't know of any others that need to be polished. The --nix-help page in stack needs similar treatment, though.

I built it, cd'd to the dir, and ran ./stack --docker-help. It did what I expected.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -90,6 +90,8 @@ Other enhancements:
 
 Bug fixes:
 
+* `stack --docker-help` is now clearer about --docker implying
+   system-ghc: true, rather than both --docker and --no-docker.
 * `stack haddock` now includes package names for all modules in the
    Haddock index page. See:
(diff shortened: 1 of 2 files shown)


mgsloan:
Thanks!

Issue #3400, opened 2017-08-31
Invoking stack in parallel fails when using built-in docker integration


SimSaladin:
In our Gitlab CI pipelines (internal, can't provide much details), we are invoking stack many times simultaneusly (build with different params, `stack path` etc.). And we are using the built-in docker integration (all commands thus start `stack --docker`. (We also have two different images we use, also simultaneusly, but that probably isn't related. Haven't tested that much.)

In this scenario, some `stack` commands fail with
```
SQLite3 returned ErrorBusy while attempting to perform step
```
Without knowing much about stack internals I suspect  `$STACK_ROOT/.stack/docker.db` that it is failing to interact with when run in parallel.

This is annoying because as it causes pipelines to fail randomly.

Yes this looks like it's more a limitation resulting from using sqlite for tracking docker container state (or whatever that is used for). But maybe there's a way around it?

### Steps to reproduce

Invoke multiple `stack --docker` commands simultaneusly, some of them should fail.

### Expected

I expected stack to work just fine with parallel execution even when using docker integration.

### Actual

Some (usually, just one) stack commands fail immediately with

`SQLite3 returned ErrorBusy while attempting to perform step.`

### Stack version
```
$ stack --version
Version 1.5.1, Git revision 600c1f01435a10d127938709556c1682ecfd694e (4861 commits) x86_64 hpack-0.17.1
```

### Method of installation

`stack upgrade`

mgsloan:
Thanks for the report!  I have also seen this on occasion, but it never really caused me much issues.  I agree it should be fixed.

SimSaladin:
That's good to hear. I could take a chance at implementing a fix, but I can only think of hackish ways around. Then again I'm not familiar with the code base so don't know what to expect. Any nice ideas?

mgsloan:
@SimSaladin Hey, that'd be great!  I believe the relevant code is in `Stack.Docker.GlobalDB`.  I think it makes sense to have a dedicated lockfile for the sql database.  For this, use `System.FileLock`.  There are already some examples of its use in the codebase.

PR #3405, opened 2017-09-01
Fix parallel invocation of `stack --docker` (#3400)

SimSaladin:

See issue #3400 for background.

Thanks to @mgsloan for your help on how to go about this! Turned out the fix was actually quite trivial.

A patched `stack` is already live in our CI. I will report findings here later.


lwm:
> A patched stack is already live in our CI. I will report findings here later.

:cake:

mgsloan:
That looks great, thanks!  Curious to here if it works well, but this looks like the right thing to me so merging.

I also opened https://github.com/commercialhaskell/stack/issues/3407 .  Not an issue with this in particular, just the use of lockfiles in general.

SimSaladin:
As expected, the pipelines run that much more smoother now, no issues since!

Issue #3410, opened 2017-09-03
Docs on master should warn that they apply to master


snoyberg:
A number of people have contacted me about confusion popping up on IRC (or even this issue tracker), regarding "out of date" documentation. In reality, the situation is that the docs on the master branch apply to the code on the master branch, which may be different behavior from what exists in the latest released version. The changes to package syntax for extensible snapshots seems to be a significant sticking point.

I've noticed that, as an example, the Kubernetes project puts a big warning at the top of their source files that explain that the docs apply to master, and point to the official documentation site.

@borsboom Could we do something similar for Stack? CC @bitemyapp

borsboom:
I'm trying to think of the best way to achieve this.  I don't know how you'd go about having some warning text in the _markdown source files_ only in one branch without that making merges really annoying (since you'd always have to be careful to add or strip the warning text).  Trying to automate something like that with git hooks seems dicey.

Another would be to have the warning in all branches, but have the tool that renders the HTML strip it out for all except the master branch.  I looked and don't see a way to get mkdocs or readthedocs to do that (even for all branches).

What might work is using CSS to hide the warning in the rendered HTML.  We can specify extra CSS stylesheets with mkdocs, so we could use a `hidden-warning` class around the warning so that it's hidden on the documentation site, but would still show up when Github renders the markdown (since Github wouldn't have that class in its CSS).  I don't see a way to override the CSS for different branches, though, so the warning would be hidden in all branches.  That's probably alright since the documentation site defaults to the `stable` branch (which corresponds to the latest released version).

What should the warning say?  Maybe something like:

> **This is the source code for the Stack documentation, and may not be correct for the released version of Stack.  Visit [docs.haskellstack.org](https://docs.haskellstack.org) to browse the documentation for the latest release.**

borsboom:
PR here: https://github.com/commercialhaskell/stack/pull/3411

bitemyapp:
I'd state here that a big warning is good, but also say that perhaps the master docs should be hidden so people stop stumbling into them. My experience has been that, uh, people don't read. Even those that do, they might not make the following leaps:

"these are master docs"

"I am running 1.4.1" (or whatever)

"1.4.1 is not master"

"these docs do not apply to me"

So you might want to add extra verbiage about the docs not being applicable unless you compiled Stack from the git repository. But I really think hiding them somehow would help a lot too.

borsboom:
[...]

PR #3411, opened 2017-09-03
Add 'not for released version' warning to docs source (#3410)

borsboom:

Unfortunately, GFM strips out many tags so I couldn't make the warning as prominent as I would've liked (e.g. bright background colour), and I didn't want to resort to embedding an image.  I also couldn't test on readthedocs.org itself since its documentation building seems to be completely broken right now (every build fails with "An unexpected error occurred").  But I did test using a local `mkdocs build` and the warning is indeed hidden, so once readthedocs.org is working again it _should_ be hidden there as well.

* [ ] Before merging to master, adjust the warning SVG URLs to use the `master` branch rather than this (`3410-docs-warning`) branch which will cease to be a valid URL after it's deleted.


borsboom:
I've made the warning more prominent by using an SVG image.

PR #3415, opened 2017-09-04
Fix readthedocs search (fixes #3376)

### LINKED ISSUE #3376: readthedocs.org search is broken

borsboom:

Tested here, searching for 'yaml' returns appropriate results: https://docs.haskellstack.org/en/3376-fix-search/README/

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -122,6 +122,10 @@ Bug fixes:
   trust the etag and not re-download.  Fixed in this version.
 * Invoking `stack --docker` in parallel now correctly locks the sqlite database.
   See [#3400](https://github.com/commercialhaskell/stack/issues/3400).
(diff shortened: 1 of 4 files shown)


lwm:
LGTM! My [test search](https://docs.haskellstack.org/en/3376-fix-search/search.html?q=extra-deps) returns the goods so I think we're set on this one. Great to get this fixed.

PR #3433, opened 2017-09-14
When using Nix, depend on gcc.

shlevy:
Fixes builds using the FFI, with no additional dependency footprint.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -93,6 +93,9 @@ Other enhancements:
   [#3360](https://github.com/commercialhaskell/stack/issues/3360)
 * Better descriptions are now available for `stack upgrade --help`. See:
   [#3070](https://github.com/commercialhaskell/stack/issues/3070)
+* When using Nix, nix-shell now depends always on gcc to prevent build errors
+  when using the FFI. As ghc depends on gcc anyway, this doesn't increase the
+  dependency footprint.
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


shlevy:
Waiting on @sellout to test

sellout:
@shlevy Tested – 👍🏾

lwm:
@shlevy, thanks for contributing a fix! You mentioned that this fixes some builds. Is there some logs/explanation from the builds that were broken? Probably someone with more Stack/Nix experience can merge this but just checking anyway 👍

mgsloan:
Thanks!

acowley:
This doesn't seem like the right default for folks on clang (e.g. darwin). At least, I never pull in `gcc`, but perhaps I've just been lucky to never trip into this problem.

shlevy:
@sellout Do you remember what caused this issue? Were you on a mac?

sellout:
Yeah, this happened to me on a Mac.

However, I just tried to reproduce the old error, and I couldn’t – I removed `gcc` from the dependencies, got rid of stack-work, created a new `nix-shell --pure` with only git, nix, and stack in it, and built everything fine.

I’m on a new laptop since this issue was opened, though, and have switched to nixpkgs-darwin from nixpkgs-unstable … and probably a few other variations along the way. So I don’t know. This was certainly affecting me three months ago.

acowley:
@sellout Do you recall any details of what sorts of FFI things you needed `gcc` for? The `ghc` nixpkgs provides on Darwin has been set to use `clang` as the C compiler for quite some time. We should try to document when `gcc` is needed for future reference, as I imagine it could be tricky to recognize the problem.

It’s happened before that CPP behavior diverged between `gcc` and `clang` in the past, leaving `ghc` in a bit of a lurch. So something like that with a particular header could plausibly arise again.

PR #3434, opened 2017-09-15
Stack installer script is not detected latest `redhatenterpriseserver` linux version

OvermindDL1:
The current install script at `https://get.haskellstack.org/` is attempting to detect for `rhel` when the latest RedHate Enterprise is detected as `redhatenterpriseserver`, thus the script is falling back to the `sloppy` install.

Adding `|redhatenterpriseserver` as an option after `rhel` fixes the issue and install fully completes properly and successfully with all dependent packages properly installed.


lwm:
Thanks!

Issue #3382, opened 2017-08-23
Only run hpack once per package.yaml


mgsloan:
Currently, hpack gets invoked a fair number of times, I count 6 times for a `stack build` on a trivial project.  This is usually pretty quick, but I'm not sure how smart it is about avoiding work.  This is particularly important if there's an option to use the `hpack` executable - https://github.com/commercialhaskell/stack/issues/3179 .

snoyberg:
I'm fairly certain this is no longer the case, please reopen if you're still seeing it with the latest `master`.

PR #3443, opened 2017-09-19
Add hpack command line option

### LINKED ISSUE #3179: New feature: add hpack command-line option
Proposal: allow one to give a path to an `hpack` executable rather than using the `hack` bundled with `stack`.

Suggestion: include a command line option such as `--hpack=<PATH_TO_HPACK>`

Use case: we ran into a couple issues with `hpack`, one of which is fixed on `hpack/master` and another which is fixed in a PR that is not yet merged. That means we have to wait for both `hpack` to accept the PR change, as well as wait for the next `stack` release that bundles those `hpack` changes. That is a double delay for release cycles. Allowing a custom `hpack` gives us the freedom to change `hpack` for our custom needs now and not have to wait for them to be released (which could be months away).

I may look into code changes according to this proposal. I am opening this issue to get feedback on the idea, as well as to double check that I didn't miss an existing option or workaround for this.

Thank you for your consideration!

scott-fleischman:
Resolves #3179.

I tested it with the `--with-hpack` option and without, and it works as I intended. I couldn't think of a good way to test this in code, so I did not add any tests to the test suite.

### Details
The core change is pretty basic.
* I added an `HpackExecutable` type with two cases--one for the bundled hpack, and one with the hpack command.
* I added `HpackExecutable` as a field to the `GlobalOpts` and `BuildConfig` types.
* I used `runCmd` to execute the command, which requires an `EnvOverride` to create a `Cmd` value.
* Then the remaining effort was in propagating the necessary information to the `hpack` function. 
  * When `HasEnvConfig` was available, I used the `hpackExecutableL` lens to retrieve the `HpackExecutable`
  * Else I passed `HpackExecutable` and `EnvOverride` as needed as function arguments.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -51,6 +51,9 @@ Behavior changes:
 
 Other enhancements:
 
+* The `with-hpack` configuration option specifies an Hpack executable to use
+  instead of the Hpack bundled with Stack. Please
+  see [#3179](https://github.com/commercialhaskell/stack/issues/3179).
 * It's now possible to skip tests and benchmarks using `--skip`
   flag
 * `GitSHA1` is now `StaticSHA256` and is implemented using the `StaticSize 64 ByteString` for improved performance.
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -528,3 +528,4 @@ Yes:
 
 * If a package directory contains an Hpack `package.yaml` file, then Stack will use it to generate a `.cabal` file when building the package.
 * You can run `stack init` to initialize a `stack.yaml` file regardless of whether your packages are declared with `.cabal` files or with Hpack `package.yaml` files.
+* You can use the `with-hpack` configuration option to specify an Hpack executable to use instead of the Hpack bundled with Stack.
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -460,6 +460,14 @@ Specify a path to gcc explicitly, rather than relying on the normal path resolut
 with-gcc: /usr/local/bin/gcc-5
 ```
 
+### with-hpack
+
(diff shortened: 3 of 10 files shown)


mgsloan:
LGTM!  I think it probably makes sense to also support this as a configuration option, similarly to `with-gcc:`.  This way you can just stick it in your `config.yaml` to easily change hpack version globally.  Do you want to add that to this PR?  If not, that's fine, this looks great to me otherwise, ready to merge.

scott-fleischman:
@mgsloan I updated the PR based on your suggestion. And thanks too! That radically simplifies the code, since all code paths to `hpack` had `HasConfig env`, so I was able to avoid all of the extra argument passing. I was able to add a couple simple tests to verify that the configuration option is parsed correctly.

One question is whether it's ok that the hpack output is simply inherited. A typical run looks like:
```
$ stack build --with-hpack=hpack
generated testhp.cabal
testhp.cabal is up-to-date
testhp.cabal is up-to-date
testhp.cabal is up-to-date
testhp.cabal is up-to-date
testhp.cabal is up-to-date
testhp-0.1.0.0: configure (exe)
Configuring testhp-0.1.0.0...
testhp-0.1.0.0: build (exe)
Preprocessing executable 'testhp' for testhp-0.1.0.0...
testhp-0.1.0.0: copy/register
...
```

Also, it's surprising that the `hpack` function gets called 6 times as part of a `stack build`! And it apparently gets called twice as part of a `stack clean`. Is that expected?

scott-fleischman:
Ok I see that the multiple hpack invocations are addressed in https://github.com/commercialhaskell/stack/issues/3382

mgsloan:
LGTM, thanks!

Issue #3424, opened 2017-09-08
local file changes: /home/user/.stack/..../ghcversion.h


phadej:
Stack current `master` seems to track out-of-root files, and somehow believes them changed all the time.
Maybe there's something wrong with my machine?

### Steps to reproduce

- Build a project with two packages, `foo` and `bar`, where `foo` uses CPP (no `TemplateHaskell`), and `bar` depends on `foo`
- change something in `bar`.

### Expected

Only package I'm working on is rebuild

### Actual

Everything is rebuild, for example:

```
ansi-pretty-0.1.2.1: unregistering (local file changes: /home/ogre/.stack/programs/x86_64-linux/ghc-8.0.2/lib/ghc-8.0.2/include/ghcversion.h /usr/include...)
...
````
### Stack version

```
$ stack --version
Version 1.5.1, Git revision 2b0392468a03174a7c5d317291c62452d771b055 x86_64 hpack-0.18.1
```

phadej:
I wiped `~/.stack`, and did `stack setup`, `stack build`, and another `stack build`. Problem persists.

Can anyone point where dependencies are gathered, and how they are checked. I propose to
- omit dependencies out of the project root (IMHO `stack clean` when env changes is ok, compared to gained simplifactions: `stack` doesn't track anything outside the project directory)
- investigate why stack thinks taht `ghcversion.h` is dirty.

jonathanknowles:
I'm seeing the same problem. On successive builds of an already-built tree, stack appears to claim that different sets of files have changed, and thus attempts to rebuild those components.

mgsloan:
I just ran into this too.  Gnarly, still not sure how this happened!  Potential fix in https://github.com/commercialhaskell/stack/pull/3456

mgsloan:
@phadej @jonathanknowles Can you please verify that this is fixed on master?  Thanks!

I think it is, so closing this.

PR #3456, opened 2017-09-26
Fix computation of new build-cache #3424

mgsloan:
Pinging @snoyberg I think the behavior here changed during the extensible snapshots stuff.  Curious if you have any idea why this bug only cropped up now, even though the involved code hasn't been changed in quite a long time.

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md (unnecessary because it's addressing a new breakage not present in last release)


mgsloan:
Also including a change to run `checkForUnlistedFiles` for non-target packages.  This should be done, because `checkForUnlistedFiles` actually has the side effect of updating the build cache.

It might make sense to omit these warnings for extra deps, with the idea being that the user probably doesn't care about these warnings for packages they didn't author.

snoyberg:
I don't have insight into what's going on here, sorry. I don't see a problem with the code change.

Issue #3432, opened 2017-09-14
Licensed versions are outputted with Show'ed version


phadej:
```
machines-directory Apache (Just (mkVersion [2,0]))
```

This should be easy to fix!

PR #3457, opened 2017-09-26
Fixes automatically shown versions for licenses #3432

khanage:
Cabal has a method to render this, so we'll just call it directly.

This was just tested manually by adding the package `machines-directory` and checking the output of `stack list-dependencies --license | grep machines`. This renders properly now.

I can some tests if people think it's a good idea, I haven't because it's really just a call to an upstream library, and the code itself lives in `IO` land (which means a refactor).


mgsloan:
Great, thanks!

PR #3474, opened 2017-10-05
Fix STACK_IN_NIX_SHELL environment variable name

wfranzini:
The correct name has one more '_':

https://github.com/NixOS/nixpkgs/blob/72f530ba3310b8304b148b74ece9e6e75309666b/pkgs/development/haskell-modules/generic-stack-builder.nix#L20

### DIFF
--- a/src/Stack/Constants.hs
+++ b/src/Stack/Constants.hs
@@ -94,7 +94,7 @@ inContainerEnvVar = stackProgNameUpper ++ "_IN_CONTAINER"
 -- although we already have STACK_IN_NIX_EXTRA_ARGS that is set in the same conditions,
 -- it can happen that STACK_IN_NIX_EXTRA_ARGS is set to empty.
 inNixShellEnvVar :: String
-inNixShellEnvVar = map toUpper stackProgName ++ "_IN_NIXSHELL"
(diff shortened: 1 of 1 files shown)


mattaudesse:
Thanks @wfranzini!

mgsloan:
Hmm, perhaps instead generic-stack-builder.nix is wrong?  This seems like breaking backwards compatibility for this env var name for no reason.  Why not fix the script?

wfranzini:
@mgsloan I assumed it's wrong in stack from the start, since STACK_IN_NIX_EXTRA_ARGS separates words using underscores and it's written in the same way in nixpkgs.

I'm not sure about backward compatibility since nix-shell exported STACK_IN_NIX_SHELL and stack looks for STACK_IN_NIXSHELL.  So probably it never worked as intended.

mgsloan:
Hmm, I suppose so.  I can't find internet references to the old name, but someone might be relying on it.  Not sure how likely that is.  Inclined to revert this and instead modify generic-stack-builder

mgsloan:
Then again, this does seem like a more conventional name and it's already merged.  Lets just go with it

Issue #3431, opened 2017-09-13
Extensible snapshots: snapshot packages not shared between snapshots, nor fully between packages


dbaynard:
Stack no longer shares packages built using different snapshots.

Even with the same resolver, git packages are re-downloaded, too — though not rebuilt! — and in the local package directory, not the global snapshot directory. The snapshot packages are shared.

(Incidentally, I wonder if there's an issue I should follow concerning extensible snapshots. It's a feature I've been anticipating for a while, and I very much liked some of the previous behaviour — in particular the package sharing).

### Steps to reproduce

1.  First, try to install pandoc using the global-project.

    Use `resolver: https://gist.github.com/dbaynard/9693107a883a6282f52f22785d0072e2/raw/cbbfa7251704add22d145dc14da9b278fdac2c77` in stack.yaml

    `stack install pandoc` (though it probably works for any package; the point is, this is a downstream package)

2.  Change to `resolver: https://gist.github.com/dbaynard/9693107a883a6282f52f22785d0072e2/raw/eeb1f7a905e06bb22c641f30923a91f96bc4ead6` which has one change (that of pandoc).

    `stack install pandoc`

3.  Then, enter a directory containing a stack.yaml pointing to the latter resolver, and no .stack-work

    `stack build --dependencies-only`

### Expected

1.  Stack builds whatever dependencies it needs, then builds pandoc.
2.  Stack recognises it has already built all the dependencies and then begins downloading and building pandoc.
3.  Stack recognises it has already built some of the dependencies and then gives messages about reusing existing packages. It builds the remaining dependencies.

### Actual

1.  Stack builds all the dependencies (then fails to build pandoc; this is not the issue here)
2.  Stack begins rebuilding all the same dependencies it built the first time.
3.  Stack downloads the snapshot packages provided as git commits, into the local .stack-work directory, but does not rebuild them. It then builds the remaining dependencies.

### Stack version

```
$ stack --version
Version 1.5.1, Git revision 2b0392468a03174a7c5d317291c62452d771b055 (5182 commits) x86_64 hpack-0.18.1
```

Configuration just sets up some personal information for new projects, and allows newer packages and system ghc.

### Method of installation

Bootstrapped the git repo with `stack install stack` from the most recent stable version.

dbaynard:
Workaround for downloading the git repos: manually rsync the `.stack-work/downloaded` directory.

lwm:
[...]

PR #3494, opened 2017-10-17
Do correct path checking in precompiled cache.

snoyberg:
The commit ed9ccc08f327bad68dd2d09a1851ce0d055c0422 changed the behavior
of the precompiled cache to store relative instead of absolute paths.
That seems to have broken the ability to properly check if a library
exists, thereby defeating precompiled caching.

The issue that discovered this is #3431, and ascribed this to extensible
snapshots. I _think_, though I'm not 100% certain, that this bug reaches
much farther than extensible snapshots, and in fact breaks all
precompiled caches.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -148,6 +148,8 @@ Bug fixes:
 * `stack build --only-dependencies` no longer builds local project packages
   that are depended on. See
   [#3476](https://github.com/commercialhaskell/stack/issues/3476).
+* Properly handle relative paths stored in the precompiled cache files. See
+  [#3431](https://github.com/commercialhaskell/stack/issues/3431).
 
 
 ## 1.5.1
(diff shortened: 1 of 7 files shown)


mgsloan:
LGTM

snoyberg:
Thanks! I've added an integration test for this as well. Once I get the green from Travis, I'll merge. (Or if someone gets to it before me, that's great too.)

dbaynard:
One and a half laptops later, I can confirm this fixes the sharing issue in #3431.

dbaynard:
I'm seeing a regression on this — including on `master`. I'm going to check the minimum example later today.

dbaynard:
Yeah, integration test failing on master and 8.4.3 (the test needs stm updating to 2.4.5.0 and then it fails properly)

PR #3495, opened 2017-10-17
Revert "Clean up --only-dependencies logic in ConstructPlan"

### LINKED ISSUE #3476: Stack builds packages when called with `--only-dependencies` in some multi-package projects
### General summary/comments
If you have a project with packages `a` and `b`, and `b` depends on `a`, `stack build --only-dependencies` will build `a`.

Bisecting, this change was introduced in https://github.com/commercialhaskell/stack/commit/d1e6c480cd46def34bfacd3aeaee953ba32c7758. I'm not sure if it was introduced intentionally or not. If it's unclear why some might prefer this behaviours absence, consider building pedantically. You don't want to to be responsible for warnings in all your dependencies, so would rather build deps first unpedantically, and then be able to build all your own packages pedantically.

### Steps to reproduce

1. Make a stack project with two packages in sub-directories.
2. Have one package depend on the other.
3. `stack build --only-dependencies`

### Expected

I expect neither of the packages to build

### Actual

The depended-on package builds

### Stack version

```
$ stack --version
Version 1.5.1, Git revision 600c1f01435a10d127938709556c1682ecfd694e (4861 commits) x86_64 hpack-0.17.1
```

rjmk:
This reverts commit d1e6c480cd46def34bfacd3aeaee953ba32c7758 and resolves #3476 

The problem is caused specifically in the `PSFiles -> Nothing` branch of `installPackages`.

As the flag was no longer being passed through, packages were not being checked for being in the wanted map. This led to the local packages being built as dependencies

I tested this by running `stack --only-dependencies --dry-run` on our stack project with multiple packages. Setting the call to `resolveDepsAndInstall`'s `treatAsDep` value to `True` causes the problem and removing it removes the problem

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -142,6 +142,9 @@ Bug fixes:
   [#3376](https://github.com/commercialhaskell/stack/issues/3376).
 * `stack clean` now works with nix.  See
   [#3468](https://github.com/commercialhaskell/stack/issues/3376).
+* `stack build --only-dependencies` no longer builds local project packages
+  that are depended on. See
+  [#3476](https://github.com/commercialhaskell/stack/issues/3476).
 
 
 ## 1.5.1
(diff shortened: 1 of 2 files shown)


mgsloan:
Hmm, I guess I did not consider the logic fully when implementing that change.  My bad!  The logic there is still rather confusing, but if this fixes the current case, then seems reasonable.

Issue #3264, opened 2017-07-13
Add --cwd parameter to stack exec


phadej:
```
--cwd                    Change current directory before executing
```

I'm happy to make a PR, this shouldn't be too hard to do.

My motivation is running benchmarks in aeson

snoyberg:
I don't have a problem with such an addition, PR definitely welcome, but I'm curious: why isn't using `cd` an option in this case?

phadej:
@snoyberg the difference isn't big, but IMHO the first variant is hacky (note `../`)

```
aeson$ (cd benchmarks && STACK_YAML=../stack-bench.yaml stack exec -- aeson-benchmark-typed encodeGeneric/direct/twitter100 +RTS -s)
```
vs
```
aeson$  STACK_YAML=stack-bench.yaml stack exec --cwd benchmarks -- aeson-benchmark-typed encodeGeneric/direct/twitter100 +RTS -s
```

khanage:
Hey I just implemented this - works fine from the console. I'm happy to make any improvements, but it is pretty straight forward :)

```
# khan at Loki in ~/Dropbox/hask/stack on git:3264-add-cwd-to-exec ● [23:26:34]
→ stack exec -- stack -- exec --cwd dist ls
setup		store-versioned

# khan at Loki in ~/Dropbox/hask/stack on git:3264-add-cwd-to-exec ● [23:26:41]
→ stack exec -- stack -- exec ls
CONTRIBUTING.md		LICENSE			Setup.hs		appveyor.yml		dist			dot-dir-locals.el	mkdocs.yml		stack-nightly.yaml	stack.yaml
ChangeLog.md		README.md		TAGS			cabal.sandbox.config	doc			etc			src			stack.cabal		test
```

avanov:
Why wasn't this flag propagated to `stack test` command?

PR #3497, opened 2017-10-17
Implement #3264 adding --cwd to exec

khanage:
Implements feature requested by #3264 - tested it locally with stack exec.

Welcome for any suggestions as to better methods to use/docs to update.

### DIFF
--- a/src/Stack/Options/ExecParser.hs
+++ b/src/Stack/Options/ExecParser.hs
@@ -75,5 +75,7 @@ execOptsExtraParser = eoPlainParser <|>
     eoCwdParser :: Parser (Maybe FilePath)
     eoCwdParser = optional
                   (strOption (long "cwd"
-                             <> help "Sets the working directory before executing")
+                             <> help "Sets the working directory before executing"
+                             <> metavar "DIR"
+                             <> completer dirCompleter)
                   )
--- a/src/Stack/Options/ExecParser.hs
+++ b/src/Stack/Options/ExecParser.hs
@@ -43,6 +43,7 @@ execOptsExtraParser = eoPlainParser <|>
                          <$> eoEnvSettingsParser
                          <*> eoPackagesParser
                          <*> eoRtsOptionsParser
+                         <*> eoCwdParser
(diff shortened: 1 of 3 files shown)


src/main/Main.hs
```diff
@@ -788,6 +789,20 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
           pkgopts <- getPkgOpts menv wc pkgs
           return (prefix ++ compilerExeName wc, pkgopts ++ args)
 
+      runWithPath path callback = case path of
+        Nothing -> callback
+        Just p | not (isValid p) -> callback
```
mgsloan: Hmm, it seems like this would silently ignore `--cwd` for invalid paths.  Is there a reason for this?  Seems to me like it should throw an error.

src/main/Main.hs
```diff
@@ -788,6 +789,20 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+        Just p                   ->
+          if isRelative p
```
mgsloan: I think it'd be cleaner to just use `System.Directory.withCurrentDirectory`.  I wish `path` had `RelOrAbs` or something, but for now `FilePath` is the best way to represent having either.

src/Stack/Options/ExecParser.hs
```diff
@@ -70,3 +71,9 @@ execOptsExtraParser = eoPlainParser <|>
     eoPlainParser = flag' ExecOptsPlain
                           (long "plain" <>
                            help "Use an unmodified environment (only useful with Docker)")
+
+    eoCwdParser :: Parser (Maybe FilePath)
+    eoCwdParser = optional
+                  (strOption (long "cwd"
```
mgsloan: Please also add 

```
               metavar "DIR" <>
               completer dirCompleter <>
```

For better help output and path completion for this option.

khanage: Done :)

mgsloan:
Looks great, thanks!

PR #3502, opened 2017-10-20
Adds integration test for #3396

### LINKED ISSUE #3396: Treat cabal file parsing failures from index as non-fatal
Some recent issues come from the fact that Stack has been (perhaps overzealously) too strict in its parsing logic. Given that we can't properly rely on upstream sources always providing input which is compatible with the deployed version of the Cabal file parser, we should relax this logic. (I made the original decision implicitly to be strict in parsing, due to past experiences with tools like packdeps and weird corner cases around BOMs, but this reasoning no longer applies.)

Here are some examples of things we should be less strict about:

* If a cabal file does not parse, ignore it. We need to figure out what that means for cases like a snapshot referring to a specific cabal file (perhaps disallow usage of that snapshot, perhaps drop that one package).
* Already implemented: don't add additional checks to flag name parsing: we'll just let in whatever the Cabal library accepts.
* Are there other data types that have additional checks not performed by Cabal?

We need to accept that there is no Cabal format specification, and as it stands now, the Cabal library implementation _is_ the library.

denibertovic:
The provided test package (dinamo), that is part of a custom snapshot has
a malformed cabal file. Specifically, it's setting the required cabal
version to an invalid number (9001). This test makes sure that
stack does not care about this if we do not actually use the package
in our project.

Closes #3396


denibertovic:
cc @snoyberg

denibertovic:
Bah, let me know if I should get rid of the style changes in `IntegrationSpec.hs` (those were automatic). The only change there is just uncommenting "*.gz" in the `toCopyRoot` function... I'm not sure why it was commented out in the first place.

snoyberg:
1. Please do roll back those style changes
2. I'm pretty sure the `base-4.10.0.0.tar.gz` file isn't necessary.
3. I think the `Main.hs` file needs to copy the `custom.index.tar.gz` file to `$HOME/.stack/indices/CustomIndex/00-index.tar.gz` before running `stack build`

denibertovic:
@snoyberg I've added the necessary steps and removed the files we don't need, as well as the formatting. 

One note though: We don't really need the invalid `dinamo` tarball nor do we need the `custom.index.tar.gz` file anymore but I thought it would be useful to keep them around if we ever needed to re-use in similar tests (so we don't have to create them from scratch again). What do you think?

test/integration/tests/3396-package-indices/Main.hs
```diff
@@ -0,0 +1,13 @@
...
+import           System.Environment (getEnv)
+import           System.FilePath    ((</>))
+
+main :: IO ()
+main = do
+  home <- getEnv "HOME"
+  createDirectory (home </> ".stack" </> "indices" </> "CustomIndex")
+  copy "CustomIndex/00-index.tar" (home </> ".stack"  </> "indices" </> "CustomIndex" </> "00-index.tar")
+  copy "CustomIndex/01-index.tar" (home </> ".stack"  </> "indices" </> "CustomIndex" </> "01-index.tar")
+  copy "CustomIndex/01-index.cache" (home </> ".stack"  </> "indices" </> "CustomIndex" </> "01-index.cache")
```
snoyberg: Better not to copy over the `cache`, especially since the format will change over time. Also, I'd be interested in whether this works without the `00-index.tar` file as well.

snoyberg:
I'd probably leave off the unnecessary tarballs for now, it shouldn't be too hard to recreate something similar.

test/integration/tests/3396-package-indices/Main.hs
denibertovic: @snoyberg good point. Removing the cache file. Also, 00-index.tar is not needed so removing that as well.

denibertovic:
@snoyberg Updated (removed tarballs that we don't need right now).

PR #3506, opened 2017-10-22
Use hpack package.yaml to build Stack

borsboom:
This migrates to using an hpack package spec to build Stack.  The `package.yaml` was generated using `hpack-convert` and then manually tweaked.


mgsloan:
LGTM other than the CI failures.  I also think it may make sense to not have library dependencies at all, instead list all library dependencies as being shared dependencies.

That's awesome that hpack-convert got it most of the way there, very handy.

src/System/Terminal.hsc
```diff
@@ -4,6 +4,8 @@ module System.Terminal
 ( getTerminalWidth
 ) where
 
+#ifdef WINDOWS
```
mgsloan: Looks like this should be #ifndef

borsboom: Thanks for catching that!  BTW I'm still iterating until I get something reliable on all the builds.  I'll squash all these commits when I'm ready and note that it's ready for real review.

borsboom:
Looks like this is going to have to wait until the next hpack release (which will include https://github.com/sol/hpack/commit/6a55600c5fd86be5ae95eb2f7b217c898519a676)

borsboom:
Sol suggested a better way to handle System.Terminal that doesn't need the new version of hpack which seems to work nicely, so merged that.

ilovezfs:
@borsboom Running `hpack` and then attempting to bootstrap as usual I get
```
stack-1.6.0.20171202-CLoQ8x2jqWPIqvZyM8cCwH failed during the configure step.
The exception was:
dieVerbatim: user error (cabal: '/usr/local/opt/ghc@8.0/bin/ghc' exited with
an error:

dist/dist-sandbox-8aadc7ad/setup/setup.hs:15:1: error:
Failed to load interface for ‘Distribution.Types.PackageName’

dist/dist-sandbox-8aadc7ad/setup/setup.hs:16:1: error:
Failed to load interface for ‘Distribution.Types.UnqualComponentName’

dist/dist-sandbox-8aadc7ad/setup/setup.hs:18:31: error:
Module ‘Distribution.Version’ does not export ‘showVersion’
)
```
Is that expected?

borsboom:
I'm guessing you're probably trying to build with an older version of the Cabal library.  We're not settings dependency bounds in the source code's `package.yaml` since package versions are set by the Stackage snapshot referenced by `stack.yaml`.  However, for purposes of bootstrapping with cabal-install, I've now uploaded [stack-1.6.0.20171202-sdist-1.tar.gz](https://github.com/commercialhaskell/stack/releases/download/v1.6.0.20171202/stack-1.6.0.20171202-sdist-1.tar.gz) to the release, which includes a `stack.cabal` file with dependency bounds.  A Cabal spec with dependency bounds will also be uploaded to Hackage for the final v1.6.1 release.

ilovezfs:
Thanks @borsboom. I'll give https://github.com/commercialhaskell/stack/releases/download/v1.6.0.20171202/stack-1.6.0.20171202-sdist-1.tar.gz a try.

PR #3511, opened 2017-10-24
Print retry information if stack needs to retry network requests

NorfairKing:
Tested this by adding `127.0.0.1       s3.amazonaws.com localhost` to my `/etc/hosts`, to make connections fail reliably (oh, the irony).

The output looks like this:

```
Retry number 0 after a total delay of 0 us
If you see this warning and stack fails, but running the command again solves the problem, please report here: https://github.com/commercialhaskell/stack/issues/3510
Progress: 4/6HttpExceptionRequest Request {
  host                 = "s3.amazonaws.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/hackage.fpcomplete.com/package/text-zipper-0.10.1.tar.gz"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HostCannotConnect "s3.amazonaws.com" [Network.Socket.connect: <socket: 12>: does not exist (Connection refused)]))
```

### DIFF
--- a/src/Network/HTTP/Download/Verified.hs
+++ b/src/Network/HTTP/Download/Verified.hs
@@ -30,7 +30,7 @@ import qualified    Data.Text.Encoding as Text
 import              Control.Monad
 import              Control.Monad.Catch (Handler (..)) -- would be nice if retry exported this itself
 import              Stack.Prelude hiding (Handler (..))
-import              Control.Retry (recovering,limitRetries,RetryPolicy,constantDelay)
+import              Control.Retry (recovering,limitRetries,RetryPolicy,constantDelay,RetryStatus(..))
 import              Crypto.Hash
 import              Crypto.Hash.Conduit (sinkHash)
 import              Data.ByteArray as Mem (convert)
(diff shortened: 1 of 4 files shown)


src/Network/HTTP/Download/Verified.hs
```diff
@@ -179,21 +179,37 @@ hashChecksToZipSink :: MonadThrow m => Request -> [HashCheck] -> ZipSink ByteStr
...
+      run $ do
+        logWarn $ Text.unwords
+          [ "Retry number"
+          , Text.pack $ show (rsIterNumber rs)
+          , "after a total delay of"
+          , Text.pack $ show (rsCumulativeDelay rs)
+          , "us"
+          ]
+        logWarn $ Text.unwords
+          [ "If you see this warning and stack fails,"
```
mgsloan: Would be good to narrow "fails" a bit, perhaps "If you see this warning, and this download doesn't succeed,"

mgsloan:
LGTM!

Not necessary, but would be good to use Stack.PrettyPrint for logging.

NorfairKing:
I can see that travis failed, but the logs don't load on any of my browsers (even with JS turned on), so I cannot see why...

mgsloan:
In the logs I'm seeing

> /home/travis/build/commercialhaskell/stack/src/Network/HTTP/Download/Verified.hs:198:28: warning: [-Wunused-matches]
> Defined but not used: ‘e’

mgsloan:
LGTM!

PR #3516, opened 2017-10-26
Support conduit-extra 1.2.0

felixonmars:

Builds and runs fine here.

### DIFF
--- a/src/Data/Attoparsec/Interpreter.hs
+++ b/src/Data/Attoparsec/Interpreter.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE CPP #-}
 {- |  This module implements parsing of additional arguments embedded in a
       comment when stack is invoked as a script interpreter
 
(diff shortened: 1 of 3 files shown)


src/Data/Attoparsec/Interpreter.hs
```diff
@@ -146,9 +146,9 @@ getInterpreterArgs file = do
 
     decodeError e =
       case e of
-        ParseError ctxs _ (Position line col) ->
+        ParseError ctxs _ (Position line col offset) ->
           if null ctxs
           then "Parse error"
           else ("Expecting " ++ intercalate " or " ctxs)
-          ++ " at line " ++ show line ++ ", column " ++ show col
+          ++ " at line " ++ show line ++ ", column " ++ show col ++ ", offset " ++ show offset
```
lwm: Fairly slow as usual but why is the offset useful in this context?

lwm:
Apologies for letting this rot, aside from needed rebase and minor comment, LGTM!

src/Data/Attoparsec/Interpreter.hs
felixonmars: I am not sure about this either, could remove it if you think it's not relevant :)

lwm: Well, do you have an example of the output? Maybe it's useful?

mgsloan:
Yeah, sorry about letting this languish, somehow I missed it.  I don't think offset should be outputted.  Is `Position` a record constructor?  It would be nice if the code would work both with the new version of conduit-combinators and the old, but not strictly necessary.

felixonmars:
Updated to remove offset from output and support both old and new conduit-extra.

src/Data/Attoparsec/Interpreter.hs
```diff
@@ -139,14 +140,18 @@ getInterpreterArgs file = do
 
     parseArgStr str =
       case P.parseOnly (argsParser Escaping) (pack str) of
-        Left err -> handleFailure ("Error parsing command specified in the \
-                        \stack options comment: " ++ err)
+        Left err -> handleFailure ("Error parsing command specified in the "
```
mgsloan: Oh man CPP + multiline strings strikes again!  Hah :'(

mgsloan:
LGTM, thanks again!

Issue #3454, opened 2017-09-25
stack script should allow setting ghc-options


mgsloan:
Currently, `stack script` only allows specifying arguments to the user's program, not to the ghc invocation.  Should probably use `--ghc-arg` as described in https://downloads.haskell.org/~ghc/master/users-guide/runghc.html , rather than passing the arguments directly.  I suggest adding the following:

* `--ghc-options`, which works just like the argument to build
* `--ghc-arg`, which allows passing it to just the script command's ghc invocation.

d-dorazio:
@mgsloan if nobody has taken this yet, I can work on it.

I'm not sure I understand what would be the difference between `--ghc-options` and `--ghc-arg` though. My understanding is:

- `stack script --resolver lts-9.2 --ghc-options -threaded -v5` would be roughly equivalent to `runghc -v5 foo`;
- `stack script --resolver lts-9.2 --ghc-arg -v5 foo` would be roughly equivalent to `runghc --ghc-arg -v5 foo`.

but I'm not sure that's 100% right.

mgsloan:
@d-dorazio Awesome!  Sorry for the delay in getting back to you on clarification.

I've thought about this a bit more, and actually I don't think it makes sense to allow providing ghc options for use when building packages (this is what I intended `--ghc-options` to be).  So, this means no need to make up a new flag name (`--ghc-arg`).

So I think a better design is to make `--ghc-options` get passed to the ghc invocation.  So, `stack script --resolver lts-9.2 --ghc-options "-threaded -v5"` would do something like `runghc --ghc-arg "-v5 foo"`;

dbaynard:
Closing due to #3519

PR #3519, opened 2017-10-26
add --ghc-options to stack script

d-dorazio:
This should address #3454.

cc: @mgsloan


mgsloan:
LGTM, thanks!!

yiding:
This seems to only work when the script is not being run with `--compile`, since `--ghc-arg` is only recognized when running via `runghc`.

Issue #3527, opened 2017-10-28
1.6 prerelease doesn't work with offline-stack


ndmitchell:
I changed the scripts at https://github.com/ndmitchell/offline-stack to work with a 1.6 prerelease. It fails. Specifically, I get the errors:

```
The following package identifiers were not found in your indices: blank-canvas-0.6.1@sha256:7ed3baf69e4aa56a4ec63545e6fbf827b0686fbb2b6f8e489a6e894b10c83631,7201
The specified revision was not found.
Possible candidates: blank-canvas-0.6.1@sha256:1ecfe227ead93e8e4c92748a5c3751f4b3cefeb14cc37f3b514da4321b84219b, blank-canvas-0.6.1@sha256:bd2953cc6e004f7aa7699f1ae15a43110ee967e35faed02a7b43b8fa4aa97fc4.
```

Quite a few commercial companies run Stack offline, using scripts based around the example script in offline-stack, so this change would be very problematic.

ndmitchell:
Failure with:

```
Version 1.6.0, Git revision 1a9d326230139b0437cf1c35192cfd5488e86062 x86_64 hpack-0.18.1
```

Worked with 1.5.1.

mgsloan:
Hmm, can you please update the project's README with a clear way to try this thing out?  I'm not really sure how to use this together with the mirror script, I don't feel like trying to cobble together scripts at the moment.

ndmitchell:
The whole point about this project is that it isn't trivial to run Stack offline, so it gives scripts that are a known good state, then you spend a week or so smashing that into your environment (the entire process is long and painful - this script just gives you the confidence that when you arrive you have something that works).

I can try and produce Windows reproduction instructions if that's of use - or are you Linux only?

borsboom:
This might be related to https://github.com/commercialhaskell/stack/issues/3514

mgsloan:
@ndmitchell I'm on linux only.  Oh yeah I totally agree that stack should support offline use better, I'm in favor of merging PRs that make this possible in a more native way.

My current workflow when I know I'm going to be offline is to ensure I've pulled everything I need.  I can certainly understand that it's annoying when working offline and you realize you need a new dep.

ndmitchell:
Note that offline above is offline in the sense your computer is never connected to the internet, but sits behind a corporate firewall.

ndmitchell:
[...]

Issue #3514, opened 2017-10-25
Missing package in indices (1.6.0)


JoeMShanahan:
Following on from @snoyberg's [call to action](https://groups.google.com/forum/#!msg/haskell-stack/SEvMu1yymWk/phVkkc9VBwAJ).

With the up to date release of Stack, our project builds correctly.

With `stack upgrade --git`, I run in to the following:

    joeshanahan@js-ubuntu:~/project$ stack --version
    Version 1.6.0, Git revision 0e756d486a5a4d6fe03e0d448bb8e706c5834ca1 x86_64 hpack-0.18.1

    joeshanahan@js-ubuntu:~/project$ stack build
    Downloaded lts-8.24 build plan.
    Didn't see Agda-2.5.2@sha256:4db0b12bc07e72fe1b180acad2a0d59ac11d9a1d45698b46cede7b634fb6bfff,27524 in your package indices.
    Updating and trying again.
    Downloading package index from https://s3.amazonaws.com/hackage.fpcomplete.com/00-index.tar.gz
    No updates to your package index were found
    Update complete
    Unable to load cabal files for snapshot
    
    ----
    Deleting cached snapshot file: /home/joeshanahan/.stack/build-plan/lts-8.24.yaml
    Recommendation: try running again. If this fails again, open an upstream issue at:
    https://github.com/fpco/lts-haskell/issues/new
    ----
    
    The following package identifiers were not found in your indices: Agda-2.5.2@sha256:4db0b12bc07e72fe1b180acad2a0d59ac11d9a1d45698b46cede7b634fb6bfff,27524
    The specified revision was not found.
    Possible candidates: Agda-2.5.2@sha256:28af7449a60bce1d020020524e72779415d04dd2a83dee57ec62eff484834282, Agda-2.5.2@sha256:16152f1ea7f461dfc48aee61a75a7f1018832e9bcbcae45eba3a7d08ff1f28c9.

With `--resolver lts-9.1` the error is slightly different:

    Didn't see avers-0.0.17.1@sha256:dae0e728b26b259e31a731fcb73a9b724c58c81a8f09e795c1918a1388ee701b,3710 in your package indices.

I'm reluctant to provide our project yaml / cabals as they're quite large and I'd need to anonymise them somewhat. I can work on getting a minimal project together than reproduces the problem if the above is not enough to go on.

JoeMShanahan:
On nightly:

```
$ stack build --resolver nightly
[...]

PR #3529, opened 2017-10-29
Note that the 00-index.tar.gz file no longer works

ndmitchell:
As per #3527 and #3514, if you custom point at a 00 index file, you get errors. Important instead to point at the 01 version. Update the example in the user manual and put it in the changelog.


mgsloan:
LGTM, thanks!

PR #3538, opened 2017-10-31
Don't put ':' into the copyright message. It breaks 'stack new' by …

aaronmcdaid:
…causing an error 'simple/package.yaml:6:30: mapping values are not allowed in this context'

This is a tiny documentation change


mgsloan:
Thanks!

Issue #3483, opened 2017-10-10
Prevent --file-watch from watching test suites


JoeMShanahan:
How appropriate would it be to allow a new argument which prevents `stack` from triggering a rebuild when a test suite changes?

During development of a web app using `yesod`, we're in the situation where I'm running the server with `yesod devel` (which uses `stack ... --file-watch` underneath).

Our selenium test suite is attached to the server `devel` is running. I'm altering some selenium tests which I have loaded with GHCi. When I change the selenium tests, `stack` notices, which signals to `devel` that a rebuild is required even though it's not! This really just slows down development.

I propose something like `stack build --file-watch --no-watch-tests` . What do you think?

snoyberg:
It may even be reasonable to have that behavior automatically implied based on which components are being built. I don't see a downside to such a change.

JonnyRa:
I was about to log an issue for this but looking at the linked issue it seems to be fixed in 1.6.

I've updated and just tested on my setup and I'm not seeing changes to the tests retriggering builds when they shouldn't now.  Woohoo!

Can probably be closed as fixed now?

Nice one!

Cheers

Jonny

lwm:
Thanks @JonnyRa, will close this off!

PR #3565, opened 2017-11-10
Don't watch files for non-library components that are not being built

tswelsh:
Attempt at a fix for #3483, following the comment there.  I'm not quite up on the internals of stack so let me know if this is the wrong direction. It raised a couple of questions about how things work, which may make it into their own issues.  One in particular: is the library component of a package always built, regardless of targets?


tswelsh:
Just realised I had the checklist etc but hadn't actually done it.  How I tested: I used a project with a single package with an executable and test suite.  Ran `stack build --file-watch`, edited the test suite `Main.hs` and observed no rebuild.  Edited the executable's main file and it did rebuild.

snoyberg:
This looks OK, and I've approved, but just doing a little more testing. I noticed some strange behavior, not sure if it's this patch or something else.

mgsloan:
Hmm, I'm stumped.  Haven't tried to reproduce it, but the change looks good.  In particular:

* This should only affect `lpFiles`, and it is only used for file watching.

* Otherwise, it seems just like an inlining of `getPackageFilesSimple`, along with selection of which components to build.

Curious!

tswelsh:
OK, I have reproduced the issue spotted by @snoyberg.  I believe the `unregistering (local file changes: ... )` is due to a call to `checkForUnlistedFiles`, called by `postBuildCheck`, within `Stack.Build.Execute.singleBuild`.  This ultimately uses `getPackageFilesSimple` which gets all files including those to do with test suites.  As the test suite files aren't in the build cache, they get added by this check and appear to be dirty the next time round, even though they aren't being used.  Phew!

I've added a commit which uses the same logic in both places, so the ultimate effect of all the changes is more or less to replace `getPackageFilesSimple` with a new, pickier function `getPackageFilesForTargets`.

snoyberg:
Looks great now, thanks!

Issue #3573, opened 2017-11-13
$locals flags should come before specific packages


ndmitchell:
Using `stack` installed with `--git` on Windows 10 64bit, version:

```
Version 1.6.0, Git revision 46121be1b96465f1164e3f84cafa19c7369da9cc x86_64 hpack-0.18.1
```

Given [stack-bug.zip](https://github.com/commercialhaskell/stack/files/1468724/stack-bug.zip), that defines in the `stack.yaml`:

```
ghc-options:
  $locals: -bob
  foo: -beep
```

When compiling `foo` I expect the flag `-bob` to go first, followed by `-beep`, so that (using more standard examples) I can turn on `-Werror` for all local packages, but then selectively turn it off with `-Wwarn` for just the package `foo`. However, the error message shows that `-bob` occurs after `-beep`, and thus cannot be overridden for a specific package. This change is important in large multi-cabal-project stack projects. The error message reads:

```
[...]

Issue #2853, opened 2016-12-16
stack --profile shouldn't force --ghc-options='-auto-all'


robsmith11:
I want to use only manual cost centers when profiling, but it appears that stack is forcing `--ghc-options='-auto-all'` whenever profiling is enabled.  Even if I pass `--ghc-options='-fno-prof-auto'`, the auto cost center are still used.

robsmith11:
Do you know of any easy workaround?  I'm trying to profile my application, but the auto cost centers in `vector` break the critical fusion optimization making profiling unusable.

mgsloan:
Not exactly an "easy" workaround, but one temporary option might be to compile stack with the `-auto-all` stuff removed.

I'm surprised that `--ghc-options='-fno-prof-auto'` doesn't do the trick.  Perhaps we will need to provide a `--profile-no-auto` flag?

duog:
I'm running into this now. The issue seems to be that stack passes `-auto-all` and `-auto-cafs` AFTER manually specified ghc-options, so I'm unable to override it with -fno-prof-auto-all etc.

tirumaraiselvan:
Hi, any plans to work further on this issue?

tuomohopia:
I'm guessing this isn't seeing any progress anymore. 

Did anyone come up with any working workaround? Would hate to have to move to cabal just because of this.

DouglasRaillard:
Hi, +1 on the issue, passing `--ghc-options=-fno-prof-auto` does not seem to do the trick, making the actual information I'm looking for lost in the noise (and quite likely completely transformed, the runtime jumps from 6 seconds to ~45s when enabling profiling).
For some reason, it even seems to profile `>>=` from attoparsec Zepto type, even though it's marked as inline.

PR #3576, opened 2017-11-14
Apply package specific "ghc-options" after more general ones #3573

mgsloan:
The order was: specific, $targets, $locals, $everything
The order now is: $everything, $locals, $targets, specific

I figured it made sense for targets flags to come after locals, since it is
often a subset.


mgsloan:
Pinging @snoyberg .  While changing the order of these ghc-options, it might make sense to also revisit the order of ghc options specified by profiling options - see https://github.com/commercialhaskell/stack/issues/2853

PR #3579, opened 2017-11-16
Upgrade to hpack 0.20.0 (fixes #3383)

### LINKED ISSUE #3383: Detect when user changed .cabal file instead of package.yaml
Use of `package.yaml` can be confusing when the user doesn't expect it - see https://github.com/commercialhaskell/stack/issues/3380 .  I think this case should somehow be detected and a noisy warning emitted.

One way to do this might be to output a copy of the `*.cabal` on hpack generation.  Then, when later re-generating, check if the cabal file changed since the last hpack generation.  If so, then the user modified it and a noisy warning should be displayed.

snoyberg:

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -1321,11 +1321,7 @@ hpack pkgDir = do
         config <- view configL
         case configOverrideHpack config of
             HpackBundled -> do
-#if MIN_VERSION_hpack(0,18,0)
-                r <- liftIO $ Hpack.hpackResult (Just $ toFilePath pkgDir)
-#else
-                r <- liftIO $ Hpack.hpackResult (toFilePath pkgDir)
-#endif
+                r <- liftIO $ Hpack.hpackResult (Just $ toFilePath pkgDir) Hpack.NoForce
                 forM_ (Hpack.resultWarnings r) prettyWarnS
                 let cabalFile = styleFile . fromString . Hpack.resultCabalFile $ r
                 case Hpack.resultStatus r of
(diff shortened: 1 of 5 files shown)


src/Stack/Package.hs
```diff
@@ -1338,6 +1334,11 @@ hpack pkgDir = do
                         , flow "was generated with a newer version of hpack,"
                         , flow "please upgrade and try again."
                         ]
+                    Hpack.ExistingCabalFileWasModifiedManually -> prettyWarnL
+                        [ cabalFile
+                        , flow "was manually modified but it is autogenerated,"
+                        , flow "please delete and modify package.yaml in the future"
```
mgsloan: I think "please delete **it**" would be clearer, right now it could read as "delete + modify package.yaml".  Perhaps it's not so ambiguous, but still best to be clear as possible

mgsloan:
LGTM just one comment

src/Stack/Package.hs
sol: Or something along the lines of
```perl
"WARNING: $cabalFile was modified manually.  Ignoring package.yaml in favor of $cabaFile.  If you want to use package.yaml instead of $cabalFile then please delete $cabalFile."
```
?

mgsloan: Hmm, perhaps!  Or it could just pre-emptively delete the cabal file! :imp:  Hah probably people wouldn't appreciate that.

Perhaps it should just error in this circumstance.  Hmm hmm, nothing seems to be quite perfect.

sol: One more note: The reason why I added `--force` to `hpack` and why `--force` is preferable to deleting the cabal file is, that `hpack` does some effort to minimize diffs if a cabal file is already present.  So `hpack --force` and `rm *.cabal && hpack` may lead to differently formatted cabal files.

snoyberg: I'll go with @sol's verbiage, I think the hpack author is likely the best to figure these things out anyway :)

PR #3585, opened 2017-11-18
Expand --help output of 'new' and 'templates' commands

### LINKED ISSUE #3584: `stack templates --help` should mention stack-templates repo
### General summary/comments

Lots of templates listed from `stack templates` provide no description, so I have some research and found [stack-templates](https://github.com/commercialhaskell/stack-templates) repository.

I think that others might want to know abot it. Forgive me if it was in documentation somewhere already, I can't find any.

I think the best place to mention the repo would be when running

    stack templates --help

kadoban:
closes #3584

Added a note to 'templates' mentioning where the templates list comes from. Also mentioned in both 'new' and 'templates' that you can specify a file or URL for a template. It was already mentioned in 'new', but only in the description of TEMPLATE_NAME, which was easy to miss.


kadoban:
Hm, that CI failure seems to be a hlint error in a file I didn't actually touch. I'll check later to make sure though.

mgsloan:
LGTM, thanks for fixing the build, that was my bad!

PR #3587, opened 2017-11-19
[WIP] colorize warn output in s/P/CheckInstall

kadoban:
Working on prettying up more log/warn/error output in various places.

### DIFF
--- a/src/Path/CheckInstall.hs
+++ b/src/Path/CheckInstall.hs
@@ -33,9 +33,7 @@ warnInstallSearchPathIssues destDir installed = do
                           , flow "executable found on the PATH environment variable is"
                           , styleFile . fromString $ exePath
                           , flow "and not the version that was just installed."
-                          ]
-                        prettyWarnL
-                          [ flow "This means that"
+                          , flow "This means that"
                           , styleFile . fromString . T.unpack $ exe
                           , "calls on the command line will not use this version."
                           ]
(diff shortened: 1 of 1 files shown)


src/Path/CheckInstall.hs
```diff
@@ -25,32 +27,29 @@ warnInstallSearchPathIssues destDir installed = do
...
-                            , "\" executable found on the PATH environment variable is "
-                            , T.pack exePath
-                            , ", and not the version that was just installed."
-                            ]
-                        logWarn $ T.concat
-                            [ "This means that \""
-                            , exe
-                            , "\" calls on the command line will not use this version."
-                            ]
+                        prettyWarnL
```
mgsloan: Would be best to combine these two into a single warn call so that it just says "Warning:" once

mgsloan:
LGTM, just the one comment

src/Path/CheckInstall.hs
kadoban: Ah, yes nice catch thanks. Will change.

mgsloan:
@kadoban Is this one good to merge?  Looks ready to me.

kadoban:
@mgsloan Ah yeah it actually is AFAIK. I was going to add more colorizing to it, but I'll just start another for other files.

mgsloan:
Rad, thanks!

PR #3592, opened 2017-11-19
Ensure cabal file warnings display once (fixes #3591)

### LINKED ISSUE #3591: Cabal file parsing warnings display multiple times
Discovered while working on #3383. The new warning displayed when the cabal file is modified displays many times. Best fix: ensure that each cabal file is only parsed once. For various reasons, it may be difficult to ensure that in the codebase right now, so instead keeping track of which cabal files have generated warnings that have been printed may be a good trade-off for now.

(Maybe a cabal file parser cache in memory would do the same job...)

snoyberg:


mgsloan:
LGTM, though it seems best to make it so that this function doesn't get called multiple times.

Issue #3549, opened 2017-11-04
upload with lower pvp-bounds removes things in 1.6.0


mfine:
### General summary/comments

Uploading with `pvp-bounds: lower` is removing things from the cabal file in 1.6.0.

### Steps to reproduce

Publish a package to hackage containing an executable with a dependency on the package's library with a stack.yaml containing `pvp-bounds: lower`.

### Expected

I expected the bounds to be set on the dependencies for the executable.

### Actual

There were missing fields:

Before:
```
library
  default-language:    Haskell2010
  hs-source-dirs:      src
  exposed-modules:     Development.Shakers
  ghc-options:         -Wall
  build-depends:       base >= 4.8 && < 5
                     , basic-prelude
                     , directory
                     , shake

executable shake-shakers
  default-language:    Haskell2010
  main-is:             Shakefile.hs
  ghc-options:         -threaded -rtsopts -with-rtsopts=-N -Wall
  build-depends:       base
                     , shakers
```
After:
```
library
    exposed-modules:
        Development.Shakers
    build-depends:
        base >=4.8 && <5,
        basic-prelude >=0.5.2,
        directory >=1.2.2.0,
        shake >=0.15.11
    default-language: Haskell2010
    hs-source-dirs: src
    ghc-options: -Wall

executable  shake-shakers
    main-is: Shakefile.hs
```

Missing `default-language`, `ghc-options`, and `build-depends`.

Here's the verbose output of sdist:

```
$ Version 1.6.0, Git revision 64794699803ff997dd9a1ff999c342c73d99bb20 (5322 commits) x86_64
Compiled with:
- Cabal-2.0.0.2
- Glob-0.9.0
- HUnit-1.6.0.0
- MonadRandom-0.5.1
- QuickCheck-2.10.1
- StateVar-1.1.0.4
- aeson-1.2.2.0
- aeson-compat-0.3.7.1
- annotated-wl-pprint-0.7.0
- ansi-terminal-0.6.3.1
- ansi-wl-pprint-0.6.8.1
- array-0.5.2.0
- asn1-encoding-0.9.5
- asn1-parse-0.9.4
- asn1-types-0.3.2
- async-2.1.1.1
- attoparsec-0.13.2.0
- attoparsec-iso8601-1.0.0.0
- auto-update-0.1.4
- base-4.10.0.0
- base-compat-0.9.3
- base-orphans-0.6
- base-prelude-1.2.0.1
- base16-bytestring-0.1.1.6
[...]

PR #3593, opened 2017-11-21
Only tweak cabal file as necessary

snoyberg:
I think this was a mistaken change in
9f06ccfdb8fe6cc1156ce04c815731e63d3717d1. Perhaps ironically, without
this commit, we end up regularly triggering #3549 instead of only doing
so when using pvp-bounds.


mgsloan:
Oops, yeah, I don't remember why I changed it to `True`, making it easier to test I suppose.  Good catch!

PR #3594, opened 2017-11-21
WIP: integration test for static linking under nix

wfranzini:
Hi,

I'm failing to build a statically linked executable using nix.

I've tried to look at the source code searching for something (i.e. a test) I've not been able to find in the documentation (a recipe).

This pull request contains two test to expose the problem I'm having.

I hope someone can point me to some docs or fix the problem :-)

Thanks.
walter

### DIFF
--- a/.travis-setup.sh
+++ b/.travis-setup.sh
@@ -15,6 +15,11 @@ fetch_stack_linux() {
   curl -sL https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack';
 }
 
+fetch_nix () {
+  mkdir /nix
+  curl https://nixos.org/nix/install | bash
+}
+
 # We need stack to generate cabal files with precise bounds, even for cabal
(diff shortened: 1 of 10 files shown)


mgsloan:
@wfranzini Hi! Did you intend to close this?

Seems like a reasonable thing to test.  Though, I'm not 100% sure if we will want to depend on nix for our integration tests, would certainly be good to have better tests for that area of things.  Should be fine as long as it doesn't add too much time.

wfranzini:
@mgsloan I closed the PR since the commits that added integration tests to travis has been pushed by mistake.  I'll open a new one in the following days.

I think it should be (optionally) tested since nix support is mentioned in the documentation.  BTW the tests does not take too much time as long as nix is already installed.

mgsloan:
Ok, sounds good!  Instead of creating a new branch / PR, one approach there is to rebase or reset your branch.  For example, to remove a commit, you can do `git reset --hard HEAD^`.  Even more flexible is using `git rebase --interactive`

Issue #3550, opened 2017-11-04
Use git version of Cabal to build stack / release version?


mgsloan:
See these issues: https://github.com/commercialhaskell/stack/issues/3549 https://github.com/haskell/cabal/issues/4863

Released Cabal has a bug that makes it so that `showGenericPackageDescription` drops quite a few fields.  Should we build stack-1.6 against a version of the Cabal library which has a fix?  If we do that, should modify the [warning message I added](https://github.com/commercialhaskell/stack/issues/3549#issuecomment-341875162) here: https://github.com/commercialhaskell/stack/commit/9f06ccfdb8fe6cc1156ce04c815731e63d3717d1

Noted by phadej [here](https://github.com/haskell/cabal/issues/4863#issuecomment-341881026):

> By
> > Any way of pick, you have to make sure to reject cabal-version: 2.1 and later, as their spec is open atm.
> I meant you should reject .cabal files with too new Cabal spec version, there are features in master which may not be in their final form.

snoyberg:
Once #3593 is merged in, I believe this bug will only strike when we're using pvp-bounds. We already have a mounting list of concerns around pvp-bounds, such as potentially being out-of-date with the new "soft bound" changes to the PVP. Given all of this, my recommendation is to add some comments to the release notes and documentation explaining that, due to issues with Cabal, pvp-bounds will likely not work, and recommend against using them. I don't believe it's wise for us to release Stack with an unofficial version of Cabal, or start copy-pasting code to work around Cabal bugs unless we absolutely have to.

mgsloan:
Makes sense to me, closing this

PR #3595, opened 2017-11-22
Add warnings to docs/release notes about pvp-bounds #3550

snoyberg:
@mgsloan @borsboom does this text reflect things appropriately?


mgsloan:
Yup :+1:

Issue #3574, opened 2017-11-13
Local extra-dep should not be considered $locals


ndmitchell:
Using `stack` installed with `--git` on Windows 10 64bit, version:

```
Version 1.6.0, Git revision 46121be1b96465f1164e3f84cafa19c7369da9cc x86_64 hpack-0.18.1
```

Given [stack-bug.zip](https://github.com/commercialhaskell/stack/files/1468752/stack-bug.zip), that defines in the `stack.yaml`:

```
packages:
- location: foo
  extra-deps: true

ghc-options:
  $locals: -bob
```

When compiling `foo` I don't expect the flag `-bob` to be applied as `foo` is an extra-dep, and thus not a local package. However, when I `stack build` I get an error about:

```
ghc.EXE: unrecognised flag: -bob
```

mgsloan:
`extra-deps` is perhaps not the best name, a bit of a historical artifact.  It would be nice to improve the stack documentation to describe the terminology properly.  Or, perhaps, change the behavior to be less surprising.

"extra-dep" essentially just means "not a target by default".  Currently, they are always considered local packages.  Once [implicit snapshots](https://github.com/commercialhaskell/stack/issues/3330) is implemented, this wouldn't be the case and the local package vs snapshot package will make even less sense.  So I anticipate that the local package distinction will get dropped.

I think `$targets` is closer to the behavior of what you're looking for, but it will also be applied in the case that you do `stack build foo`, but won't be if you just do `stack build`.

ndmitchell:
Hmm, `extra-deps` is indeed a terrible name - we've used it completely incorrectly. Can I suggest adding definitions of `$locals` and the other `$variables` to yaml_configuration.md. Note that defining custom ghc-options makes something a `$locals` (I think). I did read the docs, and came to the conclusion (perhaps from extra-deps elsewhere on the page) that the behaviour is wrong.

FWIW `$default-targets` was exactly what I wanted, but following #3573 it's not as important, and given the `extra-deps` terminology elsewhere `$default-targets` would be reasonably confusing.

snoyberg:
First things first: the correct syntax is to say `extra-dep: true` instead of `extra-deps: true`. In fact, if I run `stack build` in this directory, I see:

```
Warning: /Users/michael/Desktop/stack.yaml: Unrecognized field in PackageEntry: extra-deps
```

Changing to `extra-dep: true` gets me:

```
[...]

PR #3601, opened 2017-11-23
Don't apply $locals to extra-deps #3574

snoyberg:


mgsloan:
Ah, this is even more clearly a bug when we consider that `getPackageConfig` has no callers that invoked it with `isLocal = True`!

LGTM

PR #3608, opened 2017-11-27
Update the travis-complex example

snoyberg:
* Comment out older GHC 7.6.3 and 7.8.4 (3 version policy)
* Add in cabal 8.2.2
* Switch from lts-8 to lts-9
* Add support for running stack solver as needed
* Add a comment about hpack support


borsboom:
LGTM (although I haven't actually tried it).  Travis failure appears to just be timeout, and given that this has no code changes I think we can merge anyway.

PR #3621, opened 2017-12-04
3568 add tar files option to sdist

Asalle:

* [X] The documentation has been updated, if necessary. (Unfortunately, no significant changes we made, so I could not find any place for it in doc/)

I did not write any unit test, because I did not write any testable units, should I write any integration tests?


snoyberg:
Travis is failing due to its style build: https://travis-ci.org/commercialhaskell/stack/jobs/311334037#L629. The message is:

```
src/main/Main.hs:758:13: Warning: Use unless

Found:

  when (not $ null $ sdoptsTarPath sdistOpts)

Why not:

  unless (null $ sdoptsTarPath sdistOpts)
```

Could you update the code to make hlint happy?

Asalle:
weird, my local hlint shows me some more warnings, about code I didn't touch
do you use some kind of hlint configuration?

snoyberg:
Try running `./etc/scripts/hlint.sh`.

Asalle:
@snoyberg Hey, is it ok? :)

src/Stack/Options/SDistParser.hs
```diff
@@ -21,7 +21,8 @@ sdistOptsParser signDefault = SDistOpts <$>
   (long "sig-server" <> metavar "URL" <> showDefault <>
     value "https://sig.commercialhaskell.org" <>
     help "URL") <*>
-  buildPackageOption
+  buildPackageOption <*>
+  strOption (long "tar-files" <> help "If specified, copy all the tar to this dir")
```
snoyberg: Unless I'm mistaken, this setup means that the command line options now _require_ a `--tar-files` argument, and we get the old behavior via `--tar-files ""`. Can you instead use `optional (strOption (...))`?

Also, I think `--tar-path` may be a nicer option name.

src/Stack/SDist.hs
```diff
@@ -84,6 +84,8 @@ data SDistOpts = SDistOpts
   -- ^ The URL of the signature server
   , sdoptsBuildTarball :: Bool
   -- ^ Whether to build the tarball
+  , sdoptsTarPath :: FilePath
```
snoyberg: Then this would also switch to `Maybe FilePath`.

src/main/Main.hs
```diff
@@ -755,7 +755,14 @@ sdistCmd sdistOpts go =
             liftIO $ L.writeFile (toFilePath tarPath) tarBytes
             checkSDistTarball sdistOpts tarPath
             prettyInfoL [flow "Wrote sdist tarball to", display tarPath]
+            unless (null $ sdoptsTarPath sdistOpts) (copyTarToTarPath tarPath tarName)
             when (sdoptsSign sdistOpts) (void $ Sig.sign (sdoptsSignServerUrl sdistOpts) tarPath)
+        where
+          copyTarToTarPath tarPath tarName = liftIO $ do
+            targetDir <- resolveDir' $ sdoptsTarPath sdistOpts
```
snoyberg: I believe this will fail is the directory does not already exist (though I haven't tested it locally). How about:

I think you'll want to create the directory first with `D.createDirectoryIfMissing` and then use `resolveDir'`. Alternatively: in this case you may want to stick to `FilePath` values and use `</>` from `System.FilePath`.

ChangeLog.md
```diff
@@ -149,6 +149,8 @@ Other enhancements:
   [#3534](https://github.com/commercialhaskell/stack/issues/3534)
 * GHC bindists can now be identified by their SHA256 checksum in addition to
   their SHA1 checksum, allowing for more security in download.
+* Added new `--tar-files` option to `stack sdist`, that allows to copy
```
snoyberg: Since this isn't going to make it into the 1.6 release, can you move this to "unreleased changes"?

mgsloan:
@Asalle Ping!  Snoyberg added some feedback in comments.  No worries if you don't have time for a while, but it would be good to get this merged.

Asalle:
@mgsloan hey, got a little busy, but I'll do my best to work on this during this weekend

ChangeLog.md
Asalle: As far as I understood, it was already in section "Unreleased changes", in subsection "other enhancements" and following versions 1.5.1 are next. Or is it changelog for 1.6 in the top of the file? It's not marked anywhere.

snoyberg: It should go in this area https://github.com/Asalle/stack/blob/630a9d63f099970a3db3bf3667a0bba0457328f0/ChangeLog.md#unreleased-changes

src/Stack/Options/SDistParser.hs
Asalle: But if I make it optional and `sdoptsTarPath` will become `Maybe FilePath` I get this error, that 
```
2 of 2] Compiling Main             ( src/main/Main.hs, .stack-work/dist/x86_64-linux-nopie/Cabal-1.24.2.0/build/stack/stack-tmp/Main.o ) [Stack.Options.SDistParser changed]

stack/src/main/Main.hs:762:40: error:
    • Couldn't match type ‘Maybe FilePath’ with ‘[Char]’
      Expected type: FilePath
        Actual type: Maybe FilePath
    • In the second argument of ‘($)’, namely ‘sdoptsTarPath sdistOpts’
      In a stmt of a 'do' block:
        targetDir <- resolveDir' $ sdoptsTarPath sdistOpts
      In the second argument of ‘($)’, namely
        ‘do { targetDir <- resolveDir' $ sdoptsTarPath sdistOpts;
              targetTarPath <- (targetDir </>) <$> parseRelFile tarName;
              D.createDirectoryIfMissing True (toFilePath targetDir);
              D.copyFile (toFilePath tarPath) (toFilePath targetTarPath) }’
```
As far as I understand, it says that `resolveDir'` can accept only `String` but not `Maybe String`, what should I do in such a case?

snoyberg: You're probably going to need to pattern match in that part of the code to determine whether to use the original default path or this new configuration value.

snoyberg:
Bumping this PR during a triage @Asalle

PR #3632, opened 2017-12-08
Fix non-buildable bug in Cabal <= 1.22

snoyberg:

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -356,7 +356,7 @@ packageFromPackageDescription packageConfig pkgFlags (PackageDescriptionPair pkg
     pkgId = package pkg
     name = fromCabalPackageName (pkgName pkgId)
     deps = M.filterWithKey (const . not . isMe) (M.union
-        (packageDependencies pkg)
+        (packageDependencies (packageConfigCompilerVersion packageConfig) pkg)
         -- We include all custom-setup deps - if present - in the
         -- package deps themselves. Stack always works with the
         -- invariant that there will be a single installed package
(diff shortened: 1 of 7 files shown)


snoyberg:
@borsboom If we're keeping a maintenance branch for future 1.6 point releases, this is a could candidate for inclusion, as it's a relatively low impact change.

src/Stack/Package.hs
```diff
@@ -602,12 +602,36 @@ getBuildComponentDir Nothing = Nothing
 getBuildComponentDir (Just name) = parseRelDir (name FilePath.</> (name ++ "-tmp"))
 
 -- | Get all dependencies of the package (buildable targets only).
-packageDependencies :: PackageDescription -> Map PackageName VersionRange
-packageDependencies pkg =
+--
+-- Note that for Cabal versions 1.22 and earlier, there is a bug where
+-- Cabal requires dependencies for non-buildable components to be
+-- present. We're going to use GHC version as a proxy for Cabal
```
mgsloan: Hmm, I don't see the reasoning of using a GHC version as a proxy for the cabal version.  AFAIU you can use Cabal 2 with GHC 7.8

snoyberg: I considered going that route, but it's a much bigger diff, and gets into a bit of a circular issue, since we need to know the version of the Cabal library before we've finished calculating the build plan. The way I've written it, we'll occasionally (but rarely) get some unnecessary extra packages built if someone is using old GHC + new Cabal, but won't get any breakage (since you can't use old Cabal with a new GHC).

mgsloan: I see!  Alright fair 'nough

borsboom:
> If we're keeping a maintenance branch for future 1.6 point releases, this is a could candidate for inclusion, as it's a relatively low impact change.

The `stable` branch is perfect for this.  Feel free to retarget this PR to that branch, or cherry-pick it after merging to master.

snoyberg:
Rebased against stable. After Travis passes, I'll merge and then merge stable to master.

snoyberg:
For posterity, here's the original Cabal bug: https://github.com/haskell/cabal/issues/1725

Issue #3486, opened 2017-10-12
stack build copies unnecessary executables when asked to build just a single executable


athas:
### Steps to reproduce

1. Have a repository with several executable targets.  Say, the project is `foo`, and the executables are `exe1`, `exe2`, `exe3`.
2. Run `stack build foo:exe3`

### Expected

Stack should build the `exe3` target (and any dirty dependencies) and then *only* copy the `exe3` binary from the `.stack-work/dist/.../build/` directory to `.stack-work/install/.../bin/`.

### Actual

Stack copies *all* executables in the package, even the ones with no changes and which it was not asked to copy, which is slow if there are a lot of them.

### Stack version

```
$ stack --version
Version 1.5.1, Git revision 96e83a05fbeafb5cc3b1f52195433101c1f6ab8a x86_64 hpack-0.18.1
```

### Method of installation

Via `stack upgrade`.

mgsloan:
This is a symptom of https://github.com/commercialhaskell/stack/issues/1406 , which at its root is a cabal issue.  It seemed like this was going to get fixed in cabal, there was even a PR, but it got left out of releases for some reason.

Worth revisiting.  For now, you'll have to live with it.  One option is to add cabal flags to set `buildable: false` on particular executables.

23Skidoo:
> This is a symptom of #1406 , which at its root is a cabal issue. It seemed like this was going to get fixed in cabal, there was even a PR, but it got left out of releases for some reason.

Can someone please summarise the issue and post a ticket to the Cabal bug tracker (or post a comment in an existing thread, if there's already a ticket)?

mgsloan:
@23Skidoo Sorry for the delay on following up on this.  I've opened https://github.com/commercialhaskell/stack/issues/3653 to track looking into whether this is resolved in Cabal-2.0, and changing how things work if it does.  If not, an issue will be opened to track it.

mgsloan:
Based on ezyang's comment in  https://github.com/haskell/cabal/issues/2780#issuecomment-354490838 , it should be feasible to move forward on this at least when Cabal >= 2.0

PR #3652, opened 2017-12-14
Add "lib" as an available comptype

tfausak:


tfausak:
Wait... I'm not sure this does what I think it does. My goal is to build only the library -- not any executables. Is this the right way to do that?

mgsloan:
@tfausak Unfortunately, it doesn't quite work - see https://github.com/commercialhaskell/stack/issues/3653

Makes sense to document `pkg:lib` here, but note that there is no `pkg:lib:something-else`

tfausak:
Ah, gotcha.

Will `pkg:lib:blah` work with Cabal 2's multiple libraries per package idea?

mgsloan:
Perhaps so!  I don't think that's plumbed through stack quite yet

lwm:
I think this all boils down to https://github.com/haskell/cabal/issues/2780!? Nice maze there ... not sure I made it out the right way. If that is the blocking issue, should we close here and move over to a ticket to track when this will be possible?

mgsloan:
Tracked by https://github.com/commercialhaskell/stack/issues/3486 .  It sounds like some form of per component build is possible with Cabal 2.0, but how to do that and how it will affect efficiency is unclear.

Issue #3647, opened 2017-12-13
Unable to extract remote tgz


eddieschoute:
### General summary/comments
I am trying to add a remote dependency hosted at `http://www.mathstat.dal.ca/~selinger/quipper/downloads/quipper-0.8.tgz` but the extraction of the archive fails.

### Steps to reproduce
1. Create a new project.
2. Add to `extra-deps` in `stack.yaml`:
```
extra-deps:
- http://www.mathstat.dal.ca/~selinger/quipper/downloads/quipper-0.8.tgz
```
3. Run `stack build`

[stack.yaml.txt](https://github.com/commercialhaskell/stack/files/1556338/stack.yaml.txt)

### Expected

I expect it to extract the archive for use as a dependecy. Downloading the archive manually and extracting using `gunzip` and `tar` does work.

### Actual

I get an error message saying that the extraction failed because the file type is not supported.

```
$ stack build --verbose
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0
2017-12-13 12:02:22.545104: [debug] Checking for project config at: /Users/eddie/dev/quipper-dep/stack.yaml
@(Stack/Config.hs:842:9)
2017-12-13 12:02:22.547226: [debug] Loading project config file stack.yaml
@(Stack/Config.hs:868:13)
2017-12-13 12:02:22.548884: [debug] Decoding build plan from: /Users/eddie/.stack/build-plan/lts-9.18.yaml
@(Stack/Snapshot.hs:150:5)
2017-12-13 12:02:22.548967: [debug] Trying to decode /Users/eddie/.stack/build-plan-cache/lts-9.18.cache
@(Data/Store/VersionTagged.hs:66:5)
2017-12-13 12:02:22.553569: [debug] Success decoding /Users/eddie/.stack/build-plan-cache/lts-9.18.cache
@(Data/Store/VersionTagged.hs:70:13)
2017-12-13 12:02:22.553940: [debug] Using standard GHC build
@(Stack/Setup.hs:619:9)
2017-12-13 12:02:22.554253: [debug] Getting global package database location
@(Stack/GhcPkg.hs:46:5)
2017-12-13 12:02:22.554872: [debug] Asking GHC for its version
@(Stack/Setup/Installed.hs:98:13)
2017-12-13 12:02:22.555007: [debug] Getting Cabal package version
@(Stack/GhcPkg.hs:185:5)
[...]

PR #3655, opened 2017-12-14
Support .tar archives #3647

snoyberg:

### DIFF
--- a/src/Stack/PackageLocation.hs
+++ b/src/Stack/PackageLocation.hs
@@ -109,31 +109,34 @@ resolveSinglePackageLocation projRoot (PLArchive (Archive url subdir msha)) = do
 
         let fp = toFilePath file
 
-        let tryTargz = do
+        withBinaryFile fp ReadMode $ \h -> do
+          -- Share a single file read among all of the different
+          -- parsing attempts. We're not worried about unbounded
+          -- memory usage, as we will detect almost immediately if
+          -- this is the wrong type of file.
+          lbs <- liftIO $ L.hGetContents h
+
+          let tryTargz = do
                 logDebug $ "Trying to ungzip/untar " <> T.pack fp
-                liftIO $ withBinaryFile fp ReadMode $ \h -> do
-                    lbs <- L.hGetContents h
-                    let entries = Tar.read $ GZip.decompress lbs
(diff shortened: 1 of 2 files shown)


src/Stack/PackageLocation.hs
```diff
@@ -120,14 +120,20 @@ resolveSinglePackageLocation projRoot (PLArchive (Archive url subdir msha)) = do
                 archive <- fmap Zip.toArchive $ liftIO $ L.readFile fp
                 liftIO $  Zip.extractFilesFromArchive [Zip.OptDestination
                                                        (toFilePath dirTmp)] archive
+            tryTar = do
+                logDebug $ "Trying to untar (no ungzip) " <> T.pack fp
+                liftIO $ withBinaryFile fp ReadMode $ \h -> do
```
mgsloan: I think the current code will end up reading the file 3 times.  Though perhaps it wouldn't be as bad as that, since the decoding should fail quickly.

Might be worth pulling `withBinaryFile` + `hGetContents` out to the top level.  Could increase memory residency, but that seems fine.

I'd actually lean towards using something like `L.readFile`, but I guess the exception handling isn't so great there w.r.t. closing the handle.  This is a side note, but I'm a bit surprised the code wouldn't use a weak pointer and also close the handle of the lbs is GCed.

Note that `L.readFile` is used above in the unzip code.

snoyberg: I'd much rather have multiple file reads than potentially unbounded memory usage, though in this case unbounded is basically impossible: parsing should fail immediately. I'll update this to do one read. Note that on master, I've already changed this to use a new helper `withLazyFile` instead of explicit `withBinaryFile` and `hGetContents`.

mgsloan: Cool, makes sense!

Why merge this into stable instead of master?

snoyberg: I was thinking that it's a small enough change to include in a patch. Since we're actively planning on such a patch, I'm trying to include as many minor, backwards compatible changes as I can. You think it's better to put this one on master?

mgsloan: Seems fine to me to have it in the patch version to, just curious

PR #3657, opened 2017-12-15
Update default generated stack.yaml to extensible snapshots

dbaynard:
Stack version
```
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0 
```
uses extensible snapshots, but the `stack.yaml` generated using `stack new` describes the old syntax. This PR corrects the errors and adds some description.

I think the update is correct, but I'm not sure how best to handle the `extra-deps` case — in any case, it seemed just as easy to open a PR as an issue about it.

---

Probably not relevant:

### DIFF
--- a/src/Stack/Init.hs
+++ b/src/Stack/Init.hs
@@ -239,7 +239,7 @@ renderStackYaml p ignoredPackages dupPackages =
         [ ("user-message"     , userMsgHelp)
         , ("resolver"         , resolverHelp)
         , ("packages"         , packageHelp)
-        , ("extra-deps"       , "# Dependency packages to be pulled from upstream that are not in the resolver\n# (e.g., acme-missiles-0.3)")
+        , ("extra-deps"       , "# Dependency packages to be pulled from upstream that are not in the resolver\n# using the same syntax as packages.\n# (e.g., acme-missiles-0.3)")
         , ("flags"            , "# Override default flag values for local packages and extra-deps")
         , ("extra-package-dbs", "# Extra package databases containing global packages")
         ]
@@ -262,9 +262,12 @@ renderStackYaml p ignoredPackages dupPackages =
         , "resolver: nightly-2015-09-21"
         , "resolver: ghc-7.10.2"
         , "resolver: ghcjs-0.1.0_ghc-7.10.2"
-        , "resolver:"
-        , " name: custom-snapshot"
-        , " location: \"./custom-snapshot.yaml\""
+        , ""
+        , "The location of a snapshot can be provided as a file or url. Stack assumes"
+        , "a snapshot provided as a file might change, whereas a url resource does not."
+        , ""
+        , "resolver: ./custom-snapshot.yaml"
+        , "resolver: https://example.com/snapshots/2018-01-01.yaml"
         ]
 
     userMsgHelp = commentHelp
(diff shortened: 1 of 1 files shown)


lwm:
I've restarted that timed out build @dbaynard! I'm not sure about the extra-deps either - but rest looks good!

src/Stack/Init.hs
```diff
@@ -281,14 +284,9 @@ renderStackYaml p ignoredPackages dupPackages =
         , "   git: https://github.com/commercialhaskell/stack.git"
         , "   commit: e7b331f14bcffb8367cd58fbfc8b40ec7642100a"
         , "- location: https://github.com/commercialhaskell/stack/commit/e7b331f14bcffb8367cd58fbfc8b40ec7642100a"
-        , "  extra-dep: true"
```
mgsloan: Hmm, did this become a default? `extra-dep: true` is probably still good to recommend.  Alternatively, now they can go in extra-deps, maybe it'd be best to switch to the new syntax (see https://www.fpcomplete.com/blog/2017/07/stacks-new-extensible-snapshots)

src/Stack/Init.hs
```diff
@@ -239,7 +239,7 @@ renderStackYaml p ignoredPackages dupPackages =
         [ ("user-message"     , userMsgHelp)
         , ("resolver"         , resolverHelp)
         , ("packages"         , packageHelp)
-        , ("extra-deps"       , "# Dependency packages to be pulled from upstream that are not in the resolver\n# (e.g., acme-missiles-0.3)")
+        , ("extra-deps"       , "# Dependency packages to be pulled from upstream that are not in the resolver\n# using the same syntax as packages.\n# (e.g., acme-missiles-0.3)")
```
mgsloan: Hmm, not sure how I see how this is helpful.  It's kinda like "Specify packages by using the same syntax as packages".

mgsloan:
Looks good, not sure about a couple of changes - added comments

src/Stack/Init.hs
dbaynard: Yes; I didn't want to repeat `packageHelp`, though. Although this implementation (With the newline is unclear; it should probably be factored out. What do you think?

src/Stack/Init.hs
dbaynard: The post claims `extra-dep: true` has caused enough confusion to be worth removing, even at the expense of backwards compatilbility — whether or not it is the default. If this is indeed the case then the earlier the default switches to the new syntax, the better.

src/Stack/Init.hs
```diff
@@ -281,14 +284,9 @@ renderStackYaml p ignoredPackages dupPackages =
...
         , "   commit: e7b331f14bcffb8367cd58fbfc8b40ec7642100a"
         , "- location: https://github.com/commercialhaskell/stack/commit/e7b331f14bcffb8367cd58fbfc8b40ec7642100a"
-        , "  extra-dep: true"
         , " subdirs:"
         , " - auto-update"
         , " - wai"
-        , ""
-        , "A package marked 'extra-dep: true' will only be built if demanded by a"
-        , "non-dependency (i.e. a user package), and its test suites and benchmarks"
-        , "will not be run. This is useful for tweaking upstream packages."
```
dbaynard: Likewise, @mgsloan, given the change to `extra-dep` I thought this was confusing, too — and I'm not sure I fully understand it.

dbaynard:
@mgsloan I've responded, though I'm not wedded to any particular phrasing. It just looked like it had been missed from the updates to the rest of stack (which is itself an issue to which I have no solution).

src/Stack/Init.hs
mgsloan: :+1:

src/Stack/Init.hs
mgsloan: :+1:

src/Stack/Init.hs
mgsloan: Ahh, I see, same as the "packages field"  I get it now.

mgsloan:
LGTM, thanks!

mgsloan:
I pushed a commit adjusting the wording a bit https://github.com/commercialhaskell/stack/commit/fc7a4c2239175bd07cbc1bf3122afc7fb085cb6a

dbaynard:
Great, I'm happy to have been able to help.

Issue #3663, opened 2017-12-16
Benchmarks are run in parallel


AndreasPK:
### General summary/comments (optional)

This seems to be a regression (https://github.com/commercialhaskell/stack/issues/1913)

### Steps to reproduce

```
git clone https://github.com/snoyberg/mono-traversable/tree/96d188bd6d7ba065aeff5ed9b3519107202069d3 mono
cd mono
stack bench
stack bench
```

### Expected

That stack runs the benchmarks in sequence.

### Actual

Stack runs the benchmarks in parallel.

```
$ stack bench --verbose
PS C:\ghc\test-stack\mono-trav2> stack bench --verbose
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0
2017-12-16 23:04:33.129747: [debug] Checking for project config at: C:\ghc\test-stack\mono-trav2\stack.yaml
@(Stack\Config.hs:842:9)
2017-12-16 23:04:33.132551: [debug] Loading project config file stack.yaml
@(Stack\Config.hs:868:13)
2017-12-16 23:04:33.135088: [debug] Decoding build plan from: C:\stack\build-plan\lts-9.9.yaml
@(Stack\Snapshot.hs:150:5)
2017-12-16 23:04:33.135088: [debug] Trying to decode C:\stack\build-plan-cache\lts-9.9.cache
@(Data\Store\VersionTagged.hs:66:5)
2017-12-16 23:04:33.138676: [debug] Success decoding C:\stack\build-plan-cache\lts-9.9.cache
@(Data\Store\VersionTagged.hs:70:13)
2017-12-16 23:04:33.143665: [debug] Using standard GHC build
@(Stack\Setup.hs:619:9)
2017-12-16 23:04:33.146171: [debug] Getting global package database location
@(Stack\GhcPkg.hs:46:5)
2017-12-16 23:04:33.147175: [debug] Getting Cabal package version
@(Stack\GhcPkg.hs:185:5)
2017-12-16 23:04:33.147675: [debug] Asking GHC for its version
@(Stack\Setup\Installed.hs:98:13)
2017-12-16 23:04:33.148177: [debug] Run process: C:\Users\Andi\AppData\Local\Programs\stack\x86_64-windows\ghc-8.0.2\bin\ghc-pkg.EXE --no-user-package-db list --global
@(System\Process\Log.hs:37:3)
[...]

PR #3666, opened 2017-12-17
Never run benchmarks concurrently, always output to console #3663

mgsloan:
Also generally cleans up code related to parallel execution of tasks. Instead of
locking happening among "final tasks" (tests and benchmark running), it's now
possible to mark some tasks as work that shouldn't be done in parallel with
anything else.  This is what makes sense for benchmark running - they shouldn't
be run concurrently with either building or running tests.

Previously benchmarks and tests shared the same final task. The mechanism to
execute one task exclusively is part of Control.Concurrent.Execute. If they were
kept in the same task, then if any benchmarks were enabled, then tests would be
run without any concurrency. In order to have as much concurrency as possible,
they are now split into two different "final" tasks.

Tested via the repro in #3663

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -42,13 +42,16 @@ Bug fixes:
   arguments. See [#3658](https://github.com/commercialhaskell/stack/issues/3658).
   In particular, this makes it possible to pass `-- +RTS ... -RTS` to specify
   RTS arguments used when running the script.
-
+* Benchmarks used to be run concurrently with other benchmarks
+  and build steps. This is non-ideal because CPU usage of other processes
+  may interfere with benchmarks. It also prevented benchmark output from
+  being displayed by default. This is now fixed. See
+  [#3663](https://github.com/commercialhaskell/stack/issues/3663).
 
 ## v1.6.1.1
 
 Hackage-only release with no user facing changes (updated to build with
 newer dependency versions).
-
 
 ## v1.6.1
 
--- a/src/Control/Concurrent/Execute.hs
+++ b/src/Control/Concurrent/Execute.hs
@@ -8,32 +8,49 @@ module Control.Concurrent.Execute
     , ActionId (..)
     , ActionContext (..)
     , Action (..)
+    , Concurrency(..)
     , runActions
     ) where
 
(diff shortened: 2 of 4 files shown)


mgsloan:
I quite like that now all benchmark output will be visible.  Might consider doing this for tests too if [concurrent-tests](https://docs.haskellstack.org/en/stable/yaml_configuration/#concurrent-tests) is set to `false`.  However, as implemented here, when this option is set to false, it will still run tests concurrently with build steps, so they can't be sent to the output.

ChangeLog.md
```diff
@@ -18,6 +18,9 @@ Bug fixes:
 * Run the Cabal file checking in the `sdist` command more reliably by
   allowing the Cabal library to flatten the
   `GenericPackageDescription` itself.
+* Benchmarks used to be run concurrently with other benchmarks
```
snoyberg: I'm assuming the reason for this is to not confuse the benchmark results with other CPU activities, right? May be worth pointing that out.

src/Control/Concurrent/Execute.hs
```diff
@@ -56,26 +72,31 @@ instance Show ExecuteException where
...
         then runActions' es
         else replicateConcurrently_ threads $ runActions' es
     readTVarIO $ esExceptions es
 
+-- | Sort actions such that those that can't be run concurrently are at
+-- the end.
+sortActions :: [Action] -> [Action]
+sortActions = sortBy (compareConcurrency `on` actionConcurrency)
+  where
+    compareConcurrency ConcurrencyAllowed ConcurrencyDisallowed = LT
```
snoyberg: What about `deriving Ord` on the `Concurrency` type?

mgsloan: I considered that, but it would be easy for someone to add a constructor and not realize its impact on the usage of Ord.  I've added a note about it.

ChangeLog.md
mgsloan: Yep, done!

mgsloan:
@snoyberg I've addressed your feedback.  I'm particularly interested to know if you like the change to default to outputting benchmark results?

Issue #3658, opened 2017-12-15
Error when using RTS options with the Stack script interpreter


danclien:
### General

Using Stack as a script interpreter doesn't accept RTS options as noted in the [Stack documentation](https://docs.haskellstack.org/en/stable/GUIDE/#specifying-interpreter-options).

I'm not sure if there's a bug with Stack or an error in the documentation.

### Steps to reproduce

1. Put the following in a file:
    * The top section is copied and pasted from the [Stack documentation](https://docs.haskellstack.org/en/stable/GUIDE/#specifying-interpreter-options).
    ```
    #!/usr/bin/env stack
    {- stack
      script
      --resolver lts-6.25
      --package turtle
      --
      +RTS -s -RTS
    -}
    
    main :: IO
    main = putStrLn "Hello, world!"
    ```
2. Make the file executable
    ```
    $ chmod +x <file>
    ```
3. Execute the file
    ```
    $ ./<file>
    ```

### Expected

The script should print `Hello, world!`.

### Actual

The script errors out with the following:
```
target ‘+RTS’ is not a module name or a source file
```

#### Output
```
$ stack --version
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0
$ cat ./StackScriptTest.hs
#!/usr/bin/env stack
{- stack
  script
  --resolver lts-6.25
  --package turtle
  --
  +RTS -s -RTS
-}

main :: IO
main = putStrLn "Hello, world!"
$ ./StackScriptTest.hs
target ‘/home/dacali/projects/onefile/+RTS’ is not a module name or a source file
```

### Stack version

```
$ stack --version
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0
```

### Method of installation

* Upgraded with `stack upgrade`

mgsloan:
Hi!  Thanks for the report.  I've opened https://github.com/commercialhaskell/stack/pull/3668 which resolves this issue.

danclien:
@mgsloan, thanks for the quick response!

lwm:
Done in #3668. Thanks for the report @danclien.

PR #3668, opened 2017-12-18
Put file arguments first when stack is run as script interpreter #3658

mgsloan:

Tested via #3658 and tested that a few misc other uses of stack as an interpreter still work.  Seems to be fine, but this is potentially a change in behavior.  I think it's a positive change, though, I know of cases that it fixes, and don't know of any cases that it breaks.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,6 +18,10 @@ Bug fixes:
 * Run the Cabal file checking in the `sdist` command more reliably by
   allowing the Cabal library to flatten the
   `GenericPackageDescription` itself.
+* The script interpreter's implicit file arguments are now passed before other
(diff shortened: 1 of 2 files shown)


ChangeLog.md
```diff
@@ -18,6 +18,8 @@ Bug fixes:
 * Run the Cabal file checking in the `sdist` command more reliably by
   allowing the Cabal library to flatten the
   `GenericPackageDescription` itself.
+* Script interpreter implicit file argumenst are now passed before other
```
lwm: Typo here in arguments - also, this might not be obvious to someone coming here to find why something broke (if that might happen) - perhaps you can say 'like +RTS' or something like that, from the use case in #3658.

mgsloan: Good catch! Fixed

Issue #2689, opened 2016-10-09
[Stack User Guide] Stack Build error when adding dependencies 


pengo98:
I'm following the examples in the user guide to learn how Stack works. After adding dependencies (updating src/Lib.hs and cabal file) as suggested in the doc, I still get a build error.
### Steps to reproduce

a. Modify `src/Lib.hs` as follows.

```
{-# LANGUAGE OverloadedStrings #-}
module Lib
    ( someFunc
    ) where

import qualified Data.Text.IO as T

someFunc :: IO ()
someFunc = T.putStrLn "someFunc"

```

b. Add `text` to build-depends in Cabal file as follows.

```
name:                helloworld
version:             0.1.0.0
synopsis:            Initial project template from stack
description:         Please see README.md
homepage:            https://github.com/githubuser/helloworld#readme
license:             BSD3
license-file:        LICENSE
author:              Author name here
maintainer:          example@example.com
copyright:           2016 Author name here
category:            Web
build-type:          Simple
-- extra-source-files:
cabal-version:       >=1.10

library
  hs-source-dirs:      src
  exposed-modules:     Lib
  build-depends:       base >= 4.7 && < 5
  default-language:    Haskell2010

executable helloworld-exe
  hs-source-dirs:      app
  main-is:             Main.hs
  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
  build-depends:       base
                     , helloworld
                     , text
  default-language:    Haskell2010

test-suite helloworld-test
  type:                exitcode-stdio-1.0
  hs-source-dirs:      test
  main-is:             Spec.hs
  build-depends:       base
                     , helloworld
  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
  default-language:    Haskell2010

source-repository head
  type:     git
  location: https://github.com/githubuser/helloworld

```

c. Run command `stack build`

```
$ stack build
helloworld-0.1.0.0: configure
Configuring helloworld-0.1.0.0...
helloworld-0.1.0.0: build
Preprocessing library helloworld-0.1.0.0...
[1 of 1] Compiling Lib              ( src\Lib.hs, .stack-work\dist\b7fec021\build\Lib.o )

C:\Users\nelson.lin\Source\Haskell\helloworld\src\Lib.hs:6:1: error:
    Failed to load interface for `Data.Text.IO'
    It is a member of the hidden package `text-1.2.2.1'.
    Perhaps you need to add `text' to the build-depends in your .cabal file.
[...]

PR #3673, opened 2017-12-19
Take a pass at the user guide documentation

### LINKED ISSUE #3633: Documentation on adding dependencies doesn't mention hpack/package.yaml
https://docs.haskellstack.org/en/stable/GUIDE/#adding-dependencies mentions how to add dependencies, but it only talks about the `.cabal` file.

Since the default template has been updated to use hpack, it really should at least mention `package.yaml`. With people on older versions of `stack` (v1.5.1 for example) this is especially confusing because it just silently overwrites the .cabal file they just changed.

In v1.6.1 it's not so bad really, but it'd still be nice if they didn't have to get that warning from modifying the `.cabal` file when they really should be editing the `package.yaml` file.

lwm:
One small nitpick on the CI setup.

The rest was a pass on the user guide prompted by https://github.com/commercialhaskell/stack/issues/3633 but I kinda got carried away when I read through it. I was using a clean build from `docker run -it ubuntu:latest bash` and found a lot of the output very out of date. So, I tried to future proof the content and add links and more references to `hpack`/`package.yaml`.

Check out [the rendered version](https://github.com/lwm/stack/blob/ab352c6834b16581b2fbb00897df969afe15b906/doc/GUIDE.md).

Changes kick off [from here](https://github.com/lwm/stack/blob/ab352c6834b16581b2fbb00897df969afe15b906/doc/GUIDE.md#downloading-and-installation) and end [over here](https://github.com/lwm/stack/blob/ab352c6834b16581b2fbb00897df969afe15b906/doc/GUIDE.md#different-databases).

Should close #3633 and #2689.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -58,16 +58,12 @@ The rest of this session will demonstrate the installation procedure on a
 vanilla Ubuntu 14.04 machine.
 
 ```
-# Starting with a *really* bare machine
 michael@d30748af6d3d:~$ sudo apt-get install wget
-# Demonstrate that stack really isn't available
-michael@d30748af6d3d:~$ stack
--bash: stack: command not found
-# Install stack
-wget -qO- https://get.haskellstack.org/ | sh
-# downloading...
-michael@d30748af6d3d:~$ stack --version
-Version 0.1.3.1, Git revision 908b04205e6f436d4a5f420b1c6c646ed2b804d7
+# installing ...
+michael@d30748af6d3d:~$ wget -qO- https://get.haskellstack.org/ | sh
+# downloading ...
+michael@d30748af6d3d:~$ stack --help
+# help output ...
 ```
 
 With stack now up and running, you're good to go. Though not required, we
@@ -89,8 +85,8 @@ will contain a Haskell *package* of the same name. So let's pick a valid
 package name first:
 
 > A package is identified by a globally-unique package name, which consists
-of one or more alphanumeric words separated by hyphens. To avoid ambiguity,
-each of these words should contain at least one letter.
+> of one or more alphanumeric words separated by hyphens. To avoid ambiguity,
+> each of these words should contain at least one letter.
 
 (From the [Cabal users guide](https://www.haskell.org/cabal/users-guide/developing-packages.html#developing-packages))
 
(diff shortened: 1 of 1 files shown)


doc/GUIDE.md
```diff
@@ -267,8 +193,8 @@ test/Spec.hs
 The `app/Main.hs`, `src/Lib.hs`, and `test/Spec.hs` files are all Haskell source
 files that compose the actual functionality of our project (we won't dwell on
 them here). The `LICENSE` file has no impact on the build, but is there for
-informational/legal purposes only. The files of interest here are `Setup.hs`,
-`helloworld.cabal`, and `stack.yaml`.
+informational/legal purposes only, as well as the `README.md`. The files of interest here
```
kadoban: This is a little awkwardly worded. Maybe something like "The `LICENSE` and `README.md` files have no impact on [...]"

doc/GUIDE.md
```diff
@@ -305,21 +223,30 @@ value here says to use [LTS Haskell version
...
-*components*: a library, executables, test suites, and benchmarks. It also
-specifies additional information such as library dependencies, default language
+The files of final importance are the `helloworld.cabal` and `package.yaml`.
+
+Since Stack 1.6.1, the `package.yaml` is an alternative package format that is
+provided built-in by stack through [the hpack tool](https://github.com/sol/hpack). The
+default behaviour is to generate the `.cabal` file from this `package.yaml`, and you
+can see that from the general message header in your new `helloworld.cabal` file. It
+is therefore important to understand that your `package.yaml` drives the package configuration.
+
```
kadoban: Do you think it would be a good idea around here to explicitly mention that they usually shouldn't be modifying the .cabal file? It's certainly strongly implied, but those who aren't familiar with hpack might still not realize that exactly.

kadoban:
Those look like a bunch of really positive changes, nice work!

I didn't get to reviewing everything in depth, but what I saw looked great. I added a couple of pretty nitpicky comments if you're interested, and I'll try to review more completely in the morning.

lwm:
Thanks @kadoban! I agree and pushed some fixups. That'd be great if you get another look at it.

doc/GUIDE.md
```diff
@@ -103,94 +99,33 @@ michael@d30748af6d3d:~$ stack new helloworld new-template
 
 For this first stack command, there's quite a bit of initial setup it needs to
 do (such as downloading the list of packages available upstream), so you'll see
-a lot of output. Though your exact results may vary, below is an example of the
-sort of output you will see. Over the course of this guide a lot of the content
-will begin to make more sense:
+a lot of output. Over the course of this guide a lot of the content will begin
+to make more sense:
```
kadoban: Probably a period instead of a colon here, since the output it was pointing to is removed now.

doc/GUIDE.md
```diff
@@ -103,94 +99,33 @@ michael@d30748af6d3d:~$ stack new helloworld new-template
...
-[1 of 1] Compiling Main             ( app/Main.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/helloworld-exe/helloworld-exe-tmp/Main.o )
-Linking .stack-work/dist/x86_64-linux/Cabal-1.22.4.0/build/helloworld-exe/helloworld-exe ...
-helloworld-0.1.0.0: install
-Installing library in
-/home/michael/helloworld/.stack-work/install/x86_64-linux/lts-3.2/7.10.2/lib/x86_64-linux-ghc-7.10.2/helloworld-0.1.0.0-6urpPe0MO7OHasGCFSyIAT
-Installing executable(s) in
-/home/michael/helloworld/.stack-work/install/x86_64-linux/lts-3.2/7.10.2/bin
-Registering helloworld-0.1.0.0...
-```
+__NOTE__: GHC will be installed to your global stack root directory, so
```
kadoban: Can't think of how better to word this part, but the "global stack root directory" seems not the best wording to me for a beginner. Maybe something about "GHC will be installed to an internal stack location," or maybe just remove the word "global" ? Or do we call STACK_ROOT or whatever it is the "global stack root" elsewhere?

I might just be being picky here, feel free to ignore, haha.

doc/GUIDE.md
```diff
@@ -305,21 +223,30 @@ value here says to use [LTS Haskell version
...
+Since Stack 1.6.1, the `package.yaml` is an alternative package format that is
+provided built-in by stack through [the hpack tool](https://github.com/sol/hpack). The
+default behaviour is to generate the `.cabal` file from this `package.yaml`, and you
+can see that from the general message header in your new `helloworld.cabal` file. It
+is therefore important to understand that your `package.yaml` drives the package configuration.
+
+It is also important to remember that stack is built on top of the Cabal build system. Therefore, an
+understanding of the moving parts in Cabal are necessary. In Cabal, we have individual
+*packages*, each of which contains a single `.cabal` file. The `.cabal` file
+can define 1 or more *components*: a library, executables, test suites, and benchmarks.
```
kadoban: s/1/one/ though I guess it depends on the style guide.

doc/GUIDE.md
```diff
@@ -498,75 +378,57 @@ someFunc :: IO ()
...
 
 With that out of the way, let's dig a little bit more into these package sets,
-also known as *snapshots*. We mentioned lts-3.2, and you can get quite a bit of
-information about it at
-[https://www.stackage.org/lts-3.2](https://www.stackage.org/lts-3.2), including:
+also known as *snapshots*. We mentioned the LTS resolvers, and you can get quite a bit of
+information about it at [https://www.stackage.org/lts](https://www.stackage.org/lts), including:
 
-* The appropriate resolver value (`resolver: lts-3.2`, as we used above)
+* The appropriate resolver value (`resolver: lts-10.0`, as is currently the latest LTS)
```
kadoban: Can we word this differently so it doesn't need to be maintained? We're guaranteed to forget to update this, so it won't be the "currently latest" one anymore soon.

doc/GUIDE.md
```diff
@@ -498,75 +378,57 @@ someFunc :: IO ()
...
 * The GHC version used
 * A full list of all packages available in this snapshot
 * The ability to perform a Hoogle search on the packages in this snapshot
-* A [list of all modules](https://www.stackage.org/lts-3.2/docs) in a snapshot,
+* A [list of all modules](https://www.stackage.org/lts-10.0/docs) in a snapshot,
```
kadoban: If we link to https://www.stackage.org/lts/docs that always seems to give the newest one. Is that worth doing, or think that'll be confusing?

doc/GUIDE.md
kadoban: Might not really be possible, looking below it seems like it's going to be hard to word all of these so it doesn't specify an actual specific resolver (without it being uselessly confusing to beginners).

doc/GUIDE.md
```diff
@@ -577,46 +439,25 @@ default as well).
...
 We can also change resolvers on the command line, which can be useful in a
 Continuous Integration (CI) setting, like on Travis. For example:
 
 ```
-michael@d30748af6d3d:~/helloworld$ stack --resolver lts-3.1 build
-Downloaded lts-3.1 build plan.
-Caching build plan
-stm-2.4.4: configure
-# Rest is the same, no point copying it
+michael@d30748af6d3d:~/helloworld$ stack --resolver lts-9.18 build
```
kadoban: Might want to avoid specifying a resolver with a different GHC than the resolver above. It's a little wasteful if they won't end up using that GHC ever.

doc/GUIDE.md
```diff
@@ -737,97 +542,37 @@ Haskell packages, this will often be enough to build most packages. However,
 at times, you may find that not all dependencies required may be available in
 the Stackage snapshots.
 
-Let's simulate an unsatisfied dependency by adding acme-missiles to our
-build-depends and re-initing:
+Let's simulate an unsatisfied dependency by adding acme-missiles to the `.cabal` file
+(yackage does not currently support hpack, but you can also [hpack-convert](https://github.com/yamadapc/hpack-convert)
```
kadoban: :( that's unfortunate. I wonder if we can fix that. Think a PR would be appropriate? The reason for changing to hpack seems a bit flimsy though, for an actual piece of software to have to change just to fit as a decent example in some other documentation ...

Or maybe it's a decent learning opportunity anyway, since they'll probably have to deal with at least some packages eventually that don't have a package.yaml?

Probably beyond the scope of this documentation change anyway though.

doc/GUIDE.md
```diff
@@ -919,8 +619,8 @@ under the `extra-deps` section.
 
 #### Installing the compiler
 
-You can install the required compiler if not already installed by using the
-`--install-ghc` flag with the `stack init` command.
+stack will automatically install the compiler when you run `stack build` but you can
+manually specify the compiler by running `stack setup <GHC-VERSION>`.
```
kadoban: Think this needs to be here, the manually specifying the GHC version part? At least if it stays it seems like it should be a separate sentence to me, and maybe mention that if they're trying to change what GHC is used by a package, they should modify the resolver in stack.yaml instead of hoping that'll do it.

doc/GUIDE.md
lwm: Oh, good catch.

doc/GUIDE.md
```diff
@@ -103,94 +99,33 @@ michael@d30748af6d3d:~$ stack new helloworld new-template
...
 ```
 
-stack needs GHC in order to build your project, and `stack setup` must be run to
-check whether GHC is available (and install it if not).
+stack needs a GHC in order to build your project. stack will discover that you
+are missing it and will install it for you. You can do this manually by using
+the `stack setup` command.
 
-Having run `stack setup` successfully, `stack build` should build our project:
+You'll get intermediate download percentage statistics while the download is
```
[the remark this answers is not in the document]
lwm: It's the same!?

lwm:
Thanks to you both! I've added a few more edits and will merge this now. Somewhat of an improvement, for now.

Issue #3628, opened 2017-12-06
Specify User-Agent to get over my company's firewall


igrep:
# Problem

My colleagues and I are using stack behind a (weirdly) strong firewall, which denies http requests whose User-Agent header and the destination URL match with the preset white list, since the recent change by the infra team.
So I asked the infra team to add stack and s3.amazonaws.com (and so on) to the white list, but he declined because stack accesses to various domains without any User-Agent header!!

# Solution

If you don't have a problem, I'll make a PR to add some User-Agent header (["The Haskell Stack" looks good](https://github.com/commercialhaskell/stack/blob/2d9721356270e11a37add59b772bde7174ebeb67/src/Network/HTTP/Download.hs#L115).
So can you tell me where I should fix?

# Version of stack I'm using

```
> stack --version
Version 1.5.1, Git revision 600c1f01435a10d127938709556c1682ecfd694e x86_64 hpack-0.17.1
```

mgsloan:
Sure, this would be an appreciated PR!  There are a fair number of places that do downloads.  Searching for `Network.` should turn up most of the modules that do web requests.

igrep:
Now I'm trying to which functions to wrap.
Then, I have a question: Shouldn't I touch `Hackage.Security.Client.Repository.HttpLib.HttpClient`?
Because that module said to be copied from the other. So is it better to keep as is?

mgsloan:
@igrep You might need to make some changes to hackage-security itself, I'm not sure if it sets user-agent.

PR #3682, opened 2017-12-20
Specify User-Agent on every HTTP request (#3628)

igrep:
Fix: https://github.com/commercialhaskell/stack/issues/3628

Wrap up functions in `Network.HTTP.Client` and `Network.HTTP.Simple` to
add User-Agent to the request object.

# Testing the change

I confirmed some requests are fixed by `stack setup`, `stack install`, `stack new`.
But I haven't tested all HTTP requests made by `stack`.

# For reviewers

I'm not sure especially with the name and place of the wrapper module (and the name of its functions).  
Feel free to modify for better module structure.


lwm:
This looks like a solid approach to me! Perhaps, to help avoiding using the wrong function and not the wrapped one in the future, we can use a similar HLint trick like in https://github.com/commercialhaskell/stack/blob/master/.hlint.yaml#L44 - where a warning is produced (and the CI will fail) if it isn't used - assuming we always want these headers to go out? I assume so!

igrep:
@lwm Thanks for advice!

> Perhaps, to help avoiding using the not wrapped function, we can use a similar HLint trick like in https://github.com/commercialhaskell/stack/blob/master/.hlint.yaml#L44

You mean, should I rename functions exported by `Network.HTTP.StackClient` to differ from functions in `Network.HTTP.Simple` and `Network.HTTP.Client`?
It looks like that `hlint` can suggest only when `lhs` and `rhs` are literally different as their de-qualified names.
I'm new to `.hlint.yaml`, so tell me if I get something wrong!

If I should rename, I'm going to prefix `stack` to the new functions: `stackHttpJSON`, `stackHttpLbs`, etc.

lwm:
I'm not exactly sure either ;) I thought you could pass full import paths (like over [here](https://github.com/ndmitchell/hlint/blob/master/data/hlint.yaml#L321)) and get away with it. If you can't do that, maybe the renaming with the prefix is a better option in general - less magic!

mgsloan:
Good stuff @igrep , thanks a bunch!

Cool idea bout using hlint to reduce likelihood of regressions, @lwm!  Indeed I'm not sure if HLint can distinguish where identifiers are imported from.  Seems like it'd be a major weakness if not, I'm guessing it can do this.  I've opened https://github.com/commercialhaskell/stack/issues/3700 to track implementing the idea.

Issue #3649, opened 2017-12-13
Failure on install


LeventErkok:
Sometime in the last 10 days, stack based appveyor runs started failing for me. I see the following error message:

> C:\Users\appveyor\AppData\Roaming\stack\precompiled\i386-windows\ghc-8.0.2\1.24.2.0\YmFzZTE2LWJ5dGVzdHJpbmctMC4xLjEuNkBzaGEyNTY6ZWFkNTM1YTMwZjkzMTg5MDY4YjM0NTc0OTZlMzBlM2NiOGE1NWQ0OGIyZDQ2ZjM5MzI1MjMzZDBkOTk4N2QyYywxMTY0\FfYKX4Bw2N-fo7uL_5to7LNegZSrqOlwYTCgKLrdW7w=: openBinaryFile: does not exist (No such file or directory)

Here's the full failing log: https://ci.appveyor.com/project/LeventErkok/sbv/build/1.0.159
Here's a corresponding working log (no changes in between): https://ci.appveyor.com/project/LeventErkok/sbv/build/1.0.152

Did something change recently in stack to cause this? I couldn't find anything relevant in the release notes, nor I changed anything in my own configuration.

lwm:
> Did something change recently in stack to cause this?

Many things :)

Well, that is strange. I took a look at the logs - still not sure.

Can you reproduce it locally?

LeventErkok:
Unfortunately, I don't have access to a Windows machine; that's why I'm using Appveyor in the first place..

snoyberg:
Looks to me like a Windows file path issue. Can you try setting `STACK_ROOT` to `c:\sr` and see if that fixes it? I don't know why this would only be an issue with Stack 1.6.1

LeventErkok:
@snoyberg That did the trick! Thanks.

TerrorJack:
@snoyberg I'm experiencing the same issue with `stack-1.6.1` and `ghc-8.2.2`. Does some logic in recent `stack` generate super long filenames that makes it fail on Windows? If so, this workaround should be documented.

mgsloan:
@TerrorJack Ah, thanks for pinging this issue.  Something did change, and I've opened https://github.com/commercialhaskell/stack/pull/3702 to resolve it.  Can you or @LeventErkok please try it out and let me know if it fixes the problem?

PR #3702, opened 2017-12-23
Avoid MAX_PATH for precompiled cache #3649

mgsloan:
Regression was in 99911566a0ac41203d8130bc3ae5fa724dd32128

I have not tested this change, since I do not currently have easy access to windows.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,9 @@ Behaviour changes:
 Other enhancements:
 
 Bug fixes:
+* 1.6.1 introduced a change that made some precompiled cache files use
+  longer paths, sometimes causing builds to fail on windows. This has been
+  fixed. See [#3649](https://github.com/commercialhaskell/stack/issues/3649)
 
 
 ## v1.6.3
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -32,6 +32,7 @@ module Stack.Build.Cache
     , BuildCache(..)
     ) where
 
+import           Stack.Constants
 import           Stack.Prelude
 import           Crypto.Hash (hashWith, SHA256(..))
 import           Control.Monad.Trans.Maybe
@@ -254,6 +255,14 @@ precompiledCacheFile loc copts installedPackageIDs = do
             PLRepo r -> Just $ T.unpack (repoCommit r) ++ repoSubdirs r
 
   forM mpkgRaw $ \pkgRaw -> do
+    platformRelDir <- platformGhcRelDir
+    let precompiledDir =
+              view stackRootL ec
+          </> $(mkRelDir "precompiled")
+          </> platformRelDir
+          </> compiler
+          </> cabal
+
     pkg <-
       case parseRelDir pkgRaw of
         Just x -> return x
@@ -263,7 +272,6 @@ precompiledCacheFile loc copts installedPackageIDs = do
                  $ B64URL.encode
                  $ TE.encodeUtf8
                  $ T.pack pkgRaw
-    platformRelDir <- platformGhcRelDir
 
     -- In Cabal versions 1.22 and later, the configure options contain the
     -- installed package IDs, which is what we need for a unique hash.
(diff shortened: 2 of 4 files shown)


borsboom:
Consider rebasing this to the `stable` branch so it's included in a future minor release (if any).

mgsloan:
Good point!  I have rebased and retargetted stable

mungre:
I tried this new version on Windows and I'm still getting what appear to be path length errors building projects.

An error occurs after building a small number of dependencies, usually less than ten.  The weird thing is that when I restart the build it continues as if nothing had gone wrong, and then fails again after building a few more dependencies.  If the build is repeatedly restarted then it will eventually finish.

I deleted stack's cache (`C:\users\mungre\AppData\Roaming\stack`) before building my project.

This is the output from a couple of turns around this loop:

    >stack --version
    Version 1.6.4, Git revision bf31549797845d38c980703d00e1eb247338a16f (5466 commits) x86_64 hpack-0.20.0

    >stack build
    basement-0.0.4: download
    byteable-0.1.1: download
    byteable-0.1.1: configure
    byteable-0.1.1: build
    basement-0.0.4: configure
    byteable-0.1.1: copy/register
    byteorder-1.0.4: download
    basement-0.0.4: build
    byteorder-1.0.4: configure
    byteorder-1.0.4: build
    byteorder-1.0.4: copy/register
    bytestring-builder-0.10.8.1.0: download
    bytestring-builder-0.10.8.1.0: configure
    bytestring-builder-0.10.8.1.0: build
    bytestring-builder-0.10.8.1.0: copy/register
    basement-0.0.4: copy/register
    Progress: 4/177C:\Users\mungre\AppData\Roaming\stack\precompiled\x86_64-windows\ghc-8.2.2\2.0.1.0\Ynl0ZXN0cmluZy1idWlsZGVyLTAuMTAuOC4xLjBAc2hhMjU2OjY5ZDQ0ZWNmY2YyNDMzOTBmNTgyYTUxNTkwNWQ2MTRiMGRjNWMzMThlYWQ2MjFjYTU2NDFjZmJkMGFhNTZkZmYsMzc0Ng==\1f1C--p9_VorahgpH19QXy0W5mnh-O-BAhKtT-cKew4=: openBinaryFile: does not exist (No such file or directory)

    >stack build
    cabal-doctest-1.0.4: download
    cereal-0.5.4.0: download
    cabal-doctest-1.0.4: configure
    cabal-doctest-1.0.4: build
    cereal-0.5.4.0: configure
    cereal-0.5.4.0: build
    cabal-doctest-1.0.4: copy/register
    data-default-class-0.1.2.0: download
    data-default-class-0.1.2.0: configure
    data-default-class-0.1.2.0: build
    data-default-class-0.1.2.0: copy/register
    cereal-0.5.4.0: copy/register
    Progress: 3/173C:\Users\mungre\AppData\Roaming\stack\precompiled\x86_64-windows\ghc-8.2.2\2.0.1.0\ZGF0YS1kZWZhdWx0LWNsYXNzLTAuMS4yLjBAc2hhMjU2OjYzZTYyMTIwYjdlZmQ3MzNhNWExN2NmNTljZWI0MzI2OGU5YTkyOWM3NDgxMjcxNzJkN2Q0MmY0YTMzNmUzMjcsNTQy\wVWqBSSq0fCfcwGy8J2YefUrKJnfifuo0K3EVuDHpuc=: openBinaryFile: does not exist (No such file or directory)

mgsloan:
@mungre Curious!  Could you please add some debug logging?  Maybe something like `logInfo $ T.pack (show (toFilePath longPath)) <> " with length: " <> T.pack (show (length (toFilePath longPath)))`.  Maybe there's some other codepath that uses these paths.

mungre:
@mgsloan I extended the logging to show both the original path and the shortened path; they were always identical.  So I added logging to the case expression that checks the path length.  It always took the maxPathLength=Nothing branch.

Of course, the error lies in the one thing the compiler doesn't validate!

    diff --git a/src/Stack/Constants.hs b/src/Stack/Constants.hs
    index 81298df..75b2db4 100644
    --- a/src/Stack/Constants.hs
    +++ b/src/Stack/Constants.hs
    @@ -247,7 +247,7 @@ defaultTerminalWidth = 100
     -- | Maximum length to use in paths. Is only a 'Just' value on windows,
     -- corresponding to MAX_PATH.
     maxPathLength :: Maybe Int
    -#ifdef mings32_HOST_OS
    +#ifdef mingw32_HOST_OS
     maxPathLength = Just 260
     #else
     maxPathLength = Nothing

It seems to have been able to resume failed builds because the files it is creating are just pointers to other files (as far as I can see) and they don't actually seem to be required.

mgsloan:
@mungre LOL, my mistake on the typo.  Thanks for figuring it out!  Fixed in https://github.com/commercialhaskell/stack/commit/36c1226d19e31cc8049eadc880d0f0c773f6d9ae

mungre:
@mgsloan Thanks for fixing this.

I've just had another silly failure: it doesn't work when the path length is exactly 260 characters.  Microsoft's MAX_PATH includes room for a terminating null character, so the maximum path is actually 259 characters.  So the "Just 260" should be "Just 259".

I think this could still fail with astral plane characters (Haskell counts them as one but Windows counts them as two) but they are unlikely in this context anyway.

mgsloan:
@mungre Good catch!  I was a little worried about corner cases with the number.  Feel like opening a PR adjusting that?

mungre:
@mgsloan Yes, I'll have a go.

Issue #3696, opened 2017-12-22
"with-hpack" option is passed to Setup.hs


TerrorJack:
Running `stack-1.6.1` with `with-hpack` (either via `--with-hpack` or in `config.yaml`) results in this error:

```
generated azathoth.cabal
azathoth-0.0.1: configure (lib)
unrecognized 'configure' option `--with-hpack=C:\Users\astro\AppData\Roaming\local\bin\hpack.exe'
```

It seems that `with-hpack` has been passed to `Setup configure`.

lwm:
Oh darn, I also hit this (`--with-compiler flag`):

```
➜  stack (fixup-cabal-parsing) ✔ stack build intero
ghc-paths-0.1.0.9: configure
terminfo-0.4.1.0: download
ghc-paths-0.1.0.9: build
terminfo-0.4.1.0: configure
ghc-paths-0.1.0.9: copy/register
Progress: 2/4
--  While building custom Setup.hs for package terminfo-0.4.1.0 using:
[...]

PR #3703, opened 2017-12-23
Don't pass --with-hpack to Setup.hs configure #3696

mgsloan:
Pinging @scott-fleischman - any thoughts on why this worked with https://github.com/commercialhaskell/stack/pull/3443 ?  Maybe Cabal-1.24 was less strict about `--with-*` options?

Tested by
```
stack new issue3696
cd issue3696
stack build hpack
stack build --with-hpack=hpack
```
(reproduced the issue)

After this fix, the issue no longer occurs.


scott-fleischman:
I don't know why it worked before—the less strict idea is a reasonable guess. I imitated the with-ghc code I think initially and neglected to review this code to see if it made sense. Thanks for fixing it!

mgsloan:
Cool, no worries @scott-fleischman , just curious if there was a particular usecase for passing it into configure.

PR #3704, opened 2017-12-23
Add testing notes to CONTRIBUTING.md + some tidbits

lwm:
---

Every time I get around to actually try and fix something, I forget how to run the tests.

I must ask, do we still need `--flag stack:integration-tests`? IIRC, this was some hack to be able to load tests into GHCi? From my latest attempts here, it's quite easy to run them without all that `:cd` stuff. If we could drop that flag, then running integration tests from the CLI would be much easier to type.

View [the rendered version here](https://github.com/lwm/stack/blob/0ed202d54b3f802dc20d8149d454cdb25fe9de5b/CONTRIBUTING.md).


borsboom:
The integration tests are a lot more resource intensive and take much longer (causing CI timeouts) than the unit tests, so I do think it's good to have a flag to enable/disable them.  I run the unit tests much more regularly than the integration tests.

lwm:
OK, will leave as is then!

PR #3706, opened 2017-12-25
Port `list-dependencies` under the new ls sub command

### LINKED ISSUE #3669: Stack list-dependencies : Bring it under the new ls umbrella command
This [PR](https://github.com/commercialhaskell/stack/pull/3252) introduces new `ls` umbrella sub-command. `list-dependencies` needs to be brought as `dependencies` under `ls`.

Do you folks think, we can remove the existing `list-dependencies` or just deprecate it for the current time being ?

Also I'm assigning the issue to myself, if someone else wants to pickup and do - please let me know. (Edit: It seems I can't assign the issue to myself!)

psibi:
This PR also deprecates the existing interface. Fixes #3669 

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,8 @@ Other enhancements:
 * A new sub command `ls` has been introduced to stack to view
   local and remote snapshots present in the system. Use `stack ls
   snapshots --help` to get more details about it.
+* `list-dependencies` has been deprecated. The functionality has
+  to accessed through the new `ls dependencies` interface.
 * Specify User-Agent HTTP request header on every HTTP request.
   See [#3628](https://github.com/commercialhaskell/stack/issues/3628) for details.
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,8 +13,10 @@ Other enhancements:
 * A new sub command `ls` has been introduced to stack to view
   local and remote snapshots present in the system. Use `stack ls
   snapshots --help` to get more details about it.
-* `list-dependencies` has been deprecated. The functionality has
-  to accessed through the new `ls dependencies` interface.
+*`list-dependencies` has been deprecated. The functionality has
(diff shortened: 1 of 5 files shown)


ChangeLog.md
```diff
@@ -13,6 +13,8 @@ Other enhancements:
 * A new sub command `ls` has been introduced to stack to view
   local and remote snapshots present in the system. Use `stack ls
   snapshots --help` to get more details about it.
+* `list-dependencies` has been deprecated. The functionality has
+  to accessed through the new `ls dependencies` interface.
```
lwm: to->to be

lwm: We should also link out to the relevant issue here?

src/main/Main.hs
```diff
@@ -385,7 +385,7 @@ commandLineHandler currentDir progName isInterpreter = complicatedOptions
                     cleanCmd
                     cleanOptsParser
         addCommand' "list-dependencies"
-                    "List the dependencies"
+                    "DEPRECATED: Use ls dependencies instead. Will be removed in next major version."
```
lwm: While it's a safe bet, no other deprecation notice specifies a time :/

mgsloan: We actually haven't been so good at removing deprecated stuff.

mgsloan: I'd say we should have `list-dependencies` output a warning (to stderr) that it may be removed at some point.  Thing is, people do use it in scripts.  We should probably wait until stack 2.0 or 3.0 to remove it.  Could hide it from the --help output next version, though.

mgsloan:
LGTM, I'd like it to warn for the old list-dependencies command (but to stderr not stdout, since scripts use its stdout)

psibi:
Pushed some commits that fixes the above concerns.

mgsloan:
LGTM, thanks! :+1:

PR #3709, opened 2017-12-25
ResolverSnapshot becomes ResolverStackage.

lwm:
---

Following [this comment](https://github.com/commercialhaskell/stack/compare/master...lwm:rename-resolver-snapshot?expand=1#diff-44f21bb6d216d7afb930c705712b837eL71).

### DIFF
--- a/src/Stack/Types/BuildPlan.hs
+++ b/src/Stack/Types/BuildPlan.hs
@@ -102,7 +102,7 @@ instance Store SnapshotDef
 instance NFData SnapshotDef
 
 snapshotDefVC :: VersionConfig SnapshotDef
-snapshotDefVC = storeVersionConfig "sd-v1" "tnwWSSLerZ2XeR6XpVwj5Uh0eF4="
+snapshotDefVC = storeVersionConfig "sd-v1" "CKo7nln8EXkw07Gq-4ATxszNZiE="
 
 -- | A relative file path including a unique string for the given
 -- snapshot.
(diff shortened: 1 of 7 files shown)


src/Stack/Types/BuildPlan.hs
```diff
@@ -102,15 +102,15 @@ instance Store SnapshotDef
 instance NFData SnapshotDef
 
 snapshotDefVC :: VersionConfig SnapshotDef
-snapshotDefVC = storeVersionConfig "sd-v1" "tnwWSSLerZ2XeR6XpVwj5Uh0eF4="
+snapshotDefVC = storeVersionConfig "sd-v1" "CKo7nln8EXkw07Gq-4ATxszNZiE="
```
lwm: I've no idea what this is - I was just following the compiler messages. Is this right?

mgsloan: This does a hash of the type definitions involved in the snapshot cache.  `sd-v1` gets prefixed to the file as a version tag to ensure that the serialization matches up.  In this case it is fine to change the hash, because it is just a rename, the serialization has not changed.

We should probably have a comment on these documenting how they work.  Maybe referencing an md file.

mgsloan:
LGTM, thanks!

src/Stack/Types/BuildPlan.hs
lwm: I see, thanks! Yes, seems like one for documenting :)

Issue #3636, opened 2017-12-10
Setup fails with 'No information found for ghc-8.2.2' for OS key 'linux64-ncurses6-nopie'


natrys:
I am trying to do a fresh install, using nightly as the global resolver.

```
~$ stack --version
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0
~$ rm -r ~/.stack/
```

```
~$ stack --resolver nightly-2017-12-10 --verbose setup
Version 1.6.1, Git revision f25811329bbc40b0c21053a8160c56f923e1201b (5435 commits) x86_64 hpack-0.20.0
2017-12-10 21:50:44.177921: [debug] Checking for project config at: /home/natrys/stack.yaml
@(Stack/Config.hs:842:9)
2017-12-10 21:50:44.178174: [debug] Checking for project config at: /home/stack.yaml
@(Stack/Config.hs:842:9)
2017-12-10 21:50:44.178264: [debug] Checking for project config at: /stack.yaml
@(Stack/Config.hs:842:9)
2017-12-10 21:50:44.178344: [debug] No project config file found, using defaults.
@(Stack/Config.hs:872:13)
2017-12-10 21:50:44.179624: [debug] Using resolver: nightly-2017-12-10 specified on command line
@(Stack/Config.hs:535:7)
2017-12-10 21:50:44.180012: [debug] Run from outside a project, using implicit global project config
@(Stack/Config.hs:559:13)
2017-12-10 21:50:44.180167: [info] Writing implicit global project config file to: /home/natrys/.stack/global-project/stack.yaml
@(Stack/Config.hs:578:20)
2017-12-10 21:50:44.180243: [info] Note: You can change the snapshot via the resolver field there.
@(Stack/Config.hs:579:20)
2017-12-10 21:50:44.180293: [info] Using resolver: nightly-2017-12-10 specified on command line
@(Stack/Config.hs:626:17)
2017-12-10 21:50:44.180728: [debug] Decoding build plan from: /home/natrys/.stack/build-plan/nightly-2017-12-10.yaml
@(Stack/Snapshot.hs:150:5)
2017-12-10 21:50:44.180797: [debug] Trying to decode /home/natrys/.stack/build-plan-cache/nightly-2017-12-10.cache
@(Data/Store/VersionTagged.hs:66:5)
[...]

PR #3715, opened 2017-12-28
setup: improvements to selecting bindists on Linux

borsboom:
These changes are motivated by #3636 and #3521.

* `stack setup` looks for GHC bindists and installations by any OS key
  that is compatible (rather than only checking a single one).   This is
  relevant on Linux where different distributions may have different
  combinations of libtinfo 5/6, ncurses 5/6, and gmp 4/5, and will allow
  simpifying the setup-info metadata YAML for future GHC releases.

* `stack setup` no longer uses different GHC configure options on Linux
  distributions that use GCC with PIE enabled by default.  GHC detects
  this itself since ghc-8.0.2, and Stack's attempted workaround for older
  versions caused more problems than it solved.

Tested on platforms: macOS, Arch, Ubuntu 16.04, Sabayon (Gentoo), Fedora 27, CentOS 6, CentOS 7, and Void Linux  
Also tested: installing GHCJS, --reinstall, --ghc-bindist, --ghc-build, --no-install-ghc

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,11 @@ Major changes:
 
 Behaviour changes:
 
+* `stack setup` no longer uses different GHC configure options on Linux
+  distributions that use GCC with PIE enabled by default.  GHC detects
+  this itself since ghc-8.0.2, and Stack's attempted workaround for older
+  versions caused more problems than it solved.
+
 Other enhancements:
 
 * A new sub command `ls` has been introduced to stack to view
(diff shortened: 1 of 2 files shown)


ChangeLog.md
```diff
@@ -8,13 +8,23 @@ Major changes:
 
 Behaviour changes:
 
+* `stack setup` no longer uses different GHC configure options on Linux
```
mgsloan: I don't see this in the change.  Is it part of making the tarballs?  I'd expect it to be conditional on ghc >= 8, so that'd be consistent with it being part of tarball creation.

mgsloan:
LGTM :+1:

ChangeLog.md
borsboom: It's the removal of the code for the `nopie` flag: https://github.com/commercialhaskell/stack/pull/3715/files#diff-4968f87e674615592e572f2ec593e5b0L584

That was put in place when nopie/tinfo6/ncurses6 changes were very in flux in the distros 
 (especially Arch) and that approach ended up not being very future proof.  The new approach is to patch `configure` in the GHC bindists to detect this condition and set the GHC `settings` file correctly.

mgsloan:
@borsboom Should this be merged to stable?  Seems like a nice fix for 1.6.4 https://github.com/commercialhaskell/stack/issues/3788

borsboom:
No, it doesn't really fix anything, it just future-proofs stack-setup-2.yaml better and means we can avoid duplication.  It certainly doesn't magically make GHC < 8.0.2 support nopie distributions.

Issue #3620, opened 2017-12-03
Split off some general purpose helper libraries


snoyberg:
Discussed with @mgsloan last week. Now that the 1.6 release branch is cut, we can start to seriously consider it.

Basic idea: there are some subsets of functionality within the Stack codebase that are not Stack specific. There may be more, but at the least I see: general command line utility code (logging functions, pretty printer, `RIO` monad), and Cabal-related stuff (file parsing, index handling). We should split these off into separate packages.

Motivation: Doing this correctly will ensure that:

* We can reuse this functionality elsewhere (stackage-curator is a prime example)
* We're forced to define cleaner APIs with less breakage
* It becomes easier to test this subset of functionality
* We reduce the relative amount of code in Stack itself

Approach: doing this badly will greatly increase our maintenance burden on Stack. Instead of splitting off into separate packages that need to be release to Hackage and tracking with extra-deps, I recommend something along the lines of:

* Create a subdirectory inside the stack repo for holding these "split off" packages
* Start moving modules into packages within this new directory, but...
* Use `source-dirs` from the main `package.yaml` so that we don't need to disrupt the workflow of working on Stack
* When we're satisfied that these subpackages have a semi-stable API, we remove the `source-dirs` and start treating them as separate packages within the repo
* At the next release of Stack, we release all subpackages with the main package
* In the future, we may choose to split these packages off into separate repos if deemed stable enough

This is just a proposal for now, and certainly comes with some downside, so feedback from other contributors is strongly encouraged.

lwm:
I'm in favour - I like the approach as well. Long term, once these packages go out, we'd reduce compile times, which would be really nice. Reduce LOC can help.

> We can reuse this functionality elsewhere (stackage-curator is a prime example)

This would be great.

> We're forced to define cleaner APIs with less breakage

Is this the beginning of a shift towards stable APIs in Stack? As far as I know, Stack has never offered a stable API.

> It becomes easier to test this subset of functionality

I'd question if we're not writing more tests because it is hard. I think no one has time ;) It'd certainly be nice to create more tests for specific moving parts and this will help.

Blaisorblade:
👍 I suspect many of the .Extra packages are also candidates? Or are those a separate issue?

It would also be an excuse to rationalize some of those functions. Last I looked at this, I wished to rationalize the conduit wrappers used to launch subprocesses and capture IO.

[...]

PR #3730, opened 2017-12-31
Kick off the rio subpackage

snoyberg:
This is a first stab into implementing some of #3620. This is unfortunately a rather involved PR, and includes:

* A cleanup of the System.Process.Read, System.Process.Run, and System.Process.Log modules
* Moving over the typed-process, which provides much of our functionality built in
* Moving off of monad-logger for our logging needs
* Separating our most of Stack.Prelude, System.Process.Read, and the logging code into a sub-package called `rio`

This is _far from complete_, I intend to do a lot more polishing on these modules. The naming is inconsistent as a carry-on from the manner in which it was refactored over time. But before this balloons into something too massive, I wanted to get things started.

The majority of the changes (by lines of code) comes down to one of two things:

1. No longer having a `MonadLogger` typeclass
2. Passing the `EnvOverride` and working directory information implicitly rather than explicitly


mgsloan:
@snoyberg LGTM! :+1: Glad this stuff is getting split off

PR #3735, opened 2018-01-01
Fix off-by-one and char encoding issue #3649 #3702

mungre:
As described in #3649, stack fails on Windows when building project dependencies because the paths it generates for cached files are too long.  #3702 and 36c1226d19e31cc8049eadc880d0f0c773f6d9ae are fixes to this but contain an off-by-one error so builds still fail when a cache path is exactly 260 characters long.  This patch fixes the off-by-one error.

Additionally, the Windows 260 character path limit is counted in 16-bit utf-16 code units, not unicode characters.  This patch also addresses this.

This is only tested on Windows, but the non-Windows functionality should not have changed.

To test this I rebuilt stack and a simple yesod app and all their dependencies using this modified stack.  This had previously failed.  I also edited the CPP constant to ensure the non-Windows branch of the code compiles without error.

### DIFF
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE DeriveGeneric         #-}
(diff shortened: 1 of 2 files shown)


mgsloan:
This looks great, @mungre, thanks!  My bad on the delay in merging.

Issue #3716, opened 2017-12-28
"stack setup --verbose" should also cause verbose output of GHC configure process if possible


mgsloan:
See https://github.com/commercialhaskell/stack/issues/3712#issuecomment-354206927

>  It's too bad stack setup --verbose doesn't also cause the GHC configure step to show all output.

krisis:
I can work on this issue if no one is looking into it already.

snoyberg:
Implemented via #3740, thanks!

PR #3740, opened 2018-01-03
Log GHC configure output on `stack -v setup`

krisis:
If -v global flag is passed to stack-setup subcommand we log the GHC
configure output.
Ref: https://github.com/commercialhaskell/stack/issues/3716

This PR was (manually) tested by running `stack --verbose setup --reinstall` on a new project directory created by `stack new testConfigure new-template`.

Excerpt of relevant logs from above test.
```
2018-01-03 10:49:29.280452: [debug] Process finished in 1550ms: /home/kp/.stack/programs/x86_64-linux/ghc-7.10.3.temp/ghc-7.10.3/configure --prefix=/home/kp/.stack/programs/x86_64-linux/ghc-7.10.3/
@(subs/rio/src/RIO/Process.hs:191:3)
2018-01-03 10:49:29.280793: [debug] Run process within /home/kp/.stack/programs/x86_64-linux/ghc-7.10.3.temp/ghc-7.10.3/: /usr/bin/make install
@(subs/rio/src/RIO/Process.hs:191:3)
2018-01-03 10:49:29.304283: [info] /usr/bin/make -r --no-print-directory -f ghc.mk install BINDIST=YES NO_INCLUDE_DEPS=YES
@(Stack/Setup.hs:1096:54)
2018-01-03 10:49:29.737966: [info] "rm" -f utils/ghc-pkg/dist-install/build/Version.hs  
@(Stack/Setup.hs:1096:54)
2018-01-03 10:49:29.739872: [info] echo "module Version where"                    >> utils/ghc-pkg/dist-install/build/Version.hs
@(Stack/Setup.hs:1096:54)
2018-01-03 10:49:29.741026: [info] echo "version, targetOS, targetARCH :: String" >> utils/ghc-pkg/dist-install/build/Version.hs
@(Stack/Setup.hs:1096:54)
2018-01-03 10:49:29.742170: [info] echo "version    = \"7.10.3\""      >> utils/ghc-pkg/dist-install/build/Version.hs

```

### DIFF
--- a/src/Stack/Prelude.hs
+++ b/src/Stack/Prelude.hs
@@ -76,8 +76,7 @@ sinkProcessStderrStdout
   -> RIO env (e,o)
 sinkProcessStderrStdout name args sinkStderr sinkStdout =
   withProc name args $ \pc0 -> do
-    let pc = setStdin closed
-           $ setStdout createSource
+    let pc = setStdout createSource
            $ setStderr createSource
              pc0
     withProcess_ pc $ \p ->
@@ -87,7 +86,7 @@ sinkProcessStderrStdout name args sinkStderr sinkStdout =
 -- | Consume the stdout of a process feeding strict 'ByteString's to a consumer.
 -- If the process fails, spits out stdout and stderr as error log
 -- level. Should not be used for long-running processes or ones with
--- lots of output; for that use 'logProcessStderrStdout'.
+-- lots of output; for that use 'sinkProcessStderrStdout'.
 --
 -- Throws a 'ReadProcessException' if unsuccessful.
 sinkProcessStdout
--- a/src/Stack/Prelude.hs
+++ b/src/Stack/Prelude.hs
@@ -87,7 +87,7 @@ sinkProcessStderrStdout name args sinkStderr sinkStdout =
 -- | Consume the stdout of a process feeding strict 'ByteString's to a consumer.
 -- If the process fails, spits out stdout and stderr as error log
 -- level. Should not be used for long-running processes or ones with
--- lots of output; for that use 'sinkProcessStdoutLogStderr'.
+-- lots of output; for that use 'logProcessStderrStdout'.
 --
 -- Throws a 'ReadProcessException' if unsuccessful.
 sinkProcessStdout
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -237,7 +237,7 @@ setupEnv mResolveMissingGHC = do
             , soptsGHCJSBootOpts = ["--clean"]
             }
 
-    (mghcBin, compilerBuild, _) <- ensureCompiler sopts LevelInfo
+    (mghcBin, compilerBuild, _) <- ensureCompiler sopts
 
     -- Modify the initial environment to include the GHC path, if a local GHC
     -- is being used
@@ -376,9 +376,8 @@ addIncludeLib (ExtraDirs _bins includes libs) config = config
 -- | Ensure compiler (ghc or ghcjs) is installed and provide the PATHs to add if necessary
 ensureCompiler :: (HasConfig env, HasGHCVariant env)
                => SetupOpts
-               -> LogLevel
                -> RIO env (Maybe ExtraDirs, CompilerBuild, Bool)
-ensureCompiler sopts logLevel = do
+ensureCompiler sopts = do
     let wc = whichCompiler (soptsWantedCompiler sopts)
     when (getGhcVersion (soptsWantedCompiler sopts) < $(mkVersion "7.8")) $ do
         logWarn "Stack will almost certainly fail with GHC below version 7.8"
@@ -467,7 +466,6 @@ ensureCompiler sopts logLevel = do
                             (soptsWantedCompiler sopts)
                             (soptsCompilerCheck sopts)
                             (soptsGHCBindistURL sopts)
-                            logLevel
                     | otherwise -> do
                         recommendSystemGhc <-
                             if soptsUseSystem sopts
@@ -855,9 +853,8 @@ downloadAndInstallCompiler :: (HasConfig env, HasGHCVariant env)
                            -> CompilerVersion 'CVWanted
                            -> VersionCheck
                            -> Maybe String
-                           -> LogLevel
                            -> RIO env Tool
-downloadAndInstallCompiler ghcBuild si wanted@GhcVersion{} versionCheck mbindistURL logLevel = do
+downloadAndInstallCompiler ghcBuild si wanted@GhcVersion{} versionCheck mbindistURL = do
     ghcVariant <- view ghcVariantL
     (selectedVersion, downloadInfo) <- case mbindistURL of
         Just bindistURL -> do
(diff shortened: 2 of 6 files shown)


src/Stack/Setup.hs
```diff
@@ -1085,14 +1091,27 @@ installGHCPosix version downloadInfo _ archiveFile archiveType tempDir destDir =
...
-                    $ withProc cmd args
-                    $ try
-                    . readProcessStdout_
-                    -- Calling the ./configure script requires that stdin is
-                    -- open
-                    . setStdin (useHandleOpen stdin)
+            result <- case logLevel of
+              LevelDebug -> do
+                let logLines = CB.lines .| CL.mapM_ (logInfo . T.decodeUtf8With T.lenientDecode)
+                withWorkingDir wd
```
krisis: I didn't notice earlier that the indentation is off. I shall correct this in subsequent submissions along with other review comments.

src/Stack/Setup.hs
```diff
@@ -1085,14 +1091,27 @@ installGHCPosix version downloadInfo _ archiveFile archiveType tempDir destDir =
...
             menv' <- modifyEnvOverride menv (Map.union env)
-            result <- withWorkingDir wd
-                    $ withEnvOverride menv'
-                    $ withProc cmd args
-                    $ try
-                    . readProcessStdout_
-                    -- Calling the ./configure script requires that stdin is
-                    -- open
-                    . setStdin (useHandleOpen stdin)
+            result <- case logLevel of
```
mgsloan: Can avoid the manual plumbing of logLevel by doing something like `logLevel <- logMinLevel <$> view logOptionsL`

src/Stack/Setup.hs
mgsloan: This is good code!  There are some existing utility functions for this, though -  `Stack.Prelude.logProcessStderrStdout`.  This stuff got moved around recently in #3730

snoyberg: It appears that this is the only place in the codebase that uses `logProcessStderrStdout`. I'd recommend just removing the function entirely from `Stack.Prelude` and keeping the functionality here. No need to provide general utility functions if they're not of general utility :)

FWIW, this is something I would have done myself in #3730 had I noticed, so good catch to you both.

src/Stack/Setup.hs
```diff
@@ -1085,14 +1091,27 @@ installGHCPosix version downloadInfo _ archiveFile archiveType tempDir destDir =
...
+              LevelDebug -> do
+                let logLines = CB.lines .| CL.mapM_ (logInfo . T.decodeUtf8With T.lenientDecode)
```
snoyberg: It's slightly less efficient, but assuming it gives the output you're looking for, you could use `logDebug` here instead, and then unconditionally use this implementation, and totally bypass analyzing the `logLevel`. I'm not too concerned about the performance overhead of breaking the output from `./configure` into lines and then potentially ignoring them, so I'd lean in that direction for the code simplification it would yield.

krisis: I am not sure I follow how we can avoid analyzing `loglevel`. If `logLevel` is `LogDebug` (i.e, stack-setup was invoked with `-v`) we need to read stdout of configure script and print it in `stack` command's stdout. Otherwise, we should not print anything. If we used `logDebug` in both the cases, wouldn't users  who ran stack-setup without `-v` complain about the verbosity?

snoyberg: `logDebug` will only produce output if Stack is run with `-v`.

src/Stack/Setup.hs
mgsloan: Though, to me this sort of thing seems like a useful thing for a future rio - process.  It is not uncommon to want to involve process output in your log while still keeping your log format.

src/Stack/Setup.hs
```diff
@@ -1085,14 +1085,19 @@ installGHCPosix version downloadInfo _ archiveFile archiveType tempDir destDir =
...
-                    -- Calling the ./configure script requires that stdin is
-                    -- open
-                    . setStdin (useHandleOpen stdin)
+            result <- do
+                let logLines = CL.mapM_ (logDebug . T.decodeUtf8With T.lenientDecode)
+                withWorkingDir wd
+                  $ withEnvOverride menv'
+                  $ withProc cmd args
+                  $ try
+                  . (flip withLoggedProcess_ $ \p ->
```
snoyberg: What's the reason for the switch to `withLoggedProcess_` here, instead of say `withProcess_`? Quoting from the docs:

> This will store all output from stdout and stderr in memory for better error messages.

It seems like that's redundant with turning on the verbose debugging here.

src/Stack/Setup.hs
krisis: I am leaving `logProcessStderrStdout` as it is being used [here](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Setup.hs#L1320).

src/Stack/Setup.hs
krisis: I shall update the PR using withProcess_.

src/Stack/Setup.hs
```diff
@@ -1085,14 +1085,21 @@ installGHCPosix version downloadInfo _ archiveFile archiveType tempDir destDir =
...
-            result <- withWorkingDir wd
-                    $ withEnvOverride menv'
-                    $ withProc cmd args
-                    $ try
-                    . readProcessStdout_
-                    -- Calling the ./configure script requires that stdin is
-                    -- open
-                    . setStdin (useHandleOpen stdin)
+            result <- do
+                let logLines = CL.mapM_ (logDebug . T.decodeUtf8With T.lenientDecode)
```
snoyberg: This implementation is different from the one in `logProcessStderrStdout`, which includes a call to `CB.lines`:

```haskell
let logLines = CB.lines .| CL.mapM_ (logInfo . decodeUtf8With lenientDecode)
```

As implemented here, it will not break the output up into individual lines.

krisis:
Yes. I misunderstood your previous comment on code simplification and thought you were suggesting removal of `CB.lines`. I shall add it back.

src/Stack/Setup.hs
krisis: Another difference is that I have used `concurrently_` since we are ignoring the return values subsequently.

src/Stack/Prelude.hs
```diff
@@ -76,8 +76,7 @@ sinkProcessStderrStdout
   -> RIO env (e,o)
 sinkProcessStderrStdout name args sinkStderr sinkStdout =
   withProc name args $ \pc0 -> do
-    let pc = setStdin closed
```
snoyberg: I'm not in favor of changing this default—which IMO is the right default—for this one case, which should be considered an exception. I preferred the previous approach. If the goal is to reduce code duplication, I would recommend changing the function to take a `ProcessConfig` instead, or something along those lines.

snoyberg: Orthogonal to this PR... I think we _do_ have to change this default. I found another part of the code base which is expecting a working `stdin` in a subprocess: configuring a package with `build-type: Configure`. So let's move ahead with this as is.

PR #3747, opened 2018-01-04
Clone into tmp dir when upgrading

### LINKED ISSUE #3745: stack won't upgrade itself because it already exists
```
05:28 PM noon ∈ ~ ♫ stack upgrade --git
Cloning stack
fatal: destination path 'stack' already exists and is not an empty directory.
Received ExitFailure 128 when running
Raw command: /usr/bin/git clone https://github.com/commercialhaskell/stack stack --depth 1 --recursive --branch master

05:28 PM noon ∈ ~ ♫ stack --version
Version 1.7.0, Git revision 52c00c63750075a19f4a4ce4c07f4c87b0744e41 x86_64 hpack-0.20.0
```

```
Linux gpac 4.13.2-041302-generic #201709132057 SMP Thu Sep 14 00:59:32 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

silky:
This fixes #3745, I believe.

Maybe there's a good way to actually add a test for this?

### DIFF
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -211,7 +211,7 @@ sourceUpgrade gConfigMonoid mresolver builtHash (SourceOpts gitRepo) =
                 -- the stack repo until we're comfortable with "stack upgrade
                 -- --git" not working for earlier versions.
                 let args = [ "clone", repo , "stack", "--depth", "1", "--recursive", "--branch", branch]
-                withProc "git" args runProcess_
+                withWorkingDir tmp $ withProc "git" args runProcess_
                 return $ Just $ tmp </> $(mkRelDir "stack")
       Nothing -> do
         updateAllIndices


silky:
the CI failure appears to be due to some general timeout - https://travis-ci.org/commercialhaskell/stack/jobs/325251724

mgsloan:
LGTM.  I am fine with accepting this as is.  However, while reviewing it, it occurred to me that there are some issues with all of our invocations of git - https://github.com/commercialhaskell/stack/issues/3748

Want to take a swing at that too?  Would be greatly appreciated, but not at all necessary for getting this fix merged.

silky:
i reckon i could give that a go, but probably not for a day or two, as it'll probably take me a bit of time to figure it out and then work out how to test it, etc.

mgsloan:
We could test for the specific bug, but eh, I think in this case having a good test is more effort than it's worth.  Thanks for working on this!

Issue #3732, opened 2018-01-01
Too much recompilation when changing test targets


snoyberg:
I believe this is a side-effect of the changes in #3565, implemented for #3483. CC @tswelsh. To repro, I created two packages, `foo` and `bar`, with `bar` depending on `foo`. I then ran the commands:

```
$ stack test --no-run-tests
$ stack test --no-run-tests bar
```

Expected: the second command should do nothing, as all tests, libraries, and executables have already been run.

Actual: I saw the following output:

```
$ stack test --no-run-tests bar
bar-0.1.0.0: unregistering (missing dependencies: foo)
foo-0.1.0.0: unregistering (local file changes: app/Main.hs test/Spec.hs)
foo-0.1.0.0: build (lib + exe)
foo-0.1.0.0: copy/register
bar-0.1.0.0: configure (lib + exe + test)
```

Notice how `foo` is unregistered, due to Stack thinking that the files changed. This occurs because the test suite and executable files are no longer included _at all_ in the list of files in the package.

Instead, we should track, for each package, the files needed on a per-component basis, and only rebuild based on this. This may tie in with @ezyang's work on per-component builds, not sure.

@tswelsh are you interested in taking a stab at this one?

tswelsh:
Sure, I can take a look at it. I should be able to find some time this week hopefully.

snoyberg:
Resolved via #3750, thanks!

PR #3750, opened 2018-01-06
Use a separate build cache for each component of a package

tswelsh:

Attempt at a fix for #3732.  This introduces a separate build cache for each component of a package, instead of having one for the entire package.

Previously, if a package `foo` was built with tests enabled then the test files (say `Spec.hs`) would be recorded in the build cache.  With no further source changes, if the dirtiness of `foo` is checked in a scenario where we are not asking for `foo`'s tests (and so not mentioning `Spec.hs` at all), the dirtiness check of the source files would spot that `Spec.hs` is in the build cache but not in the files we want to compile, so a rebuild would be triggered.  With this patch, we only compare with the build cache for the relevant component.

I think this is a better situation than before, but still not perfect - dirty files are not tracked on a component basis with this, so I _think_ we can still get some useless rebuilds.  However, querying dirty files by component looked a bit more involved, and would change much of the `ConstructPlan` module.  Apart from it being more complicated, I wasn't sure if I would be stepping on @ezyang's toes with work in there.

Manual testing based on @snoyberg's example in the linked issue: a project with 2 packages, `foo` and `bar` that depends on `foo`, both with test suites.


ezyang:
I haven't gotten around to understanding how build caches work in Stack, so no complaints form me :>

tswelsh:
Maybe I don't understand either because this patch is broken :-) - my testing somehow failed to reveal that it doesn't rebuild if library source files change. Think it's easy enough to fix, but may as well hold of on reviewing until I sort it.

tswelsh:
I've rebased, hlinted and fixed the initial error, so it's ready to be looked over.  The AppVeyor failure looks unrelated.

PR #3756, opened 2018-01-09
Don't close stdin for subprocesses when running Setup

rjmac:
Instead, provide an empty stream.  Otherwise, operations which try to
use their stdin (for example, network's configure script) will fail
with a "bad file descriptor" error when they try to do something with
fd 0.

It looks like this was introduced a few days ago, in 0d2943ad291f6350049bfdde45d99bc9dc4dbaf3


snoyberg:
Thanks!

PR #3758, opened 2018-01-11
Fix package promotion to snapshot

snoyberg:
Fixes fpco/stackage#3185


test/integration/tests/stackage-3185-ignore-bounds-in-snapshot/files/stack.yaml
```diff
@@ -0,0 +1 @@
+resolver: lts-3.12
```
borsboom: Having integration tests with old resolvers will cause them to fail on Linux distributions with GCC with PIE enabled by default (which is the latest versions of most distributions now), since older GHC versions don't support it.  I'm not sure what we should do about this, since it obviously does make sense to be able to test against old snapshots sometimes.

I haven't confirmed this, but I believe also that, if `~/.stack` doesn't already happen to have the required GHC version installed, _every_ time the test is run it will re-install GHC which makes the tests take extra long.

snoyberg: Alright, new commit pushed that should address this.

Issue #3714, opened 2017-12-27
Stack doesn't find custom snapshot package


nmattia:
### General summary/comments (optional)

I cannot pull `git:` defined packages in a custom snapshot. The package version is looked up from the custom snapshot correctly, but then stack tries to find it on hackage. Stack does download the source but it is never built.

This is the (failing) lookup, as far as I can tell: https://github.com/commercialhaskell/stack/blob/master/src/Stack/Package.hs#L200

CC @mgsloan 

### Steps to reproduce

Clone [this repo](https://github.com/nmattia/stack-custom-test), run `stack build reflex`. The repo specifies a minimal `stack.yaml` with a custom snapshot:

**stack.yaml**
``` yaml
resolver: snapshot.yaml
```

**snapshot.yaml**
``` yaml
resolver: lts-9.9
name: Dummy
packages:
- git: https://github.com/reflex-frp/reflex
  commit: d9ef8457dfd140fde32d15a6959a15ddcd7e15e4
```

### Expected

Stack should have tried to build the `reflex` package.

### Actual

```
stack build reflex
Cloning d9ef8457dfd140fde32d15a6959a15ddcd7e15e4 from http://github.com/reflex-frp/reflex.git
Cloning into '/home/nicolas/projects/stack-custom-test/.stack-work/downloaded/zoCbUP3YjykH'...
remote: Counting objects: 3473, done.
remote: Total 3473 (delta 0), reused 0 (delta 0), pack-reused 3472
Receiving objects: 100% (3473/3473), 910.78 KiB | 325.00 KiB/s, done.
Resolving deltas: 100% (2130/2130), done.
Checking connectivity... done.
Didn't see reflex-0.5.0 in your package indices.
Updating and trying again.
Selected mirror https://s3.amazonaws.com/hackage.fpcomplete.com/                                 
Downloading timestamp                                                                            
No updates to your package index were found                                                      
Update complete                                                                                  
The following package identifiers were not found in your indices: reflex-0.5.0
Possible candidates: reflex-0.2, reflex-0.4.0, reflex-0.0.1, reflex-0.3.2, reflex-0.3, reflex-0.1.1, reflex-0.1.0, reflex-0.3.1.
```

See the [verbose output](https://gist.github.com/nmattia/e5d146260c841e44005070205249bd01).

### Stack version

Stack built from source at commit 8266a06536ecf87a9755e7800a3f4496fb10c61b, although reproducible with Stack 1.6.3

[...]

PR #3759, opened 2018-01-11
Do not discard package location #3714

snoyberg:


mgsloan:
LGTM!  Lets see if I understand this, though:

`splitUnmetDeps` is used for both packages in the global DB and packages in the snapshot.  For the packages in the global DB, it needs to convert them to snapshot packages.  For the packages in the snapshot, they should remain snapshot packages.  Hence, pulling `globalToSnapshot` out of `splitUnmetDeps` and using it in only one case.

snoyberg:
Almost exactly correct. There are two stages to calling splitUnmetDeps on the snapshot packages. The second time around it's adding in local packages, so I'm that case the snapshot packages graduate to local packages.

mgsloan:
I see, thanks!

PR #3761, opened 2018-01-12
Add more queries to `stack query`

kirelagin:
While looking into input-output-hk/stack2nix#84 I added a couple of queries to `stack query`, so I thought I’d share them in case someone finds them useful.

The change is pretty trivial, but if I had access to a Windows machine, I would make sure it still compiles on Windows, just in case.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -373,12 +373,14 @@ rawBuildInfo = do
     (locals, _sourceMap) <- loadSourceMap NeedTargets defaultBuildOptsCLI
     wantedCompiler <- view $ wantedCompilerVersionL.to compilerVersionText
     actualCompiler <- view $ actualCompilerVersionL.to compilerVersionText
+    globalHints <- view globalHintsL
     return $ object
         [ "locals" .= Object (HM.fromList $ map localToPair locals)
         , "compiler" .= object
             [ "wanted" .= wantedCompiler
             , "actual" .= actualCompiler
             ]
+        , "global-hints" .= globalHints
         ]
   where
     localToPair lp =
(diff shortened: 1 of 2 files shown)


mgsloan:
Hi, sorry for letting this languish!  It seems good to have this information, thanks for adding it

Not sure about the names, just because we're not going to want to change them in the future.  But I can't think of better ones, so going ahead to merge.

mgsloan:
Hmm, why is `global-hints` needed?  That seems like a bit of an internal concern.  `stack query` should probably have a `configuration.packages` field that gives a dump of the packages in the configuration.

kirelagin:
I was working on `stack2nix` and I was looking for a way to get a list of the packages bundled with the GHC version used in the resolver. Global hints seemed to be the closest I could get... Is there something better?

mgsloan:
Hmm, do you need this information before installing ghc?  If not, then using `stack exec -- ghc-pkg dump` along with `stack path --global-pkg-db` can tell you what's in the package db.

kirelagin:
Ideally, yes.

(I’m not sure now why I rejected the idea of installing the raw compiler and querying its database, I’ll have to thing about it again.)

mgsloan:
Yeah, that's why these are "hints".  It will let you use whatever global packages that come with the compiler.  I actually don't remember what they are used for.

kirelagin:
AFAIU they are used for `stack init`, at least according to [this post](https://www.fpcomplete.com/blog/2017/07/stacks-new-extensible-snapshots) by @snoyberg, which is where I got the idea of using them to approximate the set of boot packages.

snoyberg:
They're only used as a hack because we want to guess which snapshot to use, and not require the user to install 3 GHCs to do that guessing. Wherever possible, you should use the actual global database from GHC, which will handle cases like different OSes and prerelease/custom compilers properly.

PR #3763, opened 2018-01-12
Report currently building packages along with "Progress:" label

develop7:
Rationale: make user feel more "in control" when waiting for these tens to hundreds of packages to be built by providing information on what is building now.

Build progress line <details><summary>before</summary>`Progress: 3/74`</details> and <details><summary>after</summary>`Progress: 3/74; [Cabal-2.0.1.1|basement-0.0.4|old-time-1.1.0.3|simple-sendfile-0.2.26|stm-2.4.4.1|stringsearch-0.3.6.6]`</details> applying this change.

Limitations: `configure` and `copy/register` stages are not covered, would require refactoring. Since these are quite fast, covering them might not worth the effect.

Note: Documentation fixes for https://docs.haskellstack.org/en/stable/ should target the "stable" branch, not master.

Please include the following checklist in your PR:

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [x] The documentation has been updated, if necessary.

PR #3774, opened 2018-01-14
fix hlint suggestions from hlint-2.0.13

kadoban:
Looks like a new version of hlint has added some suggestions for our code, which is causing CI failures for PRs (eg #3763 and #3761)

This applies the changes hlint suggests.

They all appear to be   `(blah . foo) <$> something`  ====>  `blah . foo <$> something`

Can't say this seems like the most important suggestion ever, but it seems equally readable before and after, so this seems like the easiest way to shut hlint up unless someone has an objection.


mgsloan:
Thanks!

PR #3775, opened 2018-01-15
Add a script to download and install many stack releases

mgsloan:
When comparing behavior of different stack versions, or when my development version of stack has a bug, I find it useful to be able to run older versions.  For example, `stack-1.6.3 build`, etc.

This is a hacky script which depends on `stack >= 1.4`, `wget`, `tar`, and `hub`.  It installs binaries for all the stack releases that it can download that are greater than a minimum version (a constant in the code sets it to `>= 1.0`.

Not sure if writing this script was worth it vs the manual labor, but it only took an hour or two.  Maybe it will be worth it in the long run if it benefits others.

Probably should have used `shelly`, but ohwell.

### DIFF
--- a/etc/scripts/install-many-stack-releases.hs
+++ b/etc/scripts/install-many-stack-releases.hs
@@ -1,0 +1,163 @@
+#!/usr/bin/env stack
+{- stack script
+    --resolver lts-9.21
+    --package base
+    --package directory
+    --package filepath
+    --package process
+    --package safe
+    --package temporary
+-}
+
+-- # Usage summary
+--
+-- This is a hacky script to install many stack releases to a target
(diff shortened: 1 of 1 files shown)


snoyberg:
I'm ambivalent on this one, the external tool dependencies likely mean I'm not going to end up using it, but if it's useful for you, go for it.

mgsloan:
@snoyberg That's fair.  You might give `hub` a try, though, it's really nice.  Doing stuff like `git clone snoyberg/conduit` is great.  I also use `git checkout https://github.com/commercialhaskell/stack/pull/3775` all the time.

For something that's just a quick script mostly intended for stack developers, I figured calling out to some common tools would not be a problem.

PR #3776, opened 2018-01-15
Add rio package as a dep of stack

mgsloan:
Otherwise, "stack ghci" fails to run due to

> The following modules are present in multiple packages:
>  * Path.Extra (in stack, rio)
>  * RIO (in stack, rio)
>  * RIO.Logger (in stack, rio)
>  * RIO.Prelude (in stack, rio)
>  * RIO.Process (in stack, rio)

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -92,6 +92,7 @@ dependencies:
 - regex-applicative-text
 - resourcet
 - retry
+- rio
 - semigroups
 - split
 - stm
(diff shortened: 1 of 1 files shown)


snoyberg:
This is different from our original plan, but I'm OK with this approach. Once the external `rio` gets a bit more stable, we should switch over to that. @borsboom are you OK with this switch?

mgsloan:
Alternatively, we can add a cabal flag that switches between the approach, and turn on that flag when using ghci.  I usually develop stack within ghci.  I recommend this approach to everyone, really, drastically speeds up development iterations.

mgsloan:
The fix to `stack ghci` in #3779 makes this unnecessary, so closing.

Issue #3766, opened 2018-01-13
Duplicating dependencies in stack.yaml and package.yaml


corpix:
Hello. I have added a new dependency(`strings`) into `package.yaml` file and when run `stack build` I got this:

```
$ stack build                                                        ✖ 1 (2s 979ms)
Error: While constructing the build plan, the following exceptions were encountered:
In the dependencies for mammoth-0.0.0:
    strings must match -any, but the stack configuration has no specified version (latest matching version is 1.1)
needed since mammoth is a build target.
Some potential ways to resolve this:
  * Recommended action: try adding the following to your extra-deps in /home/user/projects/<redacted>/stack.yaml:
- strings-1.1
  * Set 'allow-newer: true' to ignore all version constraints and build anyway.
  * You may also want to try using the 'stack solver' command.
Plan construction failed.
```

> I am a haskell beginner. I understand, many haskell programmers are PhDs... but let's just keep things simple for mere mortals :)

1. This error is very confusing, the part `strings must match -any` makes me continuously screaming "wat? wat? wat?" because I don't understand what is this `-any` thing stands for
2. I don't thing this is a good idea to duplicate dependencies across configuration files(stack.yaml, package.yaml) but stack suggests me to `try adding the following to your extra-deps in .../stack.yaml`, I already have package.yaml, I don't want to "copy-paste"
3. Second recommendation is strange `Set 'allow-newer: true' to ignore all version constraints and build anyway`... we have `no specified version` but we could `ignore all version constrains`... wat?

kadoban:
Yeah, the `allow-newer` advice does not apply in this situation, so that is a tad not ideal. That could maybe be improved.

The rest is working as designed mostly. The error message is a bit hard to read at first, but you get used to it. It's a bit hard to be both concise there and also have the message make sense to stack beginners. It could likely me improved, but I have yet to think of a better wording for it.

As for why you need to duplicate the dependency in the packages.yaml and the stack.yaml file, that has to do with their different responsibilities.

The package.yaml file just declares what dependencies you have in terms of package names and their version ranges (or you are allowed to leave out the version ranges).

[...]

PR #3780, opened 2018-01-15
Adjustments to plan construction error messages #3766

mgsloan:
* Makes it so that it doesn't output '-any' to describe a version range.

* Doesn't suggest "allow-newer" if some errors aren't constraint mismatches.

* Reorders them so that the recommended action comes last.  This also looks
  better because it has a dedented listing of extra-deps.

* Gives a rough / brief reason why approaches other than 'stack solver' may be
  preferred.

New output:

2018-01-15_1449x440_scrot


mgsloan:
Great, thanks for reviewing!

Issue #3753, opened 2018-01-07
Stack does not respect ghc-options in user config


centromere:
The `ghc-options` setting only seems to be respected in the project-specific `stack.yaml` file. When it is in `~/.stack/config.yaml` or `/etc/stack/config.yaml` it is not respected.

### Steps to reproduce

```
% stack --version
Version 1.7.0, Git revision 6b9da26bf17c682a5db583f1a5484d4762467a82 x86_64 hpack-0.20.0
% cat ~/.stack/config.yaml
ghc-options:
  "$everything": "isdfsdf-split-sectionslsdjfhsljdf"
% stack new test
Downloading template "new-template" to create project "test" in test/ ...
% cd test
% stack build
Building all executables for `test' once. After a successful build of all of them, only specified executables will be rebuilt.
test-0.1.0.0: configure (lib + exe)
Configuring test-0.1.0.0...
test-0.1.0.0: build (lib + exe)
Preprocessing library for test-0.1.0.0..
Building library for test-0.1.0.0..
[1 of 2] Compiling Lib              ( src/Lib.hs, .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/Lib.o )
[2 of 2] Compiling Paths_test       ( .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/autogen/Paths_test.hs, .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/Paths_test.o )
Preprocessing executable 'test-exe' for test-0.1.0.0..
Building executable 'test-exe' for test-0.1.0.0..
[1 of 2] Compiling Main             ( app/Main.hs, .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/test-exe/test-exe-tmp/Main.o )
[2 of 2] Compiling Paths_test       ( .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/test-exe/autogen/Paths_test.hs, .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/test-exe/test-exe-tmp/Paths_test.o )
Linking .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/test-exe/test-exe ...
test-0.1.0.0: copy/register
Installing library in /Users/alex/test/.stack-work/install/x86_64-osx/lts-10.2/8.2.2/lib/x86_64-osx-ghc-8.2.2/test-0.1.0.0-LK8fulHuYNc5gAYF1YHqqE
[...]

PR #3781, opened 2018-01-15
Fix inheritance of category ghc-options from config.yaml #3753

mgsloan:

Tried the repro in #3753 and it works.

NOTE: Needs a changelog!  Haven't written one yet, since the desired semantics for ghc-options isn't yet clear to me.  It's either "shadowing" or "extending" prior ghc-options settings.  Previous semantics were "shadowing".

### DIFF
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -828,17 +828,16 @@ parseConfigMonoidObject rootDir obj = do
     optionsEverything <-
       case (Map.lookup GOKOldEverything options, Map.lookup GOKEverything options) of
         (Just _, Just _) -> fail "Cannot specify both `*` and `$everything` GHC options"
-        (Nothing, Just x) -> return x
+        (Nothing, Just x) -> return (Just x)
         (Just x, Nothing) -> do
           tell "The `*` ghc-options key is not recommended. Consider using $locals, or if really needed, $everything"
-          return x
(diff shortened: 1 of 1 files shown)


mgsloan:
An alternative approach would be disallow having config.yaml specify options for categories of packages.  Perhaps it should only allow `$everything`? Hmm hmm, tricky!

snoyberg:
I'm honestly not sure what the correct behavior here should be. Perhaps merging the options together and hoping for the best is good. Another approach may be adding a new setting which is specifically intended to be inherited by other configs, and giving a warning when the old name is used. But I'm not sure that actually prevents any misuse.

mgsloan:
The main reason not to change this is that it's a behavior change.  However, the good news is it's only a behavior change if you have a config.yaml which specifies ghc-options, which is probably not very many of them.  When they are specified, I'm guessing append semantics is preferred.

Since the effect of most (all?) ghc flags can be undone by subsequent flags, having it append may make sense.  One downside is that this could lead to "defensive flag unsetting" in package / project configs, to unset options that shouldn't be set for that particular package. Hmm hmm.

I'm thinking append is most consistent, since this is already how `$everything` flags get combined with package specific flags.

mgsloan:
I'm thinking that it makes more sense to merge the options, because they can be disabled.

mgsloan:
Also, note to myself when revisiting this in the future: needs changelog update.

PR #3786, opened 2018-01-17
Typo in cabal option --with-hc-pkg

tkvogt:
Cabal has no option --with-ghc-pkg
Probably intended was:  --with-hc-pkg
Check here: https://www.haskell.org/cabal/users-guide/search.html?q=with-hc-pkg&check_keywords=yes&area=default#
Or am I wrong? Why did it work so far?

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -843,8 +843,8 @@ ensureConfig newConfigCache pkgDir ExecuteEnv {..} announce cabal cabalfp task =
         menv <- view envOverrideL
         let programNames =
                 if eeCabalPkgVer < $(mkVersion "1.22")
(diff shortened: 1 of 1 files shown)


mgsloan:
Hey, sorry for taking so long in review! That's interesting.  Indeed the cabal docs don't seem to mention with-ghc-pkg.  However, note that those are the docs for cabal-install whereas stack is using the Cabal library, which has a different CLI.  My guess is that `with-hc-pkg` is a newer flag name.  I'm inclined to continue using the old one as it's unclear when this flag was added.

snoyberg:
Based on the response and lack of further discussion, I'm going to close. @tkvogt if you'd like to reopen, or provide a demonstration of this breaking, please add a comment.

PR #3792, opened 2018-01-18
Avoid stdout redirection in script --compile

snoyberg:
Not only did this complicate the code and add overhead, but it seems to
have confused GHC into sending incorrectly encoded output. With this
approach, GHC gets to send its output directly to stderr.

    * Left out, since this is a newly introduced regression with the refactoring of the output code


mgsloan:
:+1:

PR #3794, opened 2018-01-18
ease single directory requirement

dwaldhalm:

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -1466,7 +1466,7 @@ expectSingleUnpackedDir :: (MonadIO m, MonadThrow m) => Path Abs File -> Path Ab
 expectSingleUnpackedDir archiveFile destDir = do
     contents <- listDir destDir
     case contents of
-        ([dir], []) -> return dir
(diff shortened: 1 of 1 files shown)


kadoban:
Hm, is this the right way to fix this problem? (Not asking rhetorically, I really don't know.)

If the issue is that stack is leaving the tarball in place and that's why the test is failing, shouldn't we instead fix *that* problem so the test of if the tarball is in the right form can be run correctly? Or is that hard to do for some reason?

Or maybe there's no harm in making the check less strict, though I can think of a lot of forms of tarballs that will accidentally pass this that wouldn't really be what is expected.

dwaldhalm:
That's a fair question, but I can't answer it--I'm not familiar with stack (just a user / fan).  In this case, it seems to effect only a temp directory that is ultimately deleted.  No .tar file remains in my bin.

mgsloan:
Yeah, on one hand it seems good to check "are things as we expect them to be".  On the other hand, someone might want to include a README at the top level of their tarball or something.  So I'm fine with weakening the check.  Since this is happening in a temp directory, it will get deleted regardless.

PR #3805, opened 2018-01-24
Comment out instructions for Ubuntu package

Blaisorblade:
I just confirmed that Ubuntu 16.04 cannot upgrade to latest stack (or do installations). The problem might extend to later Ubuntu/Debian releases, so maybe we should just drop mentions of the package for now? Adding a commit for Debian.
This PR is still for discussion on what to do in docs, feel free to take over or merge as-is. I can't look more closely at this now.

### DIFF
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -102,6 +102,10 @@ If you are on OS X 10.12 ("Sierra") and encounter [GHC panic while building, see
 
 Use the [generic Linux option](#linux).
 
+There is also a
+[Ubuntu package](http://packages.ubuntu.com/search?keywords=haskell-stack&searchon=names&suite=all&section=all) for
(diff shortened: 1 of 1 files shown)


Blaisorblade:
One way to complete this PR:
- figure out what's the oldest version of stack that can still upgrade successfully?
- deprecate using the package for any distros carrying older versions. Or just document that version. Or figure out a different setup.

Blaisorblade:
So, this is what I got when upgrading: https://gist.github.com/Blaisorblade/c88757ac00238915b32379d2e0dbc63c

mgsloan:
Sorry for the delay in reviewing!  Seems like an improvement.

mgsloan:
Note that I adjusted the wording some - https://github.com/commercialhaskell/stack/commit/6d0ddf34f707dfbe73a060a0bd7bd3f58018ff30 .  I wanted to prioritize use of `--binary`.

PR #3808, opened 2018-01-25
When a cabal file can't be found, warn that extra-deps could be the problem too

### LINKED ISSUE #3806: Confusing error message when a cabal file for an extra-dep can't be found

When the cabal file for an extra-dep can't be found, stack suggests the problem is with the `packages` field, not `extra-deps`.

### Steps to reproduce

For example:

1. Add a reference to a repo without a cabal file to your extra-deps (see YAML)
2. Run command `stack build`.

```yaml
# A fork of amazonka, which is a mega-repo, so the root directory doesn't have a cabal file
extra-deps:
  - git: https://github.com/tvision-insights/amazonka.git
    commit: 9d914189042d6d31ffee912436ddfacde08f1b50
```

### Expected

The user is told which dependency the cabal file couldn't be found for, or at least that an extra-dep was the problem, or that it *could* be an extra-dep that is the problem.

### Actual

Stack implies that my `packages` field is the problem with this message:

> Stack looks for packages in the directories configured in the 'packages' variable defined in your stack.yaml

(See #2729)

### Stack version

```
$ stack --version
Version 1.6.3, Git revision b27e629b8c4ce369e3b8273f04db193b060000db (5454 commits) x86_64 hpack-0.20.0
```

### Method of installation

* Official binary, downloaded from stackage.org or fpcomplete's package repository, upgraded via `stack upgrade`

## Potential fixes:

* Quick fix: Modify the message to mention extra-deps
* Potentially better: pass the package being searched for down so that the error message can be more detailed. I think usually this will be unnecessary, because the path itself is probably much more helpful for non-git repo downloads.

MaxGabriel:
* Also: slight refactor to use only `concat` instead of ++ in the error message
* Also: mention that a package.yaml file would also be accepted
	* The code that throws this is only searching for a .cabal file, but it runs hpack on the directory immediately before that to generate a cabal file, so it's effectively looking for a package.yaml file as well

Closes #3806

* [x] The documentation has been updated, if necessary. [Not necessary]

I used `stack install` to install my changes, then ran that executable:

image

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -32,6 +32,8 @@ Other enhancements:
 * The build progress bar reports names of packages currently building.
 * `stack setup --verbose` causes verbose output of GHC configure process.
   See [#3716](https://github.com/commercialhaskell/stack/issues/3716)
+* Improve the error message when an `extra-dep` from a path or git reference can't be found
+  See [#3808](https://github.com/commercialhaskell/stack/pull/3808)
 
 Bug fixes:
 * 1.6.1 introduced a change that made some precompiled cache files use
--- a/src/Stack/Types/Package.hs
+++ b/src/Stack/Types/Package.hs
@@ -52,9 +52,10 @@ instance Show PackageException where
         ]
     show (PackageNoCabalFileFound dir) = concat
         [ "Stack looks for packages in the directories configured in"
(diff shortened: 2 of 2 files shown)


mgsloan:
LGTM, thanks! :+1:

PR #3814, opened 2018-01-28
Hpack forward compatibility

sol:


package.yaml
```diff
@@ -15,7 +15,7 @@ homepage: http://haskellstack.org
 custom-setup:
   dependencies:
   - base
-  - Cabal
+  - Cabal == 2.*
```
sol: At least it didn't work on my machine with an older `Cabal`, but feel free to skip this change.

mgsloan:
Thanks!  Cabal constraint seems good too.

PR #3816, opened 2018-01-28
Add stack install script ElementaryOS support

mspaulding06:
This small change allows the stack installer script to recognize ElementaryOS (https://elementary.io/) as a derivative of Ubuntu so that required dependency packages get installed.


mgsloan:
Thanks!

PR #3817, opened 2018-01-29
Fix a cabal warning by adding Build_stack to generated-other-modules

sol:


package.yaml
```diff
@@ -276,12 +276,11 @@ executables:
   stack:
     main: Main.hs
     source-dirs: src/main
+    generated-other-modules: Build_stack
```
sol: This requires `hpack >= 0.23.0`, older versions o `hpack` will just ignore the field with a warning.  So I guess no big issue.

In addition, this will result in `cabal-version: >= 2`, which I believe is already required by `Setup.hs` (see my other PR).

package.yaml
```diff
@@ -276,12 +276,11 @@ executables:
...
     ghc-options:
     - -threaded
     dependencies:
     - stack
-    other-modules:
-    - Paths_stack
```
sol: This is not require for `hpack >= 0.19.0`.

mgsloan:
Sorry for the delay on feedback, this LGTM, thanks!

PR #3831, opened 2018-01-31
Move over to the external rio library

snoyberg:
Not sure if we're ready to make this change, especially since rio is still an experimental library. However, it is almost entirely code extracted from Stack itself, making the experimental aspect somewhat moot.


borsboom:
I do have one concern: we're trying to keep Stack compatible with nightly snapshots, and depending on a library that could have frequent breaking changes may increase the workload (since it would require a patchlevel release to hackage for any breaking change).

I _want_ to be in favour though, and it probably won't be _that_ much extra work.

snoyberg:
In defense of this change then: it would only kick in at the next major release of Stack, by which time we can make a decision that the rio API needs to be stable.

mgsloan:
I am in favor of switching to `rio` but not committing to a stable API just yet.  I have not had time recently to give it a thorough review.

Issue #3600, opened 2017-11-23
Nix doesn't work on Windows


domenkozar:
Given `stack.yaml`:

```
...
nix:
  enable: true
...
```

The project can't be build on Windows, since stack will try to use Nix also there. While it's possible to use Nix via cygwin or WSL, that's rarely done.

I suggest we disable Nix on Windows (and optionally support forcing it if really wanted).

mgsloan:
I hear that nix on wsl can work - https://www.tweag.io/posts/2017-11-10-nix-on-wsl.html - though noidea if that would work with stack.

IMHO it'd make sense to just not have `enable: true`.  There are a few other stack.yaml options that allow you to break builds for users with configurations different than yours, for example, `extra-lib-dirs`.  Not saying this is ideal, but it's the way things are, currently.  Definitely some interesting tradeoffs between configurability and universality.

domenkozar:
@mgsloan we do default to `false` currently, but that trips people since they forget to pass `--nix` and stack doesn't warn about it. I'd like to make default configuration sane, but Windows prevents that.

mgsloan:
I see!  This is another potential usecase for https://github.com/commercialhaskell/stack/issues/3369

mgsloan:
I've opened a PR, https://github.com/commercialhaskell/stack/pull/3833 which always overrides with `enable: false` on windows.

Previously when I suggested that #3369 be implemented for this, I didn't properly consider that when using WSL it will be built as an ELF binary that thinks it's targeting linux.

mgsloan:
Fixed!

PR #3833, opened 2018-01-31
Disable nix on windows (#3600)

mgsloan:

This change is untested because I do not have convenient access to a windows box.  However, the change is trivial, it should work fine.

### DIFF
--- a/src/Stack/Config/Nix.hs
+++ b/src/Stack/Config/Nix.hs
@@ -11,18 +11,20 @@ module Stack.Config.Nix
 import Stack.Prelude
 import qualified Data.Text as T
 import Distribution.System (OS (..))
+import Stack.Constants
 import Stack.Types.Version
 import Stack.Types.Nix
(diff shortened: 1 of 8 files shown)


src/Stack/Config/Nix.hs
```diff
@@ -32,6 +34,12 @@ nixOptsFromMonoid NixOptsMonoid{..} os = do
         nixShellOptions = fromFirst [] nixMonoidShellOptions
                           ++ prefixAll (T.pack "-I") (fromFirst [] nixMonoidPath)
         nixAddGCRoots   = fromFirst False nixMonoidAddGCRoots
+    nixEnable <-
+      if nixEnable0 && osIsWindows
```
snoyberg: I can't say exactly what strikes me about this, but I find this code particularly elegant.

mgsloan: Thanks! :)

PR #3842, opened 2018-02-03
Fix #3812 on stable

### LINKED ISSUE #3812: local file archive paths in custom-snapshot should be relative to snapshot yaml
If I use local file archives in custom snapshots the relative path is resolved against either pwd or project root (difference is hard to tell as I usually call stack from project root).

This is imho unintuitive (or worse) and doesn't help towards reproducible builds.

### Steps to reproduce

For example:
1. Create a new project, e.g. `stack new custom-snapshot simple-hpack`
2. `cd custom-snapshot`
3. `mkdir -p some/path`
4. `touch some/path/archive.tar.gz`
5. Create file `some/snapshot.yml`
```
---
resolver: ghc-8.2.2
name: vendored # For user display only
packages:
- archive: ./path/archive.tar.gz
```
6. Use some/snapshot.yml as resolver
7. `stack build`

Result:
```
23:51 $ stack build
Unable to load cabal files for snapshot
InvalidUrlException "./path/archive.tar.gz" "Invalid URL"
```

If you instead have `./some/path/archive.tar.gz` stack correctly complains about the invalid archive. 

### Stack version

```
$ stack --version
Version 1.6.3, Git revision b27e629b8c4ce369e3b8273f04db193b060000db (5454 commits) x86_64 hpack-0.20.0
```

loewenstein:


test/integration/tests/1265-extensible-snapshots/Main.hs
```diff
@@ -1,8 +1,10 @@
-import StackTest
+import           StackTest
 
 main :: IO ()
 main = do
     stack ["build", "async"]
     stackErr ["build", "zlib-bindings"]
     stack ["build", "--stack-yaml", "stack-modify-lts.yaml", "async"]
+    stack ["build", "--stack-yaml", "stack-local-snapshot.yaml", "async"]
```
mgsloan: I think maybe to test this case it would need to try to build `package`, right?  Also to test that the paths are correctly local, it would need to use `withCurrentDirectory`

src/Stack/Snapshot.hs
```diff
@@ -260,17 +262,29 @@ loadResolver (ResolverCustom url loc) = do
...
+      let resolveLocalArchives sd = sd {
+            sdLocations = resolveLocalArchive <$> sdLocations sd
+          }
+          resolveLocalArchive (PLOther (PLArchive archive)) = 
+            PLOther $ PLArchive $ archive {
+              archiveUrl = T.pack $ resolveLocalFilePath (T.unpack $ archiveUrl archive)
+            }
+          resolveLocalArchive pl = pl
+          resolveLocalFilePath path =
+            if not $ isURI path && FilePath.isRelative path
```
mgsloan: Hmm, I think this condition is saying "If it's not a uri and relative"  AFAICT that should never happen.

Perhaps the logic should be `if not (isURI path) && FilePath.isRelative path`.  Also I recommend trying it manually on your real world case to ensure that it is indeed a fix.

mgsloan:
Mostly looks good, just a few comments to address

Thanks for working on this! :+1:

test/integration/tests/1265-extensible-snapshots/Main.hs
loewenstein: I think building package is not necessary (actually not possible, `package` merely contains a valid `.cabal` file).

But you are right, the file system layout does probably not require the fix... I'll add a commit to push snapshots and package one level deeper.

src/Stack/Snapshot.hs
mgsloan: Hmm, yeah I guess if `path` is absolute then the behavior of `</>` will forget about the path on the left.  Seems like perhaps the isRelative test is not necessary at all here?  Would be good to have a comment about it.

loewenstein: Hmm, I am not sure the "forget about the left path" is very obvious although it is indeed documented like that.

I'd stick to "don't touch if it is either a URI or an absolute path".

I'll submit another commit later this week and create a PR to fix master. I won't be able to test my exact use case because of #3861 though.

loewenstein:
@mgsloan also created #3928 to fix the same thing on master.

loewenstein:
@mgsloan how are the chances to get this merged and released?

Alternatively how close are we to get stack 1.7 out of the door? Would be good to get #3928 in before that.

I'd love to help getting https://github.com/commercialhaskell/stack/issues/3861 solved as well.

snoyberg:
@loewenstein Sorry for the delayed response here. As you probably saw, we're in the middle of the 1.7 release process now. It looks like these fixes are already in place on master. Is that the case?

loewenstein:
@snoyberg partly... There is #3928 that would be good to get in before a final release.

It fixes a problem with the original commit that indeed already landed on master.

snoyberg:
I just merged #3928 :)

Issue #3809, opened 2018-01-25
Use hpack executable from PATH if it is newer than what stack ships with?


sol:
The latest version of `hpack` has a `--numeric-version` option.  With this it would be possible for `stack` to check if there is a newer version of `hpack` installed on a users system and if so use that instead of the bundled `hpack`.

- Check if `hpack` is on the users `PATH`, if no use bundled `hpack`
- If yes, try `hpack --numeric-version`, if it fails use bundled `hpack`
- If it succeeds then check if it is newer than the bundled `hpack` and use it if so, otherwise use bundled `hpack`

This is just a suggestion, not a request to add?

(cc @scott-fleischman who did related stuff before)

mgsloan:
Hmm, sure, this makes sense to me!

amir:
Does this also require addition of a config to allow the override (e.g., `configAllowHpacCommandNewer`)?

quasicomputational:
+1 from me, with a switch to disable if you really want the wired-in hpack.

vlatkoB:
IMHO, `hpack` should not be included in `stack` at all, but should  always use external/system installed `hpack`.

`hpack` is a separate project, completely unrelated to `stack`'s functionality, has its own pace of release (bug fixes and new features), which will force `stack` to release more often just to keep up with `hpack`, or fall behind, increasing the maintenance burden.
(For example, latest `stack-1.6.5` packs `hpack-0.20.0`, and `hpack` is already on version 0.28.0.)

I mean, if you want to use the `hpack` functionality, you should have `hpack` installed (likewise, if you want documentation, you have to install `haddock`, otherwise `stack` reports about missing bin).

philderbeast:
Another option aside from being bundled with stack or being picked up from the environment is to use project relative tools, to have stack install a version of a tool relative to the project. These tool versions could then be specified and locked down in the project. I do this at the moment when installing dhall;

With a separate [stack-dhall.yaml](https://github.com/BlockScope/flare-timing/blob/master/stack-dhall.yaml), I can use a different resolver for tooling than I do for the project;

```
resolver: lts-12.5
flags: {}
packages: []
extra-deps:
- dhall-1.16.1
- hpack-dhall-0.3.0
- dhall-nix-1.1.5
- hnix-0.5.2
- serialise-0.2.0.0
local-bin-path: __shake-build
work-dir: .stack-work-dhall
```

```
[...]

PR #3856, opened 2018-02-08
Search for a newer user installed hpack

amir:

This is an attempt at fixing #3809. There's an open question on that issue on whether the behaviour should be configurable. I'll update, PR, the documentation and ChangeLog once that's decided.

It's been manually tested with a fresh build of stack and new versions of hpack with `--numeric-version` present.

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -42,10 +42,12 @@ module Stack.Package
 
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as C8
+import qualified Data.ByteString.Lazy.Char8 as BL8
 import           Data.List (isSuffixOf, isPrefixOf)
 import           Data.List.Extra (nubOrd)
 import qualified Data.Map.Strict as M
 import qualified Data.Set as S
+import           Data.Char (isSpace)
 import qualified Data.Text as T
 import           Data.Text.Encoding (decodeUtf8, decodeUtf8With)
 import           Data.Text.Encoding.Error (lenientDecode)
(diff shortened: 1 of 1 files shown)


src/Stack/Package.hs
```diff
@@ -1362,6 +1364,57 @@ findOrGenerateCabalFile pkgDir = do
...
+                ]
+
+  case eres of
+    Right hpackPath -> return $ HpackCommand (toFilePath hpackPath)
+    Left err -> do
+      logDebug err
+      return HpackBundled
+
+  where
+    envSettings =
```
mgsloan: Should be able to inherit EnvSettings from somewhere or perhaps use default.  At least, `esIncludeLocals` should be `True`, that will actually make a difference.

src/Stack/Package.hs
```diff
@@ -1362,6 +1364,57 @@ findOrGenerateCabalFile pkgDir = do
             _:_ -> Left $ PackageMultipleCabalFilesFound pkgDir files
       where hasExtension fp x = FilePath.takeExtension fp == "." ++ x
 
+-- | Searches user path for hpack executables newer than bundled one.
+findUserExecutableHpack :: HasConfig env => Config -> RIO env HpackExecutable
```
mgsloan: Main issue with this is that it will re-query hpack version every single time.  I think it makes sense to cache it.  The pattern of having an `IORef (Maybe ...)` in scope is used elsewhere, and I think would be good to use here.

mgsloan:
Looks good, just a couple comments.  Thanks for working on this! :+1:

snoyberg:
@amir are you interested in continuing this PR?

amir:
@snoyberg yes, I am. I'll try to address the reviewers concerns soon.

snoyberg:
Awesome, thanks for the quick response!

mihaimaruseac:
Ping on this, as it's becoming stale.

mihaimaruseac:
Very stale, should we close this?

snoyberg:
SGTM

PR #3858, opened 2018-02-10
stack build: Add --keep-tmp-files flag. Fixes #3857

### LINKED ISSUE #3857: How to keep temporary files on build failures?
In a project where my build is failing, I'm trying to inspect the intermediate files that GHC generates.

I tried

```
stack build --ghc-options=-keep-tmp-files
```

but it doesn't work, because while GHC doesn't delete its temp files, stack does.

I think it's because of this: https://github.com/commercialhaskell/stack/blob/b48f18fab2288b06464ca4a17bd21a99388afc0b/src/Stack/Build/Execute.hs#L326-L327

Is there a way I can convince stack to keep the temporary directory for inspection? If not, I'd like to feature request that.

nh2:
See #3857.

I've tested this by building stack itself with `stack build --keep-going`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -37,6 +37,11 @@ Other enhancements:
 * Nix integration is now disabled on windows even if explicitly enabled,
   since it isn't supported. See
   [#3600](https://github.com/commercialhaskell/stack/issues/3600)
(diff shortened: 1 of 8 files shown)


mgsloan:
:+1:

Issue #3073, opened 2017-03-19
Bad C-c handling during 'stack init'


sjakobi:
(This is a follow up from the discussion in #3055)

```
~/s/hackage-server (master|✔) $ stack init
Looking for .cabal or package.yaml files to use to init the project.
Using cabal packages:
- hackage-server.cabal
- tests/unpack-checks/missing-configure-0.1.0.0/missing-configure.cabal
- tests/unpack-checks/correct-package-0.1.0.0/correct-package.cabal

Selecting the best among 10 snapshots...

* Partially matches lts-8.5
    ....

Downloaded nightly-2017-03-19 build plan.    
Missing some cabal revision files, updating indices
Selected mirror https://s3.amazonaws.com/hackage.fpcomplete.com/                                 
Downloading timestamp                                                                            
Downloading snapshot                                                                             
Updating index                                                                                   
UpException user interrupt when using mirror https://s3.amazonaws.com/hackage.fpcomplete.com/      
Selected mirror http://hackage.fpcomplete.com/                                                   
Downloading timestamp                                                                            
Downloading snapshot                                                                             
Updating index                                                                                   
Updated package list downloaded                                                                  
```

Note the line `UpException user interrupt when using mirror https://s3.amazonaws.com/hackage.fpcomplete.com/` – this is where I pressed C-c.

I would have expected stack to abort then but it appears as if stack simply retries downloading the index.

mgsloan:
We don't have any such retry logic, looks like it is an upstream issue with hackage-security: https://github.com/haskell/hackage-security/issues/187

snoyberg:
I've created a patch for hackage-security which should hopefully address this:

https://github.com/haskell/hackage-security/pull/202

I've also pushed a branch `3073-hackage-security-async-exceptions` with commit 4bf68f02d901a6ffc7f4b81a22985d98435fbb14, which moves over to this patched hackage-security. It would be great to get some feedback on whether this patch fixes the issues described here.

snoyberg:
I just tested the repro above: with previous builds of Stack against unpatched `hackage-security`, hitting `Ctrl-C` during the `Updating index` phase causes the output above. With this patched version, the program exits immediately.

dcoutts:
Thanks for the contributions upstream. I'd suggest to persist with getting the contributions or equivalent fixes into the upstream package as that way everyone benefits from everyone else's improvements.

borsboom:
[...]

PR #3865, opened 2018-02-14
Include patched hackage-security for #3073

snoyberg:
NOTE: This is included via an extra-dep, which would constitute the
first time Stack would include a patched version of an upstream library.
This is due to the fact that
https://github.com/haskell/hackage-security/pull/203 is likely not going
to be merged, despite fixing issues affecting Stack. This leaves us with
(AFAICT) 4 choices at the Stack level:

1. Continue using the officially released upstream version of
   hackage-security, bugs and all
2. Fork hackage-security on Hackage, and depend on the fork
3. Inline the code from hackage-security into Stack itself, and drop the
   explicit dependency on hackage-security
4. Include hackage-security via an `extra-dep` pointing at a Git commit.
   Our official builds will use the patched version of hackage-security,
   and anyone building from Hackage will end up with the unpatched version

This PR represents approach (4). If and when the PR is merged and
released to Hackage, this becomes a non-issue. But generally speaking,
we should have a policy in Stack for handling these kinds of upstream
issues cases.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -34,6 +34,12 @@ Bug fixes:
   we were forgetting to record the result, which meant that all tests
   always ran even if they had already passed before. See
   [#3770](https://github.com/commercialhaskell/stack/pull/3770).
+* Includes a patched version of `hackage-security` which fixes both
+  some issues around asynchronous exception handling, and moves from
+  directory locking to file locking, making the update mechanism
+  resilient against SIGKILL and machine failure. See
+  [hackage-security #187](https://github.com/haskell/hackage-security/issues/187)
+  and [#3073](https://github.com/commercialhaskell/stack/issues/3073).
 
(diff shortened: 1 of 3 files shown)


snoyberg:
I've asked @hvr for clarification on the release timing of hackage-security in https://github.com/haskell/hackage-security/pull/202#issuecomment-365635050. I'd say that, if the release isn't scheduled to happen this week, we move ahead with this PR, and then update to use the Hackage version when available. (Meanwhile, I'll update this PR to use the master branch of haskell/hackage-security instead.)

snoyberg:
I've pushed a commit to rebase against the latest stable branch and include further upstream changes on hackage-security (for better or worse, we should match upstream as closely as possible). Since we're probably not getting an answer any time soon, let's merge this as soon as CI goes green.

PR #3878, opened 2018-02-20
Switch to Cabal 2.2 (prerelease)

snoyberg:

CC @phadej 

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -47,8 +47,7 @@ import           Data.List.Extra (nubOrd)
 import qualified Data.Map.Strict as M
 import qualified Data.Set as S
 import qualified Data.Text as T
-import           Data.Text.Encoding (decodeUtf8, decodeUtf8With)
-import           Data.Text.Encoding.Error (lenientDecode)
+import           Data.Text.Encoding (decodeUtf8)
 import           Distribution.Compiler
 import           Distribution.ModuleName (ModuleName)
 import qualified Distribution.ModuleName as Cabal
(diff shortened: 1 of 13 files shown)


src/Stack/Package.hs
```diff
@@ -129,14 +127,14 @@ rawParseGPD
...
+      Left (mversion, errs) -> throwM $ PackageInvalidCabalFile key
+        (fromCabalVersion <$> mversion)
+        errs
+        warnings
+      Right gpkg -> return (warnings, gpkg)
   where
-    chars = T.unpack (dropBOM (decodeUtf8With lenientDecode bs))
-
-    -- https://github.com/haskell/hackage-server/issues/351
-    dropBOM t = fromMaybe t $ T.stripPrefix "\xFEFF" t
```
snoyberg: @phadej Can we rely on Cabal stripping out the byte-order marker from the bytestring, or do we need to do something more intelligent here?

phadej: Cabal strips BOM (and warns about, if sees it). `parseGenericPackageDescription` also verifies that `ByteString` looks like valid UTF8, so you don't to do that either.

TL;DR you don't need to do any preprocessing.

snoyberg: Awesome, thanks!

mgsloan:
LGTM :+1:

snoyberg:
After fighting with version bounds in a few different places, this is not blocked on:

https://github.com/fizruk/http-api-data/issues/72

The issue: http-api-data has a `Cabal < 2.2` unnecessarily in its `custom-setup` stanza. The maintainer wants to wait until `blaze-builder` adds GHC 8.4 support before relaxing that bound. So our choices are:

* Include `allow-newer` in the `stack.yaml` files, making the released version basically unbuildable directly from Hackage
* Hold off indefinitely on the upgrade
* Try to push ahead with dropping http-api-data from the dependency chain of persistent (https://github.com/yesodweb/persistent/issues/800)
* Have Cabal as a submodule and/or a forked version on Hackage, and only pull in the relevant parsing code we need, bypassing the erroneous upper bounds

I kind of like the last option, as it makes us more disconnected from the Cabal release process over all, but it adds some maintenance overhead. And from my initial experience, there's unfortunately a very large module graph involved in the parsing code.

PR #3888, opened 2018-02-23
Initialize a new project only when no stack.yaml

gxmas:

`stack new` unpacks the project template and then initializes the unpacked project.   When a project template contains a stack.yaml, initializing the project fails due to the presence of the stack.yaml files.  Skip the initialization if the project template contains a stack.yaml.

The change was tested by using project templates that contain stack.yaml files.


snoyberg:
Merging despite Travis failure, since it's just a timeout that's already occurring on master.

PR #3890, opened 2018-02-25
Allow 'github' shorthand for extra-deps (fixes #3873)

### LINKED ISSUE #3873: Allow github shorthand for extra-deps
It's quite common to include an `extra-dep` from Github. The common way to do it, via a `git: https://github.com/...` URL, is actually _not_ the best. It's better to instead use a link to the archive, since this doesn't require downloading the entire Git history, and results in a faster transport. For example, see this commit:

https://github.com/commercialhaskell/stack/pull/3831/commits/638efeaef17a02cb00201bf85d9ef557999ae83f

I'd like to recommend that we add a convenience syntax that does the right thing, namely, replacing:

```
- https://github.com/commercialhaskell/rio/archive/09654f9fcbdcd96d0f5102796b32fdac5da7260e.tar.gz
```

with

```
- github: commercialhaskell/rio
  commit: 09654f9fcbdcd96d0f5102796b32fdac5da7260e
```

Or perhaps even:

```
- github: commercialhaskell/rio@09654f9fcbdcd96d0f5102796b32fdac5da7260e
```

And have this generate the archive link described above.

mitchellwrosen:
First, sorry to @lhcopetti for stealing this one - I neglected to notice you were interested in doing this.

This patch adds support for this syntax of `extra-deps`:

```
- github: commercialhaskell/rio
  commit: 09654f9fcbdcd96d0f5102796b32fdac5da7260e
```

with an additional optional `subdirs` field. The above would resolve to this archive:

```
https://github.com/commercialhaskell/rio/archive/09654f9fcbdcd96d0f5102796b32fdac5da7260e.tar.gz
```

Please let me know if I should tweak anything, including adding more tests. Thanks!

### DIFF
--- a/src/Data/Aeson/Extended.hs
+++ b/src/Data/Aeson/Extended.hs
@@ -150,7 +150,7 @@ instance IsString WarningParserMonoid where
 
 -- Parsed JSON value with its warnings
 data WithJSONWarnings a = WithJSONWarnings a [JSONWarning]
-    deriving Generic
+    deriving (Eq, Generic, Show)
 instance Functor WithJSONWarnings where
     fmap f (WithJSONWarnings x w) = WithJSONWarnings (f x) w
 instance Monoid a => Monoid (WithJSONWarnings a) where
@@ -160,6 +160,7 @@ instance Monoid a => Monoid (WithJSONWarnings a) where
 -- | Warning output from 'WarningParser'.
 data JSONWarning = JSONUnrecognizedFields String [Text]
                  | JSONGeneralWarning !Text
+    deriving Eq
 instance Show JSONWarning where
     show (JSONUnrecognizedFields obj [field]) =
         "Unrecognized field in " <> obj <> ": " <> T.unpack field
--- a/src/Stack/Types/BuildPlan.hs
+++ b/src/Stack/Types/BuildPlan.hs
@@ -270,7 +270,7 @@ instance subdirs ~ Subdirs => FromJSON (WithJSONWarnings (PackageLocation subdir
             , archiveHash = msha'
             }
 
-        github = withObjectWarnings "PLArchive" $ \o -> do
+        github = withObjectWarnings "PLArchive:github" $ \o -> do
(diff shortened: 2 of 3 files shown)


src/Data/Aeson/Extended.hs
```diff
@@ -150,7 +150,7 @@ instance IsString WarningParserMonoid where
 
 -- Parsed JSON value with its warnings
 data WithJSONWarnings a = WithJSONWarnings a [JSONWarning]
-    deriving Generic
+    deriving (Eq, Generic, Show)
```
snoyberg: Is this change necessary for the PR, or just to help with debugging?

src/Stack/Types/BuildPlan.hs
```diff
@@ -269,6 +270,26 @@ instance subdirs ~ Subdirs => FromJSON (WithJSONWarnings (PackageLocation subdir
             , archiveHash = msha'
             }
 
+        github = withObjectWarnings "PLArchive" $ \o -> do
```
snoyberg: It should probably say `PLArchive:github` or something like that instead, this is just used for error messages, and it would be useful to know what part of the code failed the parsing.

src/test/Stack/Types/BuildPlanSpec.hs
```diff
@@ -0,0 +1,94 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Stack.Types.BuildPlanSpec where
```
snoyberg: Nice! 👍

src/Data/Aeson/Extended.hs
mitchellwrosen: `Eq` and `Show` are required for test code like:

```haskell
decode' contents `shouldBe` Just (WithJSONWarnings expected [])
```

src/Stack/Types/BuildPlan.hs
mitchellwrosen: Done

mitchellwrosen:
Of course!

PR #3896, opened 2018-03-02
Fixed get-stack.sh script to match manual installation instructions

tjlaxs:

Debian, Fedora and CentOS were missing git and gnupg from the install dependencies. Nothing massive just git and gnupg were missing. Ubuntu already had these and the manual installation instruction suggested installing them.

I admit that I didn't test this, but now the dependencies match the ones on https://docs.haskellstack.org/en/stable/install_and_upgrade/#linux-generic

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -47,6 +47,8 @@ Other enhancements:
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


snoyberg:
Thanks!

PR #3910, opened 2018-03-10
'stack setup' can install GHC on linux-aarch64

borsboom:
GHC release team has provided linux-aarch64 bindists since GHC 8.2.1.  This adds support by `stack setup` to install them.  Note, however, that manual installation of the correct LLVM version is still required.


mgsloan:
LGTM. Should it mention to the user that LLVM needs to be installed, or perhaps the ensuing error messages will be sufficient?

borsboom:
GHC's error messages about that are pretty clear.  And the same problem already exists for ARMv7, so I think improving that clarity should be a separate issue/PR if desirable.

PR #3916, opened 2018-03-13
Don't require yaml and persistent-sqlite in snapshots

snoyberg:
The recent nightly-2018-03-13 uncovered a bug in Stack: it assumes that
these two packages are present in all snapshots to apply a historical,
hacky fix. The hacky fix has not been required for quite some time, so
we simply gate on whether we're using an older snapshot or not.

### DIFF
--- a/src/Stack/Snapshot.hs
+++ b/src/Stack/Snapshot.hs
@@ -32,6 +32,7 @@ import qualified Data.Conduit.List as CL
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Map as Map
 import qualified Data.Set as Set
+import           Data.Time (toGregorian)
 import qualified Data.Text as T
 import           Data.Text.Encoding (encodeUtf8)
 import           Data.Yaml (decodeFileEither, ParseException (AesonException))
@@ -686,10 +687,9 @@ findPackage platform compilerVersion (gpd, loc, localLoc) = do
   where
     PackageIdentifier name _version = fromCabalPackageIdentifier $ C.package $ C.packageDescription gpd
 
--- | Some hard-coded fixes for build plans, hopefully to be irrelevant over
--- time.
+-- | Some hard-coded fixes for build plans, only for hysterical raisins.
 snapshotDefFixes :: SnapshotDef -> SnapshotDef
-snapshotDefFixes sd | isStackage (sdResolver sd) = sd
+snapshotDefFixes sd | isOldStackage (sdResolver sd) = sd
     { sdFlags = Map.unionWith Map.union overrides $ sdFlags sd
     }
   where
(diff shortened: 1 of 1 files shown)


borsboom:
LGTM.  Should we make a minor release with this fix?

snoyberg:
If it's convenient, a minor release would be good.

hvr:
It would be great if a Stack point release could be done sooner than later to avoid repeating the unnecessary drama from last time which escalated into package forking and even vendetta campaigns being orchestrated against maintainers of perfectly valid packages which happen to expose bugs in Stack...

mgsloan:
If would be great if you would remove a single `-` character in their cabal file.  But you refuse.  Why?

mgsloan:
If would be great if you would remove a single - character in the cabal file of cassava that will have essentially zero impact on cabal-install builds. But you refuse. Why?

mgsloan:
If would be great if you would remove a single `-` character in the cabal file of cassava that will have essentially zero impact on cabal-install builds. But you refuse. Why?

hvr:
Because at the time it made no sense on a technical level, and the obvious thing to do was to fix the component at fault; and I didn't expect it would take Stack _several_ months to acknowledge and fix the bug; and some time later you said it yourself that it's "water under the bridge" anyway, and the bug had eventually been fixed in Stack anyway. And there's surely several other packages in the Package index possibly even predating Stack which would equally expose the incorrect flag parser in Stack, but conveniently, nobody brought that up that Stack didn't have proper QA in place to make sure it could parse all of the package index (and before you ask; the Cabal project does have such a test as part of its QA to ensure our parser is able to parse the full Hackage package index; you're welcome to steal that idea).

snoyberg:
This entire discussion is off topic to the matter at hand. Please
discontinue immediately.

hvr:
@snoyberg :-)

DanBurton:
@hvr Note that with the nightly-2018-03-16 snapshot (and onwards), `yaml` and `persistent-simple` are included, which is effectively a workaround for this issue. Users of older versions of stack will thankfully no longer encounter errors from this (because snapshots are able to accommodate old versions of stack) in the same way that they encountered errors with parsing the cassava flag (in which snapshots were not able to make accommodations for old versions of stack).

tl;dr a new release of stack is not urgently needed, due to a workaround which is now in place.
n.b. this in part was thanks to https://github.com/haskell-infra/hackage-trustees/issues/143, as well as several people graciously making package releases in order to unblock persistent-simple.

mgsloan:
Yep, the breakage was restricted to a few nightly snapshots, thanks to the
work of diligent maintainers.  It isn't even close to the magnitude of the
cassava breakage which continues to this day.

PR #3937, opened 2018-03-22
Install zlib development package

afcowie:
When building programs on Fedora Linux, you'll run into needing to build the **zlib** Haskell package in all likelihood, and that requires the **zlib-devel** Fedora package. 

The `get-stack.sh` script makes sure **gmp-devel** is present (that's the other one new users always seem to be missing). The script installs **zlib** but not **zlib-devel**. This patch changes that package set to include the latter.

AfC

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -188,7 +188,7 @@ do_debian_install() {
 # and install the necessary dependencies explicitly.
 do_fedora_install() {
   install_dependencies() {
-    dnf_install_pkgs perl make automake gcc gmp-devel libffi zlib xz tar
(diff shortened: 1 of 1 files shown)


borsboom:
Hmm... `zlib` and `gmp-devel` are needed for `stack` and `ghc` themselves to work, rather than for building particular packages, so that's why they're included.  `zlib-devel`, on the other hand, isn't required to build a basic `main = putStrLn "hello"` program.

So the question is: does it make sense to have get-stack.sh install devel libraries for commonly used dependencies?  If so, which ones?  `zlib` is obviously a good candidate, so this change does seem reasonable.

Issue #2586, opened 2016-09-12
Stack setup fails to install ghc on NixOS


tdimiduk:
Stack is failing to install ghc 7.10 for me on NixOS unstable. I have ghc 8.0.1 installed as my default ghc:
$ ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.0.1
But stack wants to use 7.10
$ stack new test
<snip>
Downloaded lts-6.17 build plan.    
Did not find .cabal file for servant-yaml-0.1.0.0 with Git SHA of 71c0a55d0a877954d9590e285c0eb861ace2d8cc
Right Nothing
* Matches lts-6.17

Selected resolver: lts-6.17
Initialising configuration using resolver: lts-6.17
Total number of user packages considered: 1
Writing configuration to file: test/stack.yaml
All done.

$ cd test
$ stack setup
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Already downloaded.                 
Running /tmp/stack-setup3673/ghc-7.10.3/configure --prefix=/home/tdimiduk/.stack/programs/x86_64-linux/ghc-7.10.3/ in directory /tmp/stack-setup3673/ghc-7.10.3/ exited with ExitFailure 1

checking for path to top of build tree... 
utils/ghc-pwd/dist-install/build/tmp/ghc-pwd-bindist: line 3: /tmp/stack-setup3673/ghc-7.10.3/utils/ghc-pwd/dist-install/build/tmp/ghc-pwd: No such file or directory
configure: error: cannot determine current directory

It looks like the ghc-pwd binary wants to dynamically link against libraries I do not have present, and thus is failing, preventing stack from installing its local version of ghc 7.10.

tdimiduk:
I have determined a workaround, I can use a nix-shell with the correct ghc:
$ nix-shell -p haskell.compiler.ghc7103
and then I can run stack build just fine on packages using lts's needing ghc 7.10.

So this issue may be unnecessary, feel free to close it, but at least this way hopefully a workaround to this issue will be googleable.

domenkozar:
This was documented in recent Stack release, but I wonder if the error message should be clearer.

tdimiduk:
I clearer error message would have saved me a chunk of time, so I would be
in favor.

mgsloan:
[...]

PR #3938, opened 2018-03-22
Enable Nix mode by default when running under NixOS

rdnetto:
Stack only works under NixOS in Nix mode - see #2586. However, it still defaults to `--no-nix`, which means that it's broken by default.

This PR changes that behaviour to always enable Nix mode when running in NixOS.

Tested by running `stack ghci` under Sabayon Linux and NixOS.


MartinPotier:
> Stack only works under NixOS in Nix mode

Well, if you have docker installed, it's really useful to use it without nix mode and with docker mode to produce compatible binaries.

Issue #2310, opened 2016-06-27
Add `stack hoogle --server` command


chrisdone:
Which would run `stack hoogle -- --server --local`. `--local` is  needed to be able to click links to local files.

ndmitchell:
I wonder if Hoogle should launch a web browser on --server, and then when the web browser disappears it could close the server. Maybe as a `stack hoogle --launch`? I'm quite happy to adjust the semantics/API of Hoogle to make this better for stack users.

chrisdone:
It could be a handy shortcut. The only reason you start the server is to see the web page after all.

ndmitchell:
Ticket raised at https://github.com/ndmitchell/hoogle/issues/175.

jhance:
Small correction, I think it should actually be `stack hoogle -- server --local`.

ndmitchell:
`stack hoogle -- server --local` works today so I think everything is good.

PR #3941, opened 2018-03-23
Introduce Hoogle server command line option

SkyWriter:
This PR introduces support for `stack hoogle --server` option to start local Hoogle server as per https://github.com/commercialhaskell/stack/issues/2310.

Here's what I did:
* [x] ~Added the corresponding option and its handling~
* [x] ~Introduced a record in the ChangeLog.md file for the changes~
* [x] ~I haven't found much info on Hoogle in the docs, so I figured no changes are necessary~

The testing was manual by issuing `stack hoogle --server` on the command line and making sure it works as expected.

As a side note: besides options mentioned in the issue I added `-p 8080` to start Hoogle on a non-privileged port. This is the default Hoogle behaviour since 5.0.13 (released on 2017-07-30), but I figured that there may be older versions in the wild. As there's no way to pass options directly to Hoogle via `stack hoogle` those older versions would try binding to 127.0.0.1:80 and fail without a real way to fix it apart from using `stack exec` to supply the `-p` option.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -49,6 +49,9 @@ Other enhancements:
   i.e. `stack build --keep-tmp-files --ghc-options=-keep-tmp-files`.
   See [#3857](https://github.com/commercialhaskell/stack/issues/3857)
 * Improved error messages for snapshot parse exceptions
+* `stack hoogle` now supports a new flag `--server` that launches local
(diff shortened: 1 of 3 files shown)


snoyberg:
Works great, thanks!

Issue #3904, opened 2018-03-06
Possibility to specify a base path to unpack a package


Unisay:
### Feature request

Currently, `stack unpack` uses current folder to unpack a package into.

```shell
$ stack --version; and stack unpack --help                                                                                                                                  19:58:47
Version 1.6.5, Git revision 24ab0d6ff07f28276e082c3ce74dfdeb1a2ca9e9 (5514 commits) x86_64 hpack-0.20.0
Usage: stack unpack PACKAGE [--help]
  Unpack one or more packages locally
```
It would be convenient to be able to specify an optional destination path to unpack into it, for example:
```shell
$ stack unpack ekg --to /tmp                                                                                                                                                     
ekg-0.4.0.14: download
Unpacked ekg-0.4.0.14 to /tmp/ekg-0.4.0.14/
```

mgsloan:
Hmm, I suppose I can see the convenience of this!  The alternative would be to `cd /tmp` and then `stack unpack ekg --stack.yaml /home/...`.

I'd merge a PR for this.

Unisay:
Currently I need this as a part of a continuous delivery pipeline build where I'd like to avoid changing directory back and forth. At the moment I workaround it by redirecting stderr to stdout, parsing output folder, then moving it to a desired location.

So I am going to try creating a PR for it.

snoyberg:
Fixed by #3947, thanks!

PR #3947, opened 2018-03-28
Add --to option to the unpack command to specify destination

SkyWriter:
This PR introduces support for `stack unpack pkgname --to /dest/dir` option to unpack pkgname into the specified directory as per #3904.

Here's what I did:

- [x] ~Added the corresponding option and its handling~
- [x] ~Introduced a record in the ChangeLog.md file for the changes~
- [x] ~Updated the docs where deemed necessary~

`--to` option is optional (pun intended), the default behaviour without it remains unchanged compared to the previous Stack versions.

The following manual tests have been conducted:
- Omitting `--to`
- Specifying `--to .`
- Specifying `--to /tmp`
- Specifying `--to /directory/i/have/no/write/access/to`
- Using Unicode as a directory name `--to "असतोमा सद्गमय ।"` (who doesn't use Sanskrit to name directories these days, right?)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -65,6 +65,8 @@ Other enhancements:
   i.e. `stack build --keep-tmp-files --ghc-options=-keep-tmp-files`.
   See [#3857](https://github.com/commercialhaskell/stack/issues/3857)
 * Improved error messages for snapshot parse exceptions
+* `stack unpack` now supports a `--to /target/directory` option to
+  specify where to unpack the package into
 
 Bug fixes:
 
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -517,6 +517,13 @@ Unpacked yackage-0.8.0 to /var/home/harendra/yackage-0.8.0/
 cueball:~$ cd yackage-0.8.0/
 ```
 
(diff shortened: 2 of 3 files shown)


SkyWriter:
P.S. I tried going the hard way with `relDirOption` and types all around, but it ended up being too strict from the user perspective (e.g. not accepting paths with ".."), so I ended up going the naive way.

SkyWriter:
Not sure what's wrong with the build. The only thing that stands out is hlint being unhappy about Docker.hs, but I've never touched the file.

snoyberg:
Rebasing against master will likely fix the style build issue.

ChangeLog.md
```diff
@@ -49,6 +49,8 @@ Other enhancements:
   i.e. `stack build --keep-tmp-files --ghc-options=-keep-tmp-files`.
   See [#3857](https://github.com/commercialhaskell/stack/issues/3857)
 * Improved error messages for snapshot parse exceptions
+* `stack unpack` now supports a `--to /target/directory` option to
```
snoyberg: I think it would be worth clarifying that the `package-version` subdirectory will still be created. It's slightly ambiguous to me right now what the expected behavior is.

Issue #233, opened 2015-06-09
Wishlist: `stack run`


andrewthad:
It would do the same thing as `cabal run`. For now, I can make due with a bash script I have, but this would be a good thing to add at some point.

chrisdone:
Also: For now you can do `stack build && stack exec foo`.

andrewthad:
I was unaware of that. Good to know.

chrisdone:
(Because stack build installs the packages into the `~/.stack-work/` dir and `stack exec` runs things with that in its PATH.)

snoyberg:
Given that `stack build && stack exec foo` works pretty well for this, do you have any objection with sticking to that as the recommended workflow? I'm concerned about using up the command namespace for something general like `run` with something that's easily achieved in another way.

cocreature:
imho the current state is fine as long as there is some documentation. I didn't understand that this works before finding this issue.

cocreature:
I added it to the transition page on the wiki.

snoyberg:
Cool, good call. I was about to ask you where you thought a good place would be to cover that.

cocreature:
I am actually not quite sure what belongs to faq and what belongs in the transition page. I imagine the questions people ask most frequently are probably related to a transition so it's hard to make that distinction.

chrisdone:
So we close this one?

snoyberg:
@cocreature I agree, it's unclear what should go where. We may need to change the break-up of the pages, it was just the first thing that came to mind when I started on it. Suggestions welcome!

@chrisdone Yes, let's close it.

johnynek:
Why if `stack exec` without args ran the first (is more than one allowed?) executable in the .cabal file? Right now, without args you get a help message (which could be there with `stack exec --help`?).

Wizek:
+1 for `stack run`. I find myself reaching for it quite a few times. I have a workaround script installed in my main dev env named `stack-run`, but whenever I install `stack` on another machine I get surprised that it's unavailable. Also would be nice for it to work like `cabal run`.

@snoyberg You mention a concern about command namespace filling up. Are you still concerned with that? If so, can you imagine any competing feature(s) that might have a claim for `stack run` that would not do the same/something similar to `cabal run`?

snoyberg:
Yes, I'm still concerned with it in general, but I also won't block the
feature if it's generally desired.

Wizek:
[...]

PR #3952, opened 2018-04-01
Introduce `stack run` command line option

rszibele:
This pull request introduces `stack run` which allows directly running the first available executable in the project similar to `cabal run`. This implements the whishlist issue: https://github.com/commercialhaskell/stack/issues/233

I have manually verified that it does indeed run the first available executable and have run the following manual tests:
* `stack exec stack run`
* `stack exec -- stack run help`
* `stack exec -- stack run -- help`

Any thoughts on the code? improvements? I'm pretty new to Haskell and stack.

### DIFF
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -28,6 +28,7 @@ import qualified Data.ByteString.Lazy as L
 import           Data.IORef.RunOnce (runOnce)
 import           Data.List
 import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
 import qualified Data.Text as T
 import           Data.Version (showVersion)
 import           RIO.Process
@@ -98,6 +99,7 @@ import           Stack.Solver (solveExtraDeps)
 import           Stack.Types.Version
 import           Stack.Types.Config
 import           Stack.Types.Compiler
+import           Stack.Types.NamedComponent
 import           Stack.Types.Nix
 import           Stack.Upgrade
 import qualified Stack.Upload as Upload
@@ -353,6 +355,10 @@ commandLineHandler currentDir progName isInterpreter = complicatedOptions
                   "Execute a command"
                   execCmd
                   (execOptsParser Nothing)
+      addCommand' "run"
+                  "Run the first available executable (alias for 'exec')"
+                  execCmd
+                  (execOptsParser $ Just ExecRun)
       addGhciCommand' "ghci"
                       "Run ghci in the context of package(s) (experimental)"
                       ghciCmd
@@ -770,6 +776,7 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
         ExecOptsPlain -> do
           (cmd, args) <- case (eoCmd, eoArgs) of
                 (ExecCmd cmd, args) -> return (cmd, args)
+                (ExecRun, args) -> return ("", args)
                 (ExecGhc, args) -> return ("ghc", args)
                 (ExecRunGhc, args) -> return ("runghc", args)
           loadConfigWithOpts go $ \lc ->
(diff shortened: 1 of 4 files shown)


AshleyYakeley:
Does this build first?

rszibele:
@AshleyYakeley No, just like with `stack exec` the `stack run` command requires the project to already be built.

AshleyYakeley:
OK, so this does _not_ implement #233, which is a `stack run` command that builds and then runs.

AshleyYakeley:
OK, so will it build the entire project, or just enough to build the executable in question?

rszibele:
I'm tending towards only building the executable that should be run. Given that the dependencies are properly defined, it shouldn't be required to build everything, at least that's my line of thinking. Thoughts?

AshleyYakeley:
I think that's sensible.

rszibele:
`stack run` now also builds the executable prior to running it, similar to `stack build :myexecutable`. It now also supports specifying an executable as the first command and if it is omitted, it will default to the first available executable in the project.

I have edited my pull request to reflect these changes.

src/main/Main.hs
```diff
@@ -845,6 +853,21 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
       getPkgOpts wc pkgs =
           map ("-package-id=" ++) <$> mapM (getPkgId wc) pkgs
 
+      getRunCmd args = do
+          pkgComponents <- liftM (map lpvComponents . Map.elems . lpProject) getLocalPackages
+          let executables = filter isCExe $ concatMap Set.toList pkgComponents
+          let argExe = if not (null args) then find (\x -> x == (CExe $ T.pack $ head args)) executables
```
snoyberg: I realize that the `not (null args)` ensures that `head args` is total. However, I still prefer to avoid partial function calls, since they can easily turn into bugs with a refactor. Would it be possible to replace this with a pattern match, e.g.:

```haskell
case args of
  [] -> Nothing
  firstArg:_ -> ...
```

src/main/Main.hs
```diff
@@ -845,6 +853,21 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+                                          else Nothing
+          let firstExe = if not (null executables) then Just $ head executables else Nothing
```
snoyberg: Same applies here, though I think you can also use `listToMaybe`.

src/main/Main.hs
```diff
@@ -845,6 +853,21 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+          let (exe, args') = if isJust argExe then (argExe, tail args) else (firstExe, args)
```
snoyberg: And this could be combined with the pattern match in the above `let` binding to avoid the partial `tail` call.

src/main/Main.hs
```diff
@@ -845,6 +853,21 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+          case exe of
+              Just (CExe exe') -> do
+                Stack.Build.build (const (return ())) Nothing defaultBuildOptsCLI{boptsCLITargets = [T.concat [T.pack ":", exe']]}
```
snoyberg: Slightly more idiomatic may be: `T.cons ':' exe'`.

src/main/Main.hs
```diff
@@ -845,6 +853,21 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
...
+                return (T.unpack exe', args')
+              _                -> liftIO $ do
+                  hPutStrLn stderr "No executables found."
```
snoyberg: This should probably use `logError` instead, which would need to be outside of `liftIO`.

src/main/Main.hs
```diff
@@ -786,6 +792,7 @@ execCmd ExecOpts {..} go@GlobalOpts{..} =
         ExecOptsPlain -> do
           (cmd, args) <- case (eoCmd, eoArgs) of
                 (ExecCmd cmd, args) -> return (cmd, args)
+                (ExecRun, args) -> return ("", args)
```
snoyberg: Can you explain why we need an empty command case here? Is this a code path which shouldn't be triggered? If so, I would rather generate an error message here, since the `exec` call below will fail anyway.

rszibele:
Hello @snoyberg, thank you for your code review, it has been very insightful. I have updated my pull request according to your suggestions. I have also added docker support, as it was the code path (`ExecOptsPlain`) which I mistakenly left out during the implementation.

I have re-tested everything and also tested it with docker enabled.

Wizek:
🎉 Hooray! Thanks guys!

PR #3978, opened 2018-04-20
Improve docker related documentation

psibi:
Brief history: Two days ago I deployed a Yesod application via Docker
and found the current documentation quite confusing. I had to use
`stack -v image container` to figure out what is being done and
understand things. One of the major thing I didn't understand was how
the presence of

```
docker:
  enable: true
```

will affect things as the `stack image container` was working with and
without it. I'm changing the documentation based on my experience. But
yeah, I'm open to changes and feedback if you think this can be
improved further!

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1678,7 +1678,7 @@ docker:
 ```
 
 You can find more details about the first integration
-method [here](./docker_integration.md).
+method [in the Docker integration documentation](./docker_integration.md).
 
 ### Nix
 
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1677,8 +1677,10 @@ docker:
     enable: true
 ```
 
-You can find more details about the first integration
-method [in the Docker integration documentation](./docker_integration.md).
+Alternatively, instead of the above configuration, you can use the
+`--docker` option to achieve the same.  You can find more details
+about the first integration method [in the Docker integration
+documentation](./docker_integration.md).
 
 ### Nix
 
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1604,21 +1604,19 @@ page](https://docs.haskellstack.org/en/stable/shell_autocompletion)
 
 ### Docker
 
-stack provides two built-in Docker integrations. Firstly, you can build your
-code inside a Docker image, which means:
+stack provides two built-in Docker integrations. The first way is to
+build your code inside a Docker image, which means:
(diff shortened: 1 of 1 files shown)


doc/GUIDE.md
```diff
@@ -1669,6 +1665,21 @@ entrypoints:
...
+configuration in `stack.yaml` informs stack to switch to the first
+integration method:
+
+```yaml
+docker:
+    enable: true
+```
+
+You can find more details about the first integration
+method [here](./docker_integration.md).
```
snoyberg: How about s/here/in the Docker integration documentation ?

doc/GUIDE.md
```diff
@@ -1654,9 +1652,7 @@ the images.
 
 Note that the executable will be built in the development environment
 and copied to the container, so the dev OS must match that of the
-container OS. This is easily accomplished using [Docker integration](docker_integration.md),
-under which the exe emitted by `stack build` will be built on the
-Docker container, not the local OS.
+container OS.
```
snoyberg: How about adding that building under Docker with `--docker` or equivalent can also solve this problem?

psibi: @snoyberg Thanks for the feedback. Sorry that it took so long for me to get back on this.  

The reason I removed this information is because - this is the sentence which caused the confusion for me. While building the code inside the container using `--docker` is valuable, introducing it now itself will confuse the user because the reader as of now is reading about the second integration method ( building code in local machine and copying the executable to the container).  This deleted sentence in the PR confused me enough that I had to trace using `-v` calls in `stack image container` to understand how things work. If you see the paragraph which I added  down in this PR, I explicitly say how to achieve the first integration method and also link a more detailed doc which explains the concept you want me to add here.

If you still want to add, I think adding it below where I explain the difference would be the right place to add. Let me know what you think.

psibi: On thinking more, I have added about the `--docker` option right below when I start explaining the first integration method. I think that should clarify things.

snoyberg: I still think it's insufficiently clear. This is giving the user a warning "you need to match our dev OS to the Docker OS or things will break." It's not immediately obvious that there's a trivial solution to this problem available (building inside Docker too). I think that should be clarified, otherwise we're just dropping a problem in the user's lap without any assistance.

psibi: @snoyberg Fair enough, I have updated the PR which hopefully makes it more clear.

Issue #3940, opened 2018-03-23
Warn when upgrading base


snoyberg:
I've seen users get confusing error messages multiple times when trying to upgrade base, often from trying to install a package like ghc-mod. We should give a warning in such situations, and recommend changing GHC versions instead.

SkyWriter:
@snoyberg , it may sound silly, but I don't suppose you could drop a few lines on how to produce an erroneous case? I'm willing to attack this, but seem to be stuck understanding the actual issue. Thanks!

snoyberg:
Great question. I'm not sure how the original people suffering from this triggered the bug, but here's an example of steps that could be taken:

1.  Create a `stack.yaml` specifying `lts-11.4`, e.g.

    ```yaml
    resolver: lts-11.4
    packages: []
    ```

2.  Try to build `ghc-mod`, e.g. `stack build ghc-mod`
3.  You'll get an error message recommending, among other things, that you upgrade your `base` version, e.g.:

    ```yaml
    - Cabal-1.24.2.0
    - base-4.9.1.0
    - cabal-helper-0.7.3.0
    - extra-1.5.3
    - haskell-src-exts-1.19.1
    - hlint-2.0.15
    - monad-journal-0.7.2
    - optparse-applicative-0.13.2.0
    ```

4. If you try to build anything, you'll keep getting error messages about necessary extra deps.

Ideally, two things should instead happen:

1. Stack should never recommend adding `base` to `extra-deps`
2. If someone put `base` in `extra-deps`, it should give a very loud warning about the fact that you almost certainly didn't mean that.

SkyWriter:
I have submitted my take on implementing the changes. Could someone possibly take a look at the code, and tell me if it's the right direction at all?

snoyberg:
The PR looks great! I've added a few comments over there, thanks for the contribution.

snoyberg:
Fixed by #3993, thanks!

PR #3993, opened 2018-04-29
Help users not to go down the road of base upgrade/downgrade, addresses #3940

SkyWriter:
Implements the following:

1. Emit a warning when someones tries to upgrade/downgrade base.
2. Show a specific message when build plan results in base upgrade/downgrade requirement.

Here's what I did or checked out:

* [x] Wrote the code
* [x] Added relevant ChangeLog.md entries
* [x] No documentation updates seemed necessary to me
* [x] Linted my changes
* [x] Did some manual tests

#### Test #1

- Created `stack.yaml` with the following contents:

      resolver: lts-11.6
      extra-deps: []

- Issued `stack build ghc-mod`
- Made sure I'm getting the right error message suggesting me to pick another GHC with the matching `base`.

#### Test #2

- Created `stack.yaml` with the following contents:

      resolver: lts-11.6
      extra-deps: ["base-4.9.1.0"]

- Issued `stack build ghc-mod`
- Made sure I'm getting the right warning suggesting me not to put `base` in `extra-deps`, and use another GHC version instead.

### DIFF
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -185,6 +185,10 @@ constructPlan :: forall env. HasEnvConfig env
 constructPlan ls0 baseConfigOpts0 locals extraToBuild0 localDumpPkgs loadPackage0 sourceMap installedMap initialBuildSteps = do
     logDebug "Constructing the build plan"
 
+    bconfig <- view buildConfigL
+    when (hasBaseInDeps bconfig) $
+      prettyWarn $ flow "You are trying to upgrade/downgrade base, which is almost certainly not what you really want. Please, consider using another GHC version if you need a certain version of base, or removing base from extra-deps." <> line
+
     econfig <- view envConfigL
     let onWanted = void . addDep False . packageName . lpPackage
     let inner = do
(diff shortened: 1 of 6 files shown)


snoyberg:
An initial request, if possible: would you be able to add the tests you mention above as integration tests, to ensure that this bug doesn't recur in the future? We have instructions on the integration test suite at https://github.com/commercialhaskell/stack/blob/master/test/integration/README.md

src/Stack/Build/ConstructPlan.hs
```diff
@@ -185,6 +185,10 @@ constructPlan :: forall env. HasEnvConfig env
 constructPlan ls0 baseConfigOpts0 locals extraToBuild0 localDumpPkgs loadPackage0 sourceMap installedMap initialBuildSteps = do
     logDebug "Constructing the build plan"
 
+    bconfig <- view buildConfigL
+    when (hasBaseInDeps bconfig) $
+      prettyWarn $ flow "You are trying to upgrade/downgrade base, which is almost certainly not what you really want. Please, consider using another GHC version if you need a certain version of base, or removing base from extra-deps." <> line
```
snoyberg: Great warning, but I think some new users may still be confused. How about including a link to the original issue so they can learn more?

SkyWriter:
@snoyberg Thanks for the review! Everything makes perfect sense. I'll do the fixes.

It's going to be my first bite of integration testing in Haskell. I hope the resulting code won't make your eyes bleed too much... 😄

SkyWriter:
@snoyberg I've submitted a few changes:
- Referenced the original issue in the error message
- Added integration test cases to replicate my manual test steps
- Resolved conflicts with the master branch

Could you please give it a look when you get a minute?

snoyberg:
Thanks!

PR #3955, opened 2018-04-06
Don't recompile in presence of sublibraries

mihaimaruseac:
Note: Documentation fixes for https://docs.haskellstack.org/en/stable/ should target the "stable" branch, not master.

Please include the following checklist in your PR:

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [x] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

------

Tested first with the `internal-libraries` package mentioned in https://github.com/commercialhaskell/stack/issues/3899#issuecomment-375935325 and made Stack not ignore the internal library in this case.

Furthermore, tested against my own `hakyll`-based repo, to reproduce #3899.

PR #4003, opened 2018-05-02
4001 excess recompilation

### LINKED ISSUE #4001: --file-watch is watching test suites again
This is a regression, the original issue was #3483.  At a guess it is due to commit b008a33b4f9499bc63ccb81f1df9441d3933494c, which looks like it now uses all package compononents in `getPackageFilesForTargets` instead of lib + other targets components.  Probably an easy fix to make it lib + sublibs + other targets (I don't know anything about sublibs, but presumably they can't be targets and should always be taken into account?).  A test would have picked this up, that's on me as I fixed the original issue without adding a test.

```
$ stack --version
Version 1.7.1, Git revision 681c800873816c022739ca7ed14755e85a579565 (5807 commits) x86_64 hpack-0.28.2
```

### Method of installation

tswelsh:
Fixes #4001.  I have assumed we always want to check internal library files for changes, but @mihaimaruseac will probably know more about this than me?  Sorry if I'm stepping on your toes by the way!

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -14,6 +14,8 @@ Other enhancements:
 Bug fixes:
 
 * `~/.stack/config.yaml` and `stack.yaml` terminating by newline
+* A regression in recompilation checking introduced in v1.7.1 has been fixed.
+  See [#4001](https://github.com/commercialhaskell/stack/issues/4001)
 
 
 ## v1.7.1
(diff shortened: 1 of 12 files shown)


mihaimaruseac:
Nice. I forgot to add the `isCInternalLib` in #3955 but I'm doing it in #3982 (https://github.com/commercialhaskell/stack/pull/3982/files#diff-726a384682e68eaa6ea53d0f5825efc1).

I think the approach you've taken is good, but will look more carefully later today, if I get some time.

mihaimaruseac:
I tested it against all the internal-libraries related bugs and it doesn't break any of the already solved ones. It solves #4001, so it's good :+1:

tswelsh:
Great!  The travis failure is unrelated, it's just a lint failure in a module I haven't touched.

PR #4015, opened 2018-05-07
Windows: use 'hIsTerminalDevice or isMinTTYHandle'

mpilgrem:
`hIsTerminalDevice` does not recognise handles to 'mintty' terminals, but `isMinTTYHandle` does. Currently, `stack` relies on `hIsTerminalDevice` in three places to detect handles to terminals and does not use `isMinTTYHandle`. This proposal introduces 'isMinTTYHandle' in those three places.

`isMinTTYHandle` was first added to package `Win32-2.5.0.0` and then made robust to an upstream (ntdll-related) bug from `Win32-2.5.3.0`. The latter constraint is added to `package.yaml`.

* [x] The documentation has been updated, if necessary. (None necessary.)

This has been tested only by building `stack` on Windows 10 and using it in native and 'mintty' terminals - I do not know what significance the terminal device-checking has.

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -243,8 +243,8 @@ runContainerAndExit getCmdArgs
      (env,isStdinTerminal,isStderrTerminal,homeDir) <- liftIO $
        (,,,)
        <$> getEnvironment
-       <*> hIsTerminalDevice stdin
-       <*> hIsTerminalDevice stderr
+       <*> hIsTerminalDeviceOrMinTTY stdin
+       <*> hIsTerminalDeviceOrMinTTY stderr
        <*> (parseAbsDir =<< getHomeDirectory)
      isStdoutTerminal <- view terminalL
      let dockerHost = lookup "DOCKER_HOST" env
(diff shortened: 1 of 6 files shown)


src/Stack/Docker.hs
```diff
@@ -243,8 +245,17 @@ runContainerAndExit getCmdArgs
      (env,isStdinTerminal,isStderrTerminal,homeDir) <- liftIO $
        (,,,)
        <$> getEnvironment
+-- hIsTerminaDevice does not recognise input or output handles to mintty
+-- terminals as terminal devices, but isMinTTYHandle does.
+#ifdef WINDOWS
+       <*> ((||) <$> hIsTerminalDevice stdin
```
snoyberg: Can I recommend that instead of including conditionals in multiple places, you define a helper function once which uses CPP, and use that in all of these places?

Additionally, it can be _marginally_ more efficient to write the code as something like:

```haskell
x <- hIsTerminalDevice stdin
if x then return True else withHandleToHANDLE stdin isMinTTYHandle
```

This will only run the second action if the first returns `False`.

mpilgrem: I've used a helper function `hIsTerminalDeviceOrMinTTY`, which I added to module `Stack.Prelude` - I could not see a better place.

PR #4017, opened 2018-05-08
Fix travis

ncaq:
fix Travis CI.

I believe that this change will not affect users and will only affect CI results.
However, I'm not sure if it really changes `optP-Wno-nonportable-include-path`.


mgsloan:
LGTM, thanks!

PR #4067, opened 2018-06-06
Change parsers to use `flag'` instead of `switch`

### LINKED ISSUE #3959: It should not matter if flags given before command or after
This happened in a wrapper script, when I'm setting default flags to pass to stack; e.g.

```
#!/bin/bash
STACK_ARGS=(
    --test --no-run-tests
    --bench --no-run-benchmarks
)
exec stack "${STACK_ARGS[@]}" "$@"
```

### Steps to reproduce

1. `stack --test --no-run-tests build`
2. `stack build --test --no-run-tests`

### Expected

Either both should build tests and not run them or neither should work (since `stack` doesn't take those flags on its own)

### Actual

The first built and ran tests. The second built but didn't run tests.

In other words, `stack --test build` and `stack --test --no-run-tests build` do the same thing.

### Stack version

```
Version 1.6.5, Git revision 24ab0d6ff07f28276e082c3ce74dfdeb1a2ca9e9 (5514 commits) x86_64 hpack-0.20.0
```

### Method of installation

From `https://github.com/commercialhaskell/stack/releases/download`

markus1189:
This fixes #3959.

The problem is that the [switch](https://hackage.haskell.org/package/optparse-applicative/docs/Options-Applicative-Builder.html#v:switch) function returns a `Just False` which gets wrapped into `First`.

This means that the value is *always* set to either `True` or `False`
and the `First` monoid means that during options parsing the order of
the flags suddenly matters as described in #3959, because `First (Just
False)` is chosen when `mappend`ing with `First (Just True)`.

The fix is to use the [flag](https://hackage.haskell.org/package/optparse-applicative-0.14.2.0/docs/Options-Applicative-Builder.html#v:flag-39-) function that does not set a default value, so we get a `First Nothing` instead.

  there should be no user relevant changes
  no relevant changes

There is a new integration test in the `3959-order-of-flags` folder that makes sure that the order does not matter.

### DIFF
--- a/src/Stack/Options/BenchParser.hs
+++ b/src/Stack/Options/BenchParser.hs
@@ -19,7 +19,7 @@ benchOptsParser hide0 = BenchmarkOptsMonoid
                                  help ("Forward BENCH_ARGS to the benchmark suite. " <>
                                        "Supports templates from `cabal bench`") <>
                                  hide))
-        <*> optionalFirst (switch (long "no-run-benchmarks" <>
+        <*> optionalFirst (flag' True (long "no-run-benchmarks" <>
                           help "Disable running of benchmarks. (Benchmarks will still be built.)" <>
                              hide))
    where hide = hideMods hide0
--- a/test/integration/tests/3959-order-of-flags/files/stack.yaml
+++ b/test/integration/tests/3959-order-of-flags/files/stack.yaml
@@ -1,0 +1,1 @@
+resolver: lts-11.6
--- a/test/integration/tests/3959-order-of-flags/files/test/Spec.hs
+++ b/test/integration/tests/3959-order-of-flags/files/test/Spec.hs
@@ -1,0 +1,2 @@
+main :: IO ()
+main = fail "this always fails for the test"
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -38,6 +38,8 @@ Bug fixes:
   was tried to be registered. This is now fixed by always building internal
   libraries. See
   [#3996](https://github.com/commercialhaskell/stack/issues/3996).
+* Order of commandline arguments does not matter anymore.
+  See [#3959](https://github.com/commercialhaskell/stack/issues/3959)
 
 
 ## v1.7.1
--- a/src/Stack/Options/TestParser.hs
+++ b/src/Stack/Options/TestParser.hs
@@ -27,12 +27,12 @@ testOptsParser hide0 =
                          help "Arguments passed in to the test suite program" <>
                          hide)))
         <*> optionalFirst
-                (switch
+                (flag' True
                     (long "coverage" <>
                      help "Generate a code coverage report" <>
                      hide))
         <*> optionalFirst
-                (switch
+                (flag' True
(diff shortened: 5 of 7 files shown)


test/integration/tests/3959-order-of-flags/files/test/Spec.hs
```diff
@@ -0,0 +1,2 @@
+main :: IO ()
+main = putStrLn "Test suite not yet implemented"
```
snoyberg: Can you make this a failing line so we're sure that the test suite isn't run?

test/integration/tests/3959-order-of-flags/files/stack.yaml
```diff
@@ -0,0 +1,65 @@
+# This file was automatically generated by 'stack init'
```
snoyberg: Can you replace this with `resolver: lts-11.6`

test/integration/tests/3959-order-of-flags/files/src/Lib.hs
```diff
@@ -0,0 +1,6 @@
+module Lib
```
snoyberg: It should work to drop this module.

test/integration/tests/3959-order-of-flags/files/bug3959.cabal
```diff
@@ -0,0 +1,61 @@
+-- This file has been generated from package.yaml by hpack version 0.28.2.
```
snoyberg: I'd recommend replacing this .cabal file with a package.yaml file instead.

test/integration/tests/3959-order-of-flags/files/app/Main.hs
```diff
@@ -0,0 +1,6 @@
+module Main where
```
snoyberg: This can be dropped too.

test/integration/tests/3959-order-of-flags/files/Setup.hs
```diff
@@ -0,0 +1,2 @@
+import Distribution.Simple
```
snoyberg: This can be dropped too.

src/Stack/Options/BenchParser.hs
```diff
@@ -19,7 +19,7 @@ benchOptsParser hide0 = BenchmarkOptsMonoid
                                  help ("Forward BENCH_ARGS to the benchmark suite. " <>
                                        "Supports templates from `cabal bench`") <>
                                  hide))
-        <*> optionalFirst (switch (long "no-run-benchmarks" <>
+        <*> optionalFirst (flag' True (long "no-run-benchmarks" <>
```
snoyberg: Can you add a comment to the changelog about the bugfix?

mihaimaruseac:
Looks good on a quick glance. Minor nit, though: can you remove the autogenerated comments in the files for the integration tests?

src/Stack/Options/BenchParser.hs
markus1189: done

snoyberg:
Thanks!

PR #4072, opened 2018-06-10
Create utilities for prompting, and use for ghci target prompt

mgsloan:

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -51,7 +51,7 @@ import           Stack.Types.Package
 import           Stack.Types.PackageIdentifier
 import           Stack.Types.PackageName
 import           Stack.Types.Runner
-import           System.IO (putStrLn, putStr, getLine)
+import           System.IO (putStrLn)
 import           System.IO.Temp (getCanonicalTemporaryDirectory)
 
 #ifndef WINDOWS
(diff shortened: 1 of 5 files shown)


src/Stack/Ghci.hs
```diff
@@ -532,11 +531,10 @@ figureOutMainFile bopts mainIsTargets targets0 packages = do
             T.pack (toFilePath mainIs)
     candidateIndices = take (length candidates) [1 :: Int ..]
     userOption = do
-      putStr "Specify main module to use (press enter to load none): "
-      option <- getLine
+      option <- prompt "Specify main module to use (press enter to load none): "
       let selected = fromMaybe
                       ((+1) $ length candidateIndices)
-                      (readMaybe option :: Maybe Int)
+                      (readMaybe (T.unpack option) :: Maybe Int)
```
snoyberg: Not a big deal, but we could use the `Data.Text.Read.decimal` function for this.

src/Stack/Prelude.hs
```diff
@@ -15,6 +15,9 @@ module Stack.Prelude
   , withProcessContext
   , stripCR
   , hIsTerminalDeviceOrMinTTY
+  , prompt
```
snoyberg: Nice additions.

PR #4075, opened 2018-06-10
get-stack: allow users to use sudo carefully

bukzor:
When requesting root permission, always show the command and never re-use cached credentials.

* [ :heavy_check_mark: ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [ :heavy_check_mark: ] The documentation has been updated, if necessary.

Note that the second sudo invocation doesn't prompt for password: (This also has #4074 applied.)
```
$ sudo apt remove libgmp-dev                          
[sudo] password for buck: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  libgmp-dev
0 upgraded, 0 newly installed, 1 to remove and 0 not upgraded.
After this operation, 1,612 kB disk space will be freed.
Do you want to continue? [Y/n] 
(Reading database ... 295761 files and directories currently installed.)
Removing libgmp-dev:amd64 (2:6.1.2+dfsg-1) ...

$ sudo whoami
root

$ cat get.haskellstack.org | sh
Detected Linux distribution: ubuntu

Installing dependencies...

dpkg-query: no packages found matching libgmp-dev
sudo apt-get install -y g++ gcc libc6-dev libffi-dev libgmp-dev make xz-utils zlib1g-dev git gnupg
[sudo] password for buck: 
```

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -127,8 +127,15 @@ print_bindist_notice() {
 
 # Adds a 'sudo' prefix if sudo is available to execute the given command
 # If not, the given command is run as is
+# When requesting root permission, always show the command and never re-use cached credentials.
 sudocmd() {
-  $(command -v sudo) "$@"
+  if command -v sudo >/dev/null; then
+    echo "sudo $@"
+    # -k: Always prompt.
+    sudo -k "$@"
+  else
+    "$@"
+  fi
 }
 
 # Install dependencies for distros that use Apt
@@ -159,7 +166,8 @@ do_ubuntu_install() {
     install_aarch64_binary
   elif is_64_bit ; then
     install_dependencies
-    print_bindist_notice
+
+print_bindist_notice
     install_64bit_standard_binary
   else
     install_dependencies
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -166,8 +166,7 @@ do_ubuntu_install() {
     install_aarch64_binary
   elif is_64_bit ; then
     install_dependencies
-
-print_bindist_notice
+    print_bindist_notice
     install_64bit_standard_binary
   else
     install_dependencies
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -131,7 +131,7 @@ print_bindist_notice() {
 sudocmd() {
   if command -v sudo >/dev/null; then
     echo "sudo $@"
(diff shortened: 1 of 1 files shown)


ncaq:
I think this is the standard specification of sudo.
If user do not like sudo to use caching, user should change sudo's settings.
I think the script should respect the user's settings.

borsboom:
Overall I like this idea.  The problem is that `get-stack.sh` is often used in scripts on systems where sudo's `NOPASSWD` setting is used to disable prompting for passwords, and I fear this change will cause that to break.

bukzor:
@borsboom The comment is an oversimplificiation, which is apparently misleading. In actuality `-k` just disables the caching system, which already isn't used in the NOPASSWD use case.

```
     -k, --reset-timestamp
                 When used in conjunction with a command or an option that may
                 require a password, this option will cause sudo to ignore the
                 user's cached credentials.  As a result, sudo will prompt for a
                 password (if one is required by the security policy) and will not
                 update the user's cached credentials.

                 Not all security policies support credential caching.
```

bukzor:
Here's a little demo to show that -k has no effect under NOPASSWD. I temporarily added `NOPASSWD` to my user beforehand. I use `-K` to kill the cache, show that commands work without a password prompt, and still work without a password prompt when using `-k`.

```
$ sudo -l
Matching Defaults entries for buck on tortle-2016:
    env_reset, mail_badpass,
    secure_path=/usr/local/sbin\:/usr/local/bin\:/usr/sbin\:/usr/bin\:/sbin\:/bin\:/snap/bin

User buck may run the following commands on tortle-2016:
    (ALL : ALL) NOPASSWD: ALL

$ sudo -K

$ sudo id
uid=0(root) gid=0(root) groups=0(root)

$ sudo -k id
uid=0(root) gid=0(root) groups=0(root)
```

bukzor:
Addressing @ncaq's complaint:

The point of this pull request is that a trustworthy script won't run root-permission commands without my explicit consent. Making use of the sudo cache system is an (unintentional, I'm sure) end-run around asking for my permission to make use of escalated privileges. With the addition of `-k`, I will get the same number of password prompts regardless of how long ago I used sudo before `curl | sh`. Similarly, if the script runs two sudo commands, it should ask for my consent for both, not just the first. I believe this is why @borsboom likes this idea.

bukzor:
Travis failures seem unrelated.

borsboom:
Thanks, your comment addresses my concern.

PR #4089, opened 2018-06-15
Have project ghc-options append to config.yaml ghc-options #3753

mgsloan:
See #3753 and the PR that this supersedes, https://github.com/commercialhaskell/stack/pull/3781

### DIFF
--- a/src/Data/Monoid/Map.hs
+++ b/src/Data/Monoid/Map.hs
@@ -1,0 +1,18 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+module Data.Monoid.Map where
+
+import qualified Data.Map as M
+import           Stack.Prelude
+
+-- | Utility newtype wrapper to make make Map's Monoid also use the
+-- element's Monoid.
+newtype MonoidMap k a = MonoidMap (Map k a)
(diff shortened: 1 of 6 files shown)


src/Data/Monoid/Map.hs
```diff
@@ -0,0 +1,18 @@
...
+
+-- | Utility newtype wrapper to make make Map's Monoid also use the
+-- element's Monoid.
+newtype MonoidMap k a = MonoidMap (Map k a)
+    deriving (Eq, Ord, Read, Show, Generic, Functor)
+
+instance (Ord k, Semigroup a) => Semigroup (MonoidMap k a) where
+    MonoidMap mp1 <> MonoidMap mp2 = MonoidMap (M.unionWith (<>) mp1 mp2)
+
+instance (Ord k, Monoid a, Semigroup a) => Monoid (MonoidMap k a) where
```
snoyberg: Is the Monoid constraint actually needed here?

mgsloan: Good point!  It's not, it was present in the code that I copied from `Stack.Build.ConstructPlan`. I actually had written my own `MonoidMap`, and was later notified due to name ambiguity that one already existed.  Yay for consistent naming decisions

I've removed the redundant constraint

PR #4092, opened 2018-06-17
Fix style warnings from hlint-2.1.6

snoyberg:


snoyberg:
Once CI passes, we should merge this to stable, and then merge stable into master, so that future PRs don't fail. See #4093.

PR #4097, opened 2018-06-19
Add the interleaved-output option/config value #3225

### LINKED ISSUE #3225: Add flag to print GHC output to stdout
Having streaming compilation output is super useful to ensure that things compile at the right speed, especially when piping into something like `ts`.

#426 added a flag that can stream to a file, but that still doesn't provide streaming terminal output.

Of course interleaved output makes things messy, so it would make sense for this flag to work only for `-j1`.

Note that (quote @mgsloan)

> If it's a local package and the last thing to build with -j1, it should output the log

but it would be great to have it also for snapshot packages.

### LINKED ISSUE #3508: Report progress on STDOUT/STDERR during compilation of individual modules
Following-up from https://github.com/commercialhaskell/stack/issues/426#issuecomment-338386315

Would it be possible to output a `.` for every compiled module. Use-cases:

* Avoid hitting timeouts [1] on some CI environments.
* Reporting progress is important for impatient devs :)

[1] https://discuss.circleci.com/t/too-long-with-no-output-while-building-docker-image/12275

snoyberg:
* Fixes #3225
* Fixes #3508

Manual experimentation only, sorry.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -25,6 +25,10 @@ Other enhancements:
   [#4068](https://github.com/commercialhaskell/stack/pull/4068).
 * Added new `--tar-dir` option to `stack sdist`, that allows to copy
   the resulting tarball to the specified directory.
+* Introduced the `--interleaved-output` command line option and
+  `build.interleaved-output` config value which causes multiple concurrent
+  builds to dump to stdout at the same time with a `packagename> ` prefix. See
+  [#3225](https://github.com/commercialhaskell/stack/issues/3225).
 
 Bug fixes:
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -27,7 +27,7 @@ Other enhancements:
   the resulting tarball to the specified directory.
 * Introduced the `--interleaved-output` command line option and
   `build.interleaved-output` config value which causes multiple concurrent
-  builds to dump to stdout at the same time with a `packagename> ` prefix. See
+  builds to dump to stderr at the same time with a `packagename> ` prefix. See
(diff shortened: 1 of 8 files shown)


snoyberg:
Note that there are some subtle UX choices to be made in `withOutputType`, such as:

* Do we dump output from non-local packages? Currently, YES
* If we're just building a single package, do we still prefix output? Currently, YES

These are easy changes to make in the code, it's really a design issue. And I'd rather get the design right the first time around, so bikeshed away.

mgsloan:
LGTM! :+1: :+1:

ChangeLog.md
```diff
@@ -25,6 +25,10 @@ Other enhancements:
   [#4068](https://github.com/commercialhaskell/stack/pull/4068).
 * Added new `--tar-dir` option to `stack sdist`, that allows to copy
   the resulting tarball to the specified directory.
+* Introduced the `--interleaved-output` command line option and
+  `build.interleaved-output` config value which causes multiple concurrent
+  builds to dump to stdout at the same time with a `packagename> ` prefix. See
```
nh2: I have observed it to dump to `stderr` instead of `stdout`.

nh2:
> Do we dump output from non-local packages? Currently, YES
> If we're just building a single package, do we still prefix output? Currently, YES

These look like good choices to me.

snoyberg:
Good catches @nh2, I think 4e6b24799d4d223bce3ce83cec6462f23e8b16e6 should address both points. Holler if not, otherwise we can merge this once CI passes.

buggymcbugfix:
Why does adding `interleaved-output: true` to my `stack.yaml` not work?

PR #4103, opened 2018-06-21
4039 namespaced templates

lylek:
* `stack new` now allows template names of the form username/foo to download from a user other than commercialstack on Github, and can be prefixed with the service`github:`, `gitlab:`, or `bitbucket:`.
* Changelog.md has been updated
* Help text for stack new has been updated
* Tested this by temporarily adding a trace in the downloadTemplate function, to see the request it was using to download the template, and trying various new and old forms of the template name. Also created a local template and made sure that still took priority over a remote download.
* I would have liked to add an integration test, but it's hard to test things that download from remote URLs. There isn't currently a way to stub that out.
* Added documentation for the web site in the `doc` subdirectory. Unfortunately I'm not sure how to build that, to make sure it looks right.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1580,7 +1580,7 @@ Selected resolver: lts-3.2
 Wrote project config to: /home/michael/my-yesod-project/stack.yaml
 ```
 
-The default `stack-templates` repository is on [Github](https://github.com),
+The default `stack-templates` repository is on [Github](https://github.com/commercialhaskell/stack-templates),
 under the user account `commercialstack`. You can download templates from a
 different Github user by prefixing the username and a slash:
 
@@ -1588,12 +1588,18 @@ different Github user by prefixing the username and a slash:
 stack new my-yesod-project yesodweb/yesod-simple
 ```
 
+Then it would be downloaded from Github, user account `yesodweb`,
+repo `stack-templates`, and file `yesod-simple.hsfiles`.
+
 You can even download templates from a service other that Github, such as
 [Gitlab](https://gitlab.com) or [Bitbucket](https://bitbucket.com):
 
 ```
 stack new my-project gitlab:user29/foo
 ```
+
+That template would be downloaded from Gitlab, user account `user29`,
+repo `stack-templates`, and file `foo.hsfiles`.
 
 If you need more flexibility, you can specify the full URL of the template:
 
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -1559,6 +1559,12 @@ yesod-postgres
 yesod-postgres-fay
 yesod-simple
 yesod-sqlite
+```
+
+You can specify one of these templates following your project name
+in the `stack new` command:
+
+```
 michael@d30748af6d3d:~$ stack new my-yesod-project yesod-simple
 Downloading template "yesod-simple" to create project "my-yesod-project" in my-yesod-project/ ...
 Using the following authorship configuration:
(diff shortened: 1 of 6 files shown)


doc/GUIDE.md
```diff
@@ -1574,13 +1580,36 @@ Selected resolver: lts-3.2
 Wrote project config to: /home/michael/my-yesod-project/stack.yaml
 ```
 
-Alternatively you can use your own templates by specifying the path:
+The default `stack-templates` repository is on [Github](https://github.com),
```
snoyberg: How about adding a link directly to the repo?

doc/GUIDE.md
```diff
@@ -1574,13 +1580,36 @@ Selected resolver: lts-3.2
...
+under the user account `commercialstack`. You can download templates from a
+different Github user by prefixing the username and a slash:
+
+```
+stack new my-yesod-project yesodweb/yesod-simple
```
snoyberg: I think it would be worth specifying exactly how this is resolved: user `yesodweb`, repo `stack-templates`, file `yesod-simple.hsfiles`.

src/Stack/Options/NewParser.hs
```diff
@@ -23,9 +23,12 @@ newOptsParser = (,) <$> newOpts <*> initOptsParser
              help "Do not create a subdirectory for the project") <*>
         optional (templateNameArgument
             (metavar "TEMPLATE_NAME" <>
-             help "Name of a template or a local template in a file or a URL.\
-                  \ For example: foo or foo.hsfiles or ~/foo or\
-                  \ https://example.com/foo.hsfiles")) <*>
+             help "Name of a template - can take the form\
+                \ [service:][username/]template with optional service name\
```
snoyberg: Perhaps more accurate would be:

```
[[service:]username/]template
```

doc/GUIDE.md
lylek: Sure, will do. There's also a direct link to the repo later in the section.

doc/GUIDE.md
lylek: Yeah, that sounds good.

src/Stack/Options/NewParser.hs
lylek: No, actually you can write github:foo and it will pull the template `foo` from GitHub with the username `commercialhaskell` in the `stack-templates` repo. So either the service or the username can be dropped.

snoyberg: Thanks for clarifying. In that case, I think it would be better to change the behavior, otherwise we'll be granting special status to the commercialhaskell account on other services when we don't necessarily control it.

lylek: Yeah, I thought that might be odd too. I'll change the code.

src/Stack/Types/TemplateName.hs
```diff
@@ -106,3 +125,24 @@ templateName (TemplateName prefix _) = prefix
...
+    ["gitlab"    , rest] -> parseRepoPathWithService Gitlab rest
+    ["bitbucket" , rest] -> parseRepoPathWithService Bitbucket rest
+    _                    -> Nothing
+
+-- | Parses a template path of the form @user/template@, assuming the default service.
+parseRepoPathWithService :: RepoService -> Text -> Maybe RepoTemplatePath
+parseRepoPathWithService service path =
+  case T.splitOn "/" path of
+    [user, name] -> Just $ RepoTemplatePath service user name
+    [name]       -> Just $ RepoTemplatePath service defaultRepoUser name
```
johnmendonca: Deleting this line and `defaultRepoUser` should change the behavior as described above regarding `commercialhaskell` on other services.

src/Stack/Types/TemplateName.hs
```diff
@@ -98,6 +112,11 @@ mkTemplateName s =
                           AbsPath (Path fp) -> [|AbsPath (Path fp)|]
                           RelPath (Path fp) -> [|RelPath (Path fp)|]
                           UrlPath fp -> [|UrlPath fp|]
+                          RepoPath (RepoTemplatePath sv u t) ->
+                            case sv of
+                                Github    -> [|RepoPath $ RepoTemplatePath Github u t|]
+                                Gitlab    -> [|RepoPath $ RepoTemplatePath Gitlab u t|]
+                                Bitbucket -> [|RepoPath $ RepoTemplatePath Bitbucket u t|]
```
johnmendonca: I was wondering about this part, since I'm not clear about what it does exactly. 

Could it do without the pattern match and just be:

```
RepoPath (RepoTemplatePath sv u t) -> [|RepoPath $ RepoTemplatePath sv u t|]
```

src/Stack/Types/TemplateName.hs
lylek: Not quite. That would prevent `commercialhaskell` being used as the default for other services. But it would also break the simple case where only a template name is supplied. Another function will have to be created to handle that case.

src/Stack/Types/TemplateName.hs
lylek: No, that would cause TemplateHaskell to complain that there is no instance of something-or-other class. (`Rep`?) The alternative is to define an instance of that class, but Michael's recommendation was that it wasn't worth it.

src/Stack/Types/TemplateName.hs
johnmendonca: Oh right, we end up here in the case where just a single name is given, thanks.

lylek:
OK, I've adjusted the pull request as discussed. Well, one slight difference: it still accepts `github:foo` as an alternative to `foo`, defaulting to the `commercialstack` user, but will no longer accept `gitlab:foo` or `bitbucket:foo` with no username.

Issue #4091, opened 2018-06-15
Unwanted Characters in Console Output


Burtannia:
Hi,

I'm getting extra characters appearing in my console output. After doing some research I discovered that there were some issues with colour codes and `ansi-terminal` on Windows. From the bug reports I read that they had been fixed in a recent commit. I therefore updated Stack to the Git version, `stack --version` now gives the following:
`Version 1.8.0, Git revision faa941a2d697b252a2f69e8ddd3f276ff44bf8b6 x86_64 hpack-0.28.2`

However, the characters still appear after the update. Adding `--color never` fixes the issue as mentioned in a [previous bug report](https://github.com/commercialhaskell/stack/issues/3797). The characters are seen both in the terminal within VS Code and the Windows Command Prompt. The characters do not appear in the terminal or within VS Code on Ubuntu.

`←[0mAccounting-0.1.0.0: unregistering (local file changes: app\Main.hs)←[0m
←[0mAccounting-0.1.0.0: build (lib + exe)←[0m
←[0mPreprocessing library for Accounting-0.1.0.0..←[0m
←[0mBuilding library for Accounting-0.1.0.0..←[0m
←[0mPreprocessing executable 'Accounting-exe' for Accounting-0.1.0.0..←[0m
←[0mBuilding executable 'Accounting-exe' for Accounting-0.1.0.0..←[0m
←[0m[2 of 2] Compiling Main             ( app\Main.hs, .stack-work\dist\5c8418a7\build\Accounting-exe\Accounting-exe-tmp\Main.o )←[0m
←[0mLinking .stack-work\dist\5c8418a7\build\Accounting-exe\Accounting-exe.exe ...←[0m
←[0mAccounting-0.1.0.0: copy/register←[0m
←[0mInstalling library in C:\Projects\Accounting\.stack-work\install\e6d3037a\lib\x86_64-windows-ghc-8.2.2\Accounting-0.1.0.0-F8ZttcOqREWElkyQQVED0s←[0m
←[0mInstalling executable Accounting-exe in C:\Projects\Accounting\.stack-work\install\e6d3037a\bin←[0m
←[0mRegistering library for Accounting-0.1.0.0..←[0m`

Stack version: 1.8.0
LTS: 11.13
Global GHC version: 8.0.2
OS: Windows 7 Professional 64-bit

snoyberg:
FWIW, I'm unable to reproduce this on my Windows 10 machine.

Burtannia:
[...]

PR #4106, opened 2018-06-21
Re #4091: Pre-Windows 10, default --color=never

mpilgrem:
Implements `--color=never` as default on Windows before Windows 10, if emulation would be required to support ANSI colour codes.

Also bumps the lts-11 resolver in `stack.yaml` to lts-11.14, in order to benefit from `ansi-terminal-0.8.0.4`.

`defaultColorWhen`, exported by new module `Stack.DefaultColorWhen`, establishes the default. That module differs, pending on whether the operating system is Windows or non-Windows (see the conditional in `package.yaml`).

`globalOptsFromMonoid` in module `Stack.Options.GlobalParser` is modified so that the default is an argument rather than a given.

A beneficial side-effect of `defaultColorWhen` on Windows is that it enables ANSI on ANSI-capable native terminals. This allows some existing code in module `Main` to be removed.

Tested on Windows 10 and some limited testing on Windows 7.

### DIFF
--- a/src/windows/Stack/DefaultColorWhen.hs
+++ b/src/windows/Stack/DefaultColorWhen.hs
@@ -1,0 +1,65 @@
+{-# LANGUAGE CPP #-}
+
+{- | This version of the module is only for Windows operating systems.
+-}
+module Stack.DefaultColorWhen
+  ( defaultColorWhen
+  ) where
+
+-- The Win32 package provides CPP macro WINDOWS_CCONV.
+#include "windows_cconv.h"
+
+import Stack.Prelude (stdout)
+import Stack.Types.Runner (ColorWhen (ColorAuto, ColorNever))
+
+import Data.Bits ((.|.), (.&.))
+import Foreign.Marshal (alloca)
+import Foreign.Ptr (Ptr)
+import Foreign.Storable (peek)
+import System.Win32.Types (BOOL, DWORD, HANDLE, iNVALID_HANDLE_VALUE,
+  nullHANDLE, withHandleToHANDLE)
+
+defaultColorWhen :: IO ColorWhen
+defaultColorWhen = withHandleToHANDLE stdout aNSISupport
+
+-- The following is based on extracts from the modules
+-- System.Console.ANSI.Windows.Foreign and System.Console.ANSI.Windows.Detect
+-- from the ansi-terminal package, simplified.
+
+-- | This function first checks if the Windows handle is valid and yields
+-- 'never' if it is not. It then tries to get a ConHost console mode for that
+-- handle. If it can not, it assumes that the handle is ANSI-enabled. If virtual
+-- termimal (VT) processing is already enabled, the handle supports 'auto'.
+-- Otherwise, it trys to enable processing. If it can, the handle supports
+-- 'auto'. If it can not, the function yields 'never'.
+aNSISupport :: HANDLE -> IO ColorWhen
+aNSISupport h =
+  if h == iNVALID_HANDLE_VALUE || h == nullHANDLE
+    then return ColorNever  -- Invalid handle or no handle
+    else do
+      tryMode <- getConsoleMode h
+      case tryMode of
+        Nothing     -> return ColorAuto  -- No ConHost mode
+        Just mode   -> if mode .&. eNABLE_VIRTUAL_TERMINAL_PROCESSING /= 0
+          then return ColorAuto  -- VT processing already enabled
+          else do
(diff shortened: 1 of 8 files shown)


dbaynard:
From a first glance, this looks great. Thank you!  It should be fairly quick to review — I'm just searching for a windows reviewer.

I suspect we may be better off adding `ansi-terminal` as an extra-dep, rather than bumping the resolver version, but that'd be a fairly straightforward change.

src/windows/Stack/DefaultColorWhen.hs
```diff
@@ -0,0 +1,83 @@
...
+-- handle. If it can not, it assumes that the handle is ANSI-enabled. If virtual
+-- termimal (VT) processing is already enabled, the handle supports 'auto'.
+-- Otherwise, it trys to enable processing. If it can, the handle supports
+-- 'auto'. If it can not, the function yields 'never'.
+aNSISupport :: HANDLE -> IO ColorWhen
+aNSISupport h =
+  if h == iNVALID_HANDLE_VALUE || h == nullHANDLE
+    then return ColorNever  -- Invalid handle or no handle
+    else do
+      tryMode <- try (getConsoleMode h) :: IO (Either SomeException DWORD)
```
snoyberg: I'd recommend using `tryAny` here: it will make the type signature unnecessary, and be more explicit about using a function that doesn't catch async exceptions.

src/windows/Stack/DefaultColorWhen.hs
```diff
@@ -0,0 +1,83 @@
...
+      case tryMode of
+        Left _     -> return ColorAuto  -- No ConHost mode
+        Right mode -> if mode .&. eNABLE_VIRTUAL_TERMINAL_PROCESSING /= 0
+          then return ColorAuto  -- VT processing already enabled
+          else do
+            let mode' = mode .|. eNABLE_VIRTUAL_TERMINAL_PROCESSING
+            trySetMode <- try (setConsoleMode h mode') :: IO (Either SomeException ())
```
snoyberg: Same here.

src/windows/Stack/DefaultColorWhen.hs
```diff
@@ -0,0 +1,83 @@
...
+
+setConsoleMode :: HANDLE -> DWORD -> IO ()
+setConsoleMode handle attributes
+  = throwIfFalse $ cSetConsoleMode handle attributes
+
+throwIfFalse :: IO Bool -> IO ()
+throwIfFalse action = do
+  succeeded <- action
+  if not succeeded
+    then getLastError >>= throw . ConsoleException
```
snoyberg: Please use `throwIO` instead of `throw`. The former is resistant to later refactorings that would cause a bottom value to hide inside something lazy.

src/windows/Stack/DefaultColorWhen.hs
snoyberg: Or did you mean to catch only a `ConsoleException` from below?

stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-11.6
+resolver: lts-11.14
```
snoyberg: Is it strictly necessary to bump the LTS version here? I tend to avoid such updates in unrelated patches.

mpilgrem:
Before turning to the results of the review, an update on testing: I have tested on Windows 7 and 10, both in a native terminal and in a mintty terminal (Cygwin) (the test being `stack new test`; introduce a small bug in the code; `stack build`, with bug reporting). It works as expected (ie no colour code corruption of output in a Windows 7 native terminal; colour in all other circumstances).

src/windows/Stack/DefaultColorWhen.hs
mpilgrem: I've now moved away from duplicating `ansi-terminal`'s use of the exception mechanism for this small extract.

src/windows/Stack/DefaultColorWhen.hs
mpilgrem: As above.

src/windows/Stack/DefaultColorWhen.hs
mpilgrem: As above.

stack.yaml
mpilgrem: Now use an extra-dep instead. (`ansi-terminal-0.8.0.4` brings the mintty fix to its `hSupportsANSI`.)

mpilgrem:
I think my further changes are responsive to all the requested changes.

PR #4108, opened 2018-06-22
Fix issues with stderr being inherited by child processes

mgsloan:
The only case where stdio should be inherited is when passing execution to
another process, so there should be no uses of runProcess

Most processes should also not inherit stdin, but I'm not sure if there are any
concrete issues with this in stack. I recall some prior issues where some
processes do not like being given a closed stdin, for example.

There is one remaining use of readProcessStdout_ in Stack.Docker which should
probably be fixed. However, I'm not sure what the correct implementation there
should be. The comment in the code makes it seem like it should log stderr while
reading stdout.


snoyberg:
I've looked through the usages closely, I _think_ they'll all be fine. I guess we'll find out with more testing.

mgsloan:
:+1: Even if these are not concurrent with the sticky logger, I think it makes sense to avoid stderr inheritance for at least the following reasons:

1. Others might copy paste such process code into places that are concurrent with sticky logger use

2. Concurrent use of processes that inherit stderr may intersperse their output.

Issue #3510, opened 2017-10-24
Stack does not retry effectively when S3 is being flaky.


NorfairKing:
This an issue that will help us to evaluate how good `stack`'s retrying strategy is when it comes to flaky connections.

jdbrandon:
Retry failed but running the command manually (not in make) succeeded.

vilinski:
Hi,
I'm using `cntlm` behind corp firewall, and it hangs periodically, like in this case:

>'stack build intero'

<details>
  <summary>Click to view terminal output</summary>

```
...
Warning: Retry number 3 after a total delay of 300000 us
         If you see this warning and stack fails to download, but running the command again solves
         the problem, please report here: https://github.com/commercialhaskell/stack/issues/3510
HttpExceptionRequest Request {
  host                 = "s3.amazonaws.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
  path                 = "/hackage.fpcomplete.com/package/haskeline-0.7.4.1.tar.gz"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (InternalException (HostCannotConnect "localhost" [Network.Socket.connect: <socket: 708>: failed (Connection refused (WSAECONNREFUSED)),Network.Socket.connect: <socket: 708>: failed (Connection refused (WSAECONNREFUSED))]))
Progress: 4/5
```
</details>

leptonyu:
Hi, 
   this is my log, what's the problem?

```
stack install stylish-haskell

Warning: Retry number 0 after a total delay of 0 us
         If you see this warning and stack fails to download, but running the command again solves the problem, please report here:
         https://github.com/commercialhaskell/stack/issues/3510
             
Warning: Retry number 0 after a total delay of 0 us
         If you see this warning and stack fails to download, but running the command again solves the problem, please report here:
         https://github.com/commercialhaskell/stack/issues/3510
             
Warning: Retry number 0 after a total delay of 0 us
         If you see this warning and stack fails to download, but running the command again solves the problem, please report here:
         https://github.com/commercialhaskell/stack/issues/3510
             
Warning: Retry number 0 after a total delay of 0 us
         If you see this warning and stack fails to download, but running the command again solves the problem, please report here:
         https://github.com/commercialhaskell/stack/issues/3510
syb-0.7: download
[...]

PR #4109, opened 2018-06-22
Changed the default retry strategy

NorfairKing:
This changes makes the default retry strategy an exponential backoff
starting from 100 ms.
This should counteract the problem that is being tracked in #3510.
This new strategy means that it will take 12.7 seconds for stack to
fail entirely when the network cable is unplugged.

I did not add any tests, as discussed with @snoyberg.

### DIFF
--- a/src/Network/HTTP/Download/Verified.hs
+++ b/src/Network/HTTP/Download/Verified.hs
@@ -30,7 +30,7 @@ import qualified    Data.Text.Encoding as Text
 import              Control.Monad
 import              Control.Monad.Catch (Handler (..)) -- would be nice if retry exported this itself
 import              Stack.Prelude hiding (Handler (..))
-import              Control.Retry (recovering,limitRetries,RetryPolicy,constantDelay,RetryStatus(..))
+import              Control.Retry (recovering,limitRetries,RetryPolicy,exponentialBackoff,RetryStatus(..))
 import              Crypto.Hash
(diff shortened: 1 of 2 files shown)


src/Network/HTTP/Download/Verified.hs
```diff
@@ -59,9 +59,20 @@ data DownloadRequest = DownloadRequest
...
+-- * 0.1s
+-- * 0.2s
+-- * 0.4s
+-- * 0.8s
+-- * 1.6s
+-- * 3.2s
+-- * 6.4s
 drRetryPolicyDefault :: RetryPolicy
-drRetryPolicyDefault = limitRetries 3 <> constantDelay onehundredMilliseconds
+drRetryPolicyDefault = limitRetries 7 <> constantDelay onehundredMilliseconds
```
dbaynard: Should this be `constantDelay` or `exponentialBackoff`?

(If the latter it'll need importing, too, around line 33)

NorfairKing: Wauw, what a silly mistake. Expect a fix in about 2 mins.

NorfairKing:
> This is quite a small change so I can merge.

I'm not sure what this means.

dbaynard:
Ah, I apologise for the ambiguity. I missed the word 'soon'. I was trying to say the PR will be merged soon.

Issue #3942, opened 2018-03-23
Self-executing scripts don't pipe plan construction errors from runhaskell to terminal


docmoxie:
### General summary/comments 

Self-executing scripts don't pipe plan construction errors from `runhaskell` to terminal.

### Steps to reproduce

Given a `stack.yaml` that uses a package that relies on another package from `extra-deps`:
```yaml
resolver: lts-8.22

packages:
- location:
    git: git@github.com:cjdev/cloudseeder.git
    commit: c7068f25c41bcc0b4bdfc9b38ee5b05b5caefae5
  extra-dep: true

extra-deps:
- optparse-applicative-0.14.0.0

flags: {}

extra-package-dbs: []
```

...and a self-executing script that references that package, called `script.hs`:
```haskell
#!/usr/bin/env stack
-- stack runhaskell --package cloud-seeder

main :: IO ()
main = putStrLn "yo"
```

### Expected

With the `- optparse-applicative-0.14.0.0` line in place, running gives the expected result:

```bash
$ ./script
yo
```
```bash
$ stack runhaskell --package cloud-seeder ./script.hs
yo
```

### Actual

However, if we remove `- optparse-applicative-0.14.0.0`, only the direct call to `runhaskell` returns useful error information:

```bash
$ ./script
Plan construction failed.
```
```bash
$ stack runhaskell --package cloud-seeder ./script.hs

Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for cloud-seeder-0.1.0.0:
    optparse-applicative-0.13.2.0 from stack configuration does not match >=0.14.0.0 (latest matching version is 0.14.2.0)
needed since cloud-seeder is a build target.

Some potential ways to resolve this:

  * Recommended action: try adding the following to your extra-deps in /Users/marnold/Code/lambda/runhaskell-test/stack.yaml:

- optparse-applicative-0.14.2.0

  * Set 'allow-newer: true' to ignore all version constraints and build anyway.

  * You may also want to try using the 'stack solver' command.

Plan construction failed.
```

### Stack version

```
$ stack --version
Version 1.6.5, Git revision 24ab0d6ff07f28276e082c3ce74dfdeb1a2ca9e9 (5514 commits) x86_64 hpack-0.20.0
```

### Method of installation

* Official binary, downloaded from stackage.org or fpcomplete's package repository

SkyWriter:
#4114 addresses this

PR #4114, opened 2018-06-24
Fixes self-executing scripts not piping plan construction errors from runhaskell to terminal, issue #3942

SkyWriter:
The issue seems to have been caused by improper logging level. Fixed that and added an integration test case.

* [x] ChangeLog.md has a relevant bugfix entry
* [x] No documentation updates are necessary, I reckon

I have tested the fix in accordance with the issue + added integration test cases.


isovector:
Looks great! Thanks!

PR #4121, opened 2018-06-29
Fix #3992 Windows 10: Work around upstream git clone issue

### LINKED ISSUE #3992: Stack 1.7.1 outputs raw ANSI codes on Windows 10
Stack 1.7.1 (Version 1.7.1, Git revision 681c800873816c022739ca7ed14755e85a579565 (5807 commits) x86_64 hpack-0.28.2) outputs 'raw' 'ANSI' codes on Windows 10 (the ConHost terminals - Command Prompt and PowerShell) rather than those codes being interpreted as console virtual terminal sequences. Earlier versions of stack (specifically, 1.7.0) behaved as expected.

I do not know how `stack` handles 'ANSI' codes, but my best guess of the source of the problem is as follows: ConHost on Windows 10 is ANSI-capable, but it is not ANSI-enabled by default; applications have to 'turn it on'. The `ansi-terminal` package seeks to do that, the first time one of its relevant functions is used (a function that behaves differently on Windows when ANSI-support is native or it is emulated; `ansi-terminal` provides emulation for legacy Windows operating systems). It seems to me that `stack` is sending ANSI codes to the terminal before it has enabled it and outside of the functions provided by `ansi-terminal`.

In that regard, I searched the `stack` source code for `System.Console.ANSI` (the main `ansi-terminal` module). It is imported in modules `Stack.Types.Runner` and `Text.PrettyPrint.Leijen.Extended`.

Unfortunately, the former module does not use explicit imports, but I think it uses only `hSupportsANSI` from `ansi-terminal`. That function does not 'turn on ANSI' (and its heuristics probably are not very reliable on Windows).

The latter module (`Text.PrettyPrint.Leijen.Extended`) does use explicit imports. It uses `hSupportsANSI` too but it also uses `setSGRCode` (in function `displayANSISimple`). `setSGRCode` is a function from `ansi-terminal` that 'turns on ANSI' on Windows, if ANSI-support is available and not yet turned on.

However, I added some `Debug.Trace.trace` functions to that `setSGRCode` usage and discovered:

(1) `stack` was sending ANSI codes to the terminal before evaluating those `setSGRCode` uses. It appears to me that `displayANSISimple` is not the only way `stack` introduces 'ANSI` codes into its output; and

(2) once it used those `setSGRCode` uses, ANSI was no longer 'raw' but triggered the expected colours. That is, as expected, once a call to the `ansi-terminal` functionality was made, that package 'turned on ANSI'. (It was `-- verbose` that triggered, at some point, the use of `displayANSISimple`.)

mpilgrem:
See comment in code.

Compiles, but not yet tested on Windows (in part because my own stack is 'up to date').

### DIFF
--- a/src/Stack/Upgrade.hs
+++ b/src/Stack/Upgrade.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE NoImplicitPrelude     #-}
 {-# LANGUAGE CPP                   #-}
 {-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE FlexibleContexts      #-}
@@ -24,6 +24,10 @@ import           Path
 import qualified Paths_stack as Paths
 import           Stack.Build
 import           Stack.Config
+-- Following import is redundant on non-Windows operating systems
+#ifdef WINDOWS
+import           Stack.DefaultColorWhen (defaultColorWhen)
+#endif
 import           Stack.Fetch
 import           Stack.PackageIndex
 import           Stack.PrettyPrint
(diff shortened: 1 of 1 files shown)


mpilgrem:
Now tested on Windows 10.

isovector:
Looks great, thanks!! Waiting to hear back from AppVeyor, but I'm happy to merge it when everything goes green!

mpilgrem:
@isovector, I think we are green. I have now also tested on Windows 7.

PR #4132, opened 2018-07-04
Use a hard-coded tool dependency map (fixes #4125)

### LINKED ISSUE #4125: stack 1.7.1 still has problems finding build tools ("The program 'happy' is required")

hledger appveyor builds are failing due to not being able to find "happy" when building pretty-show, haskell-src-exts, language-javascript. https://github.com/simonmichael/hledger/issues/832

I saw this a month ago and tried several things to fix it, and probably thought I had solved it by upgrading to stack 1.7.1, but now it seems not.

Possibly related:
https://github.com/commercialhaskell/stack/issues/595
https://github.com/commercialhaskell/stack/issues/3178
https://github.com/commercialhaskell/stack/pull/3249

### Steps to reproduce

I don't have a short repro, instead I'll point to appveyor logs. 

### Expected

Successful builds of pretty-show, haskell-src-exts, language-javascript.

### Actual

These packages failed to build due to "missing happy", even with stack 1.7.1. Eg see
https://ci.appveyor.com/project/simonmichael/hledger/build/master-460.

Some more details:

hledger build history: https://ci.appveyor.com/project/simonmichael/hledger/history. 
Only the full, >30m builds are building pretty-show etc. 

The last full build that succeeded was https://ci.appveyor.com/project/simonmichael/hledger/build/master-451 on 26 May. The stack version is not reported, but it fetched http://www.stackage.org/stack/windows-x86_64, and stack 1.7.1 was released on 27 Apr, so I would have thought it was using stack 1.7.1. Unless there was some path issue.

The next full build, 456 on 30 May, failed with `No packages found in snapshot which provide a "happy" executable, which is a build-tool dependency of "pretty-show"` warnings and `The program 'happy' version >=1.19 is required but it could not be found` errors. So did build 460 and later, and these report the stack version as definitely 1.7.1. https://ci.appveyor.com/project/simonmichael/hledger/build/master-460.

Builds 459, 462 and 465 were non-full builds that succeeded by luck, appveyor apparently picks randomly from several build instances each with their own cache state.

### Stack version

1.7.1

### Method of installation

snoyberg:
This is a fundamental change to how Stack discovers build tools, its inclusion should be discussed before we merge. We should also wait to see if it actually fixes #4125.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -28,6 +28,13 @@ Behavior changes:
   [help file](https://github.com/commercialhaskell/stack-templates/blob/master/STACK_HELP.md)
   with more information on how to discover templates. See:
   [#4039](https://github.com/commercialhaskell/stack/issues/4039)
+* Build tools are now handled in a similar way to `cabal-install`. In
+  particular, for legacy `build-tools` fields, we use a hard-coded
+  list of build tools in place of looking up build tool packages in a
+  tool map. This both brings Stack's behavior closer into line with
+  `cabal-install`, avoids some bugs, and opens up some possible
+  optimizations/laziness. See:
+  [#4125](https://github.com/commercialhaskell/stack/issues/4125).
 
 Other enhancements:
 
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -40,8 +40,6 @@ import           Stack.Build.Cache
 import           Stack.Build.Haddock
 import           Stack.Build.Installed
 import           Stack.Build.Source
-import           Stack.BuildPlan
-import           Stack.Config (getLocalPackages)
 import           Stack.Constants
 import           Stack.Package
 import           Stack.PackageDump
@@ -133,7 +131,6 @@ data Ctx = Ctx
     , baseConfigOpts :: !BaseConfigOpts
     , loadPackage    :: !(PackageLocationIndex FilePath -> Map FlagName Bool -> [Text] -> M Package)
     , combinedMap    :: !CombinedMap
-    , toolToPackages :: !(ExeName -> Map PackageName VersionRange)
     , ctxEnvConfig   :: !EnvConfig
     , callStack      :: ![PackageName]
     , extraToBuild   :: !(Set PackageName)
@@ -196,8 +193,7 @@ constructPlan ls0 baseConfigOpts0 locals extraToBuild0 localDumpPkgs loadPackage
     let inner = do
             mapM_ onWanted $ filter lpWanted locals
             mapM_ (addDep False) $ Set.toList extraToBuild0
-    lp <- getLocalPackages
-    let ctx = mkCtx econfig lp
+    let ctx = mkCtx econfig
     ((), m, W efinals installExes dirtyReason deps warnings parents) <-
         liftIO $ runRWST inner ctx M.empty
     mapM_ (logWarn . RIO.display) (warnings [])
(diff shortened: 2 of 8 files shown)


snoyberg:
Some design decision time. Stack originally (and through the 1.7 release) calculated build tools by looking at all of the packages in a snapshot, finding the executables they provide, and performing a reverse lookup for any build-tools values. This works, but:

* Requires eagerly parsing all cabal files in a snapshot, something we may want to try to avoid
* Has had some bugs in the past with overridden packages, e.g. #4125 
* Is out of sync with cabal (see next paragraph)

Starting with Cabal 2.0 (or 2.2), there's a new feature in the cabal file format called build-tool-depends, which allows you to specify which package a tool comes from, bypassing the need for the reverse lookup. cabal-install handles the legacy build-tools field with a hard-coded list of packages like `alex` and `happy`. This PR does the same for Stack, though with a slightly difference list.

This _may_ cause some existing builds to break, but that's highly doubtful, and can be worked around easily enough by switching to build-tool-depends and/or explicitly building the needed dependencies. I lean in favor of merging this because matching cabal behavior is a Good Thing, pushing towards the new build-tool-depends and more explicit deps is a Good Thing, and reducing complexity in our code base is a Good Thing 😄 .

I'd say let's give it another few days for someone to object, and to confirm it fixes #4125, and then we can merge.

Ericson2314:
N.B. I made https://github.com/haskell/cabal/issues/5412 based on the issues I had making `build-tool-depends` (and `build-tools`) the sole way to get access to binaries with Nixpkgs alone. I'm curious what Stack's experience with packages that, e.g. do `build-depends: hspec-discovery` and expect to get an executable in addition to a library.

snoyberg:
I'm not quite sure of the question, but I'll give an answer related to other bugs I've seen. I think it's a mistake for Cabal to hide the `hspec-discovery` executable when `build-depends: hspec-discovery` is specified. It breaks compatibility with how cabal worked in the past, and AFAICT doesn't add any real determinism to the process, since we've already specified that `hspec-discovery` must be installed before the package in question. Stack does not attempt to do any of this hiding, and I would strongly advocate against adding in such a restriction.

snoyberg:
As an update: this does in fact seem to solve #4125.

simonmichael:
This sounds good to me. I have an addition for the list of tools:

- `hsx2hs`, provided by http://hackage.haskell.org/package/hsx2hs. Users include the http://hackage.haskell.org/package/darcsden  package. (It is not mentioned in darcsden's cabal file, but should be.)

If a tool is not in the hard-coded list, does/should the install fail, or fall back to the old behaviour ?

snoyberg:
I'll add that, no problem. If it's not in the hard-coded list, and the executable isn't already available with that name, Stack will print a warning, and the build itself will most likely fail. Falling back to the old behavior would require all of the complicated code I mentioned which is suffering from the bugs right now and preventing performance optimizations.

Ericson2314:
@snoyberg [BTW, happy to talk about this elsewhere as I'm definitely veering off-topic.] Being able to only build the libraries components of dependencies is, I agree a small benefit. The real benefit is cross compilation: the tools must be built for the build platform, but libraries are built for the host platform. If everything is built together then either the libraries are for the wrong platform, or the tools won't run. If every `build-depends` package has its binaries built for the build platform, then we've doubled the build time for cross as a great number of packages are conservatively being built twice.

snoyberg:
OK, that motivation makes a lot more sense, I hadn't heard it before. I still think it's best to leave Stack as-is for maximum backwards compatibility, I know in particular the hspec-discovery thing has caused breakage with new-build.

PR #4138, opened 2018-07-07
'stack ghci' now asks which main module to load before building

mgsloan:
Before this change, 'stack ghci' would ask the user a question after doing a
build which may take a long time. The reason it was this way is that the
question required resolution of package files, and package file resolution
needed to come after the build due to #1180.

The solution here is to resolve the package files twice - once before the build, 
and once after. This isn't the most efficient solution possible, but it is a
much better user experience to ask the main target question before building
rather than after.

Tested this by running on a large project and observing that ghci functions properly, and that the main target question is asked before the build.

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -87,6 +87,13 @@ data GhciPkgInfo = GhciPkgInfo
     , ghciPkgPackage :: !Package
     } deriving Show
 
+-- | Loaded package description and related info.
+data GhciPkgDesc = GhciPkgDesc
+    { ghciDescPkg :: !Package
+    , ghciDescCabalFp :: !(Path Abs File)
+    , ghciDescTarget :: !Target
+    }
+
 -- Mapping from a module name to a map with all of the paths that use
 -- that name. Each of those paths is associated with a set of components
 -- that contain it. Purpose of this complex structure is for use in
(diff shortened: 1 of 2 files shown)


src/Stack/Ghci.hs
```diff
@@ -633,15 +637,59 @@ makeGhciPkgInfo buildOptsCLI sourceMap installedMap locals addPkgs mfileTargets
...
+    -> RIO env [GhciPkgInfo]
+getGhciPkgInfos sourceMap addPkgs mfileTargets localTargets = do
+    (installedMap, _, _, _) <- getInstalled
+        GetInstalledOpts
+            { getInstalledProfiling = False
+            , getInstalledHaddock   = False
+            , getInstalledSymbols   = False
+            }
+        sourceMap
+    let localLibs =
```
mihaimaruseac: I think we should add the internal libs here too, if they exist

mgsloan: Agreed!  Perhaps that can be done in a separate PR?  This bit is just a refactoring, same logic as before the change.

mihaimaruseac: Sure. I'm travelling now but can look at it in two weeks and see to it.

mgsloan:
Thanks for reviewing!

Issue #4086, opened 2018-06-14
DISCUSSION: Mark GHCJS support as deprecated?


snoyberg:
From what I can tell:

* The GHCJS codepaths are not currently being actively maintained in Stack
* Most of the GHCJS community recommends Nix for setting up the toolchain
* None of the regular Stack contributors are using the GHCJS support

Given this, I'd recommend in the next major release of Stack to mark GHCJS support as deprecated and begin printing warnings when it's used. We can later worry about when, if ever, we actually remove the support.

Another possibility could be to only remove support for installing GHCJS itself, but retain the ability to build with it.

mgsloan:
> Another possibility could be to only remove support for installing GHCJS itself, but retain the ability to build with it.

I'm in favor of this approach.  IIRC the code for using ghcjs to build is quite small and so not a big maintenance burden, but the code for installing ghcjs is complicated and error prone.  Considering that `stack setup` for ghcjs can take a *long* time, it seems sensible to encourage users to find a different approach to installation, perhaps involving prebuilt binaries.

borsboom:
When I last tried the GHCJS support in the fall, it was very frustrating to get working.  Documentation was incomplete and out of date, there didn't seem to be a good way to get it to work with recent GHC versions or Stackage snapshots, and it all just felt very hacky because you needed these special GHCJS distributions with specific versions of base packages that were difficult to generate.  To be fair, a big part of this is the fact that there wasn't any release process for GHCJS versions, it just seemed to be in perpetual development.  I'd be in favour of deprecating unless some real effort is put into making this more usable.

bitemyapp:
How's a warning that'll usually get triggered at the start?

```
$ /home/callen/work/stack/.stack-work/install/x86_64-linux/lts-11.6/8.2.2/bin/stack build
Building a GHCJS project. Note that GHCJS support in Stack is DEPRECATED and it will be removed  in a future release of Stack.
miso-0.21.1.0: unregistering (flags changed from ["-fexamples","--ghcjs"] to ["--ghcjs"])
miso-0.21.1.0: configure (lib)
Configuring miso-0.21.1.0...
miso-0.21.1.0: build (lib)
Preprocessing library miso-0.21.1.0...
[... snip ...]
[34 of 34] Compiling Miso             ( ghcjs-src/Miso.hs, .stack-work/dist/x86_64-linux/Cabal-1.24.0.0_ghcjs/build/Miso.js_o )
miso-0.21.1.0: copy/register
[...]

PR #4154, opened 2018-07-13
Warning about GHCJS use when constructing a build plan

bitemyapp:
Fixes https://github.com/commercialhaskell/stack/issues/4086

I tested it (produced the output cited in the issue above) by building [miso](https://github.com/dmjio/miso).

Only issue with how it presently works is that people might miss the warning if it's at the start rather than the end. Do we have a means of registering warnings that should get emitted at the end by `main` or similar?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,16 +27,16 @@ manual_build:
   when: manual
   script: *build
 
-docker_image_lts-10:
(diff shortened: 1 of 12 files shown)


bitemyapp:
@snoyberg How's this changelog message? It seemed to fit "major change" in my head.

PR #4160, opened 2018-07-17
Mark ghcjs support as experimental #4086

### LINKED ISSUE #4086: DISCUSSION: Mark GHCJS support as deprecated?

matchwood:
Changes deprecation warning into experimental warning and makes relevant changes to documentation.

Fixes #4086 .

### DIFF
--- a/doc/ghcjs.md
+++ b/doc/ghcjs.md
@@ -1,8 +1,8 @@
 <div class="hidden-warning"><a href="https://docs.haskellstack.org/"><img src="https://rawgit.com/commercialhaskell/stack/master/doc/img/hidden-warning.svg"></a></div>
 
-# GHCJS
+# GHCJS (experimental)
 
-To use GHCJS with stack, place a GHCJS version in the [`compiler`](yaml_configuration.md#compiler) field of `stack.yaml`.  After this, all stack commands should work with GHCJS!  In particular:
+To use GHCJS with stack, place a GHCJS version in the [`compiler`](yaml_configuration.md#compiler) field of `stack.yaml`.  After this, some stack commands should work with GHCJS.  In particular:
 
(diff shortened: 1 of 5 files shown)


doc/ghcjs.md
```diff
@@ -21,7 +21,13 @@ then you may need to install a different version of `node`. See
 
 ## Example Configurations
 
-### Recent versions of GHCJS, repacked for stack
+### GHCJS repacked for snapsnots lts-8 and lts-9
+
+Please see the [ghcjs-stack-dist repository](https://github.com/matchwood/ghcjs-stack-dist) for `lts-8` and `lts-9` configurations and refer to the [README](https://github.com/matchwood/ghcjs-stack-dist/blob/master/README.md) for common issues.
```
dbaynard: I take it we can't just have a table like for the earlier snapshots?

dbaynard:
As @bitemyapp has just changed these files, I'm requesting a review from him — mainly to check there's nothing missing.

doc/ghcjs.md
matchwood: We could but I'd prefer people engaged with the README on the ghcjs-stack-dist repo (as there are some important issues flagged up there) and I didn't want to just copy across all that info. If and when we get a better story for ghcjs support I will rewrite the docs and put everything in a table etc.

Issue #3850, opened 2018-02-06
stack new requires network to work


vincenthz:
### General summary/comments (optional)

without a network connection, `stack new` just fails to download a template, even though the template already exists in the template directory.

### Expected

`stack new` try to download (update ? refresh cache ?) the template, but fallback to the existing template

```
 ls ~/.stack/templates/ -lh
total 8.0K
-rw-------. 1 user user 3.8K Feb  5 18:58 new-template.hsfiles
-rw-------. 1 user user   42 Feb  5 18:58 new-template.hsfiles.etag
```

### Actual

```
$ stack new bla
Downloading template "new-template" to create project "bla" in bla/ ...

Warning: Retry number 0 after a total delay of 0 us
....
Warning: Retry number 3 after a total delay of 300000 us
         If you see this warning and stack fails to download, but running the command again solves the problem, please report here:
         https://github.com/commercialhaskell/stack/issues/3510
HttpExceptionRequest Request {
  host                 = "raw.githubusercontent.com"
  port                 = 443
  secure               = True
  requestHeaders       = [("If-None-Match","\"20199ac22a46c1b88a1e7e93f6555f4243430c44\"")]
  path                 = "/commercialhaskell/stack-templates/master/new-template.hsfiles"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (ConnectionFailure Network.Socket.getAddrInfo: does not exist (Try again))
```

### Stack version

```
$ stack --version
Version 1.6.3, Git revision b27e629b8c4ce369e3b8273f04db193b060000db (5454 commits) x86_64 hpack-0.20.0
```

mgsloan:
:+1: Agreed!

sudsj:
Hi, I'm sort of a beginner to Haskell and since this issue was marked newcomer friendly I hope I can contribute. Can you @mgsloan @vincenthz  give an overview of what has to be done, it would be helpful.

mgsloan:
@sudsj Hi, sounds good!  There are a few ways this could be implemented.  I think it would be best to have it do the following:

* When downloading templates, always store in `$STACK_ROOT/templates-cache`.  Should use `Network.HTTP.Download.redownload`.  Then, load from the stored file.

[...]

PR #4173, opened 2018-07-24
Use cached template if template download fails

BraulioVM:
I saw that issue #3850 was newcomer friendly and so I'm trying to fix it.

I intend to add tests once I make sure I'm doing things properly. For the time being, I have manually tested the feature by:

1. Properly downloading a valid template `$ stack new tmp https://raw.githubusercontent.com/stesta/haskell-templates/master/snap-basic-websockets.hsfiles` 
2. Removing the `tmp` directory: `$ rm -r tmp`.
3. Disabling network connectivity on my computer.
4. Trying to use the template again by executing the command in 1 again.

And it works as I intended. I haven't tried yet with a `TemplatePath` using the `RepoPath` constructor but that should be quick to try. I haven't seen what would happen if `redownload` throws a
`RedownloadInvalidResponse` instead of a `RedownloadHttpError` either, but the worst thing that I could see happening is a log error badly formatted. I'll take a look at that case too anyway.

I have tried to make sure nobody was catching a `HttpException` when calling `redownload` from any part of the codebase. That could be slightly problematic because, as I'm catching that exception myself at `redownload`, it would never propagate to the former `HttpException` catchers, which could have been doing something  _useful_ with the exception. I've grepped `redownload` and there is no function calling it that catches the exception, but I can't be sure about functions higher in the call stack. I would say this is not going to be a problem, but wanted to find out what you think about it.


test/integration/tests/3850-cached-templates-network-errors/files/simple.hsfiles
```diff
@@ -0,0 +1,284 @@
+{-# START_FILE {{name}}.cabal #-}
```
mihaimaruseac: Is this file needed somewhere?

BraulioVM: No! I left that by accident because I needed it for the first implementation of the integration test. I'll remove it.

BraulioVM:
I feel like I could do some more testing (trying with a `RepoPath` for example, which will be easy). I will do that tonight (in like 6-7 hours).

I believe I should add this to the `ChangeLog.md`. What about the docs?

mihaimaruseac:
The only relevant place in the docs on a quick scan seems to be https://docs.haskellstack.org/en/stable/GUIDE/#stack-new but even there it would be offtopic to talk about using cached templates.

BraulioVM:
I stumbled upon some issues that made stack build the whole project every time I made a change in the tests so I needed more time than expected to finish the PR. I'm not sure what it was but I fixed it by removing the `~/.stack` directory.

Anyway, tell me if there's anything missing. I will try to be quick.

PR #4175, opened 2018-07-26
Pantry - package identifier parsing fix

qrilka:

`stack build` now stops at later `undefined`

### DIFF
--- a/subs/pantry/src/Pantry/Types.hs
+++ b/subs/pantry/src/Pantry/Types.hs
@@ -442,8 +442,9 @@ data PackageTarball = PackageTarball
   }
   deriving Show
 
--- | Modified version of Cabal's parser - we don't need null version
--- in package indentifiers
+-- | This is almost a copy of Cabal's parser for package identifiers,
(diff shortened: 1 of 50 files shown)


subs/pantry/src/Pantry/Types.hs
```diff
@@ -438,9 +442,27 @@ data PackageTarball = PackageTarball
...
-parseC = Distribution.Text.simpleParse
+-- | Modified version of Cabal's parser - we don't need null version
+-- in package indentifiers
+parsePackageIdentifier :: String -> Maybe PackageIdentifier
+parsePackageIdentifier str =
+    case [p | (p, s) <- Parse.readP_to_S parser str, all isSpace s] of
+        [] -> Nothing
+        (p:_) -> Just p
+  where
+    parser = do
```
snoyberg: Looks like a better workaround for the Cabal bug. Can you add a comment explaining why we have to go through these hoops so someone in the future doesn't "simplify" this code?

qrilka: Added more detailed comments

PR #4188, opened 2018-07-29
Adapt upload integration test for Windows

koterpillar:
See #4167 for context. This is an attempt to fix the upload integration test on Windows; currently I'm still getting a hanging process when running the real test, but not via a manual `stack runghc`.

### DIFF
--- a/test/integration/tests/upload/files/FakeHackage.hs
+++ b/test/integration/tests/upload/files/FakeHackage.hs
@@ -1,8 +1,32 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-import Network.Simple.TCP
+import Control.Concurrent
+import Control.Monad
+import Network.Socket hiding (recv)
+import Network.Socket.ByteString (recv, sendAll)
+import System.Exit
+import System.IO
 
 -- | Fake server that always responds with HTTP OK
 main =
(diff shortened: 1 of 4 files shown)


koterpillar:
Now passing.

borsboom:
Awesome, passes in my environment too.  Thanks for taking care of this so quickly, I really appreciate it!

test/integration/tests/upload/files/FakeHackage.hs
```diff
@@ -1,8 +1,32 @@
...
+    withSocketsDo $ do
+        _ <- forkIO serve
+        -- Exit after a delay to ensure the process doesn't linger around
+        threadDelay 10000000
+        exitSuccess
+
+serve :: IO ()
+serve = do
+    let hints = defaultHints {addrFlags = [AI_PASSIVE], addrSocketType = Stream}
+    (addr:_) <- getAddrInfo Nothing Nothing (Just "12415")
```
Blaisorblade: `hints` seems used nowhere, what's it doing there? [Docs](http://hackage.haskell.org/package/network-2.7.0.2/docs/Network-Socket.html#v:getAddrInfo) suggest calling `getAddrInfo (Just hints) ...` to actually use it. That affects #4191!

PR #4190, opened 2018-07-29
Fix #2426: use removePathForcibly instead of removeDirRecur everywhere

### LINKED ISSUE #2426: removeDirectoryRecursive: permission denied (Access is denied.) Windows 7
This happens in **any** directory:
```
D:\>stack -v update
Version 1.1.2, Git revision c6dac65e3174dea79df54ce6d56f3e98bc060ecc (3647 commits) x86_64 hpack-0.14.0
2016-08-02 15:59:34.979211: [debug] Checking for project config at: D:\stack.yaml @(stack_9kewhubnl5WIl89fhd1ea2:Stack.Config src/Stack\Config.hs:811:9)
2016-08-02 15:59:34.980211: [debug] No project config file found, using defaults. @(stack_9kewhubnl5WIl89fhd1ea2:Stack.Config src/Stack\Config.hs:833:13)
2016-08-02 15:59:34.984212: [debug] Creating process: D:\usr\git\bin\git.exe fetch --tags --depth=1 @(stack_9kewhubnl5WIl89fhd1ea2:System.Process.Run src/System\Process\Run.hs:102:5)
Fetching package index ...fatal: Not a git repository (or any of the parent directories): .git
2016-08-02 15:59:35.011213: [warn] Process exited with ExitFailure 128: D:\usr\git\bin\git.exe fetch --tags --depth=1 @(stack_9kewhubnl5WIl89fhd1ea2:Stack.PackageIndex src/Stack\PackageIndex.hs:262:17)
Failed to fetch package index, retrying.
removeDirectoryRecursive: permission denied (Отказано в доступе.)

```

Blaisorblade:
Fix #2426.
I've left tests alone as the problem doesn't appear there (and the testing
environment is more under our control).
I have blindly modified *all* uses in the code, because
* I don't know which one triggers the issue
* in principle it might affect many/all calls
* while this *might* trigger regressions, I'm not sure which ones concretely.
  Unless the user is an admin and somehow this manages to remove non-Stack files
  that shouldn't be removed; this sounds unlikely (we'd have reports where this
  is attempted and fails) but I can't exclude it yet without at least closer
  review.

* [ ] Tested: not yet, only compiles. EDIT: running integration tests locally with `stack build --flag stack:integration-tests --test stack:stack-integration-test`.

### DIFF
--- a/src/Path/Extra.hs
+++ b/src/Path/Extra.hs
@@ -14,6 +14,7 @@ module Path.Extra
   ,pathToByteString
   ,pathToLazyByteString
   ,pathToText
+  ,removePathForcibly
   ,tryGetModificationTime
   ) where
 
@@ -28,6 +29,7 @@ import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy.Char8 as BSL
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
+import qualified System.Directory as D
 import qualified System.FilePath as FP
 
 -- | Convert to FilePath but don't add a trailing slash.
@@ -122,3 +124,14 @@ pathToText = T.pack . toFilePath
 
 tryGetModificationTime :: MonadIO m => Path Abs File -> m (Either () UTCTime)
 tryGetModificationTime = liftIO . tryJust (guard . isDoesNotExistError) . getModificationTime
+
+-- Copied from path-io
+liftD :: MonadIO m
+  => (FilePath -> IO a) -- ^ Original action
+  -> Path b t          -- ^ 'Path' argument
+  -> m a               -- ^ Lifted action
+liftD m = liftIO . m . toFilePath
+{-# INLINE liftD #-}
+
+removePathForcibly :: MonadIO m => Path b Dir -> m ()
+removePathForcibly = liftD D.removePathForcibly
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -59,7 +59,7 @@ import           Distribution.System            (OS (Windows),
 import qualified Distribution.Text as C
 import           Path
 import           Path.CheckInstall
-import           Path.Extra (toFilePathNoTrailingSep, rejectMissingFile)
+import           Path.Extra (toFilePathNoTrailingSep, rejectMissingFile, removePathForcibly)
 import           Path.IO hiding (findExecutable, makeAbsolute, withSystemTempDir)
 import qualified RIO
 import           Stack.Build.Cache
@@ -1575,7 +1575,7 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} installedMap
             -- require it, to reduce space usage in tmp (#3018).
             TTIndex{} -> do
                 let remaining = filter (\(ActionId x _) -> x == taskProvides) (Set.toList acRemaining)
-                when (null remaining) $ removeDirRecur pkgDir
+                when (null remaining) $ removePathForcibly pkgDir
             _ -> return ()
 
         return mpkgid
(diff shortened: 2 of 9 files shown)


Blaisorblade:
@mgsloan What do you think the bar should be here? Figuring out this is safe is much easier than figuring out if this actually fixes the problem or writing an integration test for it.

mgsloan:
Hmm, yeah, not sure what to do about that.  I think it would be best to only do this where necessary - seems good to err on the side of safety.

I'll defer to the judgement of other maintainers here.

mihaimaruseac:
I always prefer integration test but I feel that here the effort needed to write them is much higher than just checking that the code is safe and correct. That being said, I'll need more time to check the code, but maybe other maintainer is faster.

mpilgrem:
@Blaisorblade, By way of a test on Windows 7, I build `Version 1.8.0, Git revision e9d65e9432fad03adfbe7ae09792f836fb9f8b6d (6083 commits) x86_64 hpack-0.2
9.0` from branch `fix-2426` and then used that version to `stack upgrade --git`.

It first failed at the following point (I've included a little history from the console but deleted the 'Progress 28/37' type lines):
````
semigroupoids-5.2.2: configure
persistent-2.7.1: copy/register
profunctors-5.2.2: copy/register
semigroupoids-5.2.2: build
hpack-0.29.0: configure
hpack-0.29.0: build
persistent-sqlite-2.6.4: configure
semigroupoids-5.2.2: copy/register
persistent-sqlite-2.6.4: build
persistent-template-2.5.4: configure
persistent-template-2.5.4: build
persistent-template-2.5.4: copy/register
persistent-sqlite-2.6.4: copy/register
hpack-0.29.0: copy/register
Progress 32/37
removePathForcibly:removePathForcibly:removePathForcibly:removePathForcibly:removePathForcibly:removePathForcibly:DeleteFile "C:\\Users\\mpilgre\\AppData\\Local\\Temp\\stack12408\\semigroupoids-5.2.2\\.stack-work\\dist\\5c8418a7\\setup\\setup.exe": permission denied (The process cannot access the file because it is being used by another process.)
````

Blaisorblade:
@mpilgrem Wow, that is *useful* unlike all the other messages I've seen (or was I careless before, or is it a new bug?). Is that repeatable?

To explain my first hypothesis, remember that building a package compiles a builder program (`Setup.exe`) that actually does the build. Apparently, semigroupoids' builder is trying to remove itself! It could also be that some while the builder is running some *other* process is trying to remove it, but that seems less likely.

At first sight, little seems special about semigroupoids' [builder](https://github.com/ekmett/semigroupoids/blob/2cd36c9e53e3df85d1c94834642731fd3865dafd/Setup.lhs), except that it involves doctests (I hope that's not the problem) and is not the standard one! The standard builder uses a shared precompiled executable, which probably doesn't sit in the folder being removed, avoiding that failure.

IIUC, there's essentially no way whatsoever to remove a file in use with Win32 APIs (Windows can support Unix APIs so I suspect exceptions, but I guess we should ignore those). So we shouldn't try! But I don't understand enough the architecture to figure out what's causing this. @mgsloan who should this escalate to?

mpilgrem:
I tried `stack upgrade --git` one more time, and it failed at the same point for the same reason.

I then checked out branch `master` (actually, my local copy was a little out of date - commits up to 25 July only) and tried `stack install` and it built* to completion, first time - yielding `stack --version` `Version 1.8.0, Git revision 94d302c40d3d08dcaa0213f3f1ffdc1384842714 (6077 commits) x86_64 hpack-0.29.0`.

(* As an aside: building generates lots (10s, if not 100s) of these warnings at certain points: `ghc.EXE: addLibrarySearchPath: C:\Users\mpilgre\AppData\Local\Programs\stack\x86_64-windows\msys2-20150512\mingw64\bin (Win32 error 3): The system cannot find the path specified.` However, I think that is quite a different problem - likely peculiar to me - and they do not prevent building. I do not have `..\stack\x86_64-windows\msys2-20150512` but have `..\stack\x86_64-windows\msys2-20180531`.)

PR #4197, opened 2018-08-01
Implement loading from url with caching in SQLite

qrilka:

Now `stack build` fails on

```
warningsParserHelper = undefined
```

### DIFF
--- a/subs/pantry/src/Pantry.hs
+++ b/subs/pantry/src/Pantry.hs
@@ -100,6 +100,7 @@ module Pantry
 import RIO
 import qualified RIO.Map as Map
 import qualified RIO.ByteString as B
+import qualified RIO.ByteString.Lazy as LB
 import qualified RIO.Text as T
 import qualified RIO.List as List
 import qualified RIO.FilePath as FilePath
@@ -120,6 +121,8 @@ import qualified Hpack.Config as Hpack
 import RIO.Process
 import qualified Data.Yaml as Yaml
 import Data.Aeson.Extended (WithJSONWarnings (..), Value)
+import Network.HTTP.StackClient
+import Network.HTTP.Types (ok200)
 
 withPantryConfig
   :: HasLogFunc env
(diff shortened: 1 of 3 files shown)


subs/pantry/src/Pantry.hs
```diff
@@ -626,7 +628,32 @@ loadFromURL
   => Text -- ^ url
   -> Maybe BlobKey
   -> RIO env ByteString
-loadFromURL = undefined
+loadFromURL url Nothing = do
+  loadWithCheck url $ \_ -> return ()
+loadFromURL url (Just bkey@(BlobKey sha size)) = do
+  mcached <- withStorage $ loadURLBlob url bkey
```
snoyberg: This isn't what I had in mind. It should be using the URL-based cache for the `loadFromURL url Nothing` case, _not_ this case. For this case: it should only ever load by SHA.

qrilka: Please check out updated version

subs/pantry/src/Pantry.hs
```diff
@@ -626,7 +628,35 @@ loadFromURL
...
+loadFromURL url (Just bkey@(BlobKey sha size)) = do
+  mcached <- withStorage $ loadBlob bkey
+  case mcached of
+    Just bs -> return bs
+    Nothing -> loadWithCheck url $ \bs -> do
+      let blobSha = mkStaticSHA256FromBytes bs
+          blobSize = FileSize $ fromIntegral $ B.length bs
+      when (blobSha /= sha || blobSize /= size) $
+        throwIO $ BadKey . utf8BuilderToText $
+          "Expected SHA/size : " <> display sha <> "/" <> display size <>
```
snoyberg: Instead of including a `Text` here, the normal approach in the library would be to include the expected and actual `BlobKey`s in the data constructor, and doing the rendering in the `Show` instance.

subs/pantry/src/Pantry.hs
```diff
@@ -626,7 +628,35 @@ loadFromURL
...
+          ". Actual SHA/size: " <> display blobSha <> "/" <> display blobSize
+
+loadWithCheck
+  :: (HasPantryConfig env, HasLogFunc env)
+  => Text -- ^ url
+  -> (ByteString -> RIO env ()) -- ^ function to check downloaded blob
+  -> RIO env ByteString
+loadWithCheck url checkResponseBody = do
+  req <- parseRequest $ T.unpack url
+  res <- httpLbs req
```
snoyberg: We should probably check that this is a `200 OK` status code.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -437,6 +438,15 @@ instance Display PantryException where
     ":\n" <>
     displayShow e
 
+data PantryDownloadException = BadKey !Text
```
snoyberg: I'd recommend just putting this in with `PantryDownloadException` as another data constructor.

qrilka: Do you mean as another constructor into `PantryException`?

snoyberg: Yes

subs/pantry/src/Pantry.hs
qrilka: and now it fails because of 404 as  https://raw.githubusercontent.com/commercialhaskell/stack-templates/master/lts/10/1.yaml doesn't appear to be a valid snapshot URL

snoyberg: Typo on my part, it should be using `stackage-templates`.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -338,6 +340,11 @@ parsePackageIdentifierRevision t = maybe (throwM $ PackageIdentifierRevisionPars
       let (x, y) = T.break (== ':') t'
        in (x, ) <$> T.stripPrefix ":" y
 
+data Mismatch a = Mismatch
```
snoyberg: This could be nice for cleaning up a number of the exceptions.

PR #4220, opened 2018-08-10
[WIP] `stack freeze` command

qrilka:

Please also shortly describe how you tested your change:

- Integration test in `test/integration/tests/4220-freeze-command`
- run `stack freeze` on my `xlsx` package


qrilka:
@snoyberg could you have a glance on this initial attempt?
Should I add changelog entry in this PR? And what about documentation?
I guess some test is required?

snoyberg:
> could you have a glance on this initial attempt?

Reviewing now, I'll leave comments inline

> Should I add changelog entry in this PR? And what about documentation?

Yes and yes. And that reminds me that I need to add an issue about fixing up all the documentation on the pantry branch.

> I guess some test is required?

That would be ideal, yes. Ping me if you'd like some help on how to use the integration test suite.

src/main/Main.hs
```diff
@@ -63,6 +63,7 @@ import           Stack.Dot
 import           Stack.GhcPkg (findGhcPkgField)
 import qualified Stack.Nix as Nix
 import           Stack.FileWatch
+import           Stack.Freeze
```
snoyberg: I think you forgot to `git add` the new modules.

qrilka: Ouch, forgot the main part of it

src/Stack/Freeze.hs
```diff
@@ -0,0 +1,52 @@
...
+    { freezeTargets :: [Text]
+    -- ^ stack TARGETs to trace dependencies for
+    , freezeFlags :: !(Map (Maybe PackageName) (Map FlagName Bool))
+    -- ^ Flags to apply when calculating dependencies
+    , freezeExtraTargets :: !ExtraTargets
+    -- ^ Like the "--test" / "--bench" flag for build, affects the meaning of 'dotTargets'. 
+    }
+
+freeze :: HasEnvConfig env => FreezeOpts -> RIO env ()
+freeze freezeOpts = do
```
snoyberg: Instead of working at the fully-parsed level of a source map, it would be better to work on the original `stack.yaml` file and any custom snapshots it refers to. I can step through that with you on a call if that would be helpful.

qrilka: I just couldn't find that yet and checking out code of `stack dot` was easy

qrilka: After having a testcase changing implementation should be even easier

qrilka:
@snoyberg please check out new version including a test. Will add documentation file a bit later.

qrilka:
This version outputs pinned version of `stack.yaml` or `snapshot.yaml` in case if there are any unpinned dependencies. Maybe it's better to just make it stateless and produce output in any case? Empty output in very confusing. Or we could report something like "It is fine" :fire: :dog2: in this case

test/integration/tests/4220-freeze-command/Main.hs
```diff
@@ -0,0 +1,26 @@
+import Control.Monad (unless)
+import StackTest
+
+main :: IO ()
+main = do
+  stackCheckStdout ["freeze"] $ \stdOut -> do
+    let expected = unlines
+          [ "packages:"
```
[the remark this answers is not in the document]
qrilka: I know but `unlines` looks a bit more readable in this case imo

qrilka:
@snoyberg I've resolved conflicts, added more output and also wrote down a basic doc. Please take a look

PR #4222, opened 2018-08-11
Fix #4205 (Step 0): Annotate with Style, not [SGR]

### LINKED ISSUE #4205: Allow user specification of stack's colours

mpilgrem:
These changes do not affect the existing functionality of stack but provide a foundation for the next steps of suggestion #4205.

After the changes, Stack's styles are implemented as type `Style` in new module `Stack.PrettyPrint`, which also defines a `StyleSpec` as `(Text, [SGR])` (the first item providing the name of the key) and `Styles` as `Data.Array.IArray.Array Style StyleSpec`. Package `array` is added as a dependency in `package.yaml`, accordingly.

Given the use of `cyan` in the original instance of `Display` for `(PackageName, NamedComponent)`, a new _style_ `PkgComponent` (with key `package-component`) is added to `Style`.

Given the use of `green` in `displayMilliseconds`, it is assumed the intended style was `Good`.

Functions like `StyleWarning`, `StyleError` can be, and are, replaced throughout by `style :: Style -> ...` (eg `style Warning`, `style Error`).

The type `Runner` is modified to include field `runnerStyles :: !Styles`. New module `Stack.DefaultStyles` provides `defaultStyles`. Consequent changes are made to the plumbing.

`AnsiDoc` is replaced by `type StyleDoc = Doc StyleAnn`, throughout. In module `Stack.PrettyPrint`, use is made of the `HasRunner` environment and it is there that `SimpleDoc StyleDoc` is translated to `SimpleDoc AnsiDoc`.

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md _There is no change for users_

Tested by building on Windows 10 and using.

### DIFF
--- a/src/Stack/Types/PrettyPrint.hs
+++ b/src/Stack/Types/PrettyPrint.hs
@@ -1,0 +1,54 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-|
+For the most part, the data constructors of 'Style' do not clash with other
+names. When they do, import the module qualified. For example:
+
+> import qualified Stack.Types.PrettyPrint as PP
+-}
+module Stack.Types.PrettyPrint
+  (
+    Style (..)
+  , Styles
+  , StyleSpec
+  ) where
+
+import Data.Array.IArray (Array)
+import Data.Ix (Ix)
+import Stack.Prelude
+import System.Console.ANSI.Types (SGR)
+
+-- |A style of stack's output.
+data Style
+  = Error    -- Should be used sparingly, not to style entire long messages. For
+             -- example, it's used to style the "Error:" label for an error
+             -- message, not the entire message.
+  | Warning  -- Should be used sparingly, not to style entire long messages. For
+             -- example, it's used to style the "Warning:" label for an error
+             -- message, not the entire message.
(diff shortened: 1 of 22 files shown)


src/Stack/Types/PrettyPrint.hs
```diff
@@ -0,0 +1,54 @@
...
+  | Target   -- TODO: figure out how to describe this
+  | Module   -- Style as a module name
+  | PkgComponent    -- Style used to highlight the named component of a package.
+  deriving (Bounded, Enum, Eq, Ix, Ord, Show)
+
+-- |The first style overrides the second.
+instance Semigroup Style where
+  s <> _ = s
+
+-- |A style specfication, pairing its \'key\' with the corresponding list of
```
mattaudesse: Small typo: `specfication`

mpilgrem:
@malteneuss, thank you for your review. I've corrected the things that you identified. To answer your question above: I'm new to this, but it seemed sensible to take what I suggested in #4205 step-by-step rather than try to present it 'all together'. However, the largely independent steps all build on this common foundation. I am currently thinking about 'step 5' (the suggested `stack ls stack-colors` command).

mattaudesse:
Thanks @mpilgrem!

PR #4230, opened 2018-08-13
Fix #4205 (Step 5): Add `stack ls stack-colors`

### LINKED ISSUE #4205: Allow user specification of stack's colours

mpilgrem:
This builds on step 0 (commit https://github.com/commercialhaskell/stack/commit/68718a784a0eec52968540ea3de3885f1ae707a0). It adds `stack ls stack-colors` to the `stack ls` command, in module `Stack.Ls`. The command's functionality is provided by `listStylesCmd`. The rest of the changes add the `optparse-applicative` plumbing.

The `ls` command documentation, in module `Main`, is updated.

`ChangeLog.md` is updated.

This has been tested by building - `stack test` - on Windows 10 and using the new function with its various options.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,10 @@ Other enhancements:
   plan construction errors much faster, and avoid some unnecessary
   work when only building a subset of packages. This is especially
   useful for the curator use case.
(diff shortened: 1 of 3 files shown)


mpilgrem:
Example usage of the new command:
image

PR #4231, opened 2018-08-13
Fix #4205 (Step 3): Add `--stack-colors=STYLES`

### LINKED ISSUE #4205: Allow user specification of stack's colours

mpilgrem:
This builds on step 0 (commit https://github.com/commercialhaskell/stack/commit/68718a784a0eec52968540ea3de3885f1ae707a0). It adds the `--stack-colors=STYLES` global option. The functionality is primarily implemented via `readStyles` in module `Stack.Config`. `GlobalOptsMonoid` is extended to include `globalMonoidStyles`; a list of `(Style, StyleSpec)` that are then used to update the array that is `defaultStyles`.

24-bit colour is now a possibility and so package `colour` is added as a dependency in `package.yaml`.

`ChangeLog.md` is updated, accordingly.

Tested by building - `stack test` - on Windows 10 and using the new option.

### DIFF
--- a/src/Stack/Types/StylesUpdate.hs
+++ b/src/Stack/Types/StylesUpdate.hs
@@ -1,0 +1,139 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Stack.Types.StylesUpdate
+  (
+    StylesUpdate (..)
+  , parseStylesUpdateFromString
+  ) where
+
+import Data.Array.IArray (assocs)
+import Data.Colour.SRGB (Colour, sRGB24)
+import Data.Text as T (pack)
+import Data.Text (Text)
+import Stack.DefaultStyles (defaultStyles)
+import Stack.Prelude
+import Stack.Types.PrettyPrint (Style, StyleSpec)
+import System.Console.ANSI.Types (BlinkSpeed (..), Color (..),
+        ColorIntensity (..), ConsoleIntensity (..), ConsoleLayer (..), SGR (..),
+        Underlining (..))
+
+-- |Updates to 'Styles'
+newtype StylesUpdate = StylesUpdate { stylesUpdate :: [(Style, StyleSpec)] }
+  deriving (Eq, Show)
+
+-- |Parse a string that is a colon-delimited sequence of key=value, where 'key'
+-- is a style name and 'value' is a semicolon-delimited list of 'ANSI' SGR
+-- (Select Graphic Rendition) control codes (in decimal). Keys that are not
+-- present in 'defaultStyles' are ignored. Items in the semicolon-delimited
+-- list that are not recognised as valid control codes are ignored.
+parseStylesUpdateFromString :: String -> StylesUpdate
+parseStylesUpdateFromString s = StylesUpdate $ mapMaybe process table
+ where
+  table = do
+    w <- split ':' s
+    let (k, v') = break (== '=') w
+    case v' of
+      '=' : v -> return (T.pack k, parseCodes v)
+      _ -> []
+
+  process :: StyleSpec -> Maybe (Style, StyleSpec)
+  process (k, sgrs) = do
+    style <- lookup k styles
+    return (style, (k, sgrs))
+
+styles :: [(Text, Style)]
+styles = map (\(s, (k, _)) -> (k, s)) $ assocs defaultStyles
+
+parseCodes :: String -> [SGR]
+parseCodes [] = []
+parseCodes s = parseCodes' c
+ where
+  s' = split ';' s
+  c :: [Word8]
+  c = mapMaybe readMaybe s'
+
+parseCodes' :: [Word8] -> [SGR]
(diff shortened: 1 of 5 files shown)


mpilgrem:
Example of usage of new option:
image

mpilgrem:
@RoelofWobben, that failure was a matter of 'style' and is now fixed - `hlint` wanted some unnecessary brackets removed and I've removed them. The current red cross on the Travis build is misleading: the GHC 8.2.2 job (which worked before) fell over after 29 seconds due to a glitch somewhere else. I do not know how one gets Travis to 'try again', when it falls over. The pedantic GHC 8.2.2 jobs passed both times. (The optional macOS jobs always fail because Travis does not allow sufficient time for a build.)

mpilgrem:
I have now also tested successfully on Windows 7 (in Command Prompt - dull, no colour - and using Cygwin/mintty). (Note: in Cygwin, quotes are needed around the style specification. For example:
````
stack --stack-colors="good=91;104:file=30;103" ghci
````
The parser of the option is tolerant of unexpected things in the specification; it strips them out. So the following is the equivalent of the above:
````
stack --stack-colors="good=91;dummy;104:dummy=dummy:file=30;103" ghci
````

**Comment by mpilgrem** [comment]
I have now extended the help to warn about shells that use a semicolon as a command separator.

**Comment by RoelofWobben** [comment]
@snoyberg  can you also review this PR.

**Comment by mpilgrem** [comment]
I would add: I've weighed up (1) the 'pain' of using semicolons 'at the command line' on `bash` against (2) the benefit of using a styles specification format that is the same as that already used by GHC and GCC and I think that the benefit of consistency outweighs the pain. In part, that is because the real benefit for stack users will not be (1) 'step 3' (the `--stack-colors` option, at the command line) but (2) (to come, soon) 'step 4' (the `stack-colors:` setting in `config.yaml`), where the use of semicolons will not be problematic.

**Comment by mpilgrem** [comment]
@RoelofWobben, I apologise for a further revision, but working on 'step 4' has made me realise that it would be sensible to refactor this commit to create `newtype StylesUpdate = StylesUpdate { stylesUpdate :: [(Style, StyleSpec)] }` and put it and `parseStylesUpdateFromString` in new module `Stack.Types.StylesUpdate` (a new module is needed to avoid some circularity).

File: src/Stack/Types/StylesUpdate.hs [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+codeToColor :: Word8 -> Color
+codeToColor c
+  | c == 0 = Black
+  | c == 1 = Red
+  | c == 2 = Green
+  | c == 3 = Yellow
+  | c == 4 = Blue
+  | c == 5 = Magenta
+  | c == 6 = Cyan
+  | c == 7 = White
```
mihaimaruseac: I would replace this function with a `fromEnum` if possible to not have magic constants. `Color` is already an instance of `Bounded` and `Enum` in the proper order so we would only need to check that `c` is within bounds and then call `fromEnum`.

src/Stack/Types/StylesUpdate.hs
```diff
@@ -0,0 +1,119 @@
...
+  | c >= 40 && c <= 47 =
+    (Just $ SetColor Background Dull $ codeToColor (c - 40), cs)
+  | c == 48 = case codeToRGB cs of
+    (Nothing, cs') -> (Nothing, cs')
+    (Just color, cs') -> (Just $ SetRGBColor Background color, cs')
+  | c >= 90 && c <= 97 =
+    (Just $ SetColor Foreground Vivid $ codeToColor (c - 90), cs)
+  | c >= 100 && c <= 107 =
+    (Just $ SetColor Background Vivid $ codeToColor (c - 100), cs)
+  | otherwise = (Nothing, cs)
```
mihaimaruseac: This has a lot of magical constant but I guess there's nothing we can do here to make it more streamlined.

We can though add a link to a specification of the format so that these magic numbers have some meaning attached to them and anyone reading the code will get what they mean.

mpilgrem: I have added a comment explaining the codes (so they do not appear to be 'magic') and referencing both the underlying 'ANSI' standards and the Wikipedia article that summarises them.

src/Stack/Types/StylesUpdate.hs
mpilgrem: I have added a comment explaining the codes (so they do not appear to be 'magic') and another explaining why `toEnum` is not used (because `ansi-terminal` does not guarantee the order of the data constructors of type `Color`).

mpilgrem:
@mihaimaruseac, I think I have made changes responsive to your review.

mihaimaruseac:
Looks good now, thanks

src/Stack/Types/StylesUpdate.hs
mihaimaruseac: That's unfortunate, but I guess we can live with that.

PR #4233, opened 2018-08-15
Remove redundant code

sol:


src/test/Spec.hs
```diff
@@ -1,2 +1 @@
-{-# LANGUAGE NoImplicitPrelude #-}
```
sol: ^ this never had any effect

src/test/Test.hs
```diff
@@ -1,5 +0,0 @@
-import Test.Hspec (hspec)
-import Spec (spec)
-
-main :: IO ()
-main = hspec spec
```
sol: This is just a default test driver, so we can use the generated one.

RoelofWobben:
@borsboom  @snoyberg  can you take a look at it

PR #4237, opened 2018-08-16
[pantry] curator CLI

bergmark:
Currently runs previous steps before the selected one. Should it behave like that or should only the chosen step be performed, or do we want to toggle that with a flag?

Ping @snoyberg


snoyberg:
I think it should only run one step at a time, otherwise the ability to edit an intermediate file will be lost.

I'm also not 100% sold on optparse-generic, since we won't have full control to add flags exactly as desired, but I'm OK leaving it until we run into an actual limitation.

Issue #3515, opened 2017-10-25
stack install fails on FreeBSD 12.0-Current


noinia:
### General summary/comments (optional)

I'm having trouble building/installing packages with stack (1.5.1, installed using the official port) on FreeBSD 12.0-Current. Essentially, building any package fails with:

```
    Cabal-simple_mPHDZzAJ_1.24.2.0_ghc-8.0.2: No cabal file found.
    Please create a package description file <pkgname>.cabal
```
It does not seem to matter whether I try to build a local project using 'stack build', or install a particular package with 'stack install <package>'. Nor does it seem matter which stackage release/version of ghc I use: I've tried with lts-9.* and ghc-8.0.2 (installed by stack), lts-* and ghc installed using the port, and the nightly snapshots with ghc 8.2.1 (installed by stack). Any idea what is going on here?

### Steps to reproduce

```
rm ~/.stack
stack setup
stack install fgl --verbose
```

### Expected

stack installs the package fgl.

### Actual

See the full output [here](https://gist.github.com/noinia/ef58812e3faf995e25b7a2a00353691b)

### Stack version

```
frank@UU# stack --version
Version 1.5.1 x86_64
Compiled with:
<see full output linked above>
```

### Method of installation

* Via the freebsd port in devel/stack 

```
frank@UU# pkg info stack
stack-1.5.1
Name           : stack
Version        : 1.5.1
Installed on   : Tue Oct 24 16:00:14 2017 CEST
Origin         : devel/stack
Architecture   : FreeBSD:12:amd64
Prefix         : /usr/local
Categories     : haskell devel
Licenses       : BSD3CLAUSE
Maintainer     : tobik@FreeBSD.org
WWW            : http://www.haskellstack.org/
Comment        : Cross-platform program for developing Haskell programs
Shared Libs required:
	libgmp.so.10
	libiconv.so.2
	libcharset.so.1
Annotations    :
	repo_type      : binary
	repository     : local
Flat size      : 49.2MiB
Description    :
Stack is a cross-platform program for developing Haskell projects.
It is aimed at Haskellers both new and experienced. 

WWW: http://www.haskellstack.org/
```
system:

```
frank@UU# uname -a 
FreeBSD UU.FStaals.net 12.0-CURRENT FreeBSD 12.0-CURRENT #0 r324496: Wed Oct 11 09:22:17 CEST 2017     root@UU.FStaals.net:/usr/obj/usr/src/sys/GENERIC  amd64
```

[...]

PR #4238, opened 2018-08-17
Pull bindings-uname package into Stack and fix it to work with FreeBSD.

arrowd:
Due to https://github.com/commercialhaskell/stack/issues/3515 I wasn't able to test if Stack builds with this change, but changes made for `bindings-uname` do work.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -128,7 +128,6 @@ when:
     build-tools:
     - hsc2hs
     dependencies:
-    - bindings-uname
     - unix
 library:
   source-dirs: src/
@@ -273,6 +272,7 @@ library:
       source-dirs: src/windows/
     else:
       source-dirs: src/unix/
+      c-sources: src/unix/cbits/uname.c
 executables:
   stack:
     main: Main.hs
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -101,7 +101,7 @@ import              RIO.Process
 import              Text.Printf (printf)
 
 #if !WINDOWS
-import              Bindings.Uname (uname, release)
+import              System.Uname (uname, release)
 import              Data.List.Split (splitOn)
 import              Foreign.C (throwErrnoIfMinus1_, peekCString)
 import              Foreign.Marshal (alloca)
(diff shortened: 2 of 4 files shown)


RoelofWobben:
@mgsloan  can you take a look at this PR ?

mgsloan:
Looks like a reasonable change to me.  I think it would be better to fix the `bindings-uname` package and update the one used by stack, though.

arrowd:
I've never worked with Hackage. How can I update `bindings-uname` there?

dbaynard:
[It](https://hackage.haskell.org/bindings-uname) hasn't been updated since 2009, and there's no website, though the git repo still works. Ping the maintainer an email?

arrowd:
> Ping the maintainer an email?

Done that 8 days ago.

dbaynard:
Yeah, I _think_ this package needs a new maintainer. If you're willing to do it, there's a process on the hackage site.

The upshot of that is, I think we should merge the PR (subject to code review) and then when there's a new maintainer, migrate to the new code.

Which were the changes you made from the original `bindings-uname`?

arrowd:
> Which were the changes you made from the original bindings-uname?

Here is the patch: http://arrowd.name/uname.patch

dbaynard:
Thanks. I'd ideally like to see that as two separate commits.

> Due to #3515 I wasn't able to test if Stack builds with this change

Is this because there's no binary version of stack that works and so you can't bootstrap to test the changes? We should be getting a static build of stack again, soon — would that be helpful on that front?

It might be good to prioritize the fix to #3515 at the same time so it can be tested. That issue mentions the fix by @Blaisorblade for OpenBSD in 3cd92c8. Would it be possible to do a similar thing here?

arrowd:
It has been tested by CI, it's all green. Doesn't that count?

dbaynard:
It does count, although the integration tests aren't run by CI. I've asked maintainers to comment, but I think this will be OK to merge.

borsboom:
I think it's OK to merge.

arrowd:
I just found out that this PR also didn't make it into `stable`, so on this branch `sysRelease` still return garbage :<

I hope, this will be included in the next Stack release.

PR #4239, opened 2018-08-17
Use a real nonempty list

NorfairKing:
This was prompted by the comment that said `-- non-empty list`.

### DIFF
--- a/subs/pantry/src/Pantry/Types.hs
+++ b/subs/pantry/src/Pantry/Types.hs
@@ -91,6 +91,8 @@ import Data.Aeson (ToJSON (..), FromJSON (..), withText, FromJSONKey (..))
 import Data.Aeson.Types (ToJSONKey (..) ,toJSONKeyText, Parser)
 import Data.Aeson.Extended
 import Data.ByteString.Builder (toLazyByteString, byteString, wordDec)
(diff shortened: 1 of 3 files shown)


subs/pantry/src/Pantry/Types.hs
```diff
@@ -1272,7 +1274,8 @@ parseSnapshot mdir = withObjectWarnings "Snapshot" $ \o -> do
   snapshotHidden <- unCabalStringMap <$> (o ..:? "hidden" ..!= Map.empty)
   snapshotGhcOptions <- unCabalStringMap <$> (o ..:? "ghc-options" ..!= Map.empty)
   pure $ do
-    snapshotLocations <- fmap concat $ mapM (resolvePackageLocationImmutable mdir) unresolvedLocs
```
NorfairKing: This was `[NonEmpty a]`, so there is no way to concat to a `NonEmpty`

dbaynard:
Hi @NorfairKing, I'll get to this in the next day or so.

dbaynard:
…actually, I just checked out the travis failure. It seems there are more places in the code that are affected?

snoyberg:
I ended up implementing this myself, forgetting that there was a PR for this already. Thanks for pointing it out @NorfairKing!

PR #4240, opened 2018-08-17
Turned unit tests into property tests

NorfairKing:
Turned unit tests into property tests, they all still pass and are not significantly slower.

### DIFF
--- a/subs/pantry/package.yaml
+++ b/subs/pantry/package.yaml
@@ -83,3 +83,4 @@ tests:
     - hspec
     - exceptions
     - hedgehog
+    - QuickCheck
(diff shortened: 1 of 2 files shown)


RoelofWobben:
sorry. there are some failing test on travis. Could you take a look at it

NorfairKing:
@RoelofWobben It looks like these are hackage/sqlite related. @snoyberg It looks like the tests are flaky. Do you have any idea why?

snoyberg:
I've never seen these test failures before, interesting. It's definitely not related to this PR, we can continue investigating after merge since this is just a feature branch.

PR #4242, opened 2018-08-19
Add the ddump-dir option/config value #4225

### LINKED ISSUE #4225: Can't display -ddump-splices result.
```shell
$ stack build --ghc-options='-ddump-splices'
```

I want to display on the standard output, not to save to a file.

### Question

Why does the [extraBuildOptions](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Build/Execute.hs#L1948) function specify the `" -ddump-hi -ddump-to-file"` option by default?

```hs
extraBuildOptions wc bopts = do
    colorOpt <- appropriateGhcColorFlag
    let ddumpOpts = " -ddump-hi -ddump-to-file"
        optsFlag = compilerOptionsCabalFlag wc
        baseOpts = ddumpOpts ++ maybe "" (" " ++) colorOpt
    if toCoverage (boptsTestOpts bopts)
      then do
        hpcIndexDir <- toFilePathNoTrailingSep <$> hpcRelativeDir
        return [optsFlag, "-hpcdir " ++ hpcIndexDir ++ baseOpts]
      else
        return [optsFlag, baseOpts]
```

If remove this option, I think it will work as expected.

### Stack version

```
$ stack --version
Version 1.10.0, Git revision 3b61926935c28efd10c8e6870f927a7d8eb5523c (dirty) (6119 commits) x86_64 hpack-0.29.6
```

waddlaw:
Fixes #4225

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -1550,7 +1550,7 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} installedMap
                 parentDir <- parent <$> parseRelDir src
                 destBaseDir <- (ddumpDir </>) <$> stripProperPrefix distDir parentDir
                 -- exclude .stack-work dir
-                when (not (".stack-work" `isInfixOf` (toFilePath destBaseDir))) $ do
+                unless (".stack-work" `isInfixOf` toFilePath destBaseDir) $ do
                   ensureDir destBaseDir
                   src' <- parseRelFile src
                   copyFile src' (destBaseDir </> filename src'))
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -1534,26 +1534,26 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} installedMap
             when hasLibrary $ cabal KeepTHLoading ["register"]
 
         -- copy ddump-* files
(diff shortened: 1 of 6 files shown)


waddlaw:
@snoyberg this should be all set now :)
Is this alright? Could you please confirm it?

src/Stack/Build/Execute.hs
```diff
@@ -1532,6 +1533,28 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} installedMap
                 _ -> return ()
             when hasLibrary $ cabal KeepTHLoading ["register"]
 
+        -- copy ddump-* files
+        let enableDdumpDir = isJust $ boptsDdumpDir eeBuildOpts
+            ddumpPath      = maybe "" T.unpack $ boptsDdumpDir eeBuildOpts
+        when (buildingFinals && enableDdumpDir && not (null ddumpPath)) $ do
```
snoyberg: It looks like a pattern match would be easier to read here, e.g.:

```haskell
case T.unpack <$> boptsDdumpDir eebuildOpts of
  Just ddumpPath | buildingFinals && enableDdumpDir && not (null ddumpPath) -> do ...
  _ -> pure ()
```

waddlaw: That makes sense. That's true :)

snoyberg:
Thanks!

Issue #3922, opened 2018-03-14
Build tools not considered when constructing SourceMap


snoyberg:
Repro: create a directory with the following two files:

```yaml
# package.yaml
name: foo

dependencies:
- base
- haskell-src-exts

library: {}
```

```yaml
# stack.yaml
resolver: lts-11.0

extra-deps:
- happy-1.19.9@rev:2
```

NOTE: If someone knows of a package in a Stackage snapshot that uses either happy or alex, and compiles quicker than `haskell-src-exts`, please let me know, we can make this repro easier to trigger.

Next: observe the following session:

```
$ stack build
happy-1.19.9: configure
happy-1.19.9: build
happy-1.19.9: copy/register
haskell-src-exts-1.20.2: configure   
haskell-src-exts-1.20.2: build       
haskell-src-exts-1.20.2: copy/register
foo-0.0.0: configure (lib)           
Configuring foo-0.0.0...             
foo-0.0.0: build (lib) 
Preprocessing library for foo-0.0.0..
Building library for foo-0.0.0..
[1 of 1] Compiling Paths_foo        ( .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/autogen/Paths_foo.hs, .stack-work/dist/x86_64-osx/Cabal-2.0.1.0/build/Paths_foo.o )
foo-0.0.0: copy/register
Installing library in /Users/michael/Desktop/stack-build-tool-local/.stack-work/install/x86_64-osx/lts-11.0/8.2.2/lib/x86_64-osx-ghc-8.2.2/foo-0.0.0-J3VouLg1fVTLZlRnB5NxZt
Registering library for foo-0.0.0..
Completed 3 action(s). 
Michaels-MacBook-Pro-3:stack-build-tool-local michael$ stack build --dry-run
Would unregister locally:
foo-0.0.0
haskell-src-exts-1.20.2

Would build:
foo-0.0.0: database=local, source=/Users/michael/Desktop/stack-build-tool-local/, after: haskell-src-exts-1.20.2
haskell-src-exts-1.20.2: database=local, source=package index

No executables to be installed.
```

Expected: the `--dry-run` call should report a no-op, since all of the packages have already been built.

Actual: it insists on rebuilding `haskell-src-exts`. The reason is clear when you look at the following `--verbose` bit of output:

```
[...]

PR #4254, opened 2018-08-22
Start using the pantry package

snoyberg:
This is a massive diff. Any reviews people have are most welcome. The basic idea is:

* All code related to downloading packages, unpacking packages, parsing cabal files, and interacting with Hackage have been moved into a new package, `pantry`
* This new package currently resides inside the Stack repo, but will be moved out in the future. (Same with the `curator` package present as well.)
* The rest of the codebase has to adapt :)

For those looking to review, I'd recommend:

* Don't bother looking at individual commits, too much noise
* Check out the changes inside the `doc` directory and `ChangeLog.md` first
* Review the code in `subs/pantry` next
* As much as you'd like, review the changes in `src`


doc/pantry.md
```diff
@@ -0,0 +1,482 @@
+<div class="hidden-warning"><a href="https://docs.haskellstack.org/"><img src="https://rawgit.com/commercialhaskell/stack/master/doc/img/hidden-warning.svg"></a></div>
+
+# Pantry in Stack
+
+Beginning with Stack 1.11, Stack uses the Pantry library for its
+specification of snapshots and package locations. Under the surface,
+Pantry is geared towards reproducible build plans with
+cryptographically secure specification of packages and snapshots.
```
dbaynard: If version changes, it needs to change here, too (there are a few more places in the file.

doc/pantry.md
```diff
@@ -0,0 +1,482 @@
...
+  cabal-file:
+    size: 1717
+    sha256: 7b46e7a8b121d668351fa8a684810afadf58c39276125098485203ef274fd056
+  name: wai
+  version: 3.0.2.3
+  git: git@github.com:yesodweb/wai
+  pantry-tree:
+    size: 10299
+    sha256: ce33fddab13592c847fbd7acd1859dfcbb9aeb6c212db3cee27c909fa3f3ae44
+  commit: 2f8a8e1b771829f4a8a77c0111352ce45a14c30f
```
dbaynard: To update this, do I have to delete the incorrect information, or will it regenerate the correct information if I just update the commit hash?

doc/pantry.md
```diff
@@ -0,0 +1,482 @@
...
+- https://example.com/foo/bar/baz-0.0.2.tar.gz
+- archive: http://github.com/yesodweb/wai/archive/2f8a8e1b771829f4a8a77c0111352ce45a14c30f.zip
+  subdirs:
+  - wai
+  - warp
+- archive: ../acme-missiles-0.3.tar.gz
+  sha256: e563d8b524017a06b32768c4db8eff1f822f3fb22a90320b7e414402647b735b
+```
+
+With the `stack freeze` command, this would be replaced with:
```
dbaynard: Likewise, if I update and re-run stack freeze will that overwrite the details, here?

doc/pantry.md
```diff
@@ -0,0 +1,482 @@
...
+# parser
+hidden:
+  wai: true
+  warp: false
+
+# Set GHC options for specific packages
+ghc-options:
+  warp:
+  - -O2
+```
```
dbaynard: Haddock settings would go here, too? e.g. not to build a dependency with broken haddocks

doc/pantry.md
```diff
@@ -0,0 +1,482 @@
...
+  warp: false
+  wai: true
+drop-packages:
+- wai-extra
+compiler: ghc-8.0.1
+resolver:
+  size: 515969
+  url: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/8/21.yaml
+  sha256: 2ec73d520d3e55cb753eaca11a72a9ce95bd9ba7ccaf16de1150d0130a50a5a1
+```
```
dbaynard: Upgrading this looks like a bit of a nightmare — is there a way to undo a freeze?

doc/pantry.md
```diff
@@ -0,0 +1,482 @@
...
+  which do not specify pantry tree pointer of the package archive
+* `resolver` if it references a remote snapshot and if it does not specify
+  pantry tree pointer of its contents
+
+The command outputs to standard output new project's `stack.yaml` with these
+changes included.
+
+If a project is specified precisely enough stack tells about it and exits.
+
+### Snapshot freezing
```
dbaynard: This section is partially duplicated (`--snapshot` is introduced earlier), and has numerous typos and ambiguities. Perhaps just move the (fixed) last paragraph to the first description of snapshot freezing?

doc/pantry.md
```diff
@@ -0,0 +1,482 @@
...
+    sha256: 2ba66a092a32593880a87fb00f3213762d7bca65a687d45965778deb8694c5d1
+  name: acme-missiles
+  version: '0.3'
+  sha256: e563d8b524017a06b32768c4db8eff1f822f3fb22a90320b7e414402647b735b
+  pantry-tree:
+    size: 226
+    sha256: 614bc0cca76937507ea0a5ccc17a504c997ce458d7f2f9e43b15a10c8eaeb033
+```
+
+## Snapshots
```
dbaynard: Purescript's `psc-package` has a [verify command](https://psc-package.readthedocs.io/en/latest/maintenance.html) that ensures a snapshot (purescript: package set) builds correctly. Will this be in the curator tool?

doc/yaml_configuration.md
```diff
@@ -60,210 +60,159 @@ are currently four resolver types:
...
+    commit: 6bf765e000c6fd14e09ebdea6c4c5b1510ff5376
+    subdirs:
+      - wai-extra
 ```
 
-Note that if a local directory can be parsed as a package identifier,
-Stack will treat it as a package identifier. In other words, if you
-have a local directory named `foo-1.2.3`, instead of:
+And, in fact, the `packages` value could be left off entirely since
+it's using the default value.
```
dbaynard: It would be good to have some way to automate this?

snoyberg:
> Also with the way the repo is currently set up, these changes will hit version 1.10, not 1.11.

It will live in dev as 1.10, but will be released to production as 1.11.

> Merging this update to master complicates stack upgrade --git enormously — as it will also require a rebuild of all package information. I suspect nobody rolls back from a stack upgrade --git, though.

After this change gets in, I'm going to take a stab at #3922, which will be an even more fundamental change and necessitate breaking all preexisting compilations. This change is necessary to clean up some existing bugs, and open the way towards Backpack support. We're going to have to bite the bullet at some point of having this pain. My preference is to do it as early in a dev cycle as possible, which is why I think now's a good time.

To be clear about the downside here: after running `stack upgrade --git`, users will need to rebuild all of their package dependencies. They will also need to redownload all package contents. However:

1. This will only happen once
2. Switching back to the old version of Stack shouldn't necessitate rebuilding again (though I haven't tested that thoroughly)

There's an argument to wait on merging this into master until _after_ I've implemented #3922 to avoid breaking caches twice. However, my aversion to this is because:

* It will make the PR even more significantly larger
* It may introduce breakage with changes which will continue on master
* We'll get less testing of this code
* The affected audience of these repeated-cache-breakages is relatively small.

ChangeLog.md
```diff
@@ -7,6 +7,38 @@ Release notes:
...
+            information. @@@TODO ensure `stack freeze` actually makes
+            it in.
+    * Package contents and metadata are stored in an SQLite database
+      in place of files on the filesystem. The `pantry` library can be
+      used for interacting with these contents.
+    * Internally, Stack has changed many datatypes, including moving
+      to Cabal's definition of many data types. As a result of such
+      changes, existing cache files will in general be invalidated,
+      resulting in Stack needing to rebuild many previously cached
+      builds in the new version. Sorry :(.
```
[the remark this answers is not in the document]
snoyberg: I'm OK with such a version bump for this set of changes, especially if 2.0 ends up including #3922 and Backpack. @borsboom thoughts?

ChangeLog.md
```diff
@@ -7,6 +7,38 @@ Release notes:
...
+      dependencies. `packages` now only supports local filepaths.
+    * Addition of new configuration options for specifying a "pantry
+      tree" key, which provides more reproducibility around builds,
+      and (in the future) will be used for more efficient package
+      content downloads. You can also specify package name and version
+      for more efficient config parsing.
+          * __NOTE__ The new `stack freeze` command provides support
+            for automatically generating this additional
+            information. @@@TODO ensure `stack freeze` actually makes
+            it in.
```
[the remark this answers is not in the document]
snoyberg: Good catch, I'll remove that.

doc/ghci.md
```diff
@@ -75,3 +75,10 @@ For doing experiments which just involve packages installed in your databases,
 it may be useful to run ghci plainly like `stack exec ghci`. This will run a
 plain `ghci` in an environment which includes `GHC_PACKAGE_PATH`, and so will
 have access to your databases.
+
+*Note*: Running `stack ghci` on a pristine copy of the code doesn't currently
+build libraries
+([#2790](https://github.com/commercialhaskell/stack/issues/2790)) or internal
+libraries ([#4148](https://github.com/commercialhaskell/stack/issues/4148)).
+It is recommended to always run a `stack build` before running `stack ghci`,
+until these two issues are closed.
```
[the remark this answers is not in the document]
snoyberg: It looks like this was a doc change on stable. I'll merge stable into master to help make this diff clearer.

doc/pantry.md
snoyberg: You'd need to delete the incorrect information. I'll add a comment about that.

doc/pantry.md
snoyberg: I haven't actually changed any of the behavior of `ghc-options` in this PR. I don't know if any of these settings would apply to Haddock or not. IIRC, there's an open issue about that, right?

doc/pantry.md
snoyberg: That's an interesting idea. We _could_ provide an `unfreeze` command which removes all of the "unnecessary" information. Personally, I'd rather wait to try this out a bit and see if such a command is necessary. I would imagine this may actually be easier in practice than it first seems.

doc/pantry.md
snoyberg: Hopefully addressed in 01f9e84f2cd76d2e5dd66ba2411a3ba8fac08ca9, thanks.

doc/pantry.md
snoyberg: Yes, the goal of `curator` is to:

* Create a snapshot
* Test the snapshot
* Publish the snapshot

doc/yaml_configuration.md
snoyberg: I'm hesitant about this, since our story for writing out `stack.yaml` files is pretty bad right now given that we can preserve comments.

doc/pantry.md
dbaynard: By here, I meant in this file, not this stanza — please excuse the ambiguity.

I can track down the issue — I'm not sure off the top of my head. But from the gitter chat:

> @dbaynard
> Is there a way to avoid haddock, or pass haddock arguments, when building a dependency, in the same way it is possible to set flags on dependencies? I'm unable to build some haddocks, but I want to ignore the error.
> …
> @snoyberg 
> Funny you should ask, we have the same requirement for Stackage. So I'm adding a new (hidden) curator: configuration that allows you to disable Haddocks, tests, and benchmarks on a package-by-package basis

doc/pantry.md
dbaynard: Sure. It's easy enough to work around, and might be a nice issue for a newcomer.

doc/pantry.md
dbaynard: Yes, it is.

dbaynard:
Thanks, the changes look good. I'm grateful for the explanation of the cache-breaking behaviour.

snoyberg:
> Thanks, the changes look good. I'm grateful for the explanation of the cache-breaking behaviour.

I'm very grateful for the great review, thanks!

doc/yaml_configuration.md
```diff
@@ -894,21 +836,9 @@ Since 1.8.0
...
-```yaml
-ignore-revision-mismatch: false
-```
-
-For more information, see
-[the Github issue #3520 discussion](https://github.com/commercialhaskell/stack/issues/3520).
-
-Since 1.6.0
+This flag was introduced in Stack 1.6, and removed in Stack 1.11 with
+hte move to Pantry. You will receive a warning if this configuration
```
mihaimaruseac: the instead of hte

mihaimaruseac:
> There's an argument to wait on merging this into master until after I've implemented #3922 to avoid breaking caches twice. However, my aversion to this is because:..

I'd say we should merge this first, it's already quite large.

snoyberg:
Integration tests are passing, so this is ready to merge. There are some merge conflicts with master, which I'm resolving. Once that's done and CI is green, anyone should feel free to merge this.

Issue #4213, opened 2018-08-07
Remove some of the Stack.Types.* modules


snoyberg:
We've moved over to using the upstream Cabal types for many things on the pantry branch. Some of these modules can be deleted. Some functionality may move into pantry instead.

PR #4255, opened 2018-08-23
Migrate some remains of code to use Cabal types/functions

### LINKED ISSUE #4250: Remove the TH safe constructors for Cabal data types?
On the current master of Stack, we have proper abstract data types for `PackageName`, `Version`, `FlagName`, and `PackageIdentifier`. In order to construct these values, you need to either use a function with returns a `MonadThrow` at runtime, _or_ a Template Haskell function which parses a string literal at compile time. We're guaranteed to only have valid values.

For better or worse, I've moved us over to using Cabal's definition of these datatypes on the pantry branch. This simplifies a bunch of conversion code to and from these two representations, which was my primary motivation. However, I discovered accidentally that these types sometimes provide completely unchecked `IsString` instances.

Tying this together: we use the TH-enabled functions significantly throughout the codebase. We _do_ get some safety guarantees, but given the open-barn-door of the broken `IsString` instances, not as much. And Template Haskell does have a significant downside in terms of slowing down compilation time and forcing recompiles of modules often.

My proposal: let's remove these TH-enabled functions, and just use string literals in the codebase. Advantages: less code to manage, faster compiles. Disadvantages: some bugs can sneak past a compile. However, these kinds of bugs should hopefully be pretty minor given.

CC @chrisdone.

Related to: #4213

qrilka:

Fixes #4250 and #4213


src/Stack/SDist.hs
```diff
@@ -195,7 +193,7 @@ getCabalLbs pvpBounds mrev cabalfp = do
                   $ Cabal.packageDescription gpd'
                   }
               }
-    ident <- parsePackageIdentifierThrowing $ Cabal.display $ Cabal.package $ Cabal.packageDescription gpd''
```
snoyberg: Well that's much nicer :)

src/Stack/Types/PackageName.hs
```diff
@@ -10,20 +10,13 @@
 -- | Names for packages.
 
 module Stack.Types.PackageName
-  (PackageName
-  ,PackageNameParseFail(..)
-  ,parsePackageName
-  ,parsePackageNameThrowing
+  (parsePackageNameThrowing
```
snoyberg: How about moving this function to `pantry`? See any problem with that?

src/Stack/Types/Version.hs
```diff
@@ -15,9 +15,7 @@ module Stack.Types.Version
   ,Cabal.VersionRange -- TODO in the future should have a newtype wrapper
   ,IntersectingVersionRange(..)
   ,VersionCheck(..)
-  ,parseVersion
   ,parseVersionThrowing
```
snoyberg: Also a candidate for `pantry`.

snoyberg:
On the "remove TH pragams" from, looks like hlint has our back:

https://travis-ci.org/commercialhaskell/stack/jobs/419518030

src/Stack/Types/PackageName.hs
qrilka: Actually I've done that and then returned it back, I guess you're right and it would be better to have parsing code in 1 place

qrilka:
@snoyberg I've updated the PR

subs/pantry/src/Pantry/Types.hs
```diff
@@ -1002,12 +1005,56 @@ parsePackageIdentifier str =
 parsePackageName :: String -> Maybe PackageName
 parsePackageName = Distribution.Text.simpleParse
 
+-- | A package name parse fail.
+data PackageNameParseFail
```
snoyberg: I'd rather include these data constructors with the existing `PantryException` data type, otherwise the docs there will be wrong.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -1002,12 +1005,56 @@ parsePackageIdentifier str =
...
+  | CabalFileNameParseFail FilePath
+  | CabalFileNameInvalidPackageName FilePath
+  deriving (Typeable)
+instance Exception PackageNameParseFail
+instance Show PackageNameParseFail where
+    show (PackageNameParseFail bs) = "Invalid package name: " ++ show bs
+    show (CabalFileNameParseFail fp) = "Invalid file path for cabal file, must have a .cabal extension: " ++ fp
+    show (CabalFileNameInvalidPackageName fp) = "cabal file names must use valid package names followed by a .cabal extension, the following is invalid: " ++ fp
+
+-- | Parse a package name from a 'String'.
```
snoyberg: Please add `@since 0.1.0.0` comments to all newly exposed identifiers.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -1002,12 +1005,56 @@ parsePackageIdentifier str =
...
+parsePackageNameThrowing :: MonadThrow m => String -> m PackageName
+parsePackageNameThrowing str =
+  case parsePackageName str of
+    Nothing -> throwM $ PackageNameParseFail $ T.pack str
+    Just pn -> pure pn
+
+-- | Parse a package name from a file path.
+parsePackageNameFromFilePath :: MonadThrow m => Path a File -> m PackageName
```
snoyberg: It looks like this function is still be used in one place in the codebase, but I have a feeling it's not really needed. Let's take it out of pantry, and move its implementation directly into `Stack.Package` to avoid cluttering the public API with something that shouldn't in general be needed.

qrilka: Ok, I had some similar thoughts but didn't pursue them, I'll fix it

Issue #4204, opened 2018-08-04
Support the NO_COLOR environment variable


ajgrf:
Currently the only way to tell stack to suppress color is by passing `--color never` as an argument. A user who wishes to never see color might try to define a shell alias, but this is not always interpreted correctly by stack. For example:

    $ alias stack='stack --color never'
    $ stack foo
    Invalid argument `foo'
    File does not exist or is not a regular file `--color'
    
    ...

More importantly, this is stack-specific configuration which does not live in stack's config files. The [NO_COLOR](http://no-color.org/) environment variable is a proposed informal standard for signaling that programs should not emit colored output unless explicitly configured to. It is already used by a number of other popular tools, and might even be set before a user installs stack.

This feature would make stack nicer for users like me, who prefer less color, or whose unusual terminal colors make some of stack's output appear unreadable.

snoyberg:
Looks like this was already added, closing

PR #4257, opened 2018-08-23
Fix (in part) #4204: Add `color:` to `config.yaml`

mpilgrem:
This pull request allows the global option `--color=<WHEN>` to be set also in a yaml configuration file.

Tested on Windows 10 by building - `stack test` - and then using. Principally tested by reference to the output of the new `stack ls stack-colors` command, as follows:

* `stack ls stack-colors` gives a report with colour Examples (as expected)
* `stack --color=never ls stack-colors` gives a report without colour Examples (as expected)
* set `color: never` in `config.yaml`, `stack ls stack-colors` gives a report without colour Examples (as expected)
* set `color: never` in `config.yaml`, `stack --color=always ls stack-colors` gives a report with colour Examples (as expected)
* set `color: dummy` in `config.yaml`, `stack --color=always ls stack-colors` gives a useful error report about not being able to parse `config.yaml` (as expected)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,8 @@ Other enhancements:
   plan construction errors much faster, and avoid some unnecessary
   work when only building a subset of packages. This is especially
   useful for the curator use case.
+* Existing global option `--color=WHEN` is now also available as a
+  non-project-specific yaml configuration parameter `color:`.
 * New command `stack ls stack-colors` lists the styles and the associated 'ANSI'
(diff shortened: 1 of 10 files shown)


mpilgrem:
The red cross against the Travis CI build is only because the GHC 8.2.2 job cannot be built within Travis's time limits. The 'pedantic' GHC 8.2.2 job is successful.

mpilgrem:
Now also similarly tested successfully on Windows 7.

PR #4260, opened 2018-08-25
Fix (in part) #4204: Honour `NO_COLOR` variable

mpilgrem:
Adopt the standard proposed at http://no-color.org/, that colour should not be added by default if the `NO_COLOR` environment variable is present, by adapting `defaultColorWhen` in module `Stack.DefaultColorWhen` (which varies depending on whether the operating system is Windows or Unix-like).

* [x] The documentation has been updated, if necessary. (Not necessary.)

Tested on Windows 10 by building - `stack test` - and using the built tool, as below:
image

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,8 @@ Other enhancements:
   plan construction errors much faster, and avoid some unnecessary
   work when only building a subset of packages. This is especially
   useful for the curator use case.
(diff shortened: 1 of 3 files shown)


mpilgrem:
@mihaimaruseac, it seems that merging this was a little premature - there is a problem with the Unix build. I'm about to make a further pull request that fixes that.

mihaimaruseac:
Thanks. That should teach me not to rush things.

PR #4264, opened 2018-08-26
Fix #4205 (step 6): Alias 'colour' for 'color'

### LINKED ISSUE #4205: Allow user specification of stack's colours

mpilgrem:
This implements the final part of #4205. Although it is 'cosmetic', it is in the tradition of GHC and other Haskell tools (eg hlint) accepting alternative spellings.

This pull request allows the British English spelling of 'color' (colour) as aliases. Implemented at the command line for `--color`, `--stack-colors` and `stack ls stack-colors`. Implemented in yaml configuration files for `color:` and `stack-colors:` (in each case, the American spelling is the alternative that has priority).

The documentation in `yaml_configuration.md` is updated, accordingly.

Tested on macOS 10.13.6 and on Windows 10 by building - `stack test` - and then using the built tool, including:
* `stack --help`
* `stack ls --help`
* `stack ls stack-colours --help`
* `stack --color=always --stack-colors=file=31 ls stack-colours` (and the equivalent using `config.yaml`)
* `stack --colour=always --stack-colours=file=31 ls stack-colors` (and the equivalent using `config.yaml`)

### DIFF
--- a/src/Stack/Types/Config.hs
+++ b/src/Stack/Types/Config.hs
@@ -854,8 +854,16 @@ parseConfigMonoidObject rootDir obj = do
     configMonoidDumpLogs <- First <$> obj ..:? configMonoidDumpLogsName
     configMonoidSaveHackageCreds <- First <$> obj ..:? configMonoidSaveHackageCredsName
     configMonoidHackageBaseUrl <- First <$> obj ..:? configMonoidHackageBaseUrlName
-    configMonoidColorWhen <- First <$> obj ..:? configMonoidColorWhenName
-    configMonoidStyles <- fromMaybe mempty <$> obj ..:? configMonoidStylesName
+
+    configMonoidColorWhenUS <- obj ..:? configMonoidColorWhenUSName
+    configMonoidColorWhenGB <- obj ..:? configMonoidColorWhenGBName
+    let configMonoidColorWhen =  First $   configMonoidColorWhenUS
+                                       <|> configMonoidColorWhenGB
+
+    configMonoidStylesUS <- obj ..:? configMonoidStylesUSName
+    configMonoidStylesGB <- obj ..:? configMonoidStylesGBName
+    let configMonoidStyles = fromMaybe mempty $   configMonoidStylesUS
+                                              <|> configMonoidStylesGB
 
     return ConfigMonoid {..}
   where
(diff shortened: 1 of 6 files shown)


src/Stack/Types/Config.hs
```diff
@@ -891,8 +891,16 @@ parseConfigMonoidObject rootDir obj = do
...
+
+    configMonoidColorWhenPrimary <- obj ..:? configMonoidColorWhenName
+    configMonoidColorWhenAlternative <- obj ..:? configMonoidColorWhenAlternativeName
+    let configMonoidColorWhen =  First $   configMonoidColorWhenPrimary
+                                       <|> configMonoidColorWhenAlternative
+
+    configMonoidStylesPrimary <- obj ..:? configMonoidStylesName
+    configMonoidStylesAlternative <- obj ..:? configMonoidStylesAlternativeName
+    let configMonoidStyles = fromMaybe mempty $   configMonoidStylesPrimary
+                                              <|> configMonoidStylesAlternative
```
dbaynard: This (from line 894) would be clearer if rather than `Primary` and `Alternative` the options were `US` and `GB` (or equivalent); it wasn't obvious that `Primary` and `Alternative` referred to _spellings_.

src/Stack/Types/Config.hs
```diff
@@ -1041,9 +1049,15 @@ configMonoidIgnoreRevisionMismatchName = "ignore-revision-mismatch"
 configMonoidColorWhenName :: Text
 configMonoidColorWhenName = "color"
 
+configMonoidColorWhenAlternativeName :: Text
+configMonoidColorWhenAlternativeName = "colour"
+
```
dbaynard: Likewise here (`s/Alternative/British/`)

src/Stack/Types/Config.hs
```diff
@@ -1041,9 +1049,15 @@ configMonoidIgnoreRevisionMismatchName = "ignore-revision-mismatch"
...
 configMonoidStylesName :: Text
 configMonoidStylesName = "stack-colors"
 
+configMonoidStylesAlternativeName :: Text
+configMonoidStylesAlternativeName = "stack-colours"
+
```
dbaynard: Likewise here (`s/Alternative/British/`)

dbaynard:
Thanks, @mpilgrem. Looks good, just a few things that would make it clearer.

src/Stack/Types/Config.hs
mpilgrem: Done - I now use `US` and `GB` consistently.

src/Stack/Types/Config.hs
mpilgrem: Done - I now use `US` and `GB` consistently.

src/Stack/Types/Config.hs
mpilgrem: Done - I now use `US` and `GB` consistently.

Issue #4181, opened 2018-07-28
Stack clean command should not go out stubbornly downloading things


renzhi:
The stack clean command should be really simple: Clean up the object files in the project.

But what it does is, start up and sit there, pondering about life for a long time, and then decide that it wants to download a bunch things and rebuild. But it keeps getting [connection timeout issue](https://github.com/commercialhaskell/stack/issues/3510), even if curl can perfectly fetch the package.

Even if you have just upgraded stack, the stack clean should just be a simple command to do the clean up.

renzhi:
Here's part of a very frustrating daily life using stack:

```
xp @ λ ★ stack clean
Preparing to install GHC (tinfo6) to an isolated location.
This will not interfere with any system-level installation.
                                          
Warning: Retry number 0 after a total delay of 0 us
         If you see this warning and stack fails to download, but running the command again solves the problem,
         please report here: https://github.com/commercialhaskell/stack/issues/3510
                                          
Warning: Retry number 1 after a total delay of 100000 us
         If you see this warning and stack fails to download, but running the command again solves the problem,
         please report here: https://github.com/commercialhaskell/stack/issues/3510
                                          
Warning: Retry number 2 after a total delay of 200000 us
         If you see this warning and stack fails to download, but running the command again solves the problem,
         please report here: https://github.com/commercialhaskell/stack/issues/3510
                                          
Warning: Retry number 3 after a total delay of 300000 us
         If you see this warning and stack fails to download, but running the command again solves the problem,
         please report here: https://github.com/commercialhaskell/stack/issues/3510
Preparing to download ghc-tinfo6-8.2.2 ...
HttpExceptionRequest Request {
  host                 = "github.com"
  port                 = 443
  secure               = True
  requestHeaders       = [("User-Agent","The Haskell Stack")]
  path                 = "/commercialhaskell/ghc/releases/download/ghc-8.2.2-release/ghc-8.2.2-x86_64-fedora27-linux-patch1.tar.xz"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 ConnectionTimeout
xp @ λ ★ stack clean
HttpExceptionRequest Request {
  host                 = "raw.githubusercontent.com"
  port                 = 443
  secure               = True
  requestHeaders       = [("User-Agent","The Haskell Stack")]
[...]

Issue #3471, opened 2017-10-04
`stack clean --full` may not work properly with docker + nix


mgsloan:
To fix #2010, "stack clean" was made to no longer run within a docker or nix environment.  This broke cleaning within the nix-shell as described in #3468.   The solution in https://github.com/commercialhaskell/stack/pull/3470 is to not enter docker for `stack clean --full`, but enter into nix.  I don't think this will work properly for docker + nix.

A better solution might be to make `stack clean --full` work within the containers, by not trying to delete the `_home` dir mentioned in #2010

snoyberg:
I've reworked `stack clean` to no longer use an `EnvConfig` at all, which should fully address these concerns. Closing.

Issue #3468, opened 2017-10-03
Stack clean does not work correctly with --nix


ElvishJerricco:
When using nix integration, `stack clean` does not work. I'm not sure if there's any reason it should need GHC at all for `clean`, but it exits with a `No compiler found, ...` error message when the user does not have GHC installed in their user profile. Presumably Stack is not performing `clean` from within the nix shell, or it should not need GHC at all.

EDIT: This is a regression in Stack 1.5. This worked in 1.4.

mgsloan:
That is just part of stack's normal environment setup.  There isn't currently an environment type / code to load it that reflects "I need to use Cabal but I don't need GHC".  I'm not sure this is important, AFIACT a compiler was also needed in stack-1.4, not convinced there was a regression here.

`stack clean` is implemented by running Cabal clean (roughly kinda like `runghc Setup.hs clean`), so this does require GHC.

`stack clean --full` could potentially be implemented to not need the compiler, since it just recursively deletes the work dirs.

Leaning towards wontfix, I don't see why this matters.  Why would you ever want to clean if you don't have a compiler?  Would accept a PR that makes loading the full env unnecessary for `stack clean --full`.

ElvishJerricco:
@mgsloan the regression is that Stack should enter the Nix shell when doing `clean` with `—nix`, like it does for every other command. Entering the Nix shell would provide it with the GHC it needs. Excuse my irrelevant commentary on whether GHC should be necessary for `clean` at all =P

mgsloan:
Updated the first paragraph of my comment above.  I think the commit linked there is definitely the cause of the issue.  A couple possible solutions:

* Have `stack clean` enter nix shell, but not docker containers.  Problem with this approach is it won't work for docker + nix.
* Have `stack clean --full` not enter either nix or docker containers.  `stack clean` would enter both, which is an enhancement, current code path demands you have cabal / ghc outside of your docker / nix.  Can also maybe make `stack clean --full` skip demanding a ghc / cabal be installed.

I'll give it a shot.

mgsloan:
I've split off https://github.com/commercialhaskell/stack/issues/3469 and opened https://github.com/commercialhaskell/stack/pull/3470 .  Can you please check if that change fixes the issue?  Thanks!  The solution is not ideal: https://github.com/commercialhaskell/stack/issues/3471

ElvishJerricco:
That appears to have fixed it. Thanks!

mgsloan:
[...]

Issue #2010, opened 2016-04-07
`stack clean --full` tries to delete .stack-work/docker/_home which is mounted on docker enabled projects


dysinger:
Create a docker enabled project (builds the project in a docker container)

run `stack clean --full` and you'll see it break with an error

dysinger:
FYI: .stack-work/docker/_home is the $HOME directory of the `stack` user inside the container.  We can't delete that because when you are inside the container this directory is mounted and in use. 

The solution would be to do the `stack clean --full` outside of the docker container (not inside).

angerman:
Ran into this as well:
```
$ stack --docker --docker-image img clean --full
~/p/s/cd/.stack-work/docker/_home: removeDirectoryRecursive: resource busy (Device or resource busy)
```

mgsloan:
Fixed!

PR #4268, opened 2018-08-27
To prevent download ghc when using `stack clean`.

waddlaw:
`stack clean` always download the ghc even if it already deleted.

```shell
$ stack --version
Version 1.10.0, Git revision 65dd1d673a75d867325bb1c70f5170386900dd80 (dirty) (6415 commits) x86_64 hpack-0.29.6

$ stack clean
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Preparing to download ghc-7.10.2 ...^C
```

### DIFF
--- a/src/Stack/Runners.hs
+++ b/src/Stack/Runners.hs
@@ -129,7 +129,7 @@ withBuildConfigAndLock
     -> (Maybe FileLock -> RIO EnvConfig ())
     -> IO ()
 withBuildConfigAndLock go inner =
-    withBuildConfigExt WithDocker False go Nothing inner Nothing
+    withBuildConfigExt WithDocker WithDownloadCompiler go Nothing inner Nothing
 
 -- | See issue #2010 for why this exists. Currently just used for the
 -- specific case of "stack clean --full".
@@ -138,14 +138,14 @@ withBuildConfigAndLockNoDocker
     -> (Maybe FileLock -> RIO EnvConfig ())
     -> IO ()
 withBuildConfigAndLockNoDocker go inner =
-    withBuildConfigExt SkipDocker False go Nothing inner Nothing
+    withBuildConfigExt SkipDocker WithDownloadCompiler go Nothing inner Nothing
 
 withBuildConfigAndLockInClean
     :: GlobalOpts
     -> (Maybe FileLock -> RIO EnvConfig ())
     -> IO ()
 withBuildConfigAndLockInClean go inner =
-    withBuildConfigExt WithDocker True go Nothing inner Nothing
+    withBuildConfigExt WithDocker SkipDownloadCompiler go Nothing inner Nothing
 
 -- | See issue #2010 for why this exists. Currently just used for the
 -- specific case of "stack clean --full".
@@ -154,11 +154,11 @@ withBuildConfigAndLockNoDockerInClean
     -> (Maybe FileLock -> RIO EnvConfig ())
     -> IO ()
 withBuildConfigAndLockNoDockerInClean go inner =
-    withBuildConfigExt SkipDocker True go Nothing inner Nothing
+    withBuildConfigExt SkipDocker SkipDownloadCompiler go Nothing inner Nothing
 
 withBuildConfigExt
     :: WithDocker
-    -> Bool -- ^ If perform the stack clean
+    -> WithDownloadCompiler -- ^ bypassed download compiler if SkipDownloadCompiler.
     -> GlobalOpts
     -> Maybe (RIO Config ())
     -- ^ Action to perform before the build.  This will be run on the host
@@ -174,7 +174,7 @@ withBuildConfigExt
     -- available in this action, since that would require build tools to be
     -- installed on the host OS.
     -> IO ()
-withBuildConfigExt skipDocker isClean go@GlobalOpts{..} mbefore inner mafter = loadConfigWithOpts go $ \lc -> do
+withBuildConfigExt skipDocker downloadCompiler go@GlobalOpts{..} mbefore inner mafter = loadConfigWithOpts go $ \lc -> do
     withUserFileLock go (view stackRootL lc) $ \lk0 -> do
       -- A local bit of state for communication between callbacks:
       curLk <- newIORef lk0
@@ -192,7 +192,7 @@ withBuildConfigExt skipDocker isClean go@GlobalOpts{..} mbefore inner mafter = l
 
       let inner'' lk = do
               bconfig <- lcLoadBuildConfig lc globalCompiler
-              let bconfig' = bconfig { bcClean = isClean }
+              let bconfig' = bconfig { bcDownloadCompiler = downloadCompiler }
               envConfig <- runRIO bconfig' (setupEnv Nothing)
               runRIO envConfig (inner' lk)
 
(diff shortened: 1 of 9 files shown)


dbaynard:
Thanks, @waddlaw  — this relates to #4181, correct?

I'm going to review, now.

src/main/Main.hs
```diff
@@ -631,8 +631,8 @@ cleanCmd opts go =
   -- See issues #2010 and #3468 for why "stack clean --full" is not used
   -- within docker.
```
dbaynard: So interestingly, both #2010 and #3468 have been closed — but #3471 is still open.

@mgsloan can this be simplified, now?

src/Stack/Runners.hs
```diff
@@ -136,10 +138,27 @@ withBuildConfigAndLockNoDocker
...
+withBuildConfigAndLockNoDockerInClean
+    :: GlobalOpts
+    -> (Maybe FileLock -> RIO EnvConfig ())
+    -> IO ()
+withBuildConfigAndLockNoDockerInClean go inner =
+    withBuildConfigExt True True go Nothing inner Nothing
 
 withBuildConfigExt
     :: Bool
+    -> Bool -- ^ If perform the stack clean
```
dbaynard: I'm concerned by the boolean blindness, here. The following types would be helpful.

```haskell
data WithDocker
  = SkipDocker
  | WithDocker

data IsClean
  = NotClean
  | IsClean
```

But see my comments on whether the docker checks are still necessary.

src/Stack/Runners.hs
```diff
@@ -127,7 +129,7 @@ withBuildConfigAndLock
     -> (Maybe FileLock -> RIO EnvConfig ())
     -> IO ()
 withBuildConfigAndLock go inner =
-    withBuildConfigExt False go Nothing inner Nothing
+    withBuildConfigExt False False go Nothing inner Nothing
 
 -- | See issue #2010 for why this exists. Currently just used for the
```
dbaynard: See comment below

src/Stack/Runners.hs
```diff
@@ -136,10 +138,27 @@ withBuildConfigAndLockNoDocker
...
+
+withBuildConfigAndLockInClean
+    :: GlobalOpts
+    -> (Maybe FileLock -> RIO EnvConfig ())
+    -> IO ()
+withBuildConfigAndLockInClean go inner =
+    withBuildConfigExt False True go Nothing inner Nothing
+
+-- | See issue #2010 for why this exists. Currently just used for the
+-- specific case of "stack clean --full".
```
dbaynard: See comment below

src/Stack/Setup.hs
```diff
@@ -235,7 +235,7 @@ setupEnv mResolveMissingGHC = do
             , soptsGHCJSBootOpts = ["--clean"]
             }
 
-    (mghcBin, compilerBuild, _) <- ensureCompiler sopts
+    (mghcBin, compilerBuild, _) <- if bcClean bconfig then pure (Nothing, CompilerBuildStandard, False) else ensureCompiler sopts
```
dbaynard: This change is making the presence of a compiler optional, and so the code should reflect this — yet this implementation does not.

Subsequent code uses the `compilerBuild` option; it is faked here, making that code more fragile. It might be better that it has type `Maybe CompilerBuild` here; in this way the possible absence of a compiler is reflected in the type system. This then propagates to `EnvConfig` which I believe will need changes.

src/Stack/Runners.hs
mgsloan: Agreed, making another type that's iso to Bool is usually better.

To me the naming seems too purpose specific.  I think it makes sense to instead have the name correspond to the behavioral change.  That, is the name should specify whether or not it requires ghc to be installed.

mgsloan:
@dbaynard I don't recall the details of this part of the code, so it's best just to look at what it's doing. I  agree about avoiding use of `Bool`.  I left a comment about naming considerations - I imagine that skipping compiler installation is not unique to the clean command.

waddlaw:
@dbaynard Thank you for reviewed!

> this relates to #4181, correct?

I don't know that directly related to it. But, bypassed the ghc download at least.

> There are 2 major issues which must be resolved.

OK. I will try to fix it.

> I left a comment about naming considerations - I imagine that skipping compiler installation is not unique to the clean command.

That makes sense.

waddlaw:
@dbaynard Are you sure about this?

mihaimaruseac:
Ping on this, as it's becoming stale.

dbaynard:
Ah, thank you for bearing with me — I must have looked at this 3 or 4 times and didn't notice the new commits.

I'll check shortly, and then it may be worth looking at the possibilities for #4390.

src/Stack/Types/Config.hs
```diff
@@ -535,7 +548,7 @@ data EnvConfig = EnvConfig
     -- ^ The actual version of the compiler to be used, as opposed to
     -- 'wantedCompilerL', which provides the version specified by the
     -- build plan.
-    ,envConfigCompilerBuild :: !CompilerBuild
+    ,envConfigCompilerBuild :: !(Maybe CompilerBuild)
```
dbaynard:
```suggestion
    ,envConfigCompilerBuild :: !(Maybe !CompilerBuild)
```

I _think_ we want to force the whole thing.

waddlaw:
@dbaynard thanks!

dbaynard:
@waddlaw have you seen #4390? Would you like to try to implement that?

waddlaw:
I’m more than happy to help you :) I'll do that.

dbaynard:
Great!

qrilka:
I have note about this PR - `SkipDownloadCompiler` doesn't quite reflect what's going on when this value get passed: it turns off any compiler checks. I'm not sure what could be a more suitable name here but it's better to have something more explicit and probably some haddocks could be added

dbaynard:
Yes — any suggestions welcome. In particular, there is #4405 which builds on it. I'll note your comment in my review. So `SkipCheckCompiler` would be a more precise name?

qrilka:
Thanks, sounds good to me though word order is confusing a bit, I think `SkipCompilerCheck` reads better, what do you think?

waddlaw:
Thanks @qrilka
It naming so nice. Would you like me to contain this change into # 4405? or submit another pr?

qrilka:
@waddlaw I think separate commit in #4405 with this change will be OK

qrilka:
@dbaynard it looks like #4480 uncovered problem with this PR and I'm not yet clear how to resolve it: by default `stack clean` does shallow cleanup deleting dist directories for project packages but to get proper dist directories Stack consults `ghc-pkg` to get `Cabal` version and without GHC installed (or not put into the current $PATH) it fails.
I think we should be able to go around that if we could use global packages retrieved from https://github.com/commercialhaskell/stackage-content/blob/master/stack/global-hints.yaml but that changes Stack logic a bit. Do we have any guarantees that `global-pkg-db` won't contain other `Cabal` version?
BTW just today I have a Twitter discussion about using arbitrary Cabal with earlier Stackage LTSes - https://twitter.com/qrilka/status/1081538344214020096

qrilka:
You're mostly correct in your analysis. There's only 1 thing to note: currently Cabal version is more or less fixed for an LTS (the one shipped with GHC gets used) and Veronika (from the Twitter discussion I linked above) started a fresh issue #4488 to discuss ways around that limitation of Stack.

PR #4269, opened 2018-08-27
Make `stack setup` command differentiate between FreeBSD versions.

arrowd:
In the line

`cutOffFreeBSDSuffix = map $ head . splitOn "-RELEASE" . head . splitOn "-CURRENT" . head . splitOn "-STABLE"`

`map` isn't necessary, because these prefixes can be present only in the last item of the list. I just used `map` for brevity.

With this patch `stack setup` says

> Unable to find installation URLs for OS key: freebsd64-maj12-min0

instead of

> Unable to find installation URLs for OS key: freebsd64

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -623,6 +623,9 @@ getGhcBuilds = do
             Platform _ Cabal.OpenBSD -> do
                 releaseStr <- mungeRelease <$> sysRelease
                 useBuilds [CompilerBuildSpecialized releaseStr]
+            Platform _ Cabal.FreeBSD -> do
+                releaseStr <- mungeRelease <$> sysRelease
+                useBuilds [CompilerBuildSpecialized releaseStr]
 #endif
             _ -> useBuilds [CompilerBuildStandard]
     useBuilds builds = do
@@ -632,14 +635,18 @@ getGhcBuilds = do
         return builds
 
 #if !WINDOWS
--- | Encode an OpenBSD version (like "6.1") into a valid argument for
+-- | Encode an OpenBSD/FreeBSD version (like "6.1"/"12.0-RELEASE") into a valid argument for
 -- CompilerBuildSpecialized, so "maj6-min1". Later version numbers are prefixed
 -- with "r".
 -- The result r must be such that "ghc-" ++ r is a valid package name,
 -- as recognized by parsePackageNameFromString.
 mungeRelease :: String -> String
-mungeRelease = intercalate "-" . prefixMaj . splitOn "."
+mungeRelease = intercalate "-" . prefixMaj . splitOn "." . cutOffFreeBSDSuffix
   where
+    cutOffFreeBSDSuffix s = if s == cutted
+                               then cutted
+                               else init cutted
+        where cutted = dropWhileEnd (/='-') s
     prefixFst pfx k (rev : revs) = (pfx ++ rev) : k revs
     prefixFst _ _ [] = []
     prefixMaj = prefixFst "maj" prefixMin


src/Stack/Setup.hs
```diff
@@ -632,14 +635,15 @@ getGhcBuilds = do
...
+-- | Encode an OpenBSD/FreeBSD version (like "6.1"/"12.0-RELEASE") into a valid argument for
 -- CompilerBuildSpecialized, so "maj6-min1". Later version numbers are prefixed
 -- with "r".
 -- The result r must be such that "ghc-" ++ r is a valid package name,
 -- as recognized by parsePackageNameFromString.
 mungeRelease :: String -> String
-mungeRelease = intercalate "-" . prefixMaj . splitOn "."
+mungeRelease = intercalate "-" . prefixMaj . cutOffFreeBSDSuffix . splitOn "."
   where
+    cutOffFreeBSDSuffix = map $ head . splitOn "-RELEASE" . head . splitOn "-CURRENT" . head . splitOn "-STABLE"
```
dbaynard: Can you talk through what you intend this code to do?

dbaynard:
Thanks for the PR, @arrowd. Why does it matter that `stack build` differentiates between these versions?

arrowd:
Err, I meant `stack setup`, not `stack build`. It is required to allow stack select proper bindist version to setup. It is the same change that was done for OpenBSD in https://github.com/commercialhaskell/stack/commit/3cd92c8bdd2a8456ae186c49c7689aaad098c47c

src/Stack/Setup.hs
arrowd: This cuts off "-RELEASE", "-CURRENT" and "-STABLE" suffixes, that are present in output of `sysRelease` on FreeBSD.

arrowd:
Ping?

borsboom:
Will this require any changes to https://github.com/fpco/stackage-content/blob/master/stack/stack-setup-2.yaml?  IIRC when we added this logic for OpenBSD, we had to add more entries to the setup metadata for the various support specific versions.  If so, can you be sure to also make a PR for those changes in the `fpco/stackage-content` repo?

src/Stack/Setup.hs
borsboom: Would it be safe to just cut off the `-[A-Z]+$` regex rather than hardcode the specific `-RELEASE`, `-CURRENT`, and `-STABLE` suffixes?

arrowd: Now you mention that, right. There are `-ALPHA1`, `BETA2` and other suffixes, so it is better to just cut off using regex. What package/function should I use for that?

arrowd:
> Will this require any changes to https://github.com/fpco/stackage-content/blob/master/stack/stack-setup-2.yaml?

Yes, probably. This depends on how these GHC's are produced. There is a hack required to make GHC compiled on old FreeBSD system (10 or 11) to work on recent one.

So, who prepares these releases and how is it done? I'm maintaing [FreeBSD port](http://www.freshports.org/lang/ghc) for GHC, by the way, maybe it can be used to easily produce these releases.

borsboom:
@arrowd The official GHC bindists (from https://www.haskell.org/ghc/download.html) are built by @bgamari, and we try to stick with those as much as possible for Stack.  However, we do also accept bindists from third parties, so if you wanted to provide custom bindists to support additional FreeBSD versions, then we could add those to the setup-setup metadata.  In case it's helpful, this is the script that @bgamari uses to build the bindists: https://github.com/bgamari/ghc-utils/blob/master/rel-eng/bin-release.sh

src/Stack/Setup.hs
borsboom: It's probably easier to use standard text functions like `dropWhileEnd` rather than resorting to regexes, but we are using the `regex-applicative-text` in other parts of the codebase.

arrowd:
So, can this be comitted? @bgamari said he will provide bindists for GHC 8.6 and onwards and I can prepare one for 8.4.3.

borsboom:
I haven't seen a PR for https://github.com/fpco/stackage-content/blob/master/stack/stack-setup-2.yaml that adds entries with the version suffixes in the os key.  It looks like those will be needed in order for Stack to still be able to install FreeBSD bindists (I don't have a FreeBSD machine handy to verify that, though).

arrowd:
After thinking about this again, it turned that we don't really need the proposed change. All is needed is properly-built bindists that work on all FreeBSD versions.

Issue #4272, opened 2018-08-30
Improved recompilation avoidance


snoyberg:
Each module which enables the `TemplateHaskell` language pragma gets recompiled more often than necessary when dependencies change. It would be nice to bypass that by removing the usage of TemplateHaskell. A number of modules in the codebase are only using TemplateHaskell to safely create `Path` values, e.g. `mkRelFile` or `mkRelDir`. I recommend collecting these usages into a single module, likely `Stack.Constants`, and importing for there. That would allow us to reduce recompilation time and improve the project iteration cycle a bit.

I'm open to this being shot down, since I'm not at all the expert on fast iteration cycles :). CC @chrisdone and @bitemyapp.

chrisdone:
Agreed! :ok_hand:

snoyberg:
See #4273.

PR #4273, opened 2018-08-30
Avoid unnecessary recompilations

snoyberg:
This PR reduces the number of modules in the codebase which enable either `TemplateHaskell` or `CPP`, via a few techniques:

* Moving conditional code into separate Windows and Unix directories
* Moving all of the Store-related TH into a single module
* Moving `mkRelDir` and `mkRelFile` calls into `Stack.Constants`
* Replacing a few CPP usages with runtime checking of the OS, which I prefer using anyway (it reduces the amount of code which is only built on one OS or the other)

There should be no impact from these changes besides how it affects compile time.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -231,6 +231,7 @@ library:
   - Stack.Sig.Sign
   - Stack.Snapshot
   - Stack.Solver
+  - Stack.StoreTH
   - Stack.Types.Build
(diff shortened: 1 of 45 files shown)


snoyberg:
Note that this resolves #4272

PR #4285, opened 2018-09-03
Migrate to Cabal 2.4

snoyberg:

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -30,7 +30,7 @@ module Stack.Package
   where
 
 import qualified Data.ByteString.Lazy.Char8 as CL8
-import           Data.List (isSuffixOf, isPrefixOf, unzip)
+import           Data.List (isPrefixOf, unzip)
 import           Data.Maybe (maybe)
 import qualified Data.Map.Strict as M
 import qualified Data.Set as S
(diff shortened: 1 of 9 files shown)


src/Stack/Package.hs
```diff
@@ -732,7 +735,7 @@ resolveGlobFiles =
         mapM resolveFileOrWarn names
     matchDirFileGlob' dir glob =
         catch
-            (matchDirFileGlob_ dir glob)
+            (liftIO (matchDirFileGlob minBound cabalFileVersion dir glob))
```
quasicomputational: Stack previously used a patched version of this function from Cabal, removed in this same diff, which allowed the extensions to be a suffix rather than matching strictly exactly. This behaviour is now present in Cabal's functions, but *only* if the spec version passed in is at least 2.4, so any older packages may see a behaviour change here.

On the other hand, this is now consistent with cabal-install's behaviour and what you get from `runghc Setup.hs`, and also corresponds to what the Cabal docs say and have said for ages, so I think this ought to be OK. Even so, I'd mention it in the changelog.

src/Stack/Package.hs
```diff
@@ -693,7 +693,7 @@ packageDescModulesAndFiles pkg = do
             (mapM
                  (asModuleAndFileMap benchComponent benchmarkFiles)
                  (benchmarks pkg))
-    dfiles <- resolveGlobFiles
+    dfiles <- resolveGlobFiles (specVersion pkg)
                     (extraSrcFiles pkg
                         ++ map (dataDir pkg FilePath.</>) (dataFiles pkg))
```
quasicomputational: While I'm looking at this code I just want to point out that this will misfire if `dataDir` has an asterisk in it anywhere. That's not new in this patch. `*` isn't a legitimate path character on Windows anyway, so it's mostly a theoretical concern.

PR #4293, opened 2018-09-05
Reduce warnings

snoyberg:
This reduces the places in the codebase where we need to turn off warnings. The main move here is inlining the parts of `store` that we're using. There's an additional goal with this move: we're able to strip out the pieces of `store` we don't need to reduce the compile time a bit more. Over time, as we move more of our data serialization away from store, we'll be able to remove more of the code until, ultimately, it disappears before the next release.

This also takes a stab at cleaning up the pretty print code a bit, mostly monomorphizing it for better error messages, and renaming the `Display` typeclass to `Pretty` to avoid a double name conflict (both with Cabal and rio).


snoyberg:
Clarification: the main reason for monomorphising wasn't better error messages. I already needed a `newtype` wrapper to avoid orphans instances, so making the newtype monomorphic was a nice addition. If we split off the pretty printer code to its own library, we can include the missing instances and keep things generic.

PR #4299, opened 2018-09-08
Document an overview of builds

snoyberg:
With all of the various refactorings at play, I wanted to document what
the end goal of this whole process would be, both to guide my own work,
an to get feedback from others (either to improve the process, or
clarify it so others can participate in the maintenance of this better).


snoyberg:
@ezyang If you have any feedback on how this may affect Backpack, please let me know. As far as I know, the described procedure here should make it significantly easier to add Backpack (which is one of my explicit goals in this overhaul).

doc/build-overview.md
```diff
@@ -0,0 +1,251 @@
...
+* Fall back to the default global project
+
+This file is parsed to provide the following config values:
+
+* `resolver`
+* `compiler`
+* `packages`
+* `extra-deps`
+* `flags`
+* `ghc-options`
```
mihaimaruseac: I'd add a line specifying what's done if a `stack.yaml` file only has a subset of these, for completeness.

doc/build-overview.md
```diff
@@ -0,0 +1,251 @@
...
+  dependency from Hackage. If so, create an immutable dependency entry
+  with default flags, GHC options, and hidden status, and add this
+  package to the set of immutable package dependencies.
+* For all package names, ensure the package is in one of the four maps
+  we have, and if so add to either the dependency or project package
+  target set.
+* For all package name + component, ensure that the package is a
+  project package, and add that package + component to the set of
+  project targets.
+* Ensure that no target has been specified multiple times.
```
mihaimaruseac: I think we will need an extra consistency step for internal libraries. Sometimes stack needs to use the mangled name (`z-package-internallibname-z..`), sometimes the `package:internallibname` one. But I think this will become obvious when doing the code changes.

doc/build-overview.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Construct build plan
+
+* Applied to every target (project package or dependency)
+* Apply flags, platform, and actual GHC version to resolve
+  dependencies in any package analyzed
+* Include all library dependencies for all enabled components
+* Include all build tool dependencies for all enabled components
+  (using the fun backwards compat logic for `build-tools`)
+* Apply the logic recursively to come up with a full build plan
```
ezyang: A note here: one awkward thing about formulating build plan construction this way is that you might reach package p via an edge that says "I only need the library", and then later in the recursive traversal you reach p again where it says, "Oh, I also need the executable." So if you want to implement it this way, you have to have a notion where you can take a package that is already in your plan, and increase the components that are enabled for it. If you can truly assume that all packages support per-component builds, then you can just memoize over components and there isn't any problem; this might help.

doc/build-overview.md
```diff
@@ -0,0 +1,251 @@
...
+  database, the latter into the local database.
+
+We now have a set of tasks of packages/components to build, with full
+config information for each package, and dependencies that must be
+built first.
+
+*FIXME* There's some logic to deal with cyclic dependencies between
+test suites and benchmarks, where a task can be broken up into
+individual components versus be kept as a single task. Need to
+document this better. Currently it's the "all in one" logic.
```
ezyang: Backpack would hook into this step: when you were configuring a component, you'd also be obligated to mix-in link it (assuming that the things that it referred to have recursively already been resolved).

Then, after everything is said and done, you can go over *fully instantiated* references to packages, and add extra jobs to the build graph to build the instantiated forms of those packages.

doc/build-overview.md
snoyberg: I'll add defaulting comments

doc/build-overview.md
snoyberg: I'm going to copy your comment in as a FIXME.

doc/build-overview.md
snoyberg: Is there some other approach to build plan construction which doesn't present this issue? I'm also not quite following your last sentence, can you clarify?

ezyang: > Is there some other approach to build plan construction which doesn't present this issue?

Well, there are a few alternate approaches, but they have their own tradeoffs.

1. You could say, recursive traversal operates over *components*, not packages. So you'll "visit" packages multiple times, but you'll only visit any given component once. The downside is that you'll end up having split up a package into multiple components to build, but some packages can't be built on a per-component basis (e.g., a package with a Custom setup) so you'll have to agglomerate them all back together before you actually do a build. (This is what I was alluding to in the last sentence.)

2. You could process per-package, but say, "Actually, I'm going to assume that *all* components are requested." So you will never have to improve because you always greedily configured every component, so it must be available (or be impossible). The downside is that you spent a whole bunch of work figuring out a build plan to build everything, which is not actually what you want, so you're going to have to prune it down after the fact. You probably don't want to be recomputing this every time, so you had better cache the plan. (By the way, this is what cabal new-build does).

doc/build-overview.md
ezyang: You're still able to create a complete build graph during plan construction. Actually, I suppose you can add the instantiated packages as you're initially building the build graph, but I think it's easier to do split it into two phases, just from a code complexity perspective.

doc/build-overview.md
snoyberg: Makes sense. I think I'll leave the doc as-is for now, and update it during implementation. I have a feeling we'll end up closer to your (1).

snoyberg:
Thanks for the feedback @mihaimaruseac and @ezyang. I'm going to merge this and start using it as a blueprint for further work.

PR #4302, opened 2018-09-08
 Fix #4297, and simplify: use `hSupportsANSIWithoutEmulation`

### LINKED ISSUE #4297: Bug in module `Pantry.Repo`?
Module `Stack.Upgrade` explains (at around line 203) an upstream bug in `git clone` and the fix in `stack`:
````haskell
let args = [ "clone", repo , "stack", "--depth", "1", "--recursive", "--branch", branch]
withWorkingDir (toFilePath tmp) $ proc "git" args runProcess_
-- On Windows 10, an upstream issue with the `git clone` command
-- means that command clears, but does not then restore, the
-- ENABLE_VIRTUAL_TERMINAL_PROCESSING flag for native terminals.
-- The folowing hack re-enables the lost ANSI-capability.
when osIsWindows $ void $ liftIO defaultColorWhen
````
I think that module `Pantry.Repo` may now expose `stack` to the same upstream `git clone` bug, at around lines 87 to 91 (`commandName` is `"git"`):
```` haskell
    logInfo $ "Cloning " <> display commit <> " from " <> display url
    void $ proc
      commandName
      ("clone" : cloneArgs ++ [T.unpack url, suffix])
      readProcess_
````
I say that: (1) looking at the code above; and (2) because using (on Windows 10) a version of `stack` derived from the current 'master' branch, I experienced 'raw' ANSI codes in the output stream and I think that was just after the `"Cloning ... from"` had been output. It is difficult for me to reproduce the circumstances where that occurred.

On the assumption that the above diagnosis is correct, I will propose a pull request to include the same fix as in module `Stack.Upgrade`.

EDIT: Actually, that is not straightforward, because the fix depends on modules `Stack.Constants.osIsWindows` and `Stack.DefaultColorWhen.defaultColorWhen`. However, the next version of the `ansi-terminal` package, hopefully on Hackage soon, will include ANSI-enabling code that `stack` and `pantry` can rely upon.

mpilgrem:
This pull request is organised into three commits:

The first makes use of (new) `hSupportsANSIWithoutEmulation :: Handle -> IO (Maybe Bool)` from `ansi-terminal >= 0.8.1`. (See the documentation in that package for full details.) That utility function allows module `Stack.DefaultColorWhen` to be simplified (particularly on Windows) and, because the Unix/Windows distinction is made in `ansi-terminal`, the Unix/Windows distinction in `stack` is no longer necessary. There is no change in functionality.

The second makes use of `hSupportsANSIWithoutEmulation` directly (rather than via `defaultColorWhen`) in the existing upstream `git clone` bug fix in module `Stack.Upgrade`. There is no change in functionality.

The third applies the same `git clone` fix to module `Pantry.Repo`. On the assumption that `pantry` is to be independent of `stack`, the `osIsWindows` in `stack`'s `System.Permissions` is replicated in `pantry` (including its approach to implementing the Unix/Windows distinction).

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md. (None.)
* [x] The documentation has been updated, if necessary. (None necessary.)

Tested on Windows 10 and macOS by building, and using the built tool.


mihaimaruseac:
Thank you

PR #4306, opened 2018-09-10
Update package.yaml, not .cabal file

trotha01:

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -95,8 +95,8 @@ So to manage your library:
 2. If you need to include another library (for example the package
    [`text`](https://hackage.haskell.org/package/text)):
 
-   - Add the package `text` to the file `my-project.cabal`
-     in the section `build-depends: ...`.
+   - Add the package `text` to the file `package.yaml`
+     in the section `dependencies: ...`.
    - Run `stack build` another time.
 
 3. If you get an error that tells you your package isn't in the LTS.
--- a/doc/README.md
+++ b/doc/README.md
@@ -98,6 +98,9 @@ So to manage your library:
    - Add the package `text` to the file `package.yaml`
      in the section `dependencies: ...`.
    - Run `stack build` another time.
+   - `stack build` will update my-project.cabal for you.
+     If desired you can update the .cabal file manually
+     and stack will use .cabal instead of package.yaml.
 
 3. If you get an error that tells you your package isn't in the LTS.
    Just try to add a new version in the `stack.yaml` file in the `extra-deps` section.


mattaudesse:
Thanks @trotha01 but I think this section actually needs to remain as written since not everyone uses `package.yaml`.

Alternatively (and maybe you'd like to update the PR to show this?) we could explain how both work and note that `package.yaml`'s contents will automatically generate the cabal file at build time if present.

trotha01:
From what I understand, the default flow, for the current stable version, is to update package.yaml. Especially since it says this later in the docs:

> Since Stack 1.6.1, the package.yaml is the preferred package format that is provided built-in by stack through the hpack tool. The default behaviour is to generate the .cabal file from this package.yaml, and accordingly you should not modify the .cabal file.


Given that this is the quickstart guide, it seems like we should point people to updating package.yaml and leave manual edits to the autogenerated .cabal file be a special case (which we can link to if you like). Especially since a Warning is given if users edit the .cabal file. The warning can be confusing to new users of stack.
```
Warning: <path>/<project>.cabal was modified manually. Ignoring <path>/package.yaml in favor of the cabal
         file. If you want to use the package.yaml file instead of the cabal file, then please delete the cabal file.
```

mattaudesse:
Yeah, I think you're right that we should lean toward `package.yaml` as long as we don't confuse the people who don't have one.

Consider the case of somebody who uses the `simple` template to bootstrap a new project:

```bash
--> stack new beginner-proj simple
Downloading template "simple" to create project "beginner-proj" in beginner-proj/ ...
Looking for .cabal or package.yaml files to use to init the project.
Using cabal packages:
- beginner-proj/

Selecting the best among 14 snapshots...

* Matches lts-12.9

Selected resolver: lts-12.9
Initialising configuration using resolver: lts-12.9
Total number of user packages considered: 1
Writing configuration to file: beginner-proj/stack.yaml
All done.

--> tree beginner-proj/
beginner-proj/
├── LICENSE
├── README.md
├── Setup.hs
├── beginner-proj.cabal
├── src
│   └── Main.hs
└── stack.yaml

1 directory, 6 files
```

As long as there's some quick blurb to note the difference I think we're good:
> if you're using a `package.yaml` then ...; otherwise, ...

Thanks again for catching this.

trotha01:
I ended up adding the information as a separate bullet point since the Quickstart Guide starts with the command `stack new my-project`. Let me know if that works. Thanks for the help.

mattaudesse:
Cool, thanks @trotha01.

PR #4312, opened 2018-09-19
Add `--cabal-files` flag to `ide packages` command

DanielG:
With the `--cabal-files` flag active `stack ide packages` will print the paths to packages' cabal-files instead of just the package name. Having the full path to all packages within a Stack project will allow me to expose a more unified project representation in  [`cabal-helper`](https://github.com/DanielG/cabal-helper) in light of upcoming `cabal new-build` support.

---

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -90,6 +90,7 @@ Bug fixes:
 * Fix `subdirs` for git repos in `extra-deps` to match whole directory names.
   Also fixes for `subdirs: .`. See
   [#4292](https://github.com/commercialhaskell/stack/issues/4292)
+* Add `--cabal-files` flag to `stack ide targets` command.
 
 
 ## v1.9.0.1 (release candidate)
--- a/src/Stack/IDE.hs
+++ b/src/Stack/IDE.hs
@@ -7,7 +7,8 @@
 
 -- | Functions for IDEs.
 module Stack.IDE
-    ( listPackages
+    ( ListPackagesCmd(..)
+    , listPackages
     , listTargets
     ) where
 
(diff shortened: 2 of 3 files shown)


ChangeLog.md
```diff
@@ -90,7 +90,7 @@ Bug fixes:
 * Fix `subdirs` for git repos in `extra-deps` to match whole directory names.
   Also fixes for `subdirs: .`. See
   [#4292](https://github.com/commercialhaskell/stack/issues/4292)
-
```
snoyberg: Please keep the extra newline following the list, I believe it's needed by tooling (but not quite sure tbh).

DanielG:
Fixed that whitespace thing.

--- 

@dbaynard I didn't mean to imply Stack intearacts badly with new-build in some way, having this info just makes it a lot easier to have the same project representation for cabal and stack projects in cabal-helper without the user of the library having to worry about it.

You see new-build has this `plan.json` thing where it just writes out how every package/component got configured and where they are so I use that as the main entry point into finding out what's going on in a project. Previously that was LocalBuildInfo/`dist/setup-config`'s job but now you sort of have another layer of indirection on top of that where `plan.json` tells you where the `setup-config`'s are and incidentally it also lists the directories in the project dir where cabal files are.

Now it used to be that I just figure out where Stack's dist-dirs are by searching for the cabal file corresponding to a given Haskell source file by traversing up in the filesystem until I hit one and then using `stack path` to get the dist-dir path. In order to mirror what I do for new-build now I actually need a list of cabal packages beforehand though.

I think this approach of building a full project representation will also make it much easier for tooling to give the user more information about what's going on in the project, like what targets exist, where are the sources and so on (basically what IDEs usually do), so it's not just about making Stack conform to what new-build does. Besides it makes it possible to completely hide all the necessary multiple levels of caching from the library user, i.e. me in ghc-mod, so :)

dbaynard:
Sounds good, thanks! For a PR without an issue, it's useful to see what the PR is supposed to change.

PR #4314, opened 2018-09-21
Update git package submodules after clone and reset

tclem:
Related to https://github.com/commercialhaskell/stack/issues/1764 and a subtle fix for what was implemented in https://github.com/commercialhaskell/stack/pull/1852, this changes the logic for packages identified by git locations so that submodules are properly updated. The process now looks like this:

1. clone the repo
2. `reset --hard` to a specific commit
3. update submodules (recursively)

The problem with the existing implementation is that a hard reset (`git reset --hard`) doesn't update submodules so the state of submodules in a package will always be whatever's on the default branch.

I tested a similar version of this patch against the [v1.9.0.1 tag](https://github.com/commercialhaskell/stack/releases/tag/v1.9.0.1), but I was unable to get stack on master to properly read the `stack.yaml` of my project. To test requires a little bit of setup. You need:

- A stack project with a package identified using a [git location](https://docs.haskellstack.org/en/v1.2.0/yaml_configuration/#complex-package-locations-location)
- A package that uses submodules
- A commit SHA of that package where the submodule SHAs are different than what's on the default branch.

TODO:

* [ ] ~~The documentation has been updated, if necessary.~~ I don't believe there's anything to update here, but please let me know!

### DIFF
--- a/subs/pantry/src/Pantry/Repo.hs
+++ b/subs/pantry/src/Pantry/Repo.hs
@@ -76,21 +76,20 @@ getRepo' repo@(Repo url commit repoType' subdir) pm =
             RepoGit ->
               ( "git"
               , ["reset", "--hard", T.unpack commit]
-              , ["submodule", "update", "--init", "--recursive"]
+              , Just ["submodule", "update", "--init", "--recursive"]
               , ["archive", "-o", tarball, "HEAD"]
               )
             RepoHg ->
               ( "hg"
               , ["update", "-C", T.unpack commit]
-              , []
+              , Nothing
               , ["archive", tarball, "-X", ".hg_archival.txt"]
               )
 
+    let runCommand args = void $ proc commandName args readProcess_
+
     logInfo $ "Cloning " <> display commit <> " from " <> display url
-    void $ proc
-      commandName
-      ("clone" : [T.unpack url, suffix])
-      readProcess_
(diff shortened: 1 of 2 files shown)


dbaynard:
Hello @tclem,

Git behaviour is changing for the next stack version (after 1.9.*) — @snoyberg will know more; I've requested a review.

subs/pantry/src/Pantry/Repo.hs
```diff
@@ -71,25 +71,25 @@ getRepo' repo@(Repo url commit repoType' subdir) pm =
...
-              , ["--recursive"]
               , ["reset", "--hard", T.unpack commit]
+              , ["submodule", "update", "--init", "--recursive"]
               , ["archive", "-o", tarball, "HEAD"]
               )
             RepoHg ->
               ( "hg"
-              , []
               , ["update", "-C", T.unpack commit]
+              , []
```
snoyberg: Doesn't this mean that, for Mercurial, we'll be calling `hg` here with no arguments? I think using a `Nothing` value here and checking for that below would be appropriate.

tclem: 👍 Good suggestion. There's definitely no reason to have an extra shell out to `hg` with no arguments, I pushed another commit to fix that.

Also happy to squash things up back to a single commit, but I might wait for further feedback and do that once at the end.

PR #4317, opened 2018-09-24
Use Cabal 2.4 for the 1.9 branch

snoyberg:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,10 @@ Release notes:
 Major changes:
 
 * Upgrade to Cabal 2.4
+    * Note that, in this process, the behavior of file globbing has
+      been modified to match that of Cabal. In particular, this means
+      that for Cabal spec versions less than 2.4, `*.txt` will
+      match `foo.txt`, but not `foo.2.txt`.
 
 Behavior changes:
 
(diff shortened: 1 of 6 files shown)


ChangeLog.md
```diff
@@ -26,6 +26,7 @@ Release notes:
 Major changes:
 
 * `GHCJS` support is being downgraded to 'experimental'. A warning notifying the user of the experimental status of `GHCJS` will be displayed.
+* Upgrade to Cabal 2.4
```
borsboom: Can you move this to the "Unrelease changes" section?  As part of making the final release, I consolidate the changelogs for all the RCs, but this helps to know what's changed between RCs.

package.yaml
```diff
@@ -15,7 +15,7 @@ homepage: http://haskellstack.org
 custom-setup:
   dependencies:
   - base >=4.10 && < 5
-  - Cabal
+  - Cabal >= 2.4
```
borsboom: I don't think this constraint is needed.  We avoid putting any versions constraints in `package.yaml` since the `stack.yaml` will ensure the right version is used (and we use something like of `--pvp-bounds=both` when uploading to Hackage).

borsboom:
Merge whenever ready.

PR #4320, opened 2018-09-26
Add System.Uname as an 'other-module' in package.yaml.

Mokosha:
Additionally, we need to add a dummy System.Uname in the windows sources
directory in order to workaround an issue where hpack doesn't support the
'other-modules' directive in library conditionals.

I ran `stack build`. My sense is that `hpack` will usually accurately infer the contents of `other-modules`. However, on my machine, it was not adding `System.Uname` after the commit 9b61953e320796875df30646001b4dd1eefec1eb. This adds the module explicitly to `other-modules` and places a dummy module in the windows sources in order to avoid having "not found" errors on that platform.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -133,8 +133,6 @@ library:
   source-dirs: src/
   ghc-options:
   - -fwarn-identities
-  generated-exposed-modules:
-  - Paths_stack
   exposed-modules:
   - Control.Concurrent.Execute
   - Data.Attoparsec.Args
@@ -150,6 +148,7 @@ library:
   - Path.CheckInstall
   - Path.Extra
   - Path.Find
+  - Paths_stack  # Generated
   - Stack.Build
   - Stack.Build.Cache
   - Stack.Build.ConstructPlan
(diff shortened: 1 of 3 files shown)


Mokosha:
I believe `generated-exposed-mudules` has been [removed](https://github.com/sol/hpack#library-fields) in the latest versions of `hpack`, which is used by `stack` at `HEAD`. At least, on my machine I was receiving the warnings:

    Warning: Ignoring unknown field "generated-exposed-modules" in library section            
    Warning: Ignoring unknown field "generated-other-modules" in executable section "stack" 

If these should have remained, I'm happy to revert those lines (they weren't causing build problems for me).

quasicomputational:
`generated-*-packages` are still in hpack. Are you using an old version of either stack or hpack?

quasicomputational:
No, the hpack docs are saying that those fields were added in version 0.23 (i.e., using them with hpack 0.22 won't work, but hpack 0.24 and later will).

Are you still getting those warnings, even with that stack pre-release? It works fine for me when I manually run hpack 0.31, and my stack 1.7 installation also works fine.

Mokosha:
Wow, I totally misread that, coupled with the fact that the `stack` that I was using to build `stack` was `1.6.5` with `hpack-0.24`. Thank you for catching that, I've reverted those changes and they work fine for me.

Mokosha:
After upgrading, I noticed this change was not necessary, too. Sorry for the noise.

PR #4330, opened 2018-10-01
Remove suggestions for legacy `packages` syntax

fosskers:
`stack` version: `master` as of `a299ac652fde49d3b4b6a199c1a206220826da68`

Stack no longer parses the `location` syntax, but still suggests it to users in examples in a `stack.yaml` produced by `stack new`. 

This PR moves the suggestions to an improved comment section for `extra-deps`.

### DIFF
--- a/src/Stack/Init.hs
+++ b/src/Stack/Init.hs
@@ -234,7 +234,7 @@ renderStackYaml p ignoredPackages dupPackages =
         [ ("user-message"     , userMsgHelp)
         , ("resolver"         , resolverHelp)
         , ("packages"         , packageHelp)
-        , ("extra-deps"       , "# Dependency packages to be pulled from upstream that are not in the resolver\n# using the same syntax as the packages field.\n# (e.g., acme-missiles-0.3)")
+        , ("extra-deps"       , extraDepsHelp)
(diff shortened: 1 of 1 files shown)


src/Stack/Init.hs
```diff
@@ -274,15 +274,23 @@ renderStackYaml p ignoredPackages dupPackages =
         , "packages:"
         , "- some-directory"
         , "- https://example.com/foo/bar/baz-0.0.2.tar.gz"
```
fosskers: Is this still legal, or should this be in `extra-deps` too?

src/Stack/Init.hs
```diff
@@ -274,15 +274,23 @@ renderStackYaml p ignoredPackages dupPackages =
...
-        , "- location:"
-        , "   git: https://github.com/commercialhaskell/stack.git"
-        , "   commit: e7b331f14bcffb8367cd58fbfc8b40ec7642100a"
-        , "- location: https://github.com/commercialhaskell/stack/commit/e7b331f14bcffb8367cd58fbfc8b40ec7642100a"
         , " subdirs:"
```
fosskers: Is the indentation here correct?

fosskers:
The CI failures seem to be a Travis timeout.

PR #4338, opened 2018-10-04
Stack release process custom gpg key support

borsboom:


etc/scripts/with-vagrant.sh
```diff
@@ -4,7 +4,9 @@ set -xe
 export STACK_BUILD_DIR="$PWD"
 cd "$(dirname "$0")/../.."
 #TODO: set up gpg-agent forwarding for package signing (see http://superuser.com/questions/161973/how-can-i-forward-a-gpg-key-via-ssh-agent).
-gpg --export-secret-keys --armor dev@fpcomplete.com >"gpg-secret-key.asc~"
+
+gpg_key_name="${STACK_RELEASE_GPG_KEY_NAME:-dev@fpcomplete.com}" # :- means get from env or use default
```
borsboom: I don't think you need the `STACK_RELEASE_GPG_KEY_NAME` variable.  In my testing, `gpg --export-secret-key` also works with the 0xLONG key ID.

etc/scripts/release.hs
```diff
@@ -89,17 +92,32 @@ main =
...
+envOptionsIO = do
+    envVars <- getEnvironment
+    let modifyByEnv global (k, v) = case k of
+          "STACK_RELEASE_GPG_KEY" -> global{gGpgKey = v}
+          _ -> global
+    return $ \global -> foldl modifyByEnv global envVars
+
 -- | Additional command-line options.
 options :: [OptDescr (Either String (Global -> Global))]
 options =
```
borsboom: The description of the `--gpg-key` argument should be expanded to mention the environment variable as well, similar to how the description of the `--github-auth-token` argument does it.  E.g., the description string would become something like
```
"GPG user ID to sign distribution package with (defaults to " ++ gpgKeyEnvVar ++ " environment variable)."
```

etc/scripts/release.hs
```diff
@@ -89,17 +92,32 @@ main =
...
                     { gProjectRoot = projectRoot' }
             return $ Just $ rules global args
   where
     getStackPath global path = do
       out <- readProcess stackProgName (stackArgs global ++ ["path", "--" ++ path]) ""
       return $ trim $ fromMaybe out $ stripPrefix (path ++ ":") out
 
+-- | Env Vars to set some of the command-line options found below.
+envOptionsIO :: IO (Global -> Global)
+envOptionsIO = do
```
borsboom: Why did you take a totally different approach than is used for handling the GITHUB_AUTH_TOKEN variable?  I would have thought you could just copy that approach since it's basically doing the same thing (only difference is that there's a default).  E.g. in `main` have `gGpgKey <- fromMaybe "0x575159689BEFB442" <$> lookupEnv gpgKeyEnvVar` instead of `let gGpgKey = "0x575159689BEFB442"`.  Much shorter code, and also more efficient than iterating through all the environment variables.

While in there, the default key should be moved to a `defaultGpgKey` constant instead of being a magic value here (not sure why I didn't do that originally).

etc/scripts/release.hs
```diff
@@ -89,17 +92,32 @@ main =
...
+    envVars <- getEnvironment
+    let modifyByEnv global (k, v) = case k of
+          "STACK_RELEASE_GPG_KEY" -> global{gGpgKey = v}
```
borsboom: The name of the environment variable should be a constant (e.g. named `gpgKeyEnvVar`), similar to how `githubAuthTokenEnvVar` is a constant.

etc/scripts/README.md
```diff
@@ -21,7 +21,12 @@ To create a signed binary package, you need:
 
 - GPG installed and in the PATH (included with
   [msysgit](https://msysgit.github.io) on Windows)
-- `dev@fpcomplete.com` secret key in GPG keyring.
+- `dev@fpcomplete.com` secret key in GPG keyring. You may also use the
+  environment variables `STACK_RELEASE_GPG_KEY` and
+  `STACK_RELEASE_GPG_KEY_NAME`. The variable `STACK_RELEASE_GPG_KEY` should be
+  set to the hexadecimal identifier of the GPG key, while
```
borsboom: GPG calls this the `0xLONG` format, so you might want to write
```
set to the hexadecimal identifier (0xLONG format) of the GPG key
```

borsboom:
@damiancoding Added some review comments.

PR #4359, opened 2018-10-15
Make `stack setup` command differentiate between FreeBSD versions.

arrowd:
The upcoming FreeBSD 12.0 release have some serious ABI changes. These changes prevent GHC's compiled on FreeBSD 10 and 11 from functioning correctly on FreeBSD 12. Stack would need to differentiate pre-12 FreeBSDs and post-12 ones.

This patch makes `stack setup` to append `-ino64` suffix when running on FreeBSD 12+. Once last pre-12 FreeBSD release goes EoL, this patch will be removed as well as `-ino64` entries from `stack-setup-2.yaml`.

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -96,6 +96,7 @@ import              System.FilePath (searchPathSeparator)
 import qualified    System.FilePath as FP
 import              System.Permissions (setFileExecutable)
 import              RIO.Process
+import              RIO.List
 import              Text.Printf (printf)
 
 #if !WINDOWS
@@ -614,6 +615,13 @@ getGhcBuilds = do
                         _ -> CompilerBuildSpecialized (intercalate "-" c))
                     libComponents
 #if !WINDOWS
+            Platform _ Cabal.FreeBSD -> do
+                let getMajorVer = readMaybe <=< headMaybe . (splitOn ".")
+                majorVer <- getMajorVer <$> sysRelease
+                if majorVer >= Just (12 :: Int) then
+                  useBuilds [CompilerBuildSpecialized "ino64"]
+                else
+                  useBuilds [CompilerBuildStandard]
             Platform _ Cabal.OpenBSD -> do
                 releaseStr <- mungeRelease <$> sysRelease
                 useBuilds [CompilerBuildSpecialized releaseStr]


src/Stack/Setup.hs
```diff
@@ -614,6 +614,12 @@ getGhcBuilds = do
                         _ -> CompilerBuildSpecialized (intercalate "-" c))
                     libComponents
 #if !WINDOWS
+            Platform _ Cabal.FreeBSD -> do
+                majorVer <- (maybe (0 :: Int) id) . readMaybe . head . (splitOn ".") <$> sysRelease
+                if majorVer >= 12 then
```
dbaynard: This implicitly handles the `Nothing`, as `Just x > Nothing` for all `x`.

```suggestion
                if majorVer >= Just 12 then
```

dbaynard:
Fixes #3515?

myfreeweb:
haskell.org should actually publish -ino64 builds first, right?

arrowd:
Well, since nothing works for FreeBSD 12 already, this may be merged before -ino64 bindists are actually prepared. I plan to produce them myself, host them at my site and add them to `stack-setup-2.yaml`.

borsboom:
There are some similarities with this closed PR: #4269, which would specify exact versions of FreeBSD in the OSKey (like we do for OpenBSD).  I don't know that much about FreeBSD, but sounds like that approach may have been overkill since most GHC bindists can be compatible with many FreeBSD versions.  Overall, I think I like this better assuming it works.  @arrowd, can you offer an opinion on that old PR's approach vs. your approach?

arrowd:
> @arrowd, can you offer an opinion on that old PR's approach vs. your approach?

That PR was also done by me. In short, we don't need to differentiate between all FreeBSD versions, just pre-12 and 12+ ones. Moreover, when the last pre-12 release goes EoL, we can remove this patch and use `freebsd{32,64}` OSKeys again.

The old PR was really overkill, as it was extracting both minor and major version numbers and appended them to OSKey.

borsboom:
Ok, that makes sense.  I think this is a good approach.

arrowd:
What changes are now required to get this in?

dbaynard:
Ah, nothing — thanks for making the changes, I didn't notice you'd pushed another commit (thanks for the prompt, too!). I just want to check what's happening with travis and then I'll merge.

arrowd:
Why this change doesn't seem to be present in v1.9.3? https://github.com/commercialhaskell/stack/blob/v1.9.3/src/Stack/Setup.hs#L625

This broke stack for FreeBSD 12+ again.

dbaynard:
@arrowd Minor releases are taken from `stable` branch, not master. I'm cherry picking — just running integration tests.

arrowd:
I had an impression, that this commit was made before 1.9.0. That's a pity.

Thanks for pulling it into stable, anyways!

Issue #4364, opened 2018-10-19
1.9.1 fails to build from tarball without .git/


juhp:
### General summary/comments (optional)
Getting compilation errors in Main.hs see below for details

### Steps to reproduce
Build 1.9.1 in a cabal sandbox.

Buildlog: https://copr-be.cloud.fedoraproject.org/results/petersen/stack/fedora-rawhide-x86_64/00812150-stack/build.log.gz

(See https://copr-dist-git.fedorainfracloud.org/cgit/petersen/stack/stack.git/tree/stack.spec?id=d786c497ff6c79c885688e8bd7976d31eaf849fa#n139 for exact build steps.)

### Expected
Successful build
### Actual
Fail failed:

```
[3 of 3] Compiling Main             ( src/main/Main.hs, dist/dist-sandbox-ccff8d/build/stack/stack-tmp/Main.o )
src/main/Main.hs:131:16: error:
    * Ambiguous type variable `a0' arising from a use of `show'
      prevents the constraint `(Show a0)' from being solved.
      Probable fix: use a type annotation to specify what `a0' should be.
      These potential instances exist:
        instance [safe] Show (Doc a)
          -- Defined in `Text.PrettyPrint.Annotated.Leijen'
        instance Show a => Show (ZipList a)
          -- Defined in `Control.Applicative'
        instance (Show b, Show a) => Show (Either a b)
          -- Defined in `Data.Either'
        ...plus 124 others
        ...plus 282 instances involving out-of-scope types
        (use -fprint-potential-instances to see them all)
    * In the first argument of `(++)', namely `show commitCount'
      In the second argument of `(++)', namely
        `show commitCount ++ " commits)"'
      In the expression: " (" ++ show commitCount ++ " commits)"
    |
131 |     , [" (" ++ show commitCount ++ " commits)" | commitCount /= 1]
    |                ^^^^^^^^^^^^^^^^
src/main/Main.hs:136:33: error:
    * Exception when trying to run compile-time code:
        git: readCreateProcessWithExitCode: runInteractiveProcess: exec: does not exist (No such file or directory)
      Code: tGitInfoCwd
    * In the Template Haskell splice $$tGitInfoCwd
      In the first argument of `giCommitCount', namely `$$tGitInfoCwd'
      In the expression: giCommitCount $$tGitInfoCwd
    |
136 |     commitCount = giCommitCount $$tGitInfoCwd
    |                                 ^^^^^^^^^^^^^
[...]

PR #4368, opened 2018-10-21
Compile without a Git dir #4364

snoyberg:

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -298,7 +298,7 @@ executables:
     - condition: ! '!(flag(disable-git-info))'
       cpp-options: -DUSE_GIT_INFO
       dependencies:
-      - githash
+      - githash >= 0.1.2
       - optparse-simple
     - condition: flag(hide-dependency-versions)
       cpp-options: -DHIDE_DEP_VERSIONS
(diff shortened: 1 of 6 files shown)


package.yaml
```diff
@@ -298,7 +298,7 @@ executables:
     - condition: ! '!(flag(disable-git-info))'
       cpp-options: -DUSE_GIT_INFO
       dependencies:
-      - githash
+      - githash >= 0.1.2
```
borsboom: We omit version constraints from `package.yaml` since they're implied by `stack*.yaml`.  OK to leave for now though, since we do a sweep looking for any unnecessary constraints before release.

dbaynard:
This can be merged, right?

snoyberg:
Yes, I think so. Do you want me to do so, or did you want to?

borsboom:
I cherry-picked this change into the `release` branch and made a hackage-only stack-1.9.1.1 release with it so that I could unblock `stack` in in Stackage.  Some people still install it that way or with `cabal install`, and moreover I like that Stackage will inform me if the build breaks with newer dependencies.

PR #4371, opened 2018-10-23
Support KDE Neon distribution

FloEdelmann:
KDE Neon is based on Ubuntu, so the Ubuntu installer should work fine.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,9 @@ Behavior changes:
 
 Other enhancements:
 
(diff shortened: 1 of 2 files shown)


borsboom:
This looks fine.  Can you rebase this PR against the `stable` branch?  That way we can merge it and make it live at get.haskellstack.org right away.

FloEdelmann:
Okay, after fiddling around with `git rebase` a bit, it should be correct now.

PR #4372, opened 2018-10-23
Fix #4085: insufficient error message

### LINKED ISSUE #4085: Recommend rerunning with --verbose on failed setup
Motivation: https://stackoverflow.com/questions/50709844/is-there-an-issue-with-haskell-ghc-8-4-3-on-travis.

As can be seen from that discussion, the default output of `ExitFailure 2` is insufficient to debug the problem, which in this case was an "out of space" condition. We should either:

1. Recommend rerunning when setup fails
2. Automatically dump the output

At the very least, this applies to setting up GHC, but may apply to other parts of the codebase as well.

aleksejkozin:

---

Fixing [#4085](https://github.com/commercialhaskell/stack/issues/4085)

I've added an integration test that successfully replicates [the issue](https://stackoverflow.com/questions/50709844/is-there-an-issue-with-haskell-ghc-8-4-3-on-travis)

- I build Stack _inside_ docker container
- Mount a volume of relatively small size
- Running the command from [the issue](https://stackoverflow.com/questions/50709844/is-there-an-issue-with-haskell-ghc-8-4-3-on-travis) with stack-root pointing to the mounted volume
- The command fails and produces stderr
- Verifying the stderr output


aleksejkozin:
## Current error log looks like this
screenshot 2018-10-27 at 13 25 11

aleksejkozin:
@mihaimaruseac, @snoyberg 
Could you review this?

mihaimaruseac:
Sorry, was away from internet outside of work time but now that I'm back I should get a review on this today or at the latest tomorrow.

aleksejkozin:
@mihaimaruseac is there something wrong with the code?

src/Stack/Setup.hs
```diff
@@ -1097,29 +1097,35 @@ installGHCPosix version downloadInfo _ archiveFile archiveType tempDir destDir =
...
-                           "run in " <> pretty wd) <> line <> line <>
-                        "The following directories may now contain files, but won't be used by stack:" <> line <>
-                        "  -" <+> pretty tempDir <> line <>
-                        "  -" <+> pretty destDir <> line
-                    liftIO exitFailure
+          menv' <- modifyEnvVars menv (Map.union env)
+          let logLines lvl = CB.lines .| CL.mapM_ (lvl . displayBytesUtf8)
+              logStdout = logLines logDebug
+              logStderr = logLines logError
+          _ <-  withWorkingDir (toFilePath wd) $
```
mihaimaruseac: You can also do `void $ withWorkingDir ..` and then remove the `return ()` at the end.

mihaimaruseac:
There's nothing wrong, sorry for the delay again. Got swamped with work but now it should be ok.

One suggestion, if I may: next time, separaterefactoring of code from code addition with two different commits.

aleksejkozin:
> One suggestion, if I may: next time, separate refactoring of code from code addition with two different commits.  

Sounds reasonable, got it!

mihaimaruseac:
Thank you

PR #4387, opened 2018-11-05
Typo fix. Missing single quote.

Magicloud:
Comparing to other messages, apparently a single quote is missed here.


mattaudesse:
Thanks @Magicloud!

PR #4388, opened 2018-11-05
Remove trailing ',' from generated-other-modules.

philderbeast:
This is a one character change in a YAML list element, removing a trailing comma. I'll defer to CI to test this.

I found this possible problem when trying to generate stack's `package.yaml` from a `package.dhall` using the executable `dhall-hpack-yaml` from package hpack-dhall when adding a real world golden test for https://github.com/BlockScope/hpack-dhall/issues/24.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -271,7 +271,7 @@ executables:
     main: Main.hs
     source-dirs: src/main
     generated-other-modules:
-    - Build_stack,
+    - Build_stack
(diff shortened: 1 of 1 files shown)


philderbeast:
Of the travis builds, all passed except for the pedantic one that failed with;

> The job exceeded the maximum time limit for jobs, and has been terminated.

Issue #4009, opened 2018-05-04
Building Main/StackSetupShim triggers un-silenceable -nopie warnings with clang


patrickt:
### General summary/comments

In attempting to upgrade a library ([haskell-tree-sitter](https://github.com/tree-sitter/haskell-tree-sitter/)) to a newer Stackage nightly (`nightly-2018-05-04`, for GHC 8.4, but `nightly-2018-04-25` exhibits this problem too), the build process (with a custom Setup.hs) always generates the following errors from clang:

```
haskell-tree-sitter-0.1.0: configure (lib)
[1 of 2] Compiling Main             ( /Users/patrickt/src/haskell-tree-sitter/Setup.hs, /Users/patrickt/src/haskell-tree-sitter/.stack-work/dist/x86_64-osx/Cabal-2.2.0.1/setup/Main.o )
[2 of 2] Compiling StackSetupShim   ( /Users/patrickt/.stack/setup-exe-src/setup-shim-mPHDZzAJ.hs, /Users/patrickt/src/haskell-tree-sitter/.stack-work/dist/x86_64-osx/Cabal-2.2.0.1/setup/StackSetupShim.o )
Linking /Users/patrickt/src/haskell-tree-sitter/.stack-work/dist/x86_64-osx/Cabal-2.2.0.1/setup/setup ...
clang: warning: argument unused during compilation: '-nopie' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-nopie' [-Wunused-command-line-argument]
```

These errors also appear when compiling dependent C code in the project itself; I was able to fix them with the following stanza in my stack.yaml:

```
apply-ghc-options: everything
ghc-options:
  $everything: -optc -Wno-unused-command-line-argument
```

However, I'm unable to figure out how to silence these warnings when building the `setup` executable. Adding a `GHC_OPTIONS` stanza to my `Setup.hs` did not work, nor did adding `ccOptions = "-Wno-unused-command-line-argument"` to my Setup.hs's `libBuildInfo`. Short of editing `~/.stack/programs/x86_64-osx/ghc-8.4.1/lib/ghc-8.4.1/settings` on every machine we use, is there a way to silence these warnings? This makes upgrading to 8.4 a tricky proposition.

I am running macOS 10.13.4 with stack 1.7.1. The output of `clang --version` is

```
Apple LLVM version 9.1.0 (clang-902.0.39.1)
Target: x86_64-apple-darwin17.5.0
Thread model: posix
[...]

PR #4392, opened 2018-11-09
Doc/nopie 4009

philderbeast:

Markdown preview was used to test documentation changes. The fix I tested locally with `resolver: lts-11.22 (ghc-8.2.2)`.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -543,17 +543,15 @@ clang: warning: argument unused during compilation: '-nopie'
  [-Wunused-command-line-argument]
 ```
 
-This warning is shown when compiler support of `-no-pie` is expected.
-For `ghc-8.2.2`, the warning suppressing setting change
-would be;
+This warning is shown when compiler support of `-no-pie` is expected but unavailable.
+It's possible to bypass the warning for a specific version of GHC by modifying a global setting:
 
 ```
 # ~/.stack/programs/x86_64-osx/ghc-8.2.2/lib/ghc-8.2.2/settings
-
 -- ("C compiler supports -no-pie", "YES"),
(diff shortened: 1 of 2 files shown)


doc/faq.md
```diff
@@ -535,3 +535,25 @@ error "bad relocation (Invalid pointer diff)". The compiler picks up
...
 The workaround is to remove LLVM binaries from the `PATH`.
+
+## How do I suppress `'-nopie'` warnings with `stack build` on macOS?
+
+```
+clang: warning: argument unused during compilation: '-nopie'
+ [-Wunused-command-line-argument]
+```
+
+This warning is shown when compiler support of `-no-pie` is expected.
```
mattaudesse: There's a semicolon at the end of this blurb where I think you probably wanted `:` instead :-)

I think rewording things slightly might be helpful here, too. It could be for lack of coffee, but it took me a moment to understand what's being suggested when reading the current explanation.

Perhaps something like:
> This warning is shown when compiler support of `-no-pie` is expected but unavailable.
> It's possible to bypass the warning for a specific version of GHC by modifying a global setting:
> ```
># ~/.stack/programs/x86_64-osx/ghc-8.2.2/lib/ghc-8.2.2/settings
> -- ("C compiler supports -no-pie", "YES"),
> ++ ("C compiler supports -no-pie", "NO"),
> ```
> **Note that we're fixing `ghc-8.2.2` in this case; repeat for other versions as necessary.** You should apply this fix for the version of GHC that matches your resolver.
>
> Issue [#4009](https://github.com/commercialhaskell/stack/issues/4009) on GitHub goes into further detail.

Thanks for the PR :+1:

PR #4402, opened 2018-11-16
Fix test 4270

qrilka:

This is a fix for a test `4270-files-order`, the main point of it is in af36e04 other commits improve Stack integration tests a bit.
Integration tests succeed locally.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -97,7 +97,10 @@ Bug fixes:
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
 * Don't download ghc when using `stack clean`.
-* Support loading in GHCi definitions from symlinked C files. See
+* Support loading in GHCi definitions from symlinked C files. Without this
+  patch, Stack will try to find object files in the directory pointed to
+  by symlinks, while GCC will produce the object files in the original
+  directory. See
   [#4402](https://github.com/commercialhaskell/stack/pull/4402)
 
 ## v1.9.1
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -94,6 +94,8 @@ Bug fixes:
 * Fix for git packages to update submodules to the correct state. See
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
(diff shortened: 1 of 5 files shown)


src/Stack/Package.hs
```diff
@@ -803,12 +803,22 @@ resolveComponentFiles component build names = do
...
+    csources <-
+        forMaybeM (cSources build) $ \fp -> do
+            result <- resolveDirFile dir fp
+            case result of
+              Nothing -> do
+                warnMissingFile "File" cwd fp file
+                return Nothing
+              Just p -> return $ Just (DotCabalCFilePath p)
+    jsources <-
+        forMaybeM (targetJsSources build) $ \fp ->
```
snoyberg: Does it make sense to use `findCandidate` for the Javascript source? If so, why? Why not use the same logic as the C sources?

qrilka: oh, that's just a stupid miss from me, will fix that

ChangeLog.md
```diff
@@ -97,6 +97,8 @@ Bug fixes:
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
 * Don't download ghc when using `stack clean`.
+* Support loading in GHCi definitions from symlinked C files. See
```
snoyberg:
```suggestion
* Support loading in GHCi definitions from symlinked C files. Without this patch, Stack will try to find object files in the directory pointed to by symlinks, while GCC will produce the object files in the original directory. See
```

src/Stack/Package.hs
```diff
@@ -1227,14 +1236,16 @@ findCandidate dirs name = do
...
-        :: (MonadIO m, MonadThrow m)
-        => Path Abs Dir -> FilePath.FilePath -> m [Path Abs File]
-    resolveCandidate x y = do
-        -- The standard canonicalizePath does not work for this case
-        p <- parseCollapsedAbsFile (toFilePath x FilePath.</> y)
-        exists <- doesFileExist p
-        return $ if exists then [p] else []
+    resolveCandidate dir = fmap maybeToList . resolveDirFile dir
+
+resolveDirFile
```
snoyberg: Please add a comment explaining that symlinks will not be followed.

qrilka:
Anything else here?

snoyberg:
Good to merge after CI runs

Issue #4324, opened 2018-09-28
Stack dot should handle properly GHC boot packages


qrilka:
### General summary/comments (optional)

It looks like currently `stack dot` doesn't show dependencies of GHC boot libraries even when such a library comes from a Stackage LTS snapshot. Either the code should be improved by including those dependencies or it should be documented that `stack` doesn't track dependencies of GHC boot packages.

### Steps to reproduce

In the root of the stack repository run `stack dot --external | grep '"process"'`

### Expected

Dependencies of the `process` package will get shown (e.g. `directory` or `filepath`) besides the packages which depend on `process`.

### Actual

Current output (both for versions 1.7.1 and prerelese 1.9.0.1):
```
{rank=max; "process"; };
"Cabal" -> "process";
"cmdargs" -> "process";
"conduit-extra" -> "process";
"echo" -> "process";
"extra" -> "process";
"githash" -> "process";
"gitrev" -> "process";
"open-browser" -> "process";
"optparse-applicative" -> "process";
"rio" -> "process";
"stack" -> "process";
"streaming-commons" -> "process";
"typed-process" -> "process";
"unliftio" -> "process";
"x509-system" -> "process";
```

### Stack version

```
$ stack --numeric-version
1.7.1
$ ~/bin/stack-1.9.0.1 --numeric-version
1.9.0.1
```

### Method of installation

* Other : 1.7.1 install with `gentoo-haskell`, prerelease version downloaded from the Github release page

dbaynard:
I thought `--include-base` might help but it doesn't.

This isn't totally obvious to me — why is it valuable to display boot libraries?

We may want to add an option `--include-boot`, or modify `--include-base`, or make it default, depending on the answer.

qrilka:
Omitting some dependencies results in an incomplete dependency graph, I think Stack should be explicit about this. And actually we could add this information e.g. by using `ghc-pkg describe <pkg>`

vanceism7:
Hey guys, I've been looking at this issue and I think I was able to get those dependencies to load up in the dot graph. I'll post up a PR soon

qrilka:
@vanceism7 the current version of that code is more or less my refactoring of the previous version to use new source maps. As far as I understand it you're correct, probably you've noticed that for `wiredInPackages` we do consult `globalDumpMap` to get package information. Now we have global packages in `smGlobal` field of a `SourceMap` so we could just do that for any of those packages.

vanceism7:
[...]

PR #4405, opened 2018-11-18
Make dot and ls dependencies work without ghc installed

### LINKED ISSUE #4390: Simple commands should work without ghc installed
I believe several `stack` sub-command like `dot` and `ls dependencies` should work without `ghc` installed; after all, these commands do not require compiling anything.

What I see instead when running `stack --no-install-ghc --skip-ghc-check dot --external` on [this example project](https://github.com/advancedtelematic/quickcheck-state-machine/tree/aa38cb0aa2c6c0f61e5d38e6e5b385c4eb644ee1/example) is that `stack` complains with

    No compiler found, expected minor version match with ghc-8.2.2 (x86_64) (based on resolver setting in quickcheck-state-machine\example\stack.yaml).
    To install the correct GHC into C:\Users\<name>\AppData\Local\Programs\stack\x86_64-windows\, try running "stack setup" or use the "--install-ghc" flag.

As can bee seen above, I was trying to work around that issue by specifying `--no-install-ghc --skip-ghc-check`, but `stack` is checking for `ghc` nonetheless.

This issue seems to be somewhat related to https://github.com/commercialhaskell/stack/issues/4181 where even a simple `clean` tries to bootstrap all of `ghc`.

Background: We use the [OSS Review Toolkit](https://github.com/heremaps/oss-review-toolkit) to inspect Haskell project dependencies without building them. Right now, we still need to provide `ghc`, wasting time & space.

waddlaw:

@dbaynard fixed #4390. Could you please confirm it?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -97,7 +97,7 @@ Bug fixes:
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
 * Don't download ghc when using `stack clean`.
-* `dot` and `ls dependencies` commands is no longer require ghc installed. And, fix not respected `--no-install-ghc` flag. (See [#4390](https://github.com/commercialhaskell/stack/issues/4390))
+* `dot` and `ls dependencies` commands no longer require GHC to be installed. Also, ensures the `--no-install-ghc` flag is respected. See: [#4390](https://github.com/commercialhaskell/stack/issues/4390)
 
 ## v1.9.1
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -97,6 +97,7 @@ Bug fixes:
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
 * Don't download ghc when using `stack clean`.
+* `dot` and `ls dependencies` commands is no longer require ghc installed. And, fix not respected `--no-install-ghc` flag. (See [#4390](https://github.com/commercialhaskell/stack/issues/4390))
 
 ## v1.9.1
 
--- a/src/Stack/Runners.hs
+++ b/src/Stack/Runners.hs
@@ -267,4 +267,8 @@ withBuildConfigDot opts go f = withBuildConfig go' f
     go' =
         (if dotTestTargets opts then set (globalOptsBuildOptsMonoidL.buildOptsMonoidTestsL) (Just True) else id) $
         (if dotBenchTargets opts then set (globalOptsBuildOptsMonoidL.buildOptsMonoidBenchmarksL) (Just True) else id)
+        $
+        (set (globalOptsL.configMonoidSkipGHCCheckL) (Just True))
+        $
+        (set (globalOptsL.configMonoidInstallGHCL) (Just False))
         go
--- a/src/Stack/Runners.hs
+++ b/src/Stack/Runners.hs
@@ -269,8 +269,8 @@ withBuildConfigDot opts go = withBuildConfig (updateGlobalOpts go)
     updateGlobalOpts
       = updateOpts (dotTestTargets opts)  (globalOptsBuildOptsMonoidL.buildOptsMonoidTestsL)      (Just True)
       . updateOpts (dotBenchTargets opts) (globalOptsBuildOptsMonoidL.buildOptsMonoidBenchmarksL) (Just True)
-      . updateOpts True                   (globalOptsL.configMonoidSkipGHCCheckL)                 (Just True)
-      . updateOpts True                   (globalOptsL.configMonoidInstallGHCL)                   (Just False)
+      . set (globalOptsL.configMonoidSkipGHCCheckL) (Just True)
+      . set (globalOptsL.configMonoidInstallGHCL)   (Just False)
 
 -- helper function for update some option
 updateOpts :: Bool -> Lens' opts v -> v -> opts -> opts
--- a/src/Stack/Runners.hs
+++ b/src/Stack/Runners.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE RankNTypes #-}
 
 -- | Utilities for running stack commands.
 module Stack.Runners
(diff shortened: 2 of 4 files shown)


ChangeLog.md
```diff
@@ -97,6 +97,7 @@ Bug fixes:
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
 * Don't download ghc when using `stack clean`.
+* `dot` and `ls dependencies` commands is no longer require ghc installed. And, fix not respected `--no-install-ghc` flag. (See [#4390](https://github.com/commercialhaskell/stack/issues/4390))
```
snoyberg:
```suggestion
* `dot` and `ls dependencies` commands no longer require GHC to be installed. Also, ensures the `--no-install-ghc` flag is respected. See: [#4390](https://github.com/commercialhaskell/stack/issues/4390)
```

src/Stack/Runners.hs
```diff
@@ -261,10 +262,18 @@ munlockFile Nothing = return ()
...
-        (if dotBenchTargets opts then set (globalOptsBuildOptsMonoidL.buildOptsMonoidBenchmarksL) (Just True) else id)
-        go
+    updateGlobalOpts
+      = updateOpts (dotTestTargets opts)  (globalOptsBuildOptsMonoidL.buildOptsMonoidTestsL)      (Just True)
+      . updateOpts (dotBenchTargets opts) (globalOptsBuildOptsMonoidL.buildOptsMonoidBenchmarksL) (Just True)
+      . updateOpts True                   (globalOptsL.configMonoidSkipGHCCheckL)                 (Just True)
+      . updateOpts True                   (globalOptsL.configMonoidInstallGHCL)                   (Just False)
+
+-- helper function for update some option
+updateOpts :: Bool -> Lens' opts v -> v -> opts -> opts
```
snoyberg: I like this combinator 👍

src/Stack/Runners.hs
```diff
@@ -261,10 +262,18 @@ munlockFile Nothing = return ()
...
+withBuildConfigDot opts go = withBuildConfig (updateGlobalOpts go)
   where
-    go' =
-        (if dotTestTargets opts then set (globalOptsBuildOptsMonoidL.buildOptsMonoidTestsL) (Just True) else id) $
-        (if dotBenchTargets opts then set (globalOptsBuildOptsMonoidL.buildOptsMonoidBenchmarksL) (Just True) else id)
-        go
+    updateGlobalOpts
+      = updateOpts (dotTestTargets opts)  (globalOptsBuildOptsMonoidL.buildOptsMonoidTestsL)      (Just True)
+      . updateOpts (dotBenchTargets opts) (globalOptsBuildOptsMonoidL.buildOptsMonoidBenchmarksL) (Just True)
+      . updateOpts True                   (globalOptsL.configMonoidSkipGHCCheckL)                 (Just True)
```
snoyberg: It doesn't look like the `updateOpts` is really useful here, how about just `set (globalOptsL...) (Just True)`? I think it would be clearer that this is _always_ being set.

waddlaw: Right!

waddlaw:
Thank you for the review. updated!

dbaynard:
Great job, @waddlaw! I'll review shortly. I thought I might have a little longer to find you the next issue to tackle 🤩

ChangeLog.md
```diff
@@ -97,6 +97,7 @@ Bug fixes:
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
 * Add `--cabal-files` flag to `stack ide targets` command.
 * Don't download ghc when using `stack clean`.
+* `dot` and `ls dependencies` commands no longer require GHC to be installed. Also, ensures the `--no-install-ghc` flag is respected. See: [#4390](https://github.com/commercialhaskell/stack/issues/4390)
```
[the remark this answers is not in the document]
qrilka: I think that 2nd second sentence is misleading, the flag was respected at least in version 1.7.1 already:
```
qrilka@qdesktop ~/ws/h/stack-test-ghc-7.8.4 $ cat stack.yaml 
resolver: ghc-7.8.4
qrilka@qdesktop ~/ws/h/stack-test-ghc-7.8.4 $ stack --no-install-ghc clean
No compiler found, expected minor version match with ghc-7.8.4 (x86_64-tinfo6) (based on resolver setting in /home/qrilka/ws/h/stack-test-ghc-7.8.4/stack.yaml).
To install the correct GHC into /home/qrilka/.stack/programs/x86_64-linux/, try running "stack setup" or use the "--install-ghc" flag.
qrilka@qdesktop ~/ws/h/stack-test-ghc-7.8.4 $ stack --numeric-version
1.7.1
```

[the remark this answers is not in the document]
waddlaw: It's perfectly clear, thanks!!

waddlaw:
@dbaynard I was able to do it.`--system-ghc` flag is needed that resolve this issue.

```shell
$ stack --numeric-version
1.9.1

$ stack --no-install-ghc --skip-ghc-check --system-ghc dot --external
strict digraph deps {
"clean" [style=dashed];
{rank=max; "rts" [shape=box]; };
"base" -> "ghc-prim";
"base" -> "integer-gmp";
"base" -> "rts";
"clean" -> "base";
"ghc-prim" -> "rts";
"integer-gmp" -> "ghc-prim";
}

$ stack --no-install-ghc --skip-ghc-check dot --external
No compiler found...
```

`optsUseSystem sopts` is `Nothing` if missing `--system-ghc`.

https://github.com/commercialhaskell/stack/blob/0b68efc82ec989415df87635dac6b2a38ec5e264/src/Stack/Setup.hs#L386-L391
https://github.com/commercialhaskell/stack/blob/0b68efc82ec989415df87635dac6b2a38ec5e264/src/Stack/Setup.hs#L395-L399

In other words, `needLocal = not (any (uncurry canUseCompiler) Nothing)`.

```hs
ghci> :t any
any :: Foldable t => (a -> Bool) -> t a -> Bool

ghci> any (const True) Nothing
False
```

I think that we should just only enable these option in `dot` (and `ls`), or add new `--skip-ensure-compiler` flag. (This flag is completely skip `ensureCompiler` function)

What do you think?

qrilka:
I don't think enabling `--system-ghc` is a way to go here - it could be of a different and maybe incompatible version or there could be no system GHC at all. And the last one is the starting use case in #4390 
Also I don't think that having a CLI flag `--skip-ensure-compiler` is a good idea - it's better to have that option internal-only and disable it only for a selected number of commands which should not fail without proper GHC available

waddlaw:
Thanks @qrilka. I understand it.

My implemented `checkDownloadCompiler` is possibly wrong...?(always skipping `ensureCompiler` if `--no-install-ghc` specified)

```hs
checkDownloadCompiler _     SkipDownloadCompiler = return (Nothing, Nothing, False)
checkDownloadCompiler sopts WithDownloadCompiler
  | installIfMissing = ensureCompiler sopts
  | otherwise = return (Nothing, Nothing, False)
  where
    installIfMissing = soptsInstallIfMissing sopts
```

Hence, `stack --no-install-ghc --skip-ghc-check dot --external` works fine.

qrilka:
I think it's wrong, yes, disabling GHC installation doesn't mean that compiler check should get skipped. For `dot` we should be able to go on without a compiler so `SkipDownloadCompiler` (or `SkipCompilerCheck`) should be OK for it. At least that's my understanding of these things.

dbaynard:
There are some changes in progress (e.g. #4500) which may affect this.

waddlaw:
@dbaynard 
Sorry for the delay in work.I think I will work on it.
Is it better to work after # 4500 is merged?

dbaynard:
It should be merged soon — but if you work on it before then just take into account those changes. I _think_ the only changes left for that PR are in the test suite.

snoyberg:
I'm looking into this right now as part of my work on cleaning up `Stack.Runners`. I'm confused as to why this is supposed to work at all. `stack dot` depends on knowing information about dependencies of global packages, which in turn requires having a locally installed GHC. It seems like the only way to make this work is to ignore global packages entirely, which this PR isn't doing.

qrilka:
@snoyberg previously we didn't look into global packages and that's why I filed #4324

snoyberg:
Thanks for clarifying @qrilka. I've opened up #4632, which will supersede this PR.

Issue #4410, opened 2018-11-20
Fix or remove solver command


qrilka:
The command `stack solver` wan't updated for quite a long time. Its functionality wasn't checked for `cabal-install` versions newer than `1.24` and issues a warning in such a case. Upcoming PR for #3922 2 breaks `stack solver` and it needs to be decided whether this command will be supported (and thus needs to be properly updated) or this feature will get removed in the new Stack version

PR #4412, opened 2018-11-21
New build plan construction with source maps

### LINKED ISSUE #3922: Build tools not considered when constructing SourceMap

qrilka:

The PR implements what is described in https://github.com/commercialhaskell/stack/blob/master/doc/build-overview.md

2 integration tests fail currently, both involve `stack solver` which is to be fixed in #4410 

Fixes #3922

### DIFF
--- a/test/integration/tests/3397-ghc-solver/Main.hs
+++ b/test/integration/tests/3397-ghc-solver/Main.hs
@@ -1,3 +1,5 @@
+{--
+
 import StackTest
 
 main :: IO ()
@@ -6,3 +8,8 @@ main = do
   removeFileIgnore "issue3397.cabal"
   stack ["init", "--solver", "--resolver", "ghc-8.2.2"]
   stack ["solver", "--update-config"]
+
+// --}
+
+main :: IO ()
+main = putStrLn "This test is disabled (see https://github.com/commercialhaskell/stack/issues/4410)."
--- a/test/integration/tests/3533-extra-deps-solver/Main.hs
+++ b/test/integration/tests/3533-extra-deps-solver/Main.hs
@@ -1,3 +1,5 @@
+{--
+
 import StackTest
 import System.Directory
 
@@ -6,3 +8,8 @@ main = do
     copyFile "orig-stack.yaml" "stack.yaml"
     stack [defaultResolverArg, "solver", "--update-config"]
     stack ["build"]
+
+// --}
+
+main :: IO ()
+main = putStrLn "This test is disabled (see https://github.com/commercialhaskell/stack/issues/4410)."


src/Stack/Build/Source.hs
```diff
@@ -472,32 +504,19 @@ calcFci modTime' fp = liftIO $
...
-  => BuildOptsCLI
-  -> PackageName
-  -> Bool
-  -> Bool
+  => Map FlagName Bool
+  -> [Text]
   -> m PackageConfig
-getPackageConfig boptsCli name isTarget isLocal = do
-  bconfig <- view buildConfigL
+getPackageConfig flags ghcOptions = do
```
lehins: Although not introduced in this PR `MonadIO` shouldn't be in the type signature. Considering that this function's type signature is being changed in this PR, might as well fix it all the way ;)

src/Stack/Build/ConstructPlan.hs
```diff
@@ -220,20 +217,51 @@ constructPlan ls0 baseConfigOpts0 locals extraToBuild0 localDumpPkgs loadPackage
...
+                  , cpHaddocks = buildHaddocks
+                  }
+            in Just $ PSRemote loc version NotFromSnapshot common
+      deps <- for (smDeps sourceMap) $ \dp ->
+        case dpLocation dp of
+          PLImmutable loc -> do
+            version <- getPLIVersion loc (cpGPD $ dpCommon dp)
+            return $ PSRemote loc version (dpFromSnapshot dp) (dpCommon dp)
+          PLMutable dir -> do
+            -- FIXME this is not correct, we don't want to treat all Mutable as local
```
lehins: @qrilka this is the FIXME I had in mind when talking about an immutable package becoming mutable if it depends on some local mutable package, but here when it would become mutable it wouldn't be local and wouldn't have a local path.

qrilka: Cool, thanks

qrilka:
@lehins it appears that my assumption wasn't quite correct and build-overview.md doc is still valid and the quote that you given isn't quite applicable to the way things are organized: "mutability" of package locations does not depend on their dependency trees, that information is taken into account when a build plan gets constructed. In 718569a3b620374be52ba27b005517a6ab6089d7 I tried to make this separation of tasks for building mutable and immutable packages more explicit. As for the FIXME then it also appears not to be applicable so I've removed it.

lehins:
@qrilka Cool. That makes sense.

qrilka:
> This can be confusing `TTRemote` can be `Mutable` or `Immutable`, where the location can be only immutable `PackageLocationImmutable`. It be much cleaner IMHO if it was called `PackageLocationRemote`. That refactoring would touch many files, so I am not sure if it is really worth it, or if there are any other reasons against such a rename.
I'm already doing rename in that commit so amending it with a better name makes perfect sense, thanks

lehins:
@qrilka Yeah, sounds good to me.

src/Stack/Build/Source.hs
```diff
@@ -20,121 +21,190 @@ import qualified    Pantry.SHA256 as SHA256
...
+hashSourceMapData wc smDeps = do
+    path <- encodeUtf8 . T.pack . toFilePath <$> getCompilerPath wc
+    let compilerExe =
+            case wc of
+                Ghc -> "ghc"
+                Ghcjs -> "ghcjs"
+    info <- BL.toStrict . fst <$> proc compilerExe ["--info"] readProcess_
+    immDeps <-
+        fmap B.concat . forM (Map.elems smDeps) $ depPackageHashableContent
+    return $ SourceMapHash (SHA256.hashBytes $ B.concat [path, info, immDeps])
```
lehins: I know it's done once, nevertheless it is an expensive operation. It seems unnecessary to call `B.concat` twice. Moreover allocation another huge `ByteString` in order to compute it's hash seems wasteful. I'd suggest removing `fmap B.concat . ` from the previous line and doing: `(SHA256.hashBytesLazy $ BL.fromChunks (path:info:immDeps)` here.

src/Stack/Types/Config.hs
```diff
@@ -539,32 +529,16 @@ data EnvConfig = EnvConfig
...
-  }
-
--- | A view of a project package needed for resolving components
-data ProjectPackage = ProjectPackage
-    { ppCabalFP    :: !(Path Abs File)
-    , ppResolvedDir :: !(ResolvedPath Dir)
-    , ppGPD' :: !(IO GenericPackageDescription)
-    , ppName :: !PackageName
+--    ,envConfigSMActual :: !SMActual
+--    ,envConfigLoadedSnapshot :: !LoadedSnapshot
```
lehins: Do we still need this commented out record fields?

src/Stack/SourceMap.hs
```diff
@@ -0,0 +1,245 @@
...
+        throwM $ InvalidFlagSpecification $ Set.fromList unusedFlags
+
+checkFlagUsed ::
+       MonadIO m
+    => (PackageName, Map FlagName Bool)
+    -> FlagSource
+    -> Map PackageName ProjectPackage
+    -> Map PackageName DepPackage
+    -> m (Maybe UnusedFlags)
+checkFlagUsed (name, userFlags) source prj deps =
```
lehins: Just a bit of nitpicking, but for the sake of consistency would it be better to keep it plural: `checkUnusedFlags`?

src/Stack/SourceMap.hs
```diff
@@ -0,0 +1,245 @@
...
+    in case maybeCommon  of
+        -- Package is not available as project or dependency
+        Nothing ->
+            pure $ Just $ UFNoPackage source name
+        -- Package exists, let's check if the flags are defined
+        Just common -> do
+            gpd <- liftIO $ cpGPD common
+            let pname = pkgName $ PD.package $ PD.packageDescription gpd
+                pkgFlags = Set.fromList $ map PD.flagName $ PD.genPackageFlags gpd
+                unused = Set.difference (Map.keysSet userFlags)
```
lehins: There is a handy function `Map.withoutKeys` that could be used here instead, which would avoid unnecessary conversion to `Set` until necessary

src/Stack/Script.hs
```diff
@@ -205,20 +208,34 @@ blacklist = Set.fromList
...
+            installMap
+    return $
+        toModuleInfo (notHidden $ smDeps sourceMap) snapshotDumpPkgs <>
+        toModuleInfo (smGlobal sourceMap) globalDumpPkgs
+  where
+    notHidden = Map.filter (not . dpHidden)
+    toModuleInfo pkgs dumpPkgs =
+        let pnames = Map.keysSet pkgs `Set.difference` blacklist
+            modules =
+                Map.fromList
```
lehins: Correct me if I am wrong, but we can potentially have module names clash amongst different global packages, by converting to `Map` from list we will end up dropping information about those packages. My assumption is that it is exactly the reason for  `Set` in `ModuleInfo`. Consider this example:
* module `Data.Foo` comes from package `foo`
* module `Data.Foo` comes from package `bar`
We could still use both packages in our library by using `PackageImports`, but with approach employed here we will end up with one that is last in the list, eg. `("Data.Foo", Set.singleton "foo")` and info about `bar` is lost.

But I assume the desired result in above example would be  `Map.fromlIst [("Data.Foo", Set.fromList ["foo", "bar"])]`

src/Stack/Script.hs
```diff
@@ -205,20 +208,34 @@ blacklist = Set.fromList
...
+        getInstalled
+            GetInstalledOpts
+            { getInstalledProfiling = False
+            , getInstalledHaddock = False
+            , getInstalledSymbols = False
+            }
+            installMap
+    return $
+        toModuleInfo (notHidden $ smDeps sourceMap) snapshotDumpPkgs <>
+        toModuleInfo (smGlobal sourceMap) globalDumpPkgs
```
lehins: Same comment as below applies here when `mappend` is used. Can we have clashing module names between packages in `snapshotDumpPkgs` and `globalDumpPkgs`?

src/Stack/Script.hs
qrilka: Yeah, you're exactly right and in the previous version it was done with `mconcat`, so I'll just use `Map.fromListWith mappend`

src/Stack/Script.hs
qrilka:
```
instance Semigroup ModuleInfo where
  ModuleInfo x <> ModuleInfo y =
    ModuleInfo (Map.unionWith Set.union x y)
```

src/Stack/Hoogle.hs
```diff
@@ -104,22 +104,22 @@ hoogleCmd (args,setup,rebuild,startServer) go = withBuildConfig go $ do
...
-                                { boptsCLITargets =
-                                    pure $
-                                    either
-                                     (T.pack . packageIdentifierString)
-                                     (utf8BuilderToText . display)
-                                     hooglePackageIdentifier
-                                }))
+                      NeedTargets
+                      boptsCLI
+                      (\lk -> Stack.Build.build Nothing lk)
```
lehins: eta reduce to `(Stack.Build.build Nothing)`?

src/Stack/Hoogle.hs
```diff
@@ -61,16 +62,15 @@ hoogleCmd (args,setup,rebuild,startServer) go = withBuildConfig go $ do
     buildHaddocks =
         liftIO
             (catch
-                 (withBuildConfigAndLock
+                 (withDefaultBuildConfigAndLock
                       (set
                            (globalOptsBuildOptsMonoidL . buildOptsMonoidHaddockL)
                            (Just True)
                            go)
                       (\lk ->
```
lehins: eta reduce to `(Stack.Build.build Nothing)`?

src/Stack/Build/Installed.hs
```diff
@@ -39,17 +44,33 @@ data GetInstalledOpts = GetInstalledOpts
       -- ^ Require debugging symbols?
     }
 
+toInstallMap :: MonadIO m => SourceMap -> m InstallMap
+toInstallMap sourceMap = do
+    let loadVersion loc common = do
```
lehins: Maybe renaming to `loadLocalVersion` and removing `loc` argument? Would be slightly clearer, considering it's only ever called with `Local`

src/Stack/Build/ConstructPlan.hs
```diff
@@ -220,20 +217,49 @@ constructPlan ls0 baseConfigOpts0 locals extraToBuild0 localDumpPkgs loadPackage
...
         , loadPackage = \x y z -> runRIO econfig $ loadPackage0 x y z
-        , combinedMap = combineMap sourceMap installedMap
+        , combinedMap = combineMap sources installedMap
         , ctxEnvConfig = econfig
         , callStack = []
-        , extraToBuild = extraToBuild0
-        , wanted = wantedLocalPackages locals <> extraToBuild0
-        , localNames = Set.fromList $ map (packageName . lpPackage) locals
+        , wanted = Map.keysSet (smtTargets $ smTargets sourceMap)
+        , localNames = Map.keysSet (smProject sourceMap) -- Set.fromList $ map (packageName . lpPackage) locals
```
lehins: Remove commented out code?

lehins:
@qrilka There seems to be a conflict with `src/Stack/Config.hs` otherwise this PR looks good to me.

src/Stack/Build/Installed.hs
qrilka: The point is that it's not exclusively about locals, looks like the code got copy-pasted and `versionMaybeFromPM` shouldn't use the same code but should just use loadVersion - I'll refactor it

qrilka:
Conflict resolved

qrilka:
With a new merge with `master` pantry tests don't fail because of Mercurial but builds appear to time out.
Local integration tests show 2 fails: `3397-ghc-solver` and `3533-extra-deps-solver` - both are supposed to get resolved in #4410

PR #4430, opened 2018-12-04
[WIP] Azure pipelines Integration

psibi:

### DIFF
--- a/azure-linux-template.yml
+++ b/azure-linux-template.yml
@@ -7,26 +7,27 @@ jobs:
     vmImage: ${{ parameters.vmImage }}
   strategy:
     matrix:
-      stack-def:
-        BUILD: stack
-        STACK_YAML: stack.yaml
-      stack-lts-12:
-        BUILD: stack
-        STACK_YAML: stack-lts-12.yaml
-      nightly:
-        BUILD: stack
-        STACK_YAML: stack-nightly.yaml
+      # stack-def:
+      #   BUILD: stack
+      #   STACK_YAML: stack.yaml
+      # stack-lts-12:
+      #   BUILD: stack
+      #   STACK_YAML: stack-lts-12.yaml
+      # nightly:
+      #   BUILD: stack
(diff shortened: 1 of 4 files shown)


psibi:
@borsboom This is the latest status of the Azure Build: https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=81

* The build seems to be only failing for Windows and it seems to be unrelated to this commit.
* Note that the caching of the builds will be done properly once the PR lands into master branch (BASE_BRANCH). With caching enabled in my private fork, I was able to see time reduction to 12 minutes from around 40 minutes.
* Mercurial is installed in both Mac and Windows system to prevent the test failure happening because of that.

mattaudesse:
Is this ready to merge @borsboom + @psibi? I think I must've missed the discussion leading up to it.

Also, when the time comes, I wonder if we should use a `squash` merge for this PR since there are so many commits - what do you guys think?

psibi:
@mattaudesse The work is complete, I'm waiting for feedbacks from @borsboom 

This is the latest build report from Azure for this branch: https://dev.azure.com/commercialhaskell/73aaff64-82e1-454b-9a87-948a58086804/_build/results?buildId=81

The Windows build seems to be failing because of legitimate reasons, although I haven't looked into it deeply.

> I wonder if we should use a squash merge for this PR 
Yes, that would be the only reasonable thing to do here!

psibi:
@borsboom Thanks, I will take them as next steps.

@dbaynard Given the feedback, I think this is ready to be merged!

borsboom:
I've filed a #4440 for the failing test on Windows (which I've reproduced independently of the Azure pipeline in this PR).

dbaynard:
Does this need to be cherry-picked to `stable`, for azure to build stable, too?

borsboom:
I don't think it's mandatory to cherry-pick it into `stable`, but it would be nice to have `stable` using Azure soon too so it's worth a try after the merging to `master`.  I'll go ahead and merge this now.

Issue #4362, opened 2018-10-18
1.9.1 fails with "I don't know how to install GHC on your system configuration, please install manually"


ThomasLocke:
### General summary/comments (optional)
Using stack on Slackware Linux 14.2 (x86_64). 

stack 1.6.5 works flawlessly, 1.9.1 returns _I don't know how to install GHC on your system configuration, please install manually_ when running `stack build`

### Steps to reproduce
1. Run command `stack new myproj`
2. Run command `cd myproj`
3. Run command `stack build`

It fails at step 3.

### Expected
I expect 1.9.1 builds my projects, similar as to how 1.6.5 does it.

### Actual
I get the error _I don't know how to install GHC on your system configuration, please install manually_

```
$ stack build --verbose
2018-10-18 12:37:21.393172: [debug] Checking for project config at: /home/thomas/myproj/stack.yaml
2018-10-18 12:37:21.393378: [debug] Loading project config file stack.yaml
2018-10-18 12:37:21.395132: [debug] Decoding build plan from: /home/thomas/.stack/build-plan/lts-11.22.yaml
2018-10-18 12:37:21.395210: [debug] Trying to decode /home/thomas/.stack/build-plan-cache/lts-11.22.cache
2018-10-18 12:37:21.403246: [debug] Success decoding /home/thomas/.stack/build-plan-cache/lts-11.22.cache
2018-10-18 12:37:21.403692: [debug] Run process: /sbin/ldconfig -p
2018-10-18 12:37:21.408474: [debug] Process finished in 5ms: /sbin/ldconfig -p
2018-10-18 12:37:21.409755: [debug] Did not find shared library libtinfo.so.5
2018-10-18 12:37:21.409868: [debug] Did not find shared library libtinfo.so.6
2018-10-18 12:37:21.409968: [debug] Did not find shared library libncursesw.so.6
2018-10-18 12:37:21.410042: [debug] Found shared library libgmp.so.10 in 'ldconfig -p' output
2018-10-18 12:37:21.410142: [debug] Did not find shared library libgmp.so.3
2018-10-18 12:37:21.410184: [debug] Potential GHC builds: 
2018-10-18 12:37:21.410221: [debug] Found already installed GHC builds: 
I don't know how to install GHC on your system configuration, please install manually
```

### Stack version

```
$ stack --version
Version 1.9.1, Git revision f9d0042c141660e1d38f797e1d426be4a99b2a3c (6168 commits) x86_64 hpack-0.31.0
```

[...]

PR #4434, opened 2018-12-06
Workaround for libtinfo detection on Slackware64 14.2, solves #4362.

skapazzo:
Expanded the already present workaround for finding libtinfo on Arch linux (when `ldconfig -p` doesn't find it) to also work in 64 bit Slackware.
Basically also searches the directory `/usr/lib64/` in addition to `/usr/lib/` for libtinfo or ncurses libraries (see discussion in #4362 and others).

Do we want to add `/lib/` and `/lib64/` or any other paths to the list of checked directories?

I haven't written any haskell in a long time (I stumbled upon this issue while trying to get back into it) so any suggestions and criticisms on the code are more than welcome; it actually doesn't look really readable to me as is but considering it is a (temporary) workaround for such a limited issue I think it doesn't matter.

I don't think this change requires a changelog or documentation update since it doesn't affect users or developers in any way except solving the issue.

I checked it with hlint and run stack test without problems but I could only test the binary with Slackware linux.

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -588,16 +588,18 @@ getGhcBuilds = do
                             -- Cannot parse /usr/lib on Windows
                             return False
                         | otherwise = do
-                            -- This is a workaround for the fact that libtinfo.so.6 doesn't appear in
-                            -- the 'ldconfig -p' output on Arch even when it exists.
-                            -- There doesn't seem to be an easy way to get the true list of directories
-                            -- to scan for shared libs, but this works for our particular case.
-                            usrLib <- parseAbsDir "/usr/lib"
-                            e <- doesFileExist (usrLib </> lib)
-                            if e
-                                then logDebug ("Found shared library " <> libD <> " in /usr/lib")
-                                else logDebug ("Did not find shared library " <> libD)
-                            return e
+                            let extraPaths = ["/usr/lib","/usr/lib64"]
+                        -- This is a workaround for the fact that libtinfo.so.x doesn't appear in
+                        -- the 'ldconfig -p' output on Arch or Slackware even when it exists.
+                        -- There doesn't seem to be an easy way to get the true list of directories
+                        -- to scan for shared libs, but this works for our particular cases.
+                            usrLibs <- forM extraPaths (\p -> liftM ((,)p) $ parseAbsDir p)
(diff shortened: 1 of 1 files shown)


borsboom:
I'm fine with this approach for now, although as more workarounds are added I'd really love to move to something like is described [here](https://github.com/commercialhaskell/stack/blob/c368ba8d0a819c40bc0427221a95a2f08bf93ca3/src/Stack/Setup.hs#L548) some time.

Regarding the code, you could use TemplateHaskell to parse the directories at compile-time instead of runtime: `let extraPaths = [$(mkAbsDir "/usr/lib"),$(mkAbsDir "/usr/lib64")]`.  Then you don't need the first `forM` anymore.  And I don't think you need to return pairs from the `forM` since you can just convert the `Path` to a string using `Path.toFilePath` (`FilePath` is a type synonym for `String`).

skapazzo:
I tried to rewrite the code following these suggestions (i.e. using TemplateHaskell). I tried to keep the lines shorter in these.
Is there a rule about order of language pragmas? I added TemplateHaskell at the end here.
Are there some side effects of using mkAbsDir instead of parseAbsDir on other platforms or OSes?

The more I think about the inclusion of more paths the more I'm against it, this workaround should not be needed with newer ncurses 6 anyway.

@borsboom I looked at the possibility of a more stable solution as that one, but I wouldn't know where to start so I decided to take the shorter step for now.

Thank you all for your suggestions and help.

skapazzo:
Looking at the AppVeyor build failure it seems like there really can be a compile time problem with `mkAbsDir`.
I'd guess `System.FilePath.isValid` returns False on Windows which makes `parseAbsDir` (called by `mkAbsDir`) throw an `InvalidAbsDir` exception.
Any suggestions?

borsboom:
New version looks good.  I usually like to have LANGUAGE pragmas in alphabetical order, by the Stack codebase doesn't do this consistently and this module is already out of order, so adding it at the end is fine.

This code can be wrapped in
```
#ifndef WINDOWS
…
#endif
```
to avoid trying to compile it on Windows.

skapazzo:
This version seems to build on windows but gives an error on macOS on Azure, I couldn't read/understand the error log though.

borsboom:
LGTM, thanks!

The macOS log doesn't seem to be viewable at all.  Maybe a temporary glitch, I very much doubt it's anything to do with this change.

PR #4439, opened 2018-12-10
Fix Issue 4377 - Allows remote repo without cabal file and with package.yaml work

psibi:


dbaynard:
@NorfairKing Care to take a look — or identify somebody to review?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -375,7 +381,7 @@ parseArchive pli archive fp = do
         Left e -> throwIO $ UnsupportedTarball loc $ T.pack e
         Right safeFiles -> do
           let toSave = Set.fromList $ map (seSource . snd) safeFiles
-          blobs <-
+          (blobs :: Map FilePath BlobKey)  <- --come here
```
qrilka: What does this comment mean? Is type here required?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +393,60 @@ parseArchive pli archive fp = do
             case Map.lookup (seSource se) blobs of
               Nothing -> error $ "Impossible: blob not found for: " ++ seSource se
               Just blobKey -> pure (sfp, TreeEntry blobKey (seType se))
-
           -- parse the cabal file and ensure it has the right name
-          (cabalPath, cabalEntry@(TreeEntry cabalBlobKey _)) <- findCabalFile pli tree
+          (cabalPath, cabalEntry@(TreeEntry cabalBlobKey _), btype) <- findCabalOrHpackFile pli tree
```
qrilka: Minor nitpicking but what does "b" in `btype` mean?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +393,60 @@ parseArchive pli archive fp = do
...
+          (_warnings, gpd) <- rawParseGPD (Left pli) cabalBs
           let ident@(PackageIdentifier name _) = package $ packageDescription gpd
-          when (cabalPath /= cabalFileName name) $
+          when (cabalPath /= cabalFileName name && btype == CabalFile) $
             throwIO $ WrongCabalFileName pli cabalPath name
 
           -- It's good! Store the tree, let's bounce
-          (_tid, treeKey) <- withStorage $ storeTree ident tree cabalEntry
+          (_tid, treeKey) <- withStorage $ storeTree ident tree cabalEntry btype
+          let tentry = if btype == CabalFile
```
qrilka: "t" here looks similar like the 1st letter in `treeKey`, not clear what it means and a bit confusing

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +393,60 @@ parseArchive pli archive fp = do
...
+        let txt = unSafeFilePath sfp
+         in not ("/" `T.isInfixOf` txt) && (".cabal" `T.isSuffixOf` txt || "package.yaml" == txt)
+      isCabalFile (sfp, _) =
         let txt = unSafeFilePath sfp
-         in not ("/" `T.isInfixOf` txt) && ".cabal" `T.isSuffixOf` txt
-  case filter isCabalFile $ Map.toList m of
+         in not ("/" `T.isInfixOf` txt) && (".cabal" `T.isSuffixOf` txt)
+      isHpackFile (sfp, _) =
+        let txt = unSafeFilePath sfp
+         in not ("/" `T.isInfixOf` txt) && ("package.yaml" == txt)
```
qrilka: how could we have "/" in `txt` being equal to `"package.yaml"`?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +393,60 @@ parseArchive pli archive fp = do
...
     [] -> throwM $ TreeWithoutCabalFile loc
-    [(key, te)] -> pure (key, te)
-    xs -> throwM $ TreeWithMultipleCabalFiles loc $ map fst xs
+    [v@(key, te)] -> if isHpackFile v
+                     then pure (key, te, HPackFile)
+                     else pure (key, te, CabalFile)
+    xs -> case (filter isCabalFile xs, filter isHpackFile xs) of
+            ([],[]) -> throwM $ TreeWithoutCabalFile loc
+            ([(key,te)],_) -> return (key, te, CabalFile)
+            ([],[(key,te)]) -> return (key, te, HPackFile)
```
qrilka: isn't it covered in `if` above already? Any reason to have a duplication of logic here?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -443,3 +468,15 @@ takeSubdir subdir = mapMaybe $ \(fp, a) -> do
   where
     splitDirs = List.dropWhile (== ".") . filter (/= "") . T.splitOn "/"
     subdirs = splitDirs subdir
+
+getCabalBS :: forall env. (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+           => ByteString -- Hpack's content
+           -> RIO env ByteString
+getCabalBS hpackBs = withSystemTempDirectory "hpack-repo" $ \tmpdir -> withWorkingDir tmpdir $ do
```
qrilka: couldn't we use a more explicit name like `hpackToCabal`? Also we dealing with archives here, what's the point of "repo" in directory name?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -443,3 +468,15 @@ takeSubdir subdir = mapMaybe $ \(fp, a) -> do
...
+               B.writeFile (tmpdir FilePath.</> Hpack.packageConfig) hpackBs
+               tdir <- parseAbsDir tmpdir
+               -- (file:_) <- filter (flip hasExtension "cabal" . toFilePath) . snd <$> listDir tdir
+               -- tfile <- par
```
qrilka: to be removed?

subs/pantry/src/Pantry/HPack.hs
```diff
@@ -0,0 +1,92 @@
...
+            parsePackageName
+        _:_ -> throwIO $ MultipleCabalFilesFound pkgDir files
+      where hasExtension fp x = FilePath.takeExtension fp == "." ++ x
+
+-- getCabalBS :: ByteString -> IO ByteString
+-- getCabalBS hpackBS = do
+--   withSystemTempDirectory "temp-hpack-dir" $
+--                               \tmpdir -> withWorkingDir tmpdir $ do
+--                                            B.writeFile Hpack.packageConfig hpackBS
+--                                            hpack undefined
```
qrilka: to be removed?

subs/pantry/src/Pantry/Hackage.hs
```diff
@@ -427,7 +428,7 @@ getHackageTarballKey pir@(PackageIdentifierRevision name ver (CFIHash sha _msize
 getHackageTarballKey pir = packageTreeKey <$> getHackageTarball pir Nothing
 
 getHackageTarball
-  :: (HasPantryConfig env, HasLogFunc env)
+  :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
```
qrilka: why did this new constraint appear here?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -147,6 +148,7 @@ Tree
     key BlobId
     cabal BlobId
     cabalType FileType
+    buildType P.BuildFileType
```
qrilka: `buildType` sounds quite confusing to me though I don't have a good alternative other than just `buildFileType` which I also don't like much

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -443,3 +468,15 @@ takeSubdir subdir = mapMaybe $ \(fp, a) -> do
...
+               (_, cfile) <- findOrGenerateCabalFile tdir
+               B.readFile (fromAbsFile cfile)
+    where hasExtension fp x = FilePath.takeExtension fp == "." ++ x
```
lehins: There is already a function like that: `FilePath.isExtensionOf`

subs/pantry/src/Pantry/HPack.hs
```diff
@@ -0,0 +1,92 @@
...
+-- | Generate .cabal file from package.yaml, if necessary.
+hpack
+  :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+  => Path Abs Dir
+  -> RIO env ()
+hpack pkgDir = do
+    packageConfigRelFile <- parseRelFile Hpack.packageConfig
+    let hpackFile = pkgDir Path.</> packageConfigRelFile
+    exists <- liftIO $ doesFileExist hpackFile
+    when exists $ do
```
lehins: There is a cool thing in RIO: `whenM` and `doesFileExist` imported from `RIO.Directory` is restricted to `MonadIO`, so you can rewrite it as:

```haskell
whenM (doesFileExist hpackFile) $ do
   ...
```

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +393,60 @@ parseArchive pli archive fp = do
...
-    [(key, te)] -> pure (key, te)
-    xs -> throwM $ TreeWithMultipleCabalFiles loc $ map fst xs
+    [v@(key, te)] -> if isHpackFile v
+                     then pure (key, te, HPackFile)
+                     else pure (key, te, CabalFile)
+    xs -> case (filter isCabalFile xs, filter isHpackFile xs) of
+            ([],[]) -> throwM $ TreeWithoutCabalFile loc
+            ([(key,te)],_) -> return (key, te, CabalFile)
+            ([],[(key,te)]) -> return (key, te, HPackFile)
+            ([],xs) -> throwM $ TreeWithMultipleHPackFiles loc $ map fst xs
```
lehins: `TreeWithMultipleHPackFiles` is an impossible case, we can't have more than one `package.yaml` file.

subs/pantry/src/Pantry/Archive.hs
psibi: Fixed.

subs/pantry/src/Pantry/Archive.hs
psibi: I have renamed it to  `buildFileType` now.  Hopefully that's better.

subs/pantry/src/Pantry/Archive.hs
psibi: Fixed.

subs/pantry/src/Pantry/Archive.hs
psibi: Renamed it. Hopefully the new variable `packageCabal` makes it clear.

subs/pantry/src/Pantry/Archive.hs
psibi: Fixed it.

subs/pantry/src/Pantry/Archive.hs
psibi: Renamed it to `hpackToCabal`.

subs/pantry/src/Pantry/Hackage.hs
psibi: That's because `storeTree` function now has that constraint.

subs/pantry/src/Pantry/Storage.hs
psibi: This column itself is now removed.

subs/pantry/src/Pantry/Archive.hs
psibi: Done.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -436,30 +448,83 @@ loadHackageTarballInfo name version = do
...
+  insertBy hpackRecord
+  return ()
+
+hpackVersionId :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env) => ReaderT SqlBackend (RIO env) VersionId
+hpackVersionId = do
+  hpackSoftwareVersion <- lift $ hpackVersion
+  hid <- insertBy $ Version { versionVersion = P.VersionP hpackSoftwareVersion }
+  case hid of
+    Left (Entity versionId _) -> return versionId
+    Right key -> return key
```
qrilka: Aren't these lines almost the same as in the function `getKey` above?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -436,30 +448,83 @@ loadHackageTarballInfo name version = do
...
+
+storeHPack :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env) => TreeId -> P.TreeEntry -> ReaderT SqlBackend (RIO env) ()
+storeHPack tid tentry = do
+  vid <- hpackVersionId
+  hpackRecord <- getBy (UniqueHPack tid vid)
+  case hpackRecord of
+    Nothing -> generateHPack tid tentry vid
+    Just _ -> return ()
```
qrilka: `when (isNothing hpackRecord) $` gives you a bit more concise variant

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -476,6 +541,10 @@ storeTree (P.PackageIdentifier name version) tree@(P.TreeMap m) (P.TreeEntry (P.
           , treeEntryType = ft
           }
       pure (tid, P.TreeKey blobKey)
+  case btype of
+    P.CabalFile -> return ()
+    P.HPackFile -> storeHPack tid tentry
```
qrilka: also good place for `when` maybe?

subs/pantry/src/Pantry/Storage.hs
psibi: Fixed, thanks.

subs/pantry/src/Pantry/Storage.hs
psibi: Thanks, the new code has changed.

subs/pantry/src/Pantry/Storage.hs
psibi: Fixed.

subs/pantry/test/Pantry/HPackSpec.hs
```diff
@@ -0,0 +1,20 @@
...
+spec :: Spec
+spec = describe "Parse HPack version" $ do
+  it "Shipped hpack version" $ do
+    version <- runPantryApp $ hpackVersion
+    version `shouldSatisfy` isVersion
+
+  -- it "External hpack version" $ do
+  --   fp <- parseAbsFile "/home/sibi/.local/bin/hpack"
+  --   version <- runPantryAppWithCustomHpack fp hpackVersion
+  --   version `shouldSatisfy` isVersion
```
qrilka: do we need to have it as a comment?

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +390,56 @@ parseArchive pli archive fp = do
...
             case mbs of
-              Nothing -> throwIO $ TreeReferencesMissingBlob pli cabalPath cabalBlobKey
+              Nothing -> throwIO $ TreeReferencesMissingBlob pli buildFilePath buildFileBlobKey
               Just bs -> pure bs
-          (_warnings, gpd) <- rawParseGPD (Left pli) bs
+          cabalBs <- if buildFileType == CabalFile
+                     then return bs
+                     else do
+                       (_, cabalbs) <- hpackToCabal bs
+                       return cabalbs
```
qrilka: what about `snd <$> hpackToCabal bs`?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -436,30 +449,122 @@ loadHackageTarballInfo name version = do
...
+    => TreeId
+    -> VersionId
+    -> ReaderT SqlBackend (RIO env) (Key HPack)
+generateHPack tid vid = do
+    let hpackPath =
+            maybe
+                (error
+                     "generateHPack: Not able to convert hpack file to SafeFilePath")
+                id $
+            P.mkSafeFilePath (T.pack Hpack.packageConfig)
```
qrilka: `\f -> maybe f id` == `fromMaybe`

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -436,30 +449,122 @@ loadHackageTarballInfo name version = do
...
-  :: (HasPantryConfig env, HasLogFunc env)
+  :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
   => P.PackageIdentifier
   -> P.Tree
   -> P.TreeEntry
   -- ^ cabal file
+  -> P.BuildFileType
   -> ReaderT SqlBackend (RIO env) (TreeId, P.TreeKey)
-storeTree (P.PackageIdentifier name version) tree@(P.TreeMap m) (P.TreeEntry (P.BlobKey cabal _) cabalType) = do
+storeTree (P.PackageIdentifier name version) tree@(P.TreeMap m) (P.TreeEntry (P.BlobKey btypeSha _) fileType) btype = do
```
qrilka: did you want to use `buildFileType` instead of `btype`?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -497,40 +604,80 @@ getTreeForKey (P.TreeKey key) = do
...
-  let cabalEntry = P.TreeEntry cabalKey (treeCabalType ts)
-  pure Package
-    { packageTreeKey = P.TreeKey key
-    , packageTree = tree
-    , packageCabalEntry = cabalEntry
-    , packageIdent = ident
-    }
+    (mts :: Maybe Tree) <- get tid
+    ts <-
+        case mts of
```
qrilka: is there any point in splitting arrow and a case in 2 lines?

psibi: @qrilka That's hindent there. Is it making it unreadable ?

subs/pantry/test/Pantry/HPackSpec.hs
psibi: If we don't have it as a comment, it will fail in CI. :-) Basically, this tests needs an external hpack executable for it to work. We need to modify the CI scripts for it to actually work. So, right now I have added the basic idea of the test. We can adapt it when we actually integrate it with the CI.

subs/pantry/src/Pantry/Archive.hs
psibi: Done.

subs/pantry/src/Pantry/Storage.hs
psibi: Done.

subs/pantry/src/Pantry/Storage.hs
psibi: Done.

subs/pantry/src/Pantry.hs
```diff
@@ -1006,6 +1010,27 @@ runPantryApp f = runSimpleApp $ do
           }
         f
 
+-- | Like 'runPantryApp', but you can give the custom HPack executable to it.
+--
+-- @since 0.1.0.0
+runPantryAppWithCustomHpack :: MonadIO m => Path Abs File -> RIO PantryApp a -> m a
+runPantryAppWithCustomHpack fp f = runSimpleApp $ do
```
snoyberg: It seems strange to have a call to `runSimpleApp` here. Why not have it live in `RIO`? Concrete concern: this will disrupt any kind of logging already set up in the program.

subs/pantry/src/Pantry.hs
```diff
@@ -510,7 +512,9 @@ loadCabalFileBytes (PLIHackage pir _mtree) = getHackageCabalFile pir
 loadCabalFileBytes pl = do
   package <- loadPackage pl
   let sfp = cabalFileName $ pkgName $ packageIdent package
-      TreeEntry cabalBlobKey _ft = packageCabalEntry package
+  cabalBlobKey <- case (packageCabalEntry package) of
+                       PCHpack _ -> throwIO $ TreeWithoutCabalFile pl
```
snoyberg: This doesn't feel right. Why aren't we generating the cabal file in this case? There's no indication at the type level or the docs that this may fail. At the very least, I'd say we should return a `Maybe` value instead of using runtime exceptions, since this is an expected case.

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +390,54 @@ parseArchive pli archive fp = do
...
-          mbs <- withStorage $ loadBlob cabalBlobKey
+          (buildFilePath, buildFileEntry@(TreeEntry buildFileBlobKey _), buildFileType) <- findCabalOrHpackFile pli tree
+          mbs <- withStorage $ loadBlob buildFileBlobKey
           bs <-
             case mbs of
-              Nothing -> throwIO $ TreeReferencesMissingBlob pli cabalPath cabalBlobKey
+              Nothing -> throwIO $ TreeReferencesMissingBlob pli buildFilePath buildFileBlobKey
               Just bs -> pure bs
-          (_warnings, gpd) <- rawParseGPD (Left pli) bs
+          cabalBs <- if buildFileType == CabalFile
```
snoyberg: I'd rather see a pattern match here so that, if we add something besides Cabal and Hpack in the future, the compiler will warn of the missing case.

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +390,54 @@ parseArchive pli archive fp = do
...
           let ident@(PackageIdentifier name _) = package $ packageDescription gpd
-          when (cabalPath /= cabalFileName name) $
-            throwIO $ WrongCabalFileName pli cabalPath name
+          when (buildFilePath /= cabalFileName name && buildFileType == CabalFile) $
+            throwIO $ WrongCabalFileName pli buildFilePath name
 
           -- It's good! Store the tree, let's bounce
-          (_tid, treeKey) <- withStorage $ storeTree ident tree cabalEntry
+          (_tid, treeKey) <- withStorage $ storeTree ident tree buildFileEntry buildFileType
+          let packageCabal = if buildFileType == CabalFile
```
snoyberg: Also: pattern match

subs/pantry/src/Pantry/Archive.hs
```diff
@@ -387,41 +390,54 @@ parseArchive pli archive fp = do
...
+findCabalOrHpackFile loc (TreeMap m) = do
   let isCabalFile (sfp, _) =
         let txt = unSafeFilePath sfp
-         in not ("/" `T.isInfixOf` txt) && ".cabal" `T.isSuffixOf` txt
-  case filter isCabalFile $ Map.toList m of
+         in not ("/" `T.isInfixOf` txt) && (".cabal" `T.isSuffixOf` txt)
+      isHpackFile (sfp, _) =
+        let txt = unSafeFilePath sfp
+         in T.pack (Hpack.packageConfig) == txt
+  case filter (\f -> isCabalFile f || isHpackFile f) $ Map.toList m of
```
snoyberg: How about `mapMaybe` and have a function that categorizes the files as cabal or hpack files, so that we don't have to call `isHpackFile` again? Also, instead of having a separate `BuildFileType`, I think I'd prefer to see:

```haskell
data BuildFile
  = BFCabal !SafeFilePath !TreeEntry
  | BFHpack !TreeEntry
```

Note that we don't even need a `SafeFilePath` for hpack files since we know what it has to be (package.yaml).

subs/pantry/src/Pantry/HPack.hs
```diff
@@ -0,0 +1,113 @@
...
+                        "then please delete the cabal file."
+            HpackCommand command ->
+                withWorkingDir (toFilePath pkgDir) $
+                proc command [] runProcess_
+
+
+hpackToCabal :: forall env. (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+           => ByteString -- Hpack's content
+           -> RIO env (PackageName, ByteString)
+hpackToCabal hpackBs = withSystemTempDirectory "hpack-pkg-dir" $ \tmpdir -> withWorkingDir tmpdir $ do
```
snoyberg: This is insufficient. hpack depends on the _entire_ package contents on the filesystem, not just the hpack file. You'll need to receive the full tree and write it completely to the working directory.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -145,12 +150,20 @@ CacheUpdate
...
     name PackageNameId
     version VersionId
     UniqueTree key
 
+HPack
+   tree TreeId
+   version VersionId
+   -- Corresponding cabal file of hpack
+   cabal BlobId
+   path FilePathId
```
snoyberg: Is this the package.yaml path name, or the generated cabal file name?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -436,30 +449,121 @@ loadHackageTarballInfo name version = do
...
+                }
+    val <- insertBy hpackRecord
+    return $ getKey val
+
+
+hpackVersionId ::
+       (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+    => ReaderT SqlBackend (RIO env) VersionId
+hpackVersionId = do
+    hpackSoftwareVersion <- lift $ hpackVersion
```
snoyberg: Does this handle the case when hpack is an external program? Doesn't this assume we're always using the compiled-in version?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -436,30 +449,121 @@ loadHackageTarballInfo name version = do
...
   nameid <- getPackageNameId name
   versionid <- getVersionId version
+  let cabalid' = case buildFileType of
+                   P.HPackFile -> Nothing
+                   P.CabalFile -> Just buildid
   etid <- insertBy Tree
     { treeKey = bid
-    , treeCabal = cabalid
-    , treeCabalType = cabalType
+    , treeCabal = cabalid'
```
snoyberg: If you implement the `data BuildFile` idea I mentioned elsewhere, then these two fields will collapse into one.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -148,6 +152,21 @@ data Package = Package
   }
   deriving (Show, Eq)
 
+data PackageCabal = PCCabalFile !TreeEntry -- ^ TreeEntry of Cabal file
```
snoyberg: This data type feels wrong to me. What's the case where you need the `PCHpack`? I would have expected a data type closer to:

```haskell
data PackageCabal
  = PCCabalFile !TreeEntry
  | PCHpack
      !TreeEntry -- original hpack file, if actually needed
      !TreeEntry -- generated cabal file
      !Version -- version of hpack used
```

subs/pantry/src/Pantry/Types.hs
```diff
@@ -148,6 +152,21 @@ data Package = Package
...
+                  | PCHpack !TreeEntry -- ^ TreeEntry of HPack file
+                  | PCHpackCabalFile !TreeEntry -- ^ TreeEntry of Cabal file produced via HPack
+                    deriving (Show, Eq)
+
+getPCBlobKey :: PackageCabal -> BlobKey
```
snoyberg: Similarly, this makes me very nervous: it's a function which silently gives either a cabal file or an hpack file. We're throwing away a lot of type safety.

subs/pantry/src/Pantry/Storage.hs
psibi: Yes, it does handle the case when `hpack` is an external program in the implementation of the function `hpackVersion`. We pattern match between `HpackBundled` and `HpackCommand` for it.

snoyberg: Got it, thanks for clarifying.

subs/pantry/src/Pantry/Storage.hs
psibi: It's the generated cabal file name. For hpack based builds, `package.yaml` will anyway be present in `TreeEntry`.

subs/pantry/src/Pantry/HPack.hs
```diff
@@ -0,0 +1,113 @@
...
+
+
+hpackVersion
+  :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+  => RIO env Version
+hpackVersion = do
+  he <- view $ pantryConfigL.to pcHpackExecutable
+  case he of
+    HpackBundled -> do
+                 version <- BL.unpack <$> (proc "stack" ["--hpack-numeric-version"] readProcessStdout_)
```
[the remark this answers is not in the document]
qrilka: +1 it makes perfect sense to use `VERSION_hpack` as `--hpack-numeric-version` does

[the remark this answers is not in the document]
psibi: Thanks for pointing it out, it would be much more efficient!

subs/pantry/src/Pantry/Types.hs
psibi: Changed the data type accordingly.

subs/pantry/src/Pantry/Types.hs
psibi: This function and it's usage has been removed altogether.

subs/pantry/src/Pantry.hs
psibi: This is fixed.

subs/pantry/src/Pantry/Archive.hs
psibi: Fixed.

subs/pantry/src/Pantry/Archive.hs
psibi: Fixed.

subs/pantry/src/Pantry/HPack.hs
psibi: This has been fixed. The code has been moved to `Storage.hs` to avoid cyclic dependency.

subs/pantry/src/Pantry/Archive.hs
psibi: Done.

subs/pantry/src/Pantry/HPack.hs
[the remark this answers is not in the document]
psibi: Done in the latest push.

subs/pantry/src/Pantry.hs
psibi: I have removed this function altogether. This function was added for it's use in one of the unit testing functions. I have now exported a lens function from this module for achieving the same.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -437,30 +459,119 @@ loadHackageTarballInfo name version = do
   where
     go (Entity _ ht) = (hackageTarballSha ht, hackageTarballSize ht)
 
+getKey :: Either (Entity record) (Key record) -> Key record
+getKey dat = case dat of
```
NorfairKing: nit: not a very descriptive name.
Then again, `getKeyFromEntityOrItself` is not really better.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -729,6 +909,88 @@ sinkHackagePackageNames predicate sink = do
...
+    -> RIO env (P.PackageName, Path Abs File)
+findOrGenerateCabalFile pkgDir = do
+    hpack pkgDir
+    files <- filter (flip hasExtension "cabal" . toFilePath) . snd
+         <$> listDir pkgDir
+    -- If there are multiple files, ignore files that start with
+    -- ".". On unixlike environments these are hidden, and this
+    -- character is not valid in package names. The main goal is
+    -- to ignore emacs lock files - see
+    -- https://github.com/commercialhaskell/stack/issues/1897.
```
NorfairKing: Nit: What if there is an emacs lock file for a cabal file that has been deleted in the meantime?
I just checked the comment above again and it seems inaccurate in that situation since you _always_ ignore files that start with a dot.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -729,6 +909,88 @@ sinkHackagePackageNames predicate sink = do
...
+    let isHidden ('.':_) = True
+        isHidden _ = False
+    case filter (not . isHidden . toFilePath . filename) files of
+        [] -> throwIO $ P.NoCabalFileFound pkgDir
+        [x] -> maybe
+          (throwIO $ P.InvalidCabalFilePath x)
+          (\pn -> pure $ (pn, x)) $
+            List.stripSuffix ".cabal" (toFilePath (filename x)) >>=
+            P.parsePackageName
+        _:_ -> throwIO $ P.MultipleCabalFilesFound pkgDir files
```
NorfairKing: Nit: Could just be `_`, but this your version is more future proof.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -729,6 +909,88 @@ sinkHackagePackageNames predicate sink = do
...
+    files <- filter (flip hasExtension "cabal" . toFilePath) . snd
+         <$> listDir pkgDir
+    -- If there are multiple files, ignore files that start with
+    -- ".". On unixlike environments these are hidden, and this
+    -- character is not valid in package names. The main goal is
+    -- to ignore emacs lock files - see
+    -- https://github.com/commercialhaskell/stack/issues/1897.
+    let isHidden ('.':_) = True
+        isHidden _ = False
+    case filter (not . isHidden . toFilePath . filename) files of
```
NorfairKing: How about if `toFilePath` gives you `./foobar.cabal`? This can happen in new-ish versions of `path` with relative files, I think.

NorfairKing: Another nit: it's better to use `fromRelFile` than `toFilePath` for future-proofing.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -148,6 +155,13 @@ data Package = Package
   }
   deriving (Show, Eq)
 
+data PackageCabal = PCCabalFile !TreeEntry -- ^ TreeEntry of Cabal file
+                  | PCHpack
+                    !TreeEntry -- Original hpack file
+                    !TreeEntry -- Generated Cabal file
```
NorfairKing: Super-nit: If someone adds a field to this product and then adds an underscore to wherever it's being matched on, they won't get a type error.

subs/pantry/src/Pantry/Storage.hs
psibi: I don't have strong opinion on either of the names. While the current name isn't great, I feel it's fine. The function implementation is small enough and it isn't exported to other modules. Although, if you any other nice name for the function and describes more concisely, I think we can rename it.

subs/pantry/src/Pantry/Storage.hs
psibi: In that case we throw the exception that there is no cabal file found!

I don't think we want to use emacs lock file for a cabal file when the original file has been intentionally deleted.

subs/pantry/src/Pantry/Storage.hs
psibi: :+1:

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -146,12 +159,20 @@ CacheUpdate
 -- table.
 Tree
     key BlobId
-    cabal BlobId
+    cabal BlobId Maybe
```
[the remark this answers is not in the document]
psibi: I have added a comment in the recent push.

subs/pantry/src/Pantry/Storage.hs
psibi: With the recent push, we use `fromRelPath`.

subs/pantry/src/Pantry/Types.hs
psibi: Fixed in the latest push.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -855,6 +879,18 @@ cabalSpecLatestVersion =
     CabalSpecV2_2 -> error "this cannot happen"
     CabalSpecV2_4 -> mkVersion [2, 4]
 
+data BuildFile = BFCabal !SafeFilePath !TreeEntry
+               | BFHpack !TreeEntry -- We don't need SafeFilePath for Hpack since it has to be package.yaml file
+  deriving (Show, Eq)
+
+isHPackBuildFile :: BuildFile -> Bool
```
snoyberg: I would rather not have these two functions at all. Each one is only used once, and explicit pattern matching in those areas will likely have improvements on the code. For example, I don't like the idea of having a `ByteString` value present and then checking a `Bool` value to tell us whether to treat it as an hpack or cabal file.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -890,6 +926,11 @@ instance PersistFieldSql SafeFilePath where
 unSafeFilePath :: SafeFilePath -> Text
 unSafeFilePath (SafeFilePath t) = t
 
+safeFilePathtoPath :: Path Abs Dir -> SafeFilePath -> IO (Path Abs File)
```
snoyberg: This could probably be a `MonadThrow` constraint, right?

subs/pantry/src/Pantry/Types.hs
```diff
@@ -904,6 +945,13 @@ mkSafeFilePath t = do
 
   Just $ SafeFilePath t
 
+hpackSafeFilePath :: SafeFilePath
```
snoyberg: Please add a Haddock explaining what this value is.

subs/pantry/src/Pantry/Storage.hs
snoyberg: Please add a comment at least explaining that this is the cabal file path. Changing to `cabalPath` may be even better.

subs/pantry/src/Pantry/Types.hs
psibi: Added.

subs/pantry/src/Pantry/Types.hs
psibi: Done.

subs/pantry/src/Pantry/Types.hs
psibi: Remove both of them.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -729,6 +926,88 @@ sinkHackagePackageNames predicate sink = do
...
+  !bs <- lift $ B.readFile (fromAbsFile cfile)
+  lift $ removeDirRecur tmpDir
+  return $ (packageName, bs)
+
+hpackToCabal :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+           => ByteString -- Hpack's content
+           -> P.Tree
+           -> RIO env (P.PackageName, ByteString)
+hpackToCabal hpackBs tree = withSystemTempDirectory "hpack-pkg-dir" $ \tmpdir -> do
+               B.writeFile (tmpdir FilePath.</> Hpack.packageConfig) hpackBs
```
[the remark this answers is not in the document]
psibi: Fixed.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -148,6 +156,17 @@ data Package = Package
...
 
+data PHpack = PHpack
+    {
+      phOriginal :: !TreeEntry, -- ^ Original hpack file
+      phGenerated :: !TreeEntry, -- ^ Generated Cabal file
+      phVersion :: !Version -- ^ Version of Hpack used
+    } deriving (Show, Eq)
+
+data PackageCabal = PCCabalFile !TreeEntry -- ^ TreeEntry of Cabal file
+                  | PCHpack PHpack
```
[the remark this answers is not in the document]
psibi: Converted to strict.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -624,6 +643,7 @@ data PantryException
   | NoHackageCryptographicHash !PackageIdentifier
   | FailedToCloneRepo !Repo
   | TreeReferencesMissingBlob !PackageLocationImmutable !SafeFilePath !BlobKey
+  | TreeReferencesMissing !SafeFilePath !BlobKey
```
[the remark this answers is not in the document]
psibi: Removed the constructor entirely.

subs/pantry/src/Pantry/Storage.hs
snoyberg: I meant calling `path` instead `cabalPath` would be a clearer naming here.

subs/pantry/src/Pantry/Storage.hs
snoyberg: I'll just replace the function with `either entityKey id`, which is pretty common in persistent code.

subs/pantry/src/Pantry/Storage.hs
snoyberg: It can't happen since the `filename` call will strip off any leading directories.

PR #4444, opened 2018-12-13
Have travis build after trusty support is deprecated

dbaynard:
Targeting stable as this means all relevant branches can straightforwardly benefit from the change.

This may require a little back and forth with CI.

I suggest we also change the travis templates, too, before merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@
 # use cabal without any Stackage snapshots.
 #
 language: generic
+dist: xenial
 
 cache:
     directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@
 # use cabal without any Stackage snapshots.
 #
 language: generic
-sudo: false
 
 cache:
     directories:
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,8 @@ Behavior changes:
 
 Other enhancements:
 
+* Travis no longer builds using `sudo: false` as that behaviour is due to be deprecated.
+
 Bug fixes:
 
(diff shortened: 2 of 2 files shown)


dbaynard:
Well, the good news is it runs (and is green on some). It's timing out on a few jobs, however.

@qrilka What would you recommend?

qrilka:
I think we had timeouts like that in previous builds and they need to be addressed in a separate issue. Otherwise it looks like a good fix though I'm not an expert on Travis CI.

dbaynard:
@borsboom Can we merge this CI change?

> However, the IP addresses for the workers running your build are different between the two infrastructures. If you safelist Travis CI IP addresses for your container-based builds, please include the virtual-machine-based builds as soon as possible to reduce interruption in your CI/CD pipeline.

This is from the [travis blog](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) though I don't think it's relevant.

That said, it seems they've already done the first part of the migration

- 19 November, 2018 - Today we publish this post and are ready to answer all your questions!
- 28 November, 2018 - We will send a service email to remind folks still using sudo: false on recent builds to remind you to migrate.
- 03 December, 2018 - We will start randomly sampling projects on both travis-ci.org and travis-ci.com to move them permanently to using the virtual-machine-based infrastructure for all builds. The projects will be migrated incrementally over a few days
- 07 December, 2018 - All projects that use a Linux build environment will be fully migrated to using the same Linux infrastructure, which runs builds in virtual-machines.

So they've disabled `sudo: false`. It's now just a question of ensuring the builds pass on xenial.

PR #4446, opened 2018-12-16
Add `--cabal-files` flag to `ide packages` command (backported to stable)

DanielG:
Backport to of #4312 to `stable`. I was hoping this would have landed in a realease by now, didn't realise there was a big project happening on master.

FYI: This will be necessary for haskell-ide-engine to keep ineroperating with Stack once cabal new-build support lands there. 

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [x] The documentation has been updated, if necessary.

PR #4477, opened 2018-12-30
Make `ide packages --cabal-files` write to stdout

DanielG:
I didn't realise `logInfo` writes to stderr when writing #4446. On master the `ide packages` command already writes everything to stdout, including with the `--cabal-files` flag. This commit only changes the behaviour of the new (and unreleased) `--cabal-files` flag so as not to break anything.

### DIFF
--- a/src/Stack/IDE.hs
+++ b/src/Stack/IDE.hs
@@ -14,6 +14,7 @@ module Stack.IDE
 import qualified Data.Map as Map
 import qualified Data.Set as Set
 import qualified Data.Text as T
+import           System.IO (putStrLn)
 import           Stack.Config (getLocalPackages)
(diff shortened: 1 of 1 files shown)


DanielG:
Nevermind, I remembered that wrong -- master also does this wrong. I'll add a `--stdout` flag or something instead.

mgsloan:
This looks right to me.  The general rule of thumb is that info intended for users goes to stderr and info that is intended to be machine parseable goes to stdout.  For example, the `stack path` and `stack query` commands primarily write to stdout.

DanielG:
Indeed but my concern is that other tools might already rely on the output going to stderr which is currently the case. I'd be happy to send a couple PRs doing `s/logInfo/putStrLn/` on stable and master though if that's not a concern.

PR #4422, opened 2018-11-28
[WIP] Split raw locations

qrilka:
Fixes #4418 

Please include the following checklist in your PR:

* [ ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [ ] The documentation has been updated, if necessary.

Run some integration tests and some manual test invocations

PR #4479, opened 2019-01-01
Split raw locations

psibi:
Supersedes https://github.com/commercialhaskell/stack/pull/4422


subs/pantry/src/Pantry/Types.hs
```diff
@@ -179,7 +194,7 @@ data PantryConfig = PantryConfig
   -- ^ Want to try updating the index once during a single run for missing
   -- package identifiers. We also want to ensure we only update once at a
   -- time. Start at @True@.
-  , pcParsedCabalFilesImmutable :: !(IORef (Map PackageLocationImmutable GenericPackageDescription))
+  , pcParsedCabalFilesRawImmutable :: !(IORef (Map RawPackageLocationImmutable GenericPackageDescription))
```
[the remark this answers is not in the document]
qrilka: It's needed for `loadCabalFileRawImmutable` which is currently used in Stack, if (or when) `Stack.Snapshot` will be remove we could get rid of "Raw" here. But if you have some suggestions how to handle it already now - please show

subs/pantry/src/Pantry/Types.hs
```diff
@@ -1116,20 +1233,49 @@ instance Store OptionalSubdirs
 -- has the expected information.
```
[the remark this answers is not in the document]
qrilka: I'm not the author here but as far as I get it "config" here means either `stack.yaml` or `snapshot.yaml`. As for the package name field in it I don't know a good example when metadata could exist without a package name in it, maybe @snoyberg could explain it?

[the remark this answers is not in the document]
snoyberg: If you include a package as an archive or Git repo, you are not required to specify the name or version in the config, those are derived after parsing the cabal file.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -257,19 +285,54 @@ instance Display PackageLocation where
   display (PLImmutable loc) = display loc
   display (PLMutable fp) = fromString $ toFilePath $ resolvedAbsolute fp
 
+-- | Convert `PackageLocation` to its "raw" equivalent
+--
+-- @since 0.1.0.0
+toRawPL :: PackageLocation -> RawPackageLocation
+toRawPL (PLImmutable im) = RPLImmutable (toRawPLI im)
+toRawPL (PLMutable m) = RPLMutable m
```
NorfairKing: It's worth notice that the corresponding roundtrip loses info.
So if you start from a `RawPackageLocation`, then make it into a `PackageLocation` and then back, you'll lose information and that could make the error messages a bit annoying.
For error messages we should use `RawPackageLocation` if possible.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -257,19 +285,54 @@ instance Display PackageLocation where
...
+    (if T.null $ repoSubdir repo
+       then mempty
+       else " in subdir " <> display (repoSubdir repo))
+
 -- | Location for remote packages or archives assumed to be immutable.
 --
 -- @since 0.1.0.0
 data PackageLocationImmutable
-  = PLIHackage !PackageIdentifierRevision !(Maybe TreeKey)
+  = PLIHackage !PackageIdentifier !BlobKey !TreeKey
```
psibi: @qrilka Any reason why this is changed to `PackageIdentifier` from `PackageIdentifierRevision`. I see that in the Raw version it is `PackageIdentifierRevision`.

subs/pantry/src/Pantry/Types.hs
qrilka: Nah, that opens up an interesting situation: if for a package in `extra-deps` you don't set a concrete revision Stack is supposed to resolve it and work with that resolved exact location (maybe from a lock file which is to appear soon). If we won't keep a corresponding raw location then error message could be misleading: e.g. after Stack lock file was created package author could have created a new revision fixing dependency problem but Stack will complain about that previous resolved revision if there will be no measures against such a situation.

src/Stack/Build/ConstructPlan.hs
```diff
@@ -422,11 +423,16 @@ addDep treatAsDep' name = do
...
                             -- names. This code does not feel right.
-                            tellExecutablesUpstream
-                              name
-                              (PLIHackage (PackageIdentifierRevision name (installedVersion installed) CFILatest) Nothing)
-                              loc
-                              Map.empty
+                            let version = installedVersion installed
+                            mrev <- liftRIO $ getLatestHackageRevision name version
+                            case mrev of
+                              Nothing -> return () -- FIXME:qrilka report an error?
```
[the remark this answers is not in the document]
qrilka: Completely missed this, @psibi will you take care of it?

subs/pantry/src/Pantry/Types.hs
NorfairKing: Aha! Thanks

src/Stack/Build/ConstructPlan.hs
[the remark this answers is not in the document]
psibi: Sure.

subs/pantry/src/Pantry/Types.hs
qrilka: for exact locations we don't need `CabalFileInfo`, , `TreeKey` is enough and e.g. `CFILatest` is not an exact revision :)

psibi: Ah, I see. Thanks.

qrilka:
I think it's not quite correct to approve code which is mostly of my authorship but still overall I don't see any obvious issues with this PR
@psibi will you resolve the conflict?

psibi:
@qrilka Sure, I will fix that by tomorrow.

psibi:
@qrilka Resolved the conflicts and the CI is green too.

mihaimaruseac:
Thanks to both @psibi and @qrilka for this

Issue #4488, opened 2019-01-05
How to use stack with newer cabal version for older GHC?


vrom911:
## Problem

As far as I see Stack LTS not only determine the version of GHC, but also restricted to use only the specific Cabal version that is stuck with the particular GHC. This creates some problems with using a unique version of Cabal for different versions of GHCs, for example, using common stanzas in cabal file with `cabal: 2.4` and support both GHC-8.6.3 and GHC-8.4.4.

Stack also doesn't allow and deprecates now options: `--install-cabal` and `--upgrade-cabal`.

The question is: how is it possible to use a newer version of Cabal (eg. 2.4) for the old GHCs (e.g. 8.0.2)?

### Example

I've created a test project to illustrate the problem clearer. It can be found here:

* https://github.com/vrom911/summon-test

 I want to support GHC versions:
 * 8.0.2
 * 8.2.2
 * 8.4.4
 * 8.6.3

To assure that everything works at all versions both with Stack and Cabal I've created this Travis file:

* https://github.com/vrom911/summon-test/blob/6ac63c3ad672d24b6f61589cdd163d33a5e2d64a/.travis.yml

where I explicitly specify `- cabal: 2.0`.

Though it's ignored, so I also added this if statement (which is apparently doesn't work either):

https://github.com/vrom911/summon-test/blob/6ac63c3ad672d24b6f61589cdd163d33a5e2d64a/.travis.yml#L42-L44

And here is the output of failing Travis job for GHC-8.0.2 can be found here:

* https://travis-ci.org/vrom911/summon-test/jobs/475464765

### Stack version

```
Version 1.9.3, Git revision 40cf7b37526b86d1676da82167ea8758a854953b (6211 commits) x86_64 hpack-0.31.1
```

### Method of installation

```
curl -sSL https://get.haskellstack.org/ | sh
```

snoyberg:
Have you tried adding `Cabal` as an `extra-deps` entry? I'm not sure if it will be used for `build-type: Simple` or not, but it's certainly possible for us to add a configuration option to Stack to _force_ the newer Cabal library version to be used. The tradeoff is that it will force the (really CPU and memory intensive) build of Cabal before anything else can be built. But that sounds like exactly what you want here.

qrilka:
[...]

Issue #4475, opened 2018-12-30
DISCUSS: Drop support for older Cabal library versions


snoyberg:
Stack's design involves shelling out to the Setup.hs file for most activities it performs. IIRC, we currently support Setup.hs files compiled against Cabal versions as old as 1.18. However, keeping this support involves conditionals in the codebase. At some point, we will likely want to drop support for older versions of the Cabal library.

As we're looking at Stack 2.0 and adding support for Backpack, it looks like adding conditional support for component-based builds may add quite a bit of complexity to the codebase. I'd like to propose that we drop support for older versions of the Cabal library.

Upside: simplified code base, more thorough testing

Downside: users of older snapshots that include an older GHC and/or Cabal version will need to stick to older versions of the Stack executable

We should make a decision about this soon to allow those interested to start diving into component-based builds sooner rather than later.

juhp:
Do you want to define older? :-)
Do you mean say Cabal < 2?

snoyberg:
I was vague on that, good point. I _think_ we'd be dropping Cabal < 2, but until someone jumps into the work I don't think we'll be 100% certain which library versions will be the most important to try dropping.

qrilka:
+1 on that, I think it makes sense to check how good is support of component-based builds in Cabal 2.0 and 2.2, from what I have seen on the topic it's not quite clear if we'll be good enough already with 2.0

borsboom:
I think this is reasonable.  It's already pretty common for OS upgrades to break support for older GHC versions, so the idea of not being able to use old GHC versions indefinitely is not new.  The standard for build reproducibility that's emerged is that if your environment (stack version, OS version, etc.) stays the same then builds should be reproducible, but if any of those change then there's no guarantee (although we try not to break things unnecessarily).

Tangentially related: if we're going to break support for old GHC versions (via breaking support for older Cabal versions), I wonder if it'd be worth starting with a new `stack-setup-X.yaml` file so we can clean up some of the cruft that's accumulated to maintain compatibility with older Stack versions.  We'd add new GHC versions in both places for a while but eventually older Stack versions will stop supporting newer GHC versions anyway.

mihaimaruseac:
I'm also ok with dropping support on a major change of Stack.

mgsloan:
Would this mean dropping support for lts snapshots earlier than `lts-10.0`?  Or would the snapshot's selection of Cabal version be ignored?

snoyberg:
[...]

Issue #3049, opened 2017-03-08
Remove special global Cabal?


mgsloan:
I believe that with support for the custom-setup stanza, https://github.com/commercialhaskell/stack/pull/2866 , we're using the snapshot / extra-deps version of Cabal for building anything with a custom-setup stanza.

For everything else, we're using the globally installed version of Cabal (the one updated via `stack setup --upgrade-cabal`).  This can be nice, as you can update your Cabal version without needing to modify your configurations.  This means that you can get bugfixes / enhancements without updating your build configs.

This seems inconsistent to me, I think it might simplify things to always use the snapshot / extra-deps version of Cabal.

Edit: One tricky detail that I'm only considering now that I've typed all this up, is that in order to build Cabal and its dependencies, we'll need to use an existing version of Cabal - likely the global one.  So it seems hard to get full consistency of Cabal version used.  Thoughts on this?  Just special case Cabal and its transitive deps?

We'd get the following benefits:

1) Better build reproducibility
2) Ability to use a custom Cabal - https://github.com/commercialhaskell/stack/issues/2910 - can even have it as a local package incase you need custom modifications
3) Obviates the need for enhancing `stack setup` to specify a cabal version to install (https://github.com/commercialhaskell/stack/issues/2386 / https://github.com/commercialhaskell/stack/pull/2995)
4) Would fix Cabal version mismatch errors caused by lacking a custom-setup stanza - https://github.com/commercialhaskell/stack/issues/2924 
5) Obviates the need for a proposed `require-cabal-version` field in stack.yaml - https://github.com/commercialhaskell/stack/issues/2479
6) Removal of `stack setup --upgrade-cabal` would mean https://github.com/commercialhaskell/stack/issues/2230 can be closed (case where it wants to write to a system ghc install)

This would add an implicit dependency on Cabal to every package. So, if Cabal or anything it depends on is installed locally, *everything* must be installed locally rather than in the snapshot.

snoyberg:
Big :-1:. Building Cabal is a long, slow process. We don't want to add that to the beginning of every new snapshot usage.

I'd be OK with this applying just to custom Setup.hs scripts perhaps. But I'm really not seeing enough benefit for simple build types.

mgsloan:
Hmm, not sure if I buy that argument:

1) Package sharing between snapshots will often make recompilation of Cabal unnecessary.
2) It seems quite probable that non-trivial projects will have some transitive dependency which uses a custom build type.

[...]

PR #4494, opened 2019-01-08
Document how stack uses the Cabal library in Architecture.md

dbaynard:
Relates to #3049, #4475, #4488.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
   there may be more options here for new projects in the future (see issue
   [253](https://github.com/commercialhaskell/stack/issues/253))
 
+### Cabal-the-library
+
+The relationship between stack and Cabal-the-library is complicated.
+
+1. Stack itself builds against a version of the Cabal library, which it uses for parsing Cabal files.
+
+2. Stack wraps functionality from a version of the Cabal library in order to build packages.
+
+    i. For packages with `build-type: Simple`, Stack uses a setup executable compiled with a version of the Cabal library to perform builds. All packages using the same compiler and Cabal version are built with the same executable. These executables are cached in the `setup-exe-cache` configuration directory.
+
+    ii. For packages with `build-type: Custom`, Stack compiles `Setup.hs` against a version of the Cabal library and uses that setup executable to perform builds. The executable is stored TODO
+
+3. Packages themselves may depend on the Cabal library.
+
+There are three (nearly) corresponding versions of the Cabal library which are relevant to a build.
+
+1. The version used to compile stack, which is not necessarily present on stack users’ machines.
+
+   Stack will always be compiled using the most recent stable version of Cabal-the-library.
+
+2. The [boot version used by GHC](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Libraries/VersionHistory), which is globally available to stack.
+
+   This version is used to compile the `build-type: Simple` setup executable.
+   Build artefacts are placed in corresponding `.stack-work/dist/Cabal-xxxxx` directory.
+
+3. The snapshot version (which may be overridden using `extra-deps`).
+
+   This is the version on which a package may depend.
+
+   However, it is also used to compile any `build-type: Custom` setup executables.
+
+In addition, the Cabal version required by `custom-build` will be installed, if necessary.
+
+Note that these versions may be the same.
+
+There are a number of consequences of this design.
+
+1. Snapshot packages only depend on the GHC compiler version, not the Cabal library version (with the exception of `build-type: Custom` packages).
+
+2. The most recent stack can always read the most recent Cabal files.
+
+3. However, stack may not be able to build packages defined using those files.
+
+   This occurs when
+
+   i. The package uses `build-type: Simple` and the Cabal file format requires a more recent version of Cabal than the global (boot) version for the compiler.
+
+      e.g. Stack lts-11.22 uses GHC 8.2.2, corresponding to Cabal 2.0.1.0.
+      A library using `build-type: Simple` and SPDX license identifiers (introduced in Cabal 2.2) will not build, though stack will process the Cabal file correctly.
+
+          Cabal-simple_mPHDZzAJ_2.0.1.0_ghc-8.2.2: ./file.cabal:7: Parse of field 'license' failed.
+
+4. Stack only occasionally needs to build Cabal-the-library.
+   This is a resource intensive build, so avoiding it improves performance.
+
+5. Stack uses `ghc-pkg` to identify the Cabal version it should use for the setup executable for a build.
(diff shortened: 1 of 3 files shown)


dbaynard:
The bulk of the content may belong in a different document.

doc/faq.md
```diff
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
...
+
+The relationship between stack and Cabal-the-library is complicated.
+
+1. Stack itself builds against a version of the Cabal library, which it uses for parsing Cabal files.
+
+2. Stack wraps functionality from a version of the Cabal library in order to build packages.
+
+    i. For packages with `build-type: Simple`, Stack uses a setup executable compiled with a version of the Cabal library to perform builds. All packages using the same compiler and Cabal version are built with the same executable. These executables are cached in the `setup-exe-cache` configuration directory.
+
+    ii. For packages with `build-type: Custom`, Stack compiles `Setup.hs` against a version of the Cabal library and uses that setup executable to perform builds. The executable is stored TODO
```
snoyberg: "against a version of the Cabal library"- did you mean to include "specified by your stack.yaml file"?

doc/faq.md
```diff
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
   there may be more options here for new projects in the future (see issue
   [253](https://github.com/commercialhaskell/stack/issues/253))
 
+### Cabal-the-library
+
+The relationship between stack and Cabal-the-library is complicated.
+
+1. Stack itself builds against a version of the Cabal library, which it uses for parsing Cabal files.
```
snoyberg: Worth pointing out that this determines which version of the cabal file format Stack is able to parse.

doc/faq.md
```diff
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
...
+
+2. Stack wraps functionality from a version of the Cabal library in order to build packages.
```
snoyberg: "wraps" may be slightly confusing. Stack builds a `Setup.hs` file against a version of the Cabal library and then shells out to that executable. Stack itself does not perform any of the configure/build/copy/register stuff in-process. This allows it to more easily support newer versions of the Cabal library for performing builds, which is why _typically_ (but not always) you can use a Stack executable with a newer version of GHC than it was built for.

doc/faq.md
```diff
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
...
+
+    ii. For packages with `build-type: Custom`, Stack compiles `Setup.hs` against a version of the Cabal library and uses that setup executable to perform builds. The executable is stored TODO
+
+3. Packages themselves may depend on the Cabal library.
+
+There are three (nearly) corresponding versions of the Cabal library which are relevant to a build.
+
+1. The version used to compile stack, which is not necessarily present on stack users’ machines.
+
+   Stack will always be compiled using the most recent stable version of Cabal-the-library.
```
snoyberg: Not really true. We may attempt to do that, but we could for example release a point release of an older Stack release without upgrading to the newest Cabal library version.

doc/faq.md
```diff
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
...
+
+In addition, the Cabal version required by `custom-build` will be installed, if necessary.
+
+Note that these versions may be the same.
+
+There are a number of consequences of this design.
+
+1. Snapshot packages only depend on the GHC compiler version, not the Cabal library version (with the exception of `build-type: Custom` packages).
+
+2. The most recent stack can always read the most recent Cabal files.
```
snoyberg: "Usually"

doc/faq.md
dbaynard: I think this is covered later — so it probably makes more sense to zip these two lists of numbered paragraphs (the relationship and the versions) and have each bit of functionality in a self-contained unit.

doc/faq.md
dbaynard: ditto

doc/faq.md
```diff
@@ -57,6 +57,65 @@ directory. None of this should affect any existing Haskell tools at all.
...
+      A library using `build-type: Simple` and SPDX license identifiers (introduced in Cabal 2.2) will not build, though stack will process the Cabal file correctly.
+
+          Cabal-simple_mPHDZzAJ_2.0.1.0_ghc-8.2.2: ./file.cabal:7: Parse of field 'license' failed.
+
+4. Stack only occasionally needs to build Cabal-the-library.
+   This is a resource intensive build, so avoiding it improves performance.
+
+5. Stack uses `ghc-pkg` to identify the Cabal version it should use for the setup executable for a build.
+   The build output is stored in a directory `.stack-work/dist/Cabal-xxxxx` named for the Cabal version used to compile the setup executable.
+   Together, this means it is only possible to know for sure which Cabal version ought to be used if the corresponding compiler is installed.
```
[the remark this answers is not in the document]
dbaynard: Ah — I didn't know that `global-hints.yaml` was new. What should we say here?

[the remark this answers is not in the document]
qrilka: As far as I understand the idea behind that file is to have information about boot packages without having GHC installed, currently it's used in (to be removed) `Stack.Snapshot` and in https://github.com/commercialhaskell/stack/blob/master/src/Stack/SourceMap.hs#L196 but that's actual for `master`, not for `stable`.

doc/faq.md
dbaynard: Fixed at https://github.com/dbaynard/stack/blob/f31e9dd962aefa6c11854496ff578244618a9623/doc/faq.md#L85-L87

doc/faq.md
dbaynard: I've used your language to fix.

doc/faq.md
dbaynard: Added 'where possible' and made clear this refers to releases.

https://github.com/dbaynard/stack/blob/9900d99114967698223f43c84e4cedecc24cea97/doc/faq.md#L70-L72

dbaynard:
Two commits:

1.  Address straightforward issues.
2.  Refactor completely to fix more complicated ones.

I still have a few things left to do, here (e.g. the TODO).

dbaynard:
I need to put this in a new document, rather than the FAQ. Also, I'd like to rebase to take advantage of working CI on `stable`. I'll do all that when the content is OK.

dbaynard:
I think this is ready, in terms of content (assuming the latest addition is correct; it is what my logs suggest though I haven't dug into the code). It should go in a new file, though, rather than the FAQ (perhaps with the 'What is the relationship between stack and cabal' bit directly above, too).

dbaynard:
Actually, I moved the new material into `Architecture`, rather than a new document, and added links. I'll merge when it clears appveyor.

PR #4500, opened 2019-01-10
List global package deps in stack dot graph

### LINKED ISSUE #4324: Stack dot should handle properly GHC boot packages

vanceism7:
Changed the code for the dependency loader to read dependencies for global packages from the global dump. 

Tested the changes by doing the basic setup described in the issue (#4324)
i.e: ran the command `stack exec stack dot -- --external | grep '"process"'`


src/Stack/Dot.hs
```diff
@@ -275,24 +275,37 @@ createDepLoader sourceMap installed globalDumpMap globalIdMap loadPackageDeps pk
...
         case maybePkg of
             Just (_, Library _ _ mlicense) -> mlicense
             _ -> Nothing
     payloadFromDump dp = DotPayload (Just $ pkgVersion $ dpPackageIdent dp) (Right <$> dpLicense dp)
 
+    getDepsFromDump dump =
+      case dump of
+          Nothing ->
+            let errText = "Invariant violated: Expected to find wired-in-package "
+            in error (errText ++ packageNameString pkgName ++ " in global DB")
```
qrilka: This error branch is actual only for wired-in packages but it potentially could also get applied for non-wired-in ones. I think it's better to to move it out of this function and make this function accept `DumpPackage` without `Maybe`, probably renaming it to something like `dumpPackageDeps` (though it doesn't sound like `createDeplLoader` choice of which I'm not sure I quite understand).

qrilka: Actually I'm not sure that branching on wired-in packages makes much sense, what do you think about refactoring the code of `createDepLoader` into something like `projectPackgeDeps <|> dependencyDeps <|> globalDeps` where each of those will do a lookup into a corresponding table?

vanceism7: I was thinking about that myself. So would we just totally get rid of the `createDepLoader` function? And the `globalDeps` function would handle the case of wired in packages and global packages?

src/Stack/Dot.hs
```diff
@@ -255,37 +255,57 @@ createDepLoader :: HasEnvConfig env
...
+              pure (packageAllDeps pkg, payloadFromLocal pkg)
+            DepPackage{dpLocation=PLImmutable loc} -> do
+              let common = dpCommon dp
+              gpd <- liftIO $ cpGPD common
+              let PackageIdentifier name version = PD.package $ PD.packageDescription gpd
+                  flags = cpFlags common
+                  ghcOptions = cpGhcOptions common
+              assert (pkgName == name) (loadPackageDeps pkgName version loc flags ghcOptions)
+
+    -- If package is a wired-in-package or a global package, use info from ghc-pkg (#4324, #3084)
```
qrilka: I'm not sure that "or" here is required - wired-in packages are supposed to always come as global and we don't deal with them at the level of this new `globalDeps` function anymore, so I think it makes sense to leave only global ones in this comment

src/Stack/Dot.hs
```diff
@@ -255,37 +255,57 @@ createDepLoader :: HasEnvConfig env
...
-              deps = map (\depId -> maybe (error ("Invariant violated: Expected to find " ++ ghcPkgIdString depId ++ " in global DB"))
-                                          Stack.Prelude.pkgName
-                                          (Map.lookup depId globalIdMap))
-                         (dpDepends dp)
+createDepLoader sourceMap installed globalDumpMap globalIdMap loadPackageDeps pkgName = do
+  case projectPackageDeps <|> dependencyDeps <|> globalDeps of
+    Just x -> x
+    Nothing ->
+      if isWiredInPackage then
+        let errText = "Invariant violated: Expected to find wired-in-package "
```
qrilka: Any reason to have this `errText` as a separate variable?

qrilka: BTW I'm not sure that we need special handling for wired-in packages at all though I'm OK with leave it as it is

src/Stack/Dot.hs
qrilka: Oops, somehow I didn't get this comment into my inbox. Getting rid of `createDepLoader` looks like a sane thing to me. In principle the whole idea is just a do a dependency traversal and the idea of some special "loaders" looks like there is something else but I guess there isn't. But this kind of refactoring doesn't looks to be required for #4324 so probably could be done in a separate PR if you want to follow that route.

src/Stack/Dot.hs
vanceism7: I was just using `errText` to pretty up the code a bit, it looked a little ugly having all those string additions on one line. But if wired-in and global are basically the same thing, then I could just remove all of this. I was just worried about removing the whole invariant violated error. In the original code, wired-in-packages not existing in the `globalDumpMap` caused an error, but non-wired-in global packages were truncated. Should global packages missing from the dump map always cause error? Or should they just return an empty dependency set? This is the main thing im concerned with right now, knowing this will make it a little easier to remove `createDepLoader` as well. Thanks for all your help with this issue!

vanceism7: After looking at it again, it seems like getting rid of `createDepLoader` will be a bit harder than I thought. It doesn't necessarily need to exist, but It's helping to keep the code organized as it kind of unifies all the dependency loaders and makes the arguments they need available to them. So if it's decided that we should remove that function, I'll leave that for another PR.

src/Stack/Dot.hs
```diff
@@ -248,49 +248,57 @@ resolveDependencies limit graph loadPackageDeps = do
...
+          where
+            deps = map ghcIdToPackageName (dpDepends dump)
+            ghcIdToPackageName depId =
+              let errText = "Invariant violated: Expected to find "
+              in maybe (error (errText ++ ghcPkgIdString depId ++ " in global DB"))
+                 Stack.Prelude.pkgName
+                 (Map.lookup depId globalIdMap)
+
+    noDepsErr = error ("Invariant violated: Expected to find "
+                ++ packageNameString pkgName ++ " in global DB")
```
qrilka: This is at least confusing - you output this when package was not found in any of dependency sources but tell only about "global DB"

vanceism7: Ah darn, I didn't realize I had done that. So can we just say something like "Invariant violated: Package not found in any dependency sources"?

qrilka: I guess so, no better solutions come to my mind

vanceism7:
Hey np, I'm glad I could help out

dbaynard:
I think I've fixed the build failures. This is basically ready to merge. I rebased as #4547 has been merged, and I thought there could be some complications (there weren't).

I've tested with the #4324 example (though for me, I also get the expected outcome on 1.9.3 — very odd). It would be good to get some cases in `src/test/Stack/DotSpec.hs` that reflect the changes here. @vanceism7 Would you please add a test case into that module?

vanceism7:
Ok sure thing, I'll work on getting some test cases added in

dbaynard:
The spec tests are checked for every commit in ci, the integration tests aren't.

If you've done an integration test, push that. If you haven't, yet, comment with the details of the test you've been using and I'll take a look.

vanceism7:
I just ran the test scenario described in #4324. `stack dot --external | grep '"process"'` within the stack directory

snoyberg:
I just merged in `master` and added an integration test. Assuming no concerns with the updates, this should be ready to go now.

PR #4507, opened 2019-01-16
Fixes #4506 - Add fsnotify-0.3.0.1 to snapshot.yml

### LINKED ISSUE #4506: Old fsnotify dependency doesn't detect file changes when using Docker for Mac

`stack` fails to respond to file modification events on OSX when run inside docker. I believe this is simply due to being built against `fsnotify-0.2.1.1` rather than `fsnotify-0.3.0.1`. `fsnotify-0.3.0.1` included changes to the Linux file listener which listen for Attrib and Modify events instead of CloseWrite. The Docker for Mac osxfs file sharing emits Modify events on file changes, but does not emit CloseWrite.

[fsnotify-0.2.1.1 event list](https://github.com/haskell-fswatch/hfsnotify/blob/21df96438d60ff84c6076f3bbe8ac90342140488/src/System/FSNotify/Linux.hs#L68)

[fsnotify-0.3.0.1 event list](https://github.com/haskell-fswatch/hfsnotify/blob/729b41df0a8ef988d27154a0cda290a3a9f95270/src/System/FSNotify/Linux.hs#L83)

### Steps to reproduce

1. Run command `docker run -it --rm -v $PWD:/work -w /work flipstone/stack:v2-1.9.3 bash`
2. In the dockerized bash shell, run command `stack new fsevent-test`.
4. In the dockerized bash shell, run command `cd fsevent-test && stack build --file-watch`.
5. Edit `fsevent-test/src/Lib.hs`

(In vim, you will need to ensure `backupcopy` is set to `yes`.  Both the `no` and `auto` settings result in files being created/deleted in the directory and triggering extra filesystem events, masking the problem)

### Expected

The `stack build --file-watch` command should recompile when the file is edited.

### Actual

The `stack build --file-watch` command does not recompile when the file is edited.

### Stack version

```
$ stack --version
Version 1.9.3, Git revision 40cf7b37526b86d1676da82167ea8758a854953b (6211 commits) x86_64 hpack-0.31.1
```

### Method of installation

* Binary package downloaded from `https://github.com/commercialhaskell/stack/releases` as part of docker image build ([Dockerfile](https://github.com/flipstone/stack-lts/blob/master/v2/1.9.3/Dockerfile))

qxjit:
lts-11.22 includes fsnotify-0.2.1.1, which listens to Modify events on
Linux rather than CloseWrite events. For many applications the
difference is neglible because writing a file will result in both
events, but if you run stack inside of docker on Mac this breaks
the --file-watch option because the file sharing integration only
emits Modify events, never CloseWrite events.

fsnotify-0.3.0.1 changed to listen to Modify events rather than
CloseWrite events. Switching to the new version fixes the issue.

* [x] The documentation has been updated, if necessary. (not necessary)

I built stack against fsnotify-0.2.1.1 inside a docker container and verified that `--file-watch`
did *not* recompile when editing a file from MacOS on a newly creating stack project. Then I
rebuilt stack against fsnotify-0.3.0.1 in the same docker container and verified that `--file-watch`
*did* recompile when editing a file from MacOS.

Also ran `stack test` to ensure all the tests pass.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -108,6 +108,9 @@ Bug fixes:
   [#4402](https://github.com/commercialhaskell/stack/pull/4402)
 * Fix handling of GitHub and URL templates on Windows. See
   [commercialhaskell/stack#4394](https://github.com/commercialhaskell/stack/issues/4394)
+* Fix `--file-watch` not responding to file modifications when running
+  inside docker on Mac. See
+  [#4506](https://github.com/commercialhaskell/stack/issues/4506)
 
 ## v1.9.3
 
--- a/snapshot.yaml
+++ b/snapshot.yaml
@@ -17,6 +17,7 @@ packages:
 - cabal-doctest-1.0.6@rev:2
 - unliftio-0.2.8.0@sha256:5a47f12ffcee837215c67b05abf35dffb792096564a6f81652d75a54668224cd,2250
 - happy-1.19.9@sha256:f8c774230735a390c287b2980cfcd2703d24d8dde85a01ea721b7b4b4c82944f,4667
+- fsnotify-0.3.0.1@rev:1
 
 flags:
   cabal-install:


snoyberg:
Thanks!

qxjit:
My pleasure!

PR #4514, opened 2019-01-18
Disable AppVeyor and Travis

snoyberg:
We're moving over to Azure Pipelines, which have been more reliable.


borsboom:
Why not just delete the `travis-ci.yml`?  Because it serves as an example?

snoyberg:
I had assumed that would result in an error message from Travis, as it
would attempt to build with the default Travis settings. I'll try pushing a
committee and seeing if it breaks or not.

psibi:
@snoyberg I use the file `./travis-setup.sh` in Azure CI too. Can we not remove that ? I think that script should probably be renamed to `ci-setup.sh`.

snoyberg:
I didn't notice that, thanks @psibi. I renamed the file to `etc/scripts/ci-setup.sh`. Let's see if that works. I've also rebased the commits.

snoyberg:
Looks like this passed. @borsboom I'll leave merging this to you when you're comfortable with it.

PR #4515, opened 2019-01-18
Curator graph check

qrilka:

With a locally modified `build-constraints.yaml` (I included all grandfathered dependencies except `data-checked` as an example)  I get the following at the moment:
```
Checking dependencies in snapshot.yaml
Some warnings discovered when adding packages to snapshot (snapshot.yaml)
Setting flags for non-existent package: brick
Setting flags for non-existent package: exinst
Setting flags for non-existent package: fay
Setting flags for non-existent package: hpio
Setting flags for non-existent package: safe-money
Hiding non-existent package: fay-base
Hiding non-existent package: gtk3
Hiding non-existent package: log
Hiding non-existent package: objective
Errors in snapshot:

data-checked (not present) depended on by:
- [ ] text-latin1-0.3.1 (>=0.2). Grandfathered dependencies. @mvv. Used by: library


hw-mquery-0.2.0.0 (Grandfathered dependencies) is out of bounds for:
- [ ] hw-json-0.9.0.1 (>=0.1 && < 0.2). John Ky newhoggy@gmail.com @newhoggy. @haskell-works. Used by: library


leancheck-0.9.0 (Rudy Matela <rudy@matela.com.br> @rudymatela) is out of bounds for:
- [ ] dejafu-1.11.0.4 (>=0.6 && < 0.9). Michael Walker <mike@barrucadu.co.uk> @barrucadu. @barrucadu. Used by: library


tomland-1.0.0 (Kowainik <xrom.xkov@gmail.com> @ChShersh) is out of bounds for:
- [ ] summoner-1.2.0 (^>=0.5.0). Kowainik <xrom.xkov@gmail.com> @ChShersh. @kowainik. Used by: library

```


src/Network/HTTP/StackClient.hs
```diff
@@ -83,3 +96,52 @@ withResponse
...
+--     matches expected length
+-- * Limits the download to (close to) the expected # of bytes
+-- * Verifies that the expected # bytes were downloaded (not too few)
+-- * Verifies md5 if response includes content-md5 header
+-- * Verifies the expected hashes
+--
+-- Throws VerifiedDownloadException.
+-- Throws IOExceptions related to file system operations.
+-- Throws HttpException.
+verifiedDownload
```
snoyberg: I realize this isn't something you changed, but already existed. But question: is there any reason we wouldn't _always_ want to perform a verified download instead of an unverified download? In other words, can we merge these two functions together?

src/Network/HTTP/StackClient.hs
```diff
@@ -83,3 +96,52 @@ withResponse
...
+-- no download is performed. Otherwise, creates the parent directory, downloads
+-- to a temporary file, and on file download completion moves to the
+-- appropriate destination.
+--
+-- Throws an exception if things go wrong
+download :: HasTerm env
+         => Request
+         -> Path Abs File -- ^ destination
+         -> RIO env Bool -- ^ Was a downloaded performed (True) or did the file already exist (False)?
+download req dest = Download.download (setUserAgent req) dest
```
snoyberg: I'm not at all convinced this is a good idea, but: since the only purpose of this wrapper seems to be to set the user agent, perhaps we should have a `HasHttp`-style typeclass in `http-download` which includes a user agent setting instead. Then we know we won't ever accidentally forget to set it, and we can drop this wrapper module entirely.

src/Stack/SourceMap.hs
```diff
@@ -172,7 +133,7 @@ globalsFromDump compiler = do
 
 globalsFromHints ::
        HasConfig env
-    => ActualCompiler
```
snoyberg: The original type signature looks like a straight up bug, good change. Any idea why it looked like that before? (I'm well aware of the fact that I'm almost certainly the one who wrote it 😛.)

subs/curator/src/Curator/Snapshot.hs
```diff
@@ -92,3 +112,287 @@ traverseValidate f t = do
...
+        ( if null dpMaintainers
+          then "No maintainer"
+          else T.intercalate ", " (Set.toList dpMaintainers)
+        ) :
+        if null dpGithubPings
+        then []
+        else [T.concat (map (T.cons '@') (Set.toList dpGithubPings))]
+
+    compToText :: Component -> Text
+    compToText CompLibrary = "library"
```
snoyberg: We can defer this question to after this PR, but: how do we handle dependencies of custom-setup stanzas, foreign libraries, and sublibraries? Do they all appear as library dependencies?

qrilka: I had a thought about those but somehow forgot about it. This code doesn't handle them and I think it makes sense to fix it as a part of this PR

src/Stack/SourceMap.hs
qrilka: I guess it's about this function:
```
parseActualCompiler :: T.Text -> Either PantryException ActualCompiler
parseActualCompiler = fmap wantedToActual . parseWantedCompiler
```
So `WantedCompiler` and `ActualCompiler` get parsed the same way and I'm not sure that it makes sense to have a difference here at least while we have some definitions which use parsing `ActualCompiler` from JSON (e.g. `SetupInfo`)

src/Network/HTTP/StackClient.hs
qrilka: It looks like `download` is only used `loadGlobalHints`, in `Stack.New` we use `redownload` but that one looks to be not doing checks done in `verifiedDownload`. Maybe it's better to deal with this in a separate issue?

src/Network/HTTP/StackClient.hs
qrilka: I like the idea but at the same time it seems to add noticeable amount of code and it will resolve not very critical issue so I'd skip it in this PR

src/Network/HTTP/StackClient.hs
qrilka: Opened #4524 for that

snoyberg:
Probably worth merging in master here to skip the AppVeyor and Travis builds, which will almost certainly fail for unimportant reasons.

Issue #4256, opened 2018-08-23
Make snapshot: a synonym for resolver:


snoyberg:
I discussed this already on Twitter. Basic idea: resolver turns out to have not been a great name. That field is used for specifying a snapshot. Let's just call a spade a spade.

Implementation: in both snapshots and stack.yaml files, instead of looking for just `resolver`, we'll look for both `resolver` and `snapshot`. If both are present: it's an error. If one is present, use it. If neither are present, follow through with current behavior for when there's no resolver set.

DanBurton:
Also: update `stack new` templates to prefer `snapshot`.

Does stack have any philosophy regarding deprecation? Should support for `resolver` eventually be removed?

snoyberg:
We don't have an official policy. In the past, we've:

* Added a warning on the old syntax
* Left it in for a few versions
* Removed support for the old syntax

aleksejkozin:
Hi, guys!
I'm a new at Haskell, but already appreciate Stack a lot 
I want to help you with this task
Is it free to take?
Is it a good place to start?

snoyberg:
Yes to both questions! The best place to start would be to grep the `src` directory for the string `"resolver"`. The most important affected module is `Stack.Types.Config`.

snoyberg:
No worries, thanks for the update.

aleksejkozin:
@snoyberg can you tell me if I'm thinking in the right direction?

I want to add `snapshot` support into stack.yaml
To achieve it I want to amend a function `parseProjectAndConfigMonoid` (because it contains `"resolver"`):
https://github.com/commercialhaskell/stack/blob/cf9b94d1899ec1fc487e4dccda3fee5b75254a1c/src/Stack/Types/Config.hs#L1452

My plan is:
 - add several unit tests to [ConfigSpec.hs](/commercialhaskell/stack/blob/master/src/test/Stack/ConfigSpec.hs)
`describe "parseProjectAndConfigMonoid"`
`it "parses snapshot using 'resolver'"`
`it "parses snapshot using 'snapshot'"`
`it "throws if both 'resolver' and 'snapshot' are present"`
- write actual code so the tests will pass

For now I'm thinking of how to write the tests:
I want to create a temp predefined stack.yaml and load it with `parseProjectAndConfigMonoid`
It will give me `Project` object from which I can get `projectResolver :: SnapshotLocation`
After that I will check if `projectResolver` is equal to a correct value

If that is a correct line of thought then I have a question:
[...]

PR #4521, opened 2019-01-20
Show resolver being used when `stack ghci` is invoked outside of a project directory

### LINKED ISSUE #3651: Show resolver being used when `stack ghci` is invoked outside of a project directory
When I run `stack ghci` outside of a project directory I get output like:

```
bash$ stack ghci
everything-0.0.1: initial-build-steps (lib)
Configuring GHCi with the following packages: everything
GHCi, version 8.0.2: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /Users/erantapaa/.ghc/ghci.conf
Loaded GHCi configuration from /private/var/folders/.../ghci-script
Prelude>
```

It would be very helpful to display the resolver being used in this scenario.

wafelj:
Fix for https://github.com/commercialhaskell/stack/issues/3651.

Approach: read `Project` from `Config`, get `SnapshotLocation` and compiler info from it and use that to get resolver using the `loadResolver` function. To my understanding it is theoretically possible that there is no project information in config (it's wrapped in `Maybe`). But I'm assuming this never actually happens because if I try to delete the `~/.stack/global-project/stack.yaml`, a new one is recreated when `stack ghci` is run again. This is why I assumed it's safe to use `fromJust`. If there is a better approach, I'd love to hear about it.

Disclaimer: I'm new to Haskell and to stack, and this is my first open-source contribution. Any feedback is most appreciated!


src/Stack/Ghci.hs
```diff
@@ -186,8 +189,12 @@ ghci opts@GhciOpts{..} = do
               figureOutMainFile bopts mainIsTargets localTargets pkgs0
     -- Build required dependencies and setup local packages.
     stackYaml <- view stackYamlL
+
+    mproject <- view $ configL.to configMaybeProject
+
     buildDepsAndInitialSteps opts (map (T.pack . packageNameString . fst) localTargets)
-    targetWarnings stackYaml localTargets nonLocalTargets mfileTargets
+
```
mgsloan: Please don't insert newlines here unless it's logically splitting things into sections (also above)

src/Stack/Ghci.hs
```diff
@@ -836,13 +843,14 @@ checkForDuplicateModules pkgs = do
...
-  :: HasRunner env
+  :: HasEnvConfig env
   => Path Abs File
   -> [(PackageName, (Path Abs File, Target))]
   -> [PackageName]
   -> Maybe (Map PackageName [Path Abs File], [Path Abs File])
+  -> Maybe (Project, Path Abs File)
   -> RIO env ()
-targetWarnings stackYaml localTargets nonLocalTargets mfileTargets = do
+targetWarnings stackYaml localTargets nonLocalTargets mfileTargets mproject = do
```
mgsloan: Since the context has been changed from `HasRunner` to `HasEnvConfig`, there is no need to pass in `mproject`, instead `view $ configL.to configMaybeProject` can be done in this function.  Similarly, no need to pass in `stackYaml`, that can also be queried from the reader environment.

mgsloan:
The code related to resolvers and snapshots has changed quite a bit since I last looked, so I can't give direct guidance.  However, it seems like `loadResolver` does fetching and parsing that I believe should have already happened.  So, I think it'd be nice to avoid the redundant call, or cache it.  I'll leave that decision up to the more active maintainers, though.

src/Stack/Ghci.hs
wafelj: Removed.

src/Stack/Ghci.hs
wafelj: Makes sense, done.

src/Stack/Ghci.hs
```diff
@@ -853,10 +861,16 @@ targetWarnings stackYaml localTargets nonLocalTargets mfileTargets = do
       , "."
       , flow "It can still be useful to specify these, as they will be passed to ghci via -package flags."
       ]
-  when (null localTargets && isNothing mfileTargets) $
+
+  when (null localTargets && isNothing mfileTargets) $ do
+      let project = fst $ fromJust mproject
```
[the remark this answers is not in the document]
wafelj: I agree, ignoring the `Nothing` case is pretty ugly. I was thinking about this some more and decided to try something else: simply not print the resolver version if the project is `Nothing`. Would this work? I can do the `error` instead, but it's basically the same thing I had before, except it has a more useful error message.

wafelj:
@mgsloan thanks for your review! I was digging through the `HasEnvConfig` environment, trying to see if and where `SnapshotDefinition` is available, but I could not find it. `SnapshotLocation` inside `Project` inside `Config` was the closest, so I went with that. Definitely interested in other suggestions - my understanding of the project is not very deep yet.

mgsloan:
@wafelj Welcome!  Yeah I also dug around `HasEnvConfig` looking for snapshot name info.  Couldn't seem to find it, which seems surprising since `EnvConfig` has `BuildConfig`, and info about the snapshot needs to be loaded to have a `BuildConfig`.  So, yeah, not sure why that's not there ¯\_(ツ)_/¯

Looks to me like `bcSnapshot :: LoadedSnapshot` could simply be added to `BuildConfig` and then populated in `loadbuildConfig`.

qrilka:
@mgsloan @wafelj there was a large refactoring in #4412 and now Stack doesn' use `LoadedSnapshot` anymore. I think snapshot name could be added into `SMWanted` which later gets transformed into `SourceMap` - snapshot gets loaded in `Stack.Config.loadBuildConfig` but its name gets ignored (only compiler/dependencies get used).

wafelj:
@mgsloan @qrilka thanks. I looked into that and changed the code - now it passes the name from `SMWanted` to `SMActual` to `SourceMap`. Then I load `SourceMap` from `EnvConfig` and read the name from that. Is this what you had in mind?

Also, two questions:
1) `SMWanted` is available in in `BuildConfig`. Would it make sense to read it directly from that in `targetWarnings` function instead of passing it through all those other types?
2) naming - is `smwName` etc. descriptive enough or would `smwResolverName` or `smwSnapshotName` etc. be better?

qrilka:
1. The name should be the same in either of `SMXXX` or `SourceMap` so using `BuildConfig` should be OK
2. I think we should use longer field name as it's not a name of a source map itself. Also it makes sense to follow https://github.com/commercialhaskell/stack/issues/4256 and use "snapshot".
@snoyberg should we also switch to use "snapshot" instead of "resolver" in Stack output?

snoyberg:
> should we also switch to use "snapshot" instead of "resolver" in Stack output?

Sounds like a good idea to me.

wafelj:
@qrilka changed to read name directly from BuildConfig and changed the field to smwSnapshotName. Also changed output to snapshot instead of resolver. Looks good?

PR #4525, opened 2019-01-21
Better GHC version display in Azure CI for Stack builds

psibi:

### DIFF
--- a/.azure/azure-linux-template.yml
+++ b/.azure/azure-linux-template.yml
@@ -45,8 +45,11 @@ jobs:
           export PATH=$HOME/.local/bin:/opt/ghc/$GHCVER/bin:/opt/happy/1.19.5/bin:/opt/alex/3.1.7/bin:$PATH
           ;;
       esac
-      if ! [ "$BUILD" = style ]; then echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"; fi
-         set -ex
+      if [ "$BUILD" = stack ];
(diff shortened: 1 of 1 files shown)


psibi:
Looking at the build CI, I don't think this solves the problem. The `stack setup` command is run after the above check is done.

.azure/azure-linux-template.yml
```diff
@@ -45,8 +45,11 @@ jobs:
           export PATH=$HOME/.local/bin:/opt/ghc/$GHCVER/bin:/opt/happy/1.19.5/bin:/opt/alex/3.1.7/bin:$PATH
           ;;
       esac
-      if ! [ "$BUILD" = style ]; then echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"; fi
-         set -ex
+      if [ "$BUILD" = stack ];
+      then echo "$(stack ghc -- --version)";
+      else echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]";
+      fi
+      set -ex
```
mihaimaruseac: There's a semantic change here. Previously the second `echo` command would be run if the first one failed and only in that case.

Now, the second `echo` command is run whenever `$BUILD$` is not `stack`

psibi:
Closing this PR. I have removed the version display in the config files altogether as it just adds confusion. Specific version of GHC is determine based on the `STACK_YAML` environment variable you pass for the file.

PR #4528, opened 2019-01-22
No FK checks during migrations and better error if migration fails

### LINKED ISSUE #4519: almost all stack actions result in SQLite error ...
`stack build`, `stack ghci`, `stack upgrade ...` all fail ... it's a reasonably unfortunate situation.

i'm running the latest from master:

```
07:02 PM noon ∈ ~ ♫ stack --version
Version 1.10.0, Git revision ac20e0da239d2e3f6a7cb4bb67ea33b2a7c95849 x86_64 hpack-0.31.1
```

most unfortunately, it also affects `stack upgrade`:

```
07:02 PM noon ∈ ~ ♫ stack upgrade  
SQLite3 returned ErrorConstraint while attempting to perform step: FOREIGN KEY constraint failed
```

and when running again:

```
07:02 PM noon ∈ ~ ♫ stack upgrade
SQLite3 returned ErrorBusy while attempting to perform step: database is locked
```

( as an aside, i feel like the master branch is much more unstable in the last 6 months or so; i used to upgrade all the time; now, whenever i do i hit some weird bug. maybe the "stack upgrade --git" option could hit a more stable branch, if master isn't stable anymore? )

qrilka:

Tested it manually with a DB created with Stack sources before #4439 was merged

Fixes #4519 

@snoyberg is it OK to reference `$STACK_ROOT` in Pantry error messages? It doesn't look clean to me but I'm not sure how to deal with it better in this case

### DIFF
--- a/subs/pantry/src/Pantry/Storage.hs
+++ b/subs/pantry/src/Pantry/Storage.hs
@@ -223,7 +223,7 @@ initStorage fp inner = do
     (createSqlitePoolFromInfo (sqinfo True) 1)
     (liftIO . destroyAllResources) $ \pool -> inner (P.Storage pool)
   where
-    wrapMigrationFailure = handle (throwIO . MigrationFailure)
+    wrapMigrationFailure = handle (throwIO . MigrationFailure fp)
     sqinfo fk = set extraPragmas ["PRAGMA busy_timeout=2000;"]
            $ set fkEnabled fk
            $ mkSqliteConnectionInfo (fromString $ toFilePath fp)
--- a/subs/pantry/src/Pantry/Storage.hs
+++ b/subs/pantry/src/Pantry/Storage.hs
@@ -84,7 +84,7 @@ import Data.Pool (destroyAllResources)
 import Pantry.HPack (hpackVersion, hpack)
 import Conduit
 import Data.Acquire (with)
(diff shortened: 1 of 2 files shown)


subs/pantry/src/Pantry/Storage.hs
```diff
@@ -215,15 +215,17 @@ initStorage
...
-    inner (P.Storage pool)
+    migrates <- wrapMigrationFailure $ runSqlPool (runMigrationSilent migrateAll) pool
+    forM_ migrates $ \mig -> logDebug $ "Migration executed: " <> display mig
+  bracket
+    (createSqlitePoolFromInfo (sqinfo True) 1)
+    (liftIO . destroyAllResources) $ \pool -> inner (P.Storage pool)
   where
-    sqinfo = set extraPragmas ["PRAGMA busy_timeout=2000;"]
-           $ set fkEnabled True
+    wrapMigrationFailure = handle (throwIO . MigrationFailure)
```
snoyberg: How about including the `FilePath` above in the `MigrationFailure` data constructor, and replacing the `$STACK_ROOT` bit with the actual file path they would need to delete?

qrilka: Fixed

Issue #3377, opened 2017-08-22
Difference in behavior between `stack script` and `stack script --compile`


mgsloan:
When looking into https://github.com/commercialhaskell/stack/issues/3372 , I noticed a difference in behavior between `stack script` and `stack script --compile`:

* `stack script` uses `runghc`, run in the current directory.
* `stack script --compile` uses `ghc` run in the directory of the source file, and then runs the result in the current directory.

It certainly makes sense to run the code in the current directory, this allows these scripts to be used as commandline tools.  However, this means that `ghc` itself is run in different places depending on the option.  This at least affects:

* Importing modules
* Use of reading relative paths from TH

Importing modules could be supported by using `-i`, but this does not solve the TH problem.  I believe a solution there would require using something other than `runghc`.

PR #4538, opened 2019-01-25
Have "stack script" set import search path #3377

mgsloan:

This issue was brought to my attention by https://github.com/rcook .  The primary problem is that stack scripts can only import modules if the script is run from the same directory as the modules.  However, I realized that this implies a gnarlier issue, which is that since `-i` isn't being passed to `runghc`, it will use the work dir as a search dir for imports.

This also addresses part of the problem described in #3377 , which was a difference in how imports are treated when `--compile` is used.

Here's a concrete example.  The contents of `~/tmp/hmm.hs` is

```haskell
#!/usr/bin/env stack
-- stack --resolver lts-13.0 script

import Yargh

main :: IO ()
main = yargh
```

The contents of `~/tmp/Yargh.hs` is

```haskell
module Yargh where

yargh :: IO ()
yargh = putStrLn "works!"
```

The contents of `~/Yargh.hs` is

```haskell
module Yargh where

yargh :: IO ()
yargh = putStrLn "YARGH, broken!
```

Check out what happens when I run the script from different directories:

```
mgsloan@treetop:~/tmp$ ./hmm.hs                                                                               
works!                                                                                                        
mgsloan@treetop:~/tmp$ cd ../                                                                                 
mgsloan@treetop:~$ ./tmp/hmm.hs                                                                               
YARGH, broken!      
```

This PR resolves the issue.  After applying this patch, it behaves like this:

```
mgsloan@treetop:~/tmp$ ./hmm.hs                                                                               
works!                                                                                                        
mgsloan@treetop:~/tmp$ cd ../                                                                         
mgsloan@treetop:~$ ./tmp/hmm.hs                                                                               
works!                      
```

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -46,6 +46,12 @@ Major changes:
 Behavior changes:
 * `stack.yaml` now supports `snapshot`: a synonym for `resolver`. See [#4256](https://github.com/commercialhaskell/stack/issues/4256)
 
+* `stack script` now passes `-i -idir` in to the `ghc`
+  invocation. This makes it so that the script can import local
+  modules, and fixes an issue where `.hs` files in the current
+  directory could affect interpretation of the script. See
+  [#4538](https://github.com/commercialhaskell/stack/pull/4538)
+
 Other enhancements:
 
 * Defer loading up of files for local packages. This allows us to get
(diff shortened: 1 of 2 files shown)


rcook:
Thanks!

PR #4545, opened 2019-01-26
Binary module interface

### LINKED ISSUE #4541: Compilation not terminating due to giant .dump-hi (4gb) file when using type families
Hello guys!

Related fixed issue: https://github.com/commercialhaskell/stack/issues/4027
Related GHC issue: https://ghc.haskell.org/trac/ghc/ticket/8095#comment:58

My team is experiencing issues when compiling a project that extensively use type families (type level sql with https://github.com/morphismtech/squeal).
We found that the big trouble comes from GHC trying to dump a text file when compiling each unit.
Here are the investigations:
- Stack is asking GHC to dump the `.hi` file as pretty printed text (see https://stackoverflow.com/a/46399538).
- GHC seems to unfold the types regardless of their size (seems like GHC unfolding thresholds options does not apply to types, see GHC ticket).

Our context:
- We moved our plain SQL queries+migrations to type level using Squeal (involving types families a lot)
- GHC unfolding types resulting in very very long types in the `.hi` file
- Stack asking GHC to dump the interface, resulting in writing a 4gb .txt (head of the file https://gist.githubusercontent.com/haitlahcen/3f1e83cdc78d48ed2100d51a0ea05844/raw/c49218be742cd3d2450a8f13683a04339b1ab9e7/hi.txt) for each module that import the type level queries.

My questions:
- Could I PR stack to directly read the binary format ? (this would improve compilation performance a lot, relying only on the GHC `.hi` file). I've already an implementation that can read the dependencies/templatehaskell in the .hi file for ghc7-863. The drawback would be that we would have to support the protocol for each new release of GHC.
- Could we have an option to avoid the dump ?

Current solution:
- Use the `-fomit-interface-pragmas` that avoid ALL unfolding. This is not very good because we lose inlining.

Thanks for reading.

haitlahcen:
This PR try to fix #4541 by implementing the binary format of the module interface files.
This is mostly reverse engineering of https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/IfaceFiles. 
I did not found any strong documentation that would give the complete binary spec.

I have tested the implementation with https://gist.github.com/haitlahcen/de33a593added86f86fc9a126d5b4164 and `test.sh`:
```bash
#! /usr/bin/env nix-shell
#! nix-shell --pure -i bash -p haskell.compiler.{GHC_VERSION_TESTED}
ghc -fforce-recomp Test.hs && ghc --show-iface Test.hi
```
Using the extracted implementation https://gist.github.com/haitlahcen/6232e957608d742a16581789e31dc918

GHC versions passing `./test.sh && nix-shell --pure --run "cabal run ./Test.hi"`
- ghc7103
- ghc802
- ghc822
- ghc843
- ghc844
- ghc861
- ghc862
- ghc863

I would also like to know if you guys consider this major change so that I update the changelog acordingly!


dbaynard:
@mgsloan would you recommend a reviewer, please?

@haitlahcen Thanks for the PR!

I do have one major concern. The interface changes for every major GHC version, and so this will need to be revisited for every major version; it surely means old stack versions will not work with new GHC versions? I may be wrong, however — I don't know how stack handles this at the moment. Will a change to the dumped output format mean this is an issue already?

> I did not found any strong documentation that would give the complete binary spec.

The page you linked refers to [the GHC wiki](https://wiki.haskell.org/GHC/As_a_library). [GHC is available as a library on hackage](https://hackage.haskell.org/package/ghc). I don't know whether current GHC can read interface files from old GHC versions. I suspect there's a package which handles this (like for template-haskell) — perhaps one of [these packages](http://packdeps.haskellers.com/reverse/ghc) does something.

---

In any case, I'm glad it works, for now. Are you able to use your fork, then?

borsboom:
What is the behaviour if you use it with an unsupported (new) GHC version?  We do need some kind of reasonable fallback to support new versions, especially prerelease GHC versions, without having to release a new `stack` version.  If I'm reading the code right, I think it'll just not track TH dependent files in this case, which I think is fine.

haitlahcen:
@dbaynard 

> I do have one major concern. The interface changes for every major GHC version, and so this will need to be revisited for every major version; it surely means old stack versions will not work with new GHC versions?

The older version will still work but the new interface will not be deserialized (which is only used to check for dirtiness btw).

> Will a change to the dumped output format mean this is an issue already?

Yes, [this already happened when they added a hash after the path of the TH dependencies](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Package.hs#L1124).

> I don't know whether current GHC can read interface files from old GHC versions.

GHC is not supporting older versions of the protocol. I think the main reason is that interface files are generated at compile time and were meant to be used temporarily (destroyed right after compilation).

> The page you linked refers to the GHC wiki. GHC is available as a library on hackage. I don't know whether current GHC can read interface files from old GHC versions.

I have not found any package implementing the protocols, but I might have missed something !
The GHC package is not usable because:
- It only deserialize the latest version
- Loading a file through the package requires an effectful "compilation context" (here we just want to read the dependencies with no further processing).

> In any case, I'm glad it works, for now. Are you able to use your fork, then?

Indeed, I've successfully compiled Stack using the patched version !

@borsboom 

>What is the behaviour if you use it with an unsupported (new) GHC version? We do need some kind of reasonable fallback to support new versions, especially prerelease GHC versions, without having to release a new stack version. If I'm reading the code right, I think it'll just not track TH dependent files in this case, which I think is fine.

You are right, the current behavior just ignore the deserialization failure, hence not following deps dirtiness.

dbaynard:
@haitlahcen Thanks for the detailed responses — they've certainly addressed my concerns. We should be able to review it, soon.

> Indeed, I've successfully compiled Stack using the patched version !

I take it then that you are able to use your patched version of stack in production. Thanks for contributing back!

dbaynard:
See https://neilmitchell.blogspot.com/2019/02/announcing-ghc-lib.html — it may be worth considering from a maintenance perspective. Good timing! I wonder whether @ndmitchell saw this issue…

ndmitchell:
@dbaynard I did not. So `ghc-lib` would certainly allow you to load one particular `.hi` file format from one particular GHC version across many GHC versions. But you need the slightly harder problem of leading N different versions across many GHC versions - which you might be able to depend on multiple ghc-lib versions simultaneously. However, it's likely a lot more complex than the code above, and still wouldn't work with pre-releases.

From what I saw, the use of the information obtained by parsing the .hi files is purely for warnings? If that's the case, shouldn't the failure to parse the .hi files be a warning, and then just ignore the other warnings, and it would work for pre-releases just fine.

Do not that [Weeder](https://github.com/ndmitchell/weeder) also piggy backs off the .hi files that Stack creates, so if Stack could still create them with some flag that would be helpful for Weeder. Alternatively, Weeder could move to dumping the .hi-dump files itself on demand. As a consequence, this change is a major change.

src/Stack/Build/Execute.hs
```diff
@@ -1974,7 +1974,7 @@ extraBuildOptions :: (HasEnvConfig env, HasRunner env)
                   => WhichCompiler -> BuildOpts -> RIO env [String]
 extraBuildOptions wc bopts = do
     colorOpt <- appropriateGhcColorFlag
-    let ddumpOpts = " -ddump-hi -ddump-to-file"
+    let ddumpOpts = " -keep-hi-files"
```
snoyberg: This is the default behavior of GHC according to the manual. I don't mind being explicit, but could you add a comment explaining the reason for reinforcing the default?

src/Stack/Package.hs
```diff
@@ -1106,54 +1108,33 @@ getDependencies component dotCabalPath =
...
-                  -- the path. See
-                  -- https://github.com/yesodweb/yesod/issues/1551
-                  TLE.decodeUtf8 .
-                  CL8.takeWhile (/= '\"') <$>
-            mapMaybe (CL8.stripPrefix "addDependentFile \"") dumpHI
-    thDepsResolved <- liftM catMaybes $ forM thDeps $ \x -> do
-        mresolved <- liftIO (forgivingAbsence (resolveFile dir x)) >>= rejectMissingFile
-        when (isNothing mresolved) $
+  dir <- asks (parent . ctxFile)
+  content <- liftIO $ CL8.readFile dumpHIPath
```
snoyberg: I'd rather not using lazy I/O here at all. _If_ we are going to use lazy I/O, I'd much rather use the `withBinaryFile` approach which ensures that the handle is closed at the end. But doesn't the `binary` package provide a streaming parse interface?

I realize this isn't the way the original code was written...

src/Stack/ModuleInterface.hs
```diff
@@ -0,0 +1,425 @@
...
+    l <-
+        if i == 0xff
+            then getWord32be
+            else pure (fromIntegral i :: Word32)
+    List <$> replicateM (fromIntegral l) f
+
+getTuple :: Get a -> Get b -> Get (a, b)
+getTuple f g = (,) <$> f <*> g
+
+getFastString :: Get FastString
```
snoyberg: Any reason to not use a `ByteString` here?

src/Stack/ModuleInterface.hs
```diff
@@ -0,0 +1,425 @@
+module Stack.ModuleInterface
```
snoyberg: General question: is this code copied directly from somewhere else?

src/Stack/Build/Execute.hs
haitlahcen: I did not know that! Will change it!

src/Stack/Package.hs
haitlahcen: By `streaming interface` are you talking about [the pipes-binary package?](http://hackage.haskell.org/package/pipes-binary-0.4.2) If so, do you want me to use it here?

src/Stack/ModuleInterface.hs
haitlahcen: No particular reason at all, I will change it.

src/Stack/ModuleInterface.hs
haitlahcen: It's very bad for me to hear that... I've written all the code, which include reversing the different protocol versions.

haitlahcen:
@snoyberg Thanks for the review, I'll take some time to add the according tests.

src/Stack/ModuleInterface.hs
snoyberg: I didn't mean to say anything bad, sorry! I was asking because of things like the `FastString` newtype.

src/Stack/Package.hs
snoyberg: I meant a function like [`decodeFileOrFail`](https://www.stackage.org/haddock/lts-13.7/binary-0.8.6.0/Data-Binary.html#v:decodeFileOrFail).

src/Stack/ModuleInterface.hs
haitlahcen: Ohh, it's okay, I thought that you were looking for plagiarism.. The names are indeed inspired by the GHC code (mnemonic for me while I was coding)

src/Stack/Package.hs
haitlahcen: Ok let's use that

snoyberg:
Hey @haitlahcen, just bumping. Are you still planning on adding the tests here?

haitlahcen:
@snoyberg indeed, just busy right now but will definitely do it

snoyberg:
Cool, thanks for the quick response 👍

snoyberg:
I'm going to recommend we move ahead with merging this as-is, even without the additional tests, and rely on integration tests to catch bugs. A future PR to improve the unit tests will definitely be appreciated.

snoyberg:
Looks like there are some integration test failures on Windows:

https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=1205&view=logs

```
2019-04-09T10:59:36.9699824Z Running: D:\a\1\s\.stack-work\install\0c58dc80\bin\stack.exe run
2019-04-09T10:59:36.9700512Z [1 of 2] Compiling Main             ( D:\tmp\home3608\.stack\setup-exe-src\setup-Z6RU0evB.hs, D:\tmp\home3608\.stack\setup-exe-src\setup-Z6RU0evB.o )
2019-04-09T10:59:36.9701115Z [2 of 2] Compiling StackSetupShim   ( D:\tmp\home3608\.stack\setup-exe-src\setup-shim-Z6RU0evB.hs, D:\tmp\home3608\.stack\setup-exe-src\setup-shim-Z6RU0evB.o )
2019-04-09T10:59:36.9701788Z Linking D:\tmp\home3608\.stack\setup-exe-cache\i386-windows\tmp-Cabal-simple_Z6RU0evB_2.0.1.0_ghc-8.2.2.exe ...
2019-04-09T10:59:36.9753664Z [0mfile-embed-0.0.10.1: configure[0m
2019-04-09T10:59:36.9754725Z [0mfile-embed-0.0.10.1: build[0m
2019-04-09T10:59:36.9755178Z [0mfile-embed-0.0.10.1: copy/register[0m
2019-04-09T10:59:36.9755671Z [0mBuilding all executables for `watched-files' once. After a successful build of all of them, only specified executables will be rebuilt.[0m
2019-04-09T10:59:36.9756144Z [0mwatched-files-0.0.0: configure (exe)[0m
2019-04-09T10:59:36.9756557Z [0mConfiguring watched-files-0.0.0...[0m
2019-04-09T10:59:36.9757349Z [0mwatched-files-0.0.0: build (exe)[0m
2019-04-09T10:59:36.9757797Z [0mPreprocessing executable 'watched' for watched-files-0.0.0..[0m
2019-04-09T10:59:36.9758223Z [0mBuilding executable 'watched' for watched-files-0.0.0..[0m
2019-04-09T10:59:36.9758692Z [0m[1 of 2] Compiling Main             ( Main.hs, .stack-work\dist\010ee936\build\watched\watched-tmp\Main.o )[0m
2019-04-09T10:59:36.9759223Z [0m[2 of 2] Compiling Paths_watched_files ( .stack-work\dist\010ee936\build\watched\autogen\Paths_watched_files.hs, .stack-work\dist\010ee936\build\watched\watched-tmp\Paths_watched_files.o )[0m
2019-04-09T10:59:36.9759730Z [0mLinking .stack-work\dist\010ee936\build\watched\watched.exe ...[0m
2019-04-09T10:59:36.9760277Z [0m
2019-04-09T10:59:36.9760652Z [33mWarning:[0m Failed to decode module interface:
2019-04-09T10:59:36.9761091Z          [36mD:\tmp\stack-integration-watched-files3608\.stack-work\dist\010ee936\build\watched\watched-tmp\Main.hi[0m
2019-04-09T10:59:36.9761548Z          Decoding failure: [91mInvalid magic: 1face[0m[0m
2019-04-09T10:59:36.9762061Z [0mwatched-files-0.0.0: copy/register[0m
2019-04-09T10:59:36.9762556Z [0mInstalling executable watched in D:\tmp\stack-integration-watched-files3608\.stack-work\install\f20c5383\bin[0m
2019-04-09T10:59:36.9763022Z [0mCompleted 2 action(s).[0m
2019-04-09T10:59:36.9763460Z Running: D:\a\1\s\.stack-work\install\0c58dc80\bin\stack.exe run
2019-04-09T10:59:36.9763856Z [0m
2019-04-09T10:59:36.9764838Z [33mWarning:[0m Failed to decode module interface:
2019-04-09T10:59:36.9765327Z          [36mD:\tmp\stack-integration-watched-files3608\.stack-work\dist\010ee936\build\watched\watched-tmp\Main.hi[0m
2019-04-09T10:59:36.9765770Z          Decoding failure: [91mInvalid magic: 1face[0m[0m
2019-04-09T10:59:36.9766222Z foofooMain.hs: Expected: "bar"
2019-04-09T10:59:36.9766665Z Actual:  "foo"
2019-04-09T10:59:36.9767065Z CallStack (from HasCallStack):
2019-04-09T10:59:36.9767536Z   error, called at D:\a\1\s\test\integration\tests\watched-files\Main.hs:9:23 in main:Main
```

haitlahcen:
@snoyberg I apologize for the time it take, let me one more week to introduce the changes you asked as well as implementing basic test cases

snoyberg:
No problem. I've pushed the commits I was working on which implement some of what we've discussed.

haitlahcen:
@snoyberg I've updated the PR and added some dummy `.hi` files for the latest versions. However I'm not sure if this is the way you wanted to test the interface files. WDYT?

haitlahcen:
I have no clue why the windows build is not passing 🤔

snoyberg:
My best guess is that it's discovering a legitimate bug in the parsing code, most likely an architecture-specific parsing. There's probably somewhere in the .hi file which is architecture-specific while your parsing code is using a fixed word size, or vice-versa. Windows builds take place in 32-bit mode.

haitlahcen:
I have no clue why the windows build is not passing :think

haitlahcen:
This is very strange, the output say that the `magic` number is `iface64` while the detected arch is `i386` 
```
  src/test\Stack\ModuleInterfaceSpec.hs:22: 
  1) Stack.ModuleInterface, should succesfully deserialize interface from, GHC822
       uncaught exception: IOException of type UserError (user error (Invalid magic: got: 1face64, expected: 1face))

  src/test\Stack\ModuleInterfaceSpec.hs:22: 
  2) Stack.ModuleInterface, should succesfully deserialize interface from, GHC844
       uncaught exception: IOException of type UserError (user error (Invalid magic: got: 1face64, expected: 1face))

  src/test\Stack\ModuleInterfaceSpec.hs:22: 
  3) Stack.ModuleInterface, should succesfully deserialize interface from, GHC864
       uncaught exception: IOException of type UserError (user error (Invalid magic: got: 1face64, expected: 1face))
```

haitlahcen:
Your guess correct, I missed a difference between the x32/x64 magic number.
Moreover, as I compiled the interface file locally (x64), the CI can't deserialize them in windows (since it would try to deserialize x32 interface). I'm confident in saying that it is safe to bypass the deserialization tests for x32 since the only difference is in the [magic](https://github.com/ghc/ghc/blob/1c2c2d3dfd4c36884b22163872feb87122b4528d/compiler/iface/BinIface.hs#L113) number and a [pointer](https://github.com/ghc/ghc/blob/1c2c2d3dfd4c36884b22163872feb87122b4528d/compiler/iface/BinIface.hs#L125), the rest of the serialized binary being the same.

snoyberg:
I think I see the problem here: you're assuming that the architecture Stack itself is built with will match the architecture of the `.hi` files it's parsing. That's not necessarily true, especially on Windows: you can use a 32-bit GHC with a 64-bit Stack, and vice versa. I'd instead recommend either:

1. Completely ignoring the build architecture, and relying on the magic number to guide the parsing, or
2. Receiving the architecture as a parameter to the parser, and observing the actual `Platform` selected for compiling

Personally, I'd prefer (1), as I don't see a downside to it.

It would be a good idea to include some 32-bit `hi` files as well. If you don't have the ability to generate any, let me know, I can spin up my Windows machine and send some over.

haitlahcen:
I don't see any drawback in using the magic number to know the architecture, let's do that ! And as you said, I don't have the ability to create the x32 version of the files, it would be great if you could do it :)!

test/files/iface/Test.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE CPP #-}
+
+module Main where
+
+import X
+import GHC.Types
+import Control.Monad
+import Language.Haskell.TH
```
snoyberg: How about adding a `qAddDependentFile` call here, since that's the primary feature in `hi` files we need to use?

haitlahcen: The `Test.h` is already listed in the `addDependentFile` section because of the `include`, but I can also add this!

snoyberg:
32-bit hifiles for GHCs 7.10-8.4. No 32-bit GHC for 8.6.4 unfortunately.

[windows-i386-hifiles.tar.gz](https://github.com/commercialhaskell/stack/files/3090178/windows-i386-hifiles.tar.gz)

haitlahcen:
@snoyberg I've added your files, could you check whether this is good to be merged before I rewrite a cleaner git history please ?

snoyberg:
I'll look it over in a bit, and I'm going to kick off a set of integration tests as well.

snoyberg:
Sorry for the failed tests, you can ignore those. This is where you can see the actual integration test results. Still pending at time of writing: https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=1409

haitlahcen:
The results looks good!

package.yaml
```diff
@@ -183,6 +184,7 @@ library:
   - Stack.IDE
   - Stack.Init
   - Stack.Ls
+  - Stack.ModuleInterface
```
snoyberg: Please also add a ChangeLog entry explaining the change here.

src/Stack/Package.hs
```diff
@@ -27,56 +27,62 @@ module Stack.Package
   ,applyForceCustomBuild
   ) where
 
-import qualified Data.ByteString.Lazy.Char8 as CL8
```
snoyberg: I typically ask not to apply stylish-haskell changes like this in PRs, it makes it difficult to see what has actually changed.

src/Stack/Package.hs
```diff
@@ -27,56 +27,62 @@ module Stack.Package
...
+import           System.FilePath                        (replaceExtension)
+import qualified System.FilePath                        as FilePath
 import           System.IO.Error
-import           RIO.Process
-import           RIO.PrettyPrint
-import qualified RIO.PrettyPrint as PP (Style (Module))
 
-data Ctx = Ctx { ctxFile :: !(Path Abs File)
-               , ctxDistDir :: !(Path Abs Dir)
+data Ctx = Ctx { ctxFile        :: !(Path Abs File)
```
snoyberg: Also not a fan of forced alignment of records, it makes diffs larger than necessary

snoyberg:
@haitlahcen Did you notice my last comments?

snoyberg:
Just noticed your comments @ndmitchell.

> From what I saw, the use of the information obtained by parsing the .hi files is purely for warnings?

No, the primary purpose is to determine which files are depended upon by Template Haskell, so that proper dirtiness checking can occur.

> If that's the case, shouldn't the failure to parse the .hi files be a warning, and then just ignore the other warnings, and it would work for pre-releases just fine.

Yes, failure to parse should be a warning, agreed. I just checked the current code, and from what I can tell that's exactly what occurs.

> Do not that Weeder also piggy backs off the .hi files that Stack creates, so if Stack could still create them with some flag that would be helpful for Weeder.

Do you mean the `.hi` files, or the `.hi-dump` files? If the former: we're definitely still generating them. If the latter: this change does break things for weeder currently, though it's relatively trivial to add back the behavior yourself with:

```
--ghc-options "-ddump-to-file -ddump-hi"
```

> Alternatively, Weeder could move to dumping the .hi-dump files itself on demand.

Also alternatively: we could take this `.hi`  parsing code, extract it to a separate library, and both Weeder and Stack could depend upon it. That would bypass the need for either tool to deal with the slow text file generation and parsing.

> As a consequence, this change is a major change.

Can you elaborate on what you mean here by "major change?"

ndmitchell:
Weeder parses the `.hi-dump` files. Given the domain of Weeder, I am less concerned about slowness, and the information I grab from `.hi` files may be a superset of what Stack uses. If a nice `.hi` file parsing library appears I'll use it, otherwise not a big deal.

> As a consequence, this change is a major change.

Major in the sense that there is some fallout. But it's not an issue.

snoyberg:
Thanks for the clarification @ndmitchell. I've created an updated PR including the changes here, as well as:

* Separating the `.hi` file parsing code into a new sub-package called `hi-file-parser`
* Removing the extraneous whitespace changes in `Stack.Package`

Issue #4480, opened 2019-01-03
`stack clean` fails due to missing ghc-pkg (on latest from master)


silky:
looks suspiciously related to https://github.com/commercialhaskell/stack/issues/4466

to repro

1. use stack from https://github.com/commercialhaskell/stack/commit/5144dd57bd78553f5051601295a2eb215d1db5a6
1. go into a project
2. run `stack clean`

error:

```
> stack clean
Executable named ghc-pkg not found on path: [...]
```

version

```
11:21 AM noon ∈ Moon-Tile (master*) ♫ stack --version
Version 1.10.0, Git revision 5144dd57bd78553f5051601295a2eb215d1db5a6 x86_64 hpack-0.31.1
```

mihaimaruseac:
Confirmed that I can reproduce this

qrilka:
It appears to be a bit of a clash between #4412 and #4268 
To fix this properly I guess we need some good way to solve #4390 as after #4412 the most of the commands became more dependent on GHC being available

dbaynard:
Fix is probably to change the behaviour of `stack clean` when GHC is not present. #4466 is probably a coincidence.

We don't have `EnvConfig` without GHC present, and `stack clean` currently relies on `EnvConfig` in order to identify the `Cabal` library present (there's a documentation PR about these precise details following a conversation with @snoyberg).

The sensible thing would be:

-   if there's only one `Cabal` directory in `dist` then use that as the target (as now)
-   if there are multiple `Cabal` directories, either
    -   prompt the user for which to delete, or
    -   prompt the user to install ghc, or
    -   select a dist directory to clean (unexpected behaviour), or
    -   download GHC (undesirable and unexpected behaviour), or
    -   delete multiple dist directories (probably undesirable behaviour), or
    -   present an error, and introduce flags to control behaviour (seems unnecessarily complicated).

I'd favour the prompts but that leads to issues in scripts.

For now, does running `stack setup` work around the issue?

mihaimaruseac:
Prompts with a default and flags for the other would help the scripts.

qrilka:
@dbaynard I don't understand why do you say that deleting the only one `Cabal` directory in `dist` is the way `stack clean` currently works, see e.g.:
```
qrilka@qdesktop ~/ws/h/stack-build-test $ ls -l $(dirname `stack path --dist-dir`)
итого 8
[...]

PR #4547, opened 2019-01-28
Globals (GHC boot packages) pruning

### LINKED ISSUE #4510: Rebuilding packages from the global database with modified dependencies
This applies to Stack master, specifically commit 9900d99114967698223f43c84e4cedecc24cea97. Repro (on a non-Windows system). With the following `stack.yaml`:

```yaml
resolver: ghc-8.6.3
packages: []
extra-deps:
- time-1.8.0.1
```

This replaces the global `time` package with a different version. Running `stack build unix` causes both `time` and `unix` to be compiled, which makes sense, since `unix` depends on `time`.

Running `stack build unix` a second time should be a no-op, but instead, it rebuilds:

```
$ stack build unix
unix-2.7.2.2: configure
^C
```

This appears to be because the `SourceMap` knows nothing about `unix` (since it's a global package and not present at all in snapshot), and therefore Stack is ignoring the already-present copy in the snapshot database.

Possible solution: update the SourceMap to contain information about global packages.

CC @qrilka

qrilka:

Tested manually, plan to run integration tests.
Failure to build global packages as targets results in 2 errors (using `stack.yaml` from #4510 with `time` overriden):
```
$ stack build directory --dry-run

Can't use GHC boot package directory when it has an overriden dependency, you need to add the following as explicit dependencies to the
project: 
 unix

  * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some working build configuration. This can be
    convenient when dealing with many complicated constraint errors, but results may be unpredictable.


Plan construction failed.
```
or
```
$ stack build process --dry-run

In the dependencies for process-1.6.5.0:
    directory must match >=1.1 && <1.4, but the stack configuration has no specified version  (latest matching version is 1.3.3.2)
    unix must match >=2.5 && <2.9, but the stack configuration has no specified version  (latest matching version is 2.7.2.2)
needed since process is a build target.

  * Consider trying 'stack solver', which uses the cabal-install solver to attempt to find some working build configuration. This can be
    convenient when dealing with many complicated constraint errors, but results may be unpredictable.

  * Recommended action: try adding the following to your extra-deps in /home/qrilka/ws/h/stack4510/stack.yaml:

directory-1.3.3.2@sha256:79e15a62f90cb9f20f46b2b56fe67d00d71ed54cfad07ea337695f9ebb74baa4,2829
unix-2.7.2.2@sha256:4ef1d010d70a4a07a717e853d4a440c105dad38c6199316e320fdd4c48dacd34,3495

Plan construction failed.
```
Fixes #4510

Also I plan to update curator to take these details into account (probably as a separate PR)

### DIFF
--- a/src/Stack/SourceMap.hs
+++ b/src/Stack/SourceMap.hs
@@ -10,10 +10,12 @@ module Stack.SourceMap
     , loadGlobalHints
     , toActual
     , checkFlagsUsedThrowing
+    , globalCondCheck
     ) where
 
 import qualified Data.Conduit.List as CL
 import qualified Distribution.PackageDescription as PD
+import Distribution.System (Platform(..))
 import Pantry
 import qualified RIO
 import qualified RIO.Map as Map
@@ -146,10 +148,11 @@ toActual smw downloadCompiler ac = do
         case downloadCompiler of
             WithDownloadCompiler -> globalsFromDump ac
             SkipDownloadCompiler -> globalsFromHints (actualToWanted ac)
+    check <- globalCondCheck
     (prunedGlobals, keptGlobals) <-
-        partitionReplacedDependencies allGlobals (Map.keysSet $ smwDeps smw)
+        partitionReplacedDependencies allGlobals (Map.keysSet $ smwDeps smw) check
     let globals = Map.map GlobalPackage keptGlobals <>
-                  Map.fromSet (const ReplacedGlobalPackage) prunedGlobals
+                  Map.map ReplacedGlobalPackage prunedGlobals
     return
         SMActual
         { smaCompiler = ac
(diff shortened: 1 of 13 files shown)


src/Stack/SourceMap.hs
```diff
@@ -111,26 +113,26 @@ getPLIVersion (PLIRepo _ pm) = pkgVersion $ pmIdent pm
...
             ( pkgName $ dpPackageIdent d
-            , GlobalPackage (pkgVersion $ dpPackageIdent d))
+            , pkgVersion $ dpPackageIdent d)
     toGlobals <$> ghcPkgDump (whichCompiler compiler) [] pkgConduit
 
 globalsFromHints ::
        HasConfig env
     => WantedCompiler
-    -> RIO env (Map PackageName GlobalPackage)
+    -> RIO env (Map PackageName Version)
```
snoyberg: I'm almost tempted to say we should have a special `GlobalVersion` newtype wrapper to drive the point home further, but this is good.

qrilka:
I have fixed integration test `345-override-bytestring` and added this pruning to curator. It looks like `build-constraints.yaml` doesn't include `parsec` (while `mtl` is overriden) - PR on https://github.com/commercialhaskell/stackage will follow

qrilka:
This package pruning implementation partially relies on information in `global-hints.yaml` but as e.g. Michael demonstrated in his blog post packages shipped with GHC could have no (or bad) corresponding version on Hackage so I'm changing it to to use correct package information from `ghc-pkg`

qrilka:
I have changed the code quite a lot, also it looks like the refactoring I promised almost a month ago https://github.com/commercialhaskell/stack/issues/4480#issuecomment-452037476 could be quite helpful in improving code around `SkipCompilerDownload`.
@snoyberg could I ask you for another look?

PR #4554, opened 2019-02-03
Use batched ghc-pkg unregister implemented in GHC 8.0.1

### LINKED ISSUE #2662: Use new batch ghc-pkg unregister
I've added a feature to `ghc-pkg unregister` (https://ghc.haskell.org/trac/ghc/ticket/12637) that allows you to `ghc-pkg unregister pkg1 pkg2 pkg3`.

This can make stack's auto-unregistering much faster, especially when going through docker (where each `ghc-pkg` command takes up to a few seconds; on large programs this can take very long as a result).

It'll be available with GHC 8.2.1.

qrilka:

Tested manually

Fixes #2662

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -41,6 +41,7 @@ import           Data.Conduit.Process.Typed
                      runProcess_, getStdout, getStderr, createSource)
 import qualified Data.Conduit.Text as CT
 import           Data.List hiding (any)
+import           Data.List.Split (chunksOf)
 import qualified Data.Map.Strict as M
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
@@ -584,14 +585,12 @@ executePlan' installedMap0 targets plan ee@ExecuteEnv {..} = do
         [] -> return ()
         ids -> do
             localDB <- packageDatabaseLocal
-            forM_ ids $ \(id', (ident, reason)) -> do
-                logInfo $
-                    fromString (packageIdentifierString ident) <>
-                    ": unregistering" <>
-                    if T.null reason
-                        then ""
-                        else " (" <> RIO.display reason <> ")"
-                unregisterGhcPkgId wc cv localDB id' ident
+            case cv of
+                ACGhc v | v >= mkVersion [8, 0, 1] ->
+                    for_ (chunksOf 500 ids) $ unregisterPackages wc localDB
+                _ ->
+                  for_ ids $ \(gid, (ident, reason)) -> do
+                      unregisterSinglePackage cv localDB gid ident reason
 
     liftIO $ atomically $ modifyTVar' eeLocalDumpPkgs $ \initMap ->
         foldl' (flip Map.delete) initMap $ Map.keys (planUnregisterLocal plan)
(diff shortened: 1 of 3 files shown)


src/Stack/GhcPkg.hs
```diff
@@ -11,6 +11,7 @@ module Stack.GhcPkg
   ,findGhcPkgField
   ,createDatabase
   ,unregisterGhcPkgId
```
dbaynard: Maybe change this to

```suggestion
  ,unregisterSingleGhcPkgId
```

(and rename the function at line 151, where I can't comment) as it is clearer than the only difference being the plural suffix 's' and is consistent with `unregisterSinglePackage` vs `unregisterPackages`.

(note: github still haven't fixed the bug with indentation in their suggestions.)

src/Stack/GhcPkg.hs
```diff
@@ -167,6 +168,21 @@ unregisterGhcPkgId wc cv pkgDb gid ident = do
...
+                    -> [GhcPkgId]
+                    -> RIO env ()
+unregisterGhcPkgIds wc pkgDb gids = do
+    eres <- ghcPkg wc [pkgDb] args
+    case eres of
+        Left e -> logWarn $ displayShow e
+        Right _ -> return ()
+  where
+    args = "unregister" : "--user" : "--force" :
+        concatMap (\gid -> ["--ipid", ghcPkgIdString gid]) gids
```
dbaynard: If you expand the aforementioned `case`, this (duplicate) function then becomes unnecessary, as you can just rename `unregisterGhcPkgId`.

src/Stack/Build/Execute.hs
```diff
@@ -584,14 +585,12 @@ executePlan' installedMap0 targets plan ee@ExecuteEnv {..} = do
...
-                    if T.null reason
-                        then ""
-                        else " (" <> RIO.display reason <> ")"
-                unregisterGhcPkgId wc cv localDB id' ident
+            case cv of
+                ACGhc v | v >= mkVersion [8, 0, 1] ->
+                    for_ (chunksOf 500 ids) $ unregisterPackages wc localDB
+                _ ->
+                  for_ ids $ \(gid, (ident, reason)) -> do
+                      unregisterSinglePackage cv localDB gid ident reason
```
dbaynard: We actually have 3 cases here:

-  GHC >= 8.0.1 uses `unregisterPackages`
-  GHC < 8.0.1 && >= 7.9 uses `unregisterSinglePackage` with a `GhcPkgId`
-  GHC < 7.9 uses `unregisterSinglePackage` with an identifier.

It may be worth reflecting this here, to make clear that this is the case.
Or not — if it's better to focus on two cases: new vs old.

dbaynard: Actually, if you do the `case` here, you don't have to pass the `ActualCompiler` which greatly simplifies the `unregister*` functions.

src/Stack/Build/Execute.hs
```diff
@@ -661,6 +660,38 @@ executePlan' installedMap0 targets plan ee@ExecuteEnv {..} = do
...
+    -> Text
+    -> RIO env ()
+unregisterSinglePackage cv localDB gid ident reason = do
+    let wc = view whichCompilerL cv
+    logInfo $
+        fromString (packageIdentifierString ident) <> ": unregistering" <>
+        if T.null reason
+            then ""
+            else " (" <> RIO.display reason <> ")"
+    unregisterGhcPkgId wc cv localDB gid ident
```
dbaynard: With the change to the `case` above, this whole function could simply call `unregisterPackages` like

```haskell
  unregisterPackages wc localDB [(gid, (ident, reason))]
```

qrilka:
@dbaynard I've used your suggestions, please take another look

dbaynard:
Thanks, @qrilka. I'll check it out tomorrow.

src/Stack/Build/Execute.hs
```diff
@@ -661,6 +655,35 @@ executePlan' installedMap0 targets plan ee@ExecuteEnv {..} = do
...
+unregisterPackages wc ac localDB ids = do
+    let logReason ident reason =
+            logInfo $
+            fromString (packageIdentifierString ident) <> ": unregistering" <>
+            if T.null reason
+                then ""
+                else " (" <> RIO.display reason <> ")"
+    case ac of
+        ACGhc v | v >= mkVersion [8, 0, 1] -> do
+                let batchSize = 500
```
nh2: Add a comment on the choice of `500`? I assume max command line argument length?

If yes, you may want to adjust this depending on the OS, e.g. IIRC Windows can only do 8 KB long invocations, and on Linux it's `getconf ARG_MAX`, which returns to `2097152` on my Ubuntu, and according to my strace is computed as

```c
getrlimit(RLIMIT_STACK, {rlim_cur=8192*1024, rlim_max=RLIM64_INFINITY}) = 0
```

qrilka: Yeah, good catch, Windows was my main thought, I remember older its versions supporting only 255 characters in paths...

qrilka: I've added a comment and switched to 100 when using Windows

nh2: Nice

dbaynard:
@qrilka I pushed a commit. If you're happy with it, go ahead — otherwise feel free to revert.

dbaynard:
Looks like CI didn't run on my last commit; got an hlint warning. I'll sort it.

PR #4561, opened 2019-02-04
Respect targets while printing dependency tree

### LINKED ISSUE #4552: ls dependencies --tree fails when targets are provided

I've built a stack from the git master and tried to test the thing.

### Steps to reproduce

```bash
$ stack ls dependencies aeson --tree
```

### Expected

Dependency tree for `aeson` package.

### Actual

```
Packages
Map.!: given key is not an element in the map
  error, called at libraries/containers/Data/Map/Internal.hs:610:17 in containers-0.6.0.1:Data.Map.Internal
```

### Stack version

```bash
$ stack --version
Version 1.10.0 (6854 commits) x86_64 hpack-0.31.1
```

### Method of installation

* git clone https://github.com/commercialhaskell/stack && stack install --resolver snapshot-nightly.yaml

akshaymankar:

Fixes #4552


dbaynard:
Thanks for the PR! I'll review shortly.

dbaynard:
Actually, I'll rebase & force push first — I'm not confident on github's handling of review comments in the presence of force pushes, and I want to run tests on latest master. Thanks!

src/Stack/Dot.hs
```diff
@@ -140,28 +140,37 @@ listDependencies opts = do
...
     do
       liftIO $ Text.putStrLn "Packages"
-      liftIO $ printTree opts 0 [] pkgs resultGraph
+      liftIO $ printTree opts 0 [] (treeRoots opts pkgs) resultGraph
     else
       void (Map.traverseWithKey go (snd <$> resultGraph))
       where go name payload = liftIO $ Text.putStrLn $ listDepsLine opts name payload
 
+treeRoots :: ListDepsOpts -> Set PackageName -> Set PackageName
+treeRoots opts defaults = let targets = dotTargets $ listDepsDotOpts opts
```
dbaynard: These `defaults` are actually the set of local packages, right? Calling them `defaults` obscures the reason for ignoring them in the `else`.

src/Stack/Dot.hs
```diff
@@ -140,28 +140,37 @@ listDependencies opts = do
   if listDepsTree opts then
     do
       liftIO $ Text.putStrLn "Packages"
-      liftIO $ printTree opts 0 [] pkgs resultGraph
```
dbaynard: So the current line ignores the targets completely — am I reading that correctly?

src/Stack/Dot.hs
```diff
@@ -140,28 +140,37 @@ listDependencies opts = do
...
+                           else Set.fromList $ map (mkPackageName . Text.unpack) targets
+
 printTree :: ListDepsOpts
           -> Int
           -> [Int]
           -> Set PackageName
           -> Map PackageName (Set PackageName, DotPayload)
           -> IO ()
-printTree opts depth remainingDepsCounts packages dependencyMap =
+printTree opts depth remainingDepsCounts packages dependencyMap = do
```
dbaynard: This is unnecessary — you only have a single expression.

```suggestion
printTree opts depth remainingDepsCounts packages dependencyMap =
```

dbaynard: Also I'm going to have to take a break until tomorrow; I think this function could be clearer.

src/Stack/Dot.hs
akshaymankar: Yes, that is correct.

src/Stack/Dot.hs
dbaynard: That ambiguity makes sense — thank you for pointing it out. I meant whatever `localNames` means here.

https://github.com/commercialhaskell/stack/blob/1bbed48cf22ecc7a3fa3ed230e1cc6354fdf809f/src/Stack/Dot.hs#L89-L104

So this should be changed to `projectNames` too, then?

qrilka: that would be a better name but it's a bit confusing too, I think something like `projectPackages` or `projectPackageNames` would be better, also it makes sense to change the haddock above as well

dbaynard: `projectPackages` is clear, as they are clearly names from the types.

I'll keep a look out for any `local`s.

dbaynard:
Ah but we can't call it `projectPackages` because of name shadowing 🤦 (hence the build failures). I've included the changes in 7e204b8 in my WIP.

dbaynard:
I'm finding testing to be a real pain, here. I wrote a short test suite, but because the tree is printed immediately after it has been generated it needs an io-capture like wrapper. I'm considering moving to use `rio-prettyprint`. WIP.

snoyberg:
For now, if you're able to add an integration test that tests this codepath, I'd be able to review and merge, and we can deal with refactoring and unit tests after a merge.

PR #4563, opened 2019-02-05
Ignore autogenerated cabal files

qrilka:

As we now have proper `hpack` support in `master` cabal files should be OK with autogenerated cabal files.


qrilka:
@dbaynard I don't think so - as far as I know we don't build Stack with the version from `master` in CI

dbaynard:
Forgive me — I was unclear earlier. Yes, new major versions aren't pulled directly from master — but I meant in terms of tests, if there's a regression and it turns out we need the cabal files, we will see build failures in CI.

PR #4573, opened 2019-02-06
Don't error out on missing latest revision while constructing a build plan

### LINKED ISSUE #4572: No package revision found for: PackageName "ghc"
I am not sure if the following would be considered to be a `stack` issue or not. It flows from https://github.com/haskell/haskell-ide-engine/issues/1074

The Haskell IDE Engine (HIE) builds with stack 1.9.3 under GHC 8.6.2. Command `stack --stack-yaml=stack-8.6.2.yaml ls dependencies` includes in its output:
```
ghc mkVersion [8,6,2]
```
`ghc-8.6.2` is the library that comes with GHC 8.6.2 (https://downloads.haskell.org/~ghc/8.6.2/docs/html/users_guide/8.6.1-notes.html#included-libraries)

However, the versions of `ghc` on Hackage are limited to 8.2.1, 8.2.2, 8.4.1, 8.4.3, and 8.6.1 (https://hackage.haskell.org/package/ghc).

In trying to build with with `stack --version` `Version 1.10.0, Git revision 48f4befb67860a24fc8c3310e6d55ace9fd35dfe (dirty) (6868 commits) x86_64 hpack-0.31.1`, the command:
```
stack --stack-yaml=stack-8.6.2.yaml install
```
yields error:
```
No package revision found for: PackageName "ghc"
  error, called at src/Stack\Build\ConstructPlan.hs:434:42 in stack-1.10.0-Gyq3czmxkWV9B1iWoXQn3F:Stack.Build.ConstructPlan
```
That reference to `Stack.Build.ConstructPlan.hs` leads to function `getHackagePackageVersions`.

My own assumption is that either (1) `stack` should 'know' that package `ghc` 'comes with' GHC, even if it is not on Hackage; or (2) that all versions of `ghc` that come with GHC ought to be on Hackage.

qrilka:

Tested manually.
Actually I've fixed it earlier in https://github.com/commercialhaskell/stack/tree/curator-stackage-build-with-stack as I caught it during Stackage build. And it was me who added it as a part of #4479 - sorry
Fixes #4572

### DIFF
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -426,9 +426,15 @@ addDep treatAsDep' name = do
                                 askPkgLoc = liftRIO $ do
                                   mrev <- getLatestHackageRevision name version
                                   case mrev of
-                                    Nothing -> error $ "No package revision found for: " <> show name
+                                    Nothing -> do
+                                      -- this could happen for GHC boot libraries missing from Hackage
+                                      logWarn $ "No latest package revision found for: " <>
+                                          fromString (packageNameString name) <> ", dependency callstack: " <>
+                                          displayShow (map packageNameString $ callStack ctx)
+                                      return Nothing
                                     Just (_rev, cfKey, treeKey) ->
-                                      return $ PLIHackage (PackageIdentifier name version) cfKey treeKey
+                                      return . Just $
+                                          PLIHackage (PackageIdentifier name version) cfKey treeKey
                             tellExecutablesUpstream name askPkgLoc loc Map.empty
                             return $ Right $ ADRFound loc installed
                         Just (PIOnlySource ps) -> do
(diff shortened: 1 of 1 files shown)


src/Stack/Build/ConstructPlan.hs
```diff
@@ -426,9 +426,15 @@ addDep treatAsDep' name = do
                                 askPkgLoc = liftRIO $ do
                                   mrev <- getLatestHackageRevision name version
                                   case mrev of
-                                    Nothing -> error $ "No package revision found for: " <> show name
+                                    Nothing -> do
+                                      -- this could happen for GHC boot libraries missing from Hackage
```
dbaynard: I found a comment in another part of the code which cleared up my misunderstanding the other day.

https://github.com/commercialhaskell/stack/blob/12d11a64f70cc41ec3709486df4ce5073e1620be/src/Stack/Build/Target.hs#L315-L318

So if stack _doesn't_ find a revision, that's not a problem. If it _does_ find a revision, it will use that. That's correct, then?

qrilka:
@dbaynard I've fixed it earlier yesterday while doing Stackage builds

src/Stack/Build/ConstructPlan.hs
qrilka: Now when you ask it that way I'm not completely sure that it's 100% correct logic - the reason is that this `PIOnlyInstalled` should appear only for global libraries (GHC boot packages) and for those we don't have guarantees about proper versions/revisions on Hackage. From the other side as far as I understand we don't have any executables in those packages so practically `tellExecutablesUpstream` should be NOOP.

PR #4582, opened 2019-02-11
Curator stackage build with stack

### LINKED ISSUE #4217: Write a new stackage-curator tool
Some basic ideas: https://github.com/snoyberg/pantry/blob/master/PLANS.md#stackage

qrilka:

Tested using the latest Stackage snapshot builds.

Fixes #4217 

This mostly changes `curator` code but also adds extra tweaks into Stack's code to deal properly with expected test and benchmark failures


qrilka:
The last rebuild took
```
real    3m14.387s
user    1m5.144s
sys     0m9.860s
```
so it looks more or less OK and #4550 should improve it a bit more

qrilka:
@snoyberg looks like I found a problem which is a bit related to removing of `InstalledCache` but it's not about mutable dependencies (it looks to be that we shouldn't have problems specific to them). When we switch between different implicit snapshots (e.g. by adding `--profile`) we could loose information about dirtiness of local files and we have a following failure scenario:
  1) You have package compiled with and without profiling
  2) You change some files in that package
  3) You build it without profiling
  4) Then when you build it with profiling there's no information about dirty files so you have Stack's build results for files you had in 1)
I will work on fixing this (and will add a test for that of course)

qrilka:
Fixed rebuilds and added 2 integration tests

PR #4589, opened 2019-02-14
doc/yaml_configuration.md: clarify setup-info behaviour

borsboom:
@afcowie How does this clarification look?

### DIFF
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -629,16 +629,18 @@ setup-info:
         url: "https://example.com/ghc-7.10.2-i386-unknown-mingw32-foo.tar.xz"
 ```
 
-Or without using `ghc-variant`:
+Or you can point to external setup-info:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


afcowie:
That's great. Two suggestions:

- You might tweak it to say "if you need to replace it..."

- link the the canonical URL of the "`default`", being _[stack-setup-2.yaml](https://github.com/commercialhaskell/stackage-content/blob/master/stack/stack-setup-2.yam)_ (aka _<https://github.com/commercialhaskell/stackage-content/raw/master/stack/stack-setup-2.yaml>_ as pulled by the `stack` binary. Figuring that path out [to subsequently patch it, see [`patch-setup`](https://github.com/oprdyn/docker.haskell/blob/master/files/usr/local/bin/patch-setup)] was a lot of work).

otherwise, many thanks.

AfC

borsboom:
Alright, I've added another commit with your suggestions.

PR #4602, opened 2019-03-04
Add documentation for lock files

snoyberg:

### DIFF
--- a/doc/lock_files.md
+++ b/doc/lock_files.md
@@ -235,7 +235,7 @@ complicated update procedure?
 3. For each incomplete package location:
     * Lookup the value in the map created in (1)
     * If present: use that completed information
-    * Otherwise: complete the information using the procedure in
-      "creation"
+    * Otherwise: complete the information using the same completion
+      procedure from Pantry as in "creation"
 
 This should minimize the number of changes to packages incurred.
--- a/doc/lock_files.md
+++ b/doc/lock_files.md
@@ -1,0 +1,239 @@
+<div class="hidden-warning"><a href="https://docs.haskellstack.org/"><img src="https://cdn.jsdelivr.net/gh/commercialhaskell/stack/doc/img/hidden-warning.svg"></a></div>
+
+# Lock Files
+
+Stack attempts to provide reproducible build plans. This involves
+reproducibly getting the exact same contents of source packages and
+configuration options (like cabal flags and GHC options) for a given
+set of input files. There are a few problems with making this work:
+
+* Entering all of the information to fully provide reproducibility is
+  tedious. This would include things like Hackage revisions, hashes of
+  remote tarballs, etc. Users don't want to enter this information.
+* Many operations in Stack rely upon a "snapshot hash," which
+  transitively includes the completed information for all of these
+  dependencies. If any of that information is missing when parsing the
+  `stack.yaml` file or snapshot files, it could be expensive for Stack
+  to calculate it.
+
+To address this, we follow the (fairly standard) approach of having a
+_lock file_. The goal of the lock file is to cache completed
+information about all packages and snapshot files so that:
+
+* These files can be stored in source control
+* Users on other machines can reuse these lock files and get identical
+  build plans
+* Rerunning `stack build` in the future is deterministic in the build
+  plan, not depending on mutable state in the world like Hackage
+  revisions
+    * **NOTE** If, for example, a tarball available remotely is
+      deleted or the hash changes, it will not be possible for Stack
+      to perform the build. However, by deterministic, we mean it
+      either performs the same build or fails, never accidentally
+      doing something different.
+* Stack can quickly determine the build plan in the common case of no
+  changes to `stack.yaml` or snapshot files
+
+This document explains the contents of a lock file, how they are used,
+and how they are created and updated.
+
+## stack.yaml and snapshot files
+
+Relevant to this discussion, the `stack.yaml` file specifies:
+
+* Resolver (the parent snapshot)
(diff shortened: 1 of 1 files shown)


doc/lock_files.md
```diff
@@ -0,0 +1,239 @@
...
+complicated update procedure?
+
+1. Create a map from original package location to completed package
+   location in the lock file
+2. Load up each snapshot file
+3. For each incomplete package location:
+    * Lookup the value in the map created in (1)
+    * If present: use that completed information
+    * Otherwise: complete the information using the procedure in
+      "creation"
```
qrilka: The only problem is that "creation" doesn't tell much about the procedure

doc/lock_files.md
```diff
@@ -0,0 +1,239 @@
...
+of requiring that all snapshot files be fully parsed on each Stack
+invocation, we can store information in the lock file and bypass
+parsing of the additional files in the common case of no changes.
+
+## Lock file contents
+
+The lock file contains the following information:
+
+* The full snapshot definition information, including completed
+  package locations
```
psibi: Can you clarify here about which "complete package locations" you mean ? Do you mean 
* completed package locations of the packages in the snapshot file ?
* Or complete package locations of the packages in the `extra-deps` file (or both) ?

doc/lock_files.md
```diff
@@ -0,0 +1,239 @@
...
+      sha256: XXXX # QUESTION: do we really need hashes of local files?
+      size: XXXX
+  - original:
+      lts-13.9
+    completed:
+      size: 496662
+      url: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/13/9.yaml
+      sha256: 83de9017d911cf7795f19353dba4d04bd24cd40622b7567ff61fc3f7223aa3ea
+
+compiler: ghc-X.Y.Z
```
psibi: What about global ghc option ? Shouldn't it be also present ?

qrilka: what do you mean by global here?

psibi: @qrilka Probably global is the wrong word here. `stack.yaml` obtains `ghc-option` for the package. Should that option be present here in the lock file ?

qrilka: GHC options get applied per package, why do we need them here?

doc/lock_files.md
```diff
@@ -0,0 +1,239 @@
...
+
+Stack attempts to provide reproducible build plans. This involves
+reproducibly getting the exact same contents of source packages and
+configuration options (like cabal flags and GHC options) for a given
+set of input files. There are a few problems with making this work:
+
+* Entering all of the information to fully provide reproducibility is
+  tedious. This would include things like Hackage revisions, hashes of
+  remote tarballs, etc. Users don't want to enter this information.
+* Many operations in Stack rely upon a "snapshot hash," which
```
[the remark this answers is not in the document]
snoyberg: This is the hash of a snapshot used internally for file directories. I don't want to add a long description of it to this document since it's not directly related to lock files.

doc/lock_files.md
snoyberg: I mean both.

doc/lock_files.md
```diff
@@ -0,0 +1,239 @@
...
+
+snapshots:
+  # Starts with the snapshot specified in stack.yaml,
+  # then continues with the snapshot specified in each
+  # subsequent snapshot file
+  - original:
+      foo.yaml # raw content specified in a snapshot file
+    completed:
+      file: foo.yaml
+      sha256: XXXX # QUESTION: do we really need hashes of local files?
```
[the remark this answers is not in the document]
snoyberg: That's essentially what all of this is coming down to: do we trust modification times exclusively, or only use them as an optimization? Thinking on it more and reading your feedback, I think we need the hashes. I'll update.

doc/lock_files.md
snoyberg: There may be one corner case we need to consider: `ghc-options` applied to all packages. However, I think that will be addressed by the approach described here.

doc/lock_files.md
```diff
@@ -0,0 +1,239 @@
...
+* Writing out the new `stack.yaml.lock` file
+
+## Dirtiness checking
+
+If the `stack.yaml.lock` file exists, its last modification time is
+compared against the last modification time of the `stack.yaml` file
+and any local snapshot files. If any of those files is more recent
+than the `stack.yaml` file, then the update procedure is
+triggered. Otherwise, the `stack.yaml.lock` file can be used as the
+definition of the snapshot.
```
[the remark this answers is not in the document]
snoyberg: Snapshot files never contain mutable package information, only immutable. The mutable packages are stored exclusively in the `stack.yaml` file. We _will_ need to parse the `stack.yaml` in every case regardless.

doc/lock_files.md
snoyberg: If there's anything ambiguous about the procedure, please ask. I thought the procedure is fairly straightforward based on existing code we have in both Stack and Pantry.

doc/lock_files.md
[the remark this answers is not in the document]
qrilka: Oh, right, I completely forgot that thing somehow

doc/lock_files.md
qrilka: I think it could be a bit confusing for a person not quite familiar with the details in Pantry - maybe it should say something like "using the same completion procedure from Pantry as in "creation""?

snoyberg: Good call, updated.

PR #4604, opened 2019-03-06
Docker: fix pull missing images with `--docker-auto-pull` (fixes #4598)

### LINKED ISSUE #4598: Pull specific images using --docker-auto-pull option
### Expected

Pull a suitable image automatically.

### Actual

```
$ stack --docker-auto-pull install
Error: No such object: fpco/stack-build:lts-13.8
Received ExitFailure 1 when running
Raw command: /snap/bin/docker inspect fpco/stack-build:lts-13.8
Standard output:

[]
```

### Stack version

```
$ stack --version
Version 1.9.3, Git revision 40cf7b37526b86d1676da82167ea8758a854953b (6211 commits) x86_64 hpack-0.31.1
```

### Method of installation

borsboom:

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -640,6 +640,8 @@ inspects :: (HasProcessContext env, HasLogFunc env)
 inspects [] = return Map.empty
 inspects images =
   do maybeInspectOut <-
+       -- not using 'readDockerProcess' as the error from a missing image
+       -- needs to be recovered.
        try (BL.toStrict . fst <$> proc "docker" ("inspect" : images) readProcess_)
      case maybeInspectOut of
        Right inspectOut ->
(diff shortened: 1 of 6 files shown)


src/Stack/Docker.hs
```diff
@@ -639,7 +639,8 @@ inspects :: (HasProcessContext env, HasLogFunc env)
          => [String] -> RIO env (Map String Inspect)
 inspects [] = return Map.empty
 inspects images =
-  do maybeInspectOut <- try (readDockerProcess ("inspect" : images))
+  do maybeInspectOut <-
+       try (BL.toStrict . fst <$> proc "docker" ("inspect" : images) readProcess_)
```
dbaynard: Is it worth fixing `readDockerProcess` rather than overriding it? I notice that there's a FIXME in the below code.

https://github.com/commercialhaskell/stack/blob/b01c526c733d30ab2df8fa296e0e02f6314c15be/src/Stack/Docker.hs#L844-L850

borsboom: That was what I tried first, but the current `readDockerProcess` behaviour of passing stderr straight through is desirable for some docker commands such as `docker pull` (so you can see the fancy pull progress that Docker outputs).

snoyberg: If that's the case, I think the FIXME can be removed (and thanks for noticing that David).

PR #4614, opened 2019-03-08
Implemented "ignore-expiry" flag as in "cabal --ignore-expiry

l984:

* [X ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [X ] The documentation has been updated, if necessary.

This change simply exposes the same functionality which cabal already implements and exposes using "--ignore-expiry" flag.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -118,6 +118,11 @@ Other enhancements:
   [#4480](https://github.com/commercialhaskell/stack/issues/4480).
 * Add `stack purge` as a shortcut for `stack clean --full`. See
   [#3863](https://github.com/commercialhaskell/stack/issues/3863).
+* Add an optional `ignore-expiry` flag to the `hackage-security`
+  section of the `~/.stack/config.yaml`. It allows to disable timestamp
+  expiration verification just like `cabal --ignore-expiry` does.
+  The flag is not enabled by default so that the default functionality
+  is not changed.
 
 Bug fixes:
 
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -354,6 +354,7 @@ package-indices:
     - aa315286e6ad281ad61182235533c41e806e5a787e0b6d1e7eef3f09d137d2e9
     - fe331502606802feac15e514d9b9ea83fee8b6ffef71335479a2e68d84adc6b0
     key-threshold: 3 # number of keys required
+    ignore-expiry: no
 ```
 
 If you provide a replacement index which does not mirror Hackage, it
--- a/subs/pantry/src/Pantry.hs
+++ b/subs/pantry/src/Pantry.hs
@@ -254,6 +254,7 @@ defaultHackageSecurityConfig = HackageSecurityConfig
       ]
   , hscKeyThreshold = 3
   , hscDownloadPrefix = "https://s3.amazonaws.com/hackage.fpcomplete.com/"
+  , hscIgnoreExpiry = False
   }
 
 -- | Returns the latest version of the given package available from
--- a/subs/pantry/src/Pantry/Hackage.hs
+++ b/subs/pantry/src/Pantry/Hackage.hs
@@ -82,7 +82,7 @@ updateHackageIndex
 updateHackageIndex mreason = gateUpdate $ do
     for_ mreason logInfo
     pc <- view pantryConfigL
-    let HackageSecurityConfig keyIds threshold url = pcHackageSecurity pc
+    let HackageSecurityConfig keyIds threshold url ignoreExpiry = pcHackageSecurity pc
     root <- view hackageDirL
     tarball <- view hackageIndexTarballL
     baseURI <-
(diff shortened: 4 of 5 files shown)


dbaynard:
Hi @l984, thanks for the PR.

We usually ask for people to raise issues before requesting a PR, so would you kindly provide answers to the following questions:

-   What behaviour does this PR change?
-   What is the current behaviour?
-   What is the new behaviour?

In addition, you've ticked the box saying 'The documentation has been updated if necessary' but you haven't updated any documentation, and it is necessary. If this PR is to be merged it will need that.

There are also no tests; it would be good to see an integration test for this.

Thanks!

l984:
Hi and thank you for the quick response!

## Original motivation

My original idea was to implement local stack/hackage partial mirror
so that I can (re-)test automatic haskell stack installation without direct
access to Internet and without a need to (re-)download tons of data.

However, stack/hackage package management system apparently implements
some expiration logic which not only prevents from using an outdated
mirror/repository but does not even expose any way to disable expiration
logic.

This PR introduces a new optional `"ignore-expiry"` boolean flag to the
`~/.stack/config.yaml` file just for this very purpose.

## How the problem looks without the PR

Without this PR when using stack with outdated mirror stack complains
on timestamps. After brief examination I saw that the problem was
related to the file `~/.stack/indices/Hackage/timestamp.json` which puts
limits on expiration time. In my case I can see this:

```
... "expires":"2019-03-09T18:16:28Z" ...
```

Honestly I do not really understand the motives behind these
restrictions about timestamps. They do not act like a mere "http cache
expiry" but like a real "kill switch" by making repository totally
useless after expiration.

The PR solves this problem.

## References to the documentaion about timestamp/expiry logic

Documentation about timestamp and expiration time is probably
related to the keywords and sections about _Timestamp role_,
`timestamp.json` and _Update the root metadata file_ in [The Update
Framework Specification.](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md)

Expiry logic is implemented by the [checkForUpdates](http://hackage.haskell.org/package/hackage-security-0.5.3.0/docs/Hackage-Security-Client.html) and documentation
for this function explicitly refers to the TUF spec.

Finally, here is cabal documentation about [`ignore-expiry`](https://www.haskell.org/cabal/users-guide/nix-local-build.html) flag.

## Free software

Without any possibility to disable expiration checks the user of
stack/hackage can not use long-living mirrors and has to fully rely on
regular updates on the hackage server. If I understand correctly, any
disruption in hackage update process will result to users not being
able to install haskell stack.

From my point of view this approach takes away control from the user
of the software and gives full control to the publisher. This is very
much in the spirit of Google Play, Android Updates and not in the
spirit of free software.

## More things to be done about this PR

### Documentation

The PR already has an update to the documentation - see the change in
the `doc/yaml_configuration.md` file. I could document it a bit more
but the whole secion `package-indices` is quite brief so I thought
that `ignore-expiry` should have the same coverage as other fields in
this section.

### Tests

It would be difficult to make proper testcases - I would certainly
need help with this. Thank you in advance.

_Thank you!_

dbaynard:
Hi @l984, thank you for answering the questions.

Going through your responses, I realised that it would be helpful to have a quick repro (rather than having to wait for the cache to expire). Can you reproduce this by editing your timestamp file?

> This PR introduces a new optional `ignore-expiry` boolean flag to the `~/.stack/config.yaml` file just for this very purpose.

This seems reasonable. The scoping looks appropriate; too. I'm slightly concerned that it might result in non-reproducible builds (as a user may make changes to their local hackage, necessary for a package to compile, and other users will not be able to compile that package) but I don't think that is changed by this PR.

> References to the documentaion about timestamp/expiry logic

Super helpful — thanks.

Shall we add [The hackage security readme](https://github.com/haskell/hackage-security/blob/master/README.md), too, which explains the detail (timestamp ensures that the hackage snapshot as a whole is up to date).

Can you edit the timestamp and have the hackage snapshot work, in current stack/cabal-installl?

> Honestly I do not really understand the motives behind these restrictions about timestamps.

I don't understand the restrictions. I think the the link I've put just above might help — by my reading, it's for cache invalidation (including preventing 'freeze attacks').

> If I understand correctly, any disruption in hackage update process will result to users not being able to install haskell stack.

If by 'hackage update process' you mean any disruption to you updating your copy, then you're not quite correct. Stack is aware of hackage mirrors, which is useful on the uncommon occasions that the main hackage mirror is down (it used to use the fpcomplete mirror by default — I believe that's been changed, now).

The update may fail because of local connectivity problems. In that case, will stack (or cabal-install) use the cached version, or refuse because it has expired?

If by 'hackage update process' you mean that hackage itself is no longer updated, then yes there will be problems, but I would not expect them to differ from those with other similar public services.

> The PR already has an update to the documentation - see the change in the `doc/yaml_configuration.md` file. I could document it a bit more…

Yes, please.

> It would be difficult to make proper testcases - I would certainly need help with this.

If updating the timestamp manually fixes the issue (and indeed triggers it) that's a way to proceed.

Issue #4536, opened 2019-01-25
Script takes ~13x longer to start when comparing development version of stack with 1.9.3


mgsloan:
### Steps to reproduce

Contents of `~/tmp/hmm.hs`:

```haskell
#!/usr/bin/env stack
-- stack --resolver lts-13.0 script

main :: IO ()
main = putStrLn "hmm
```

Development version of stack:

```
mgsloan@treetop:~/tmp$ $(which stack) --version
Version 1.10.0, Git revision 570172b8ca0d5fd4a382abadb2951493616b0387 (6826 commits) x86_64 hpack-0.31.1
mgsloan@treetop:~/tmp$ time ./hmm.hs
hmm

real    0m3.513s
user    0m3.299s
sys     0m0.251s
```

I also had to wait quite a while for it to parse ~100,000 cabal files (but that is not included in this particular run)

Stack-1.9.3:

```
mgsloan@treetop:~/tmp$ $(which stack) --version                                                               
Version 1.9.3, Git revision 40cf7b37526b86d1676da82167ea8758a854953b (6211 commits) x86_64 hpack-0.31.1       
mgsloan@treetop:~/tmp$ time ./hmm.hs                                                                          
hmm                                                                                                           
                                                                                                              
real    0m0.281s
user    0m0.260s
sys     0m0.047s
```

silky:
i think this isn't limited to just the `script` option; it's also slow just thinking about building, but failing:

```
03:53 PM noon ∈ frontend (master*) ♫ time stack build
No package revision found for: PackageName "diagrams-miso"
CallStack (from HasCallStack):
  error, called at src/Stack/Build/ConstructPlan.hs:434:42 in stack-1.10.0-HaI8Xr4wHFrAFWXvSRDn8R:Stack.Build.ConstructPlan
stack build  1.92s user 0.09s system 103% cpu 1.956 total
03:54 PM noon ∈ frontend (master*) ♫ stack --version
Version 1.10.0, Git revision 78790810649eaca7ac020d1acb9b77eab98c60ec (dirty) x86_64 hpack-0.31.1
```

mgsloan:
@silky Yep, seems to be so.  When building stack with itself, I see a *much* larger delay before the first line of output than when using 1.9.3

qrilka:
Some of the noticed slowdown gets addressed in #4557 and #4562 improves speed of `stack build`.
Also #4550 should improve the situation

mgsloan:
[...]

PR #4624, opened 2019-03-14
[WIP]Caching exposed modules for Hackage packages in Pantry DB

qrilka:

Tested manually.

### DIFF
--- a/subs/pantry/package.yaml
+++ b/subs/pantry/package.yaml
@@ -62,6 +62,7 @@ dependencies:
 - resourcet
 - rio-prettyprint
 - mtl
+- extra
 
 # FIXME remove when we drop store
 - integer-gmp
--- a/subs/pantry/src/Pantry/Storage.hs
+++ b/subs/pantry/src/Pantry/Storage.hs
@@ -47,10 +47,15 @@ module Pantry.Storage
   , loadCabalBlobKey
   , hpackToCabal
   , countHackageCabals
+  , storeExposedModules
+  , loadHasCachedExposedModules
+  , storeHasCachedExposedModules
+  , loadHackageExposedModules
 
     -- avoid warnings
   , BlobId
   , HackageCabalId
+  , HackageExposedModulesId
   , HackageTarballId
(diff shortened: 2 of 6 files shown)


qrilka:
Fixes #4536

qrilka:
@snoyberg could you take a look at this variant? On my machine for `stack script --resolver=lts-13.0 hmm.hs` from #4536 I get the following:

| version | time|
|-----------|--------|
| `master` | 4.4s|
| this PR, migrated, not cached | 5.4s|
| this PR, cached modules | 1.2s |
| 1.9.3 | 0.4s |

From logs it looks like lock files could give a bit more improvement and probably we could get something like 2x slowdown comparing to the current stable version. Also I have an idea to store module list as a blob (maybe even in `hackage_cabal` table) - this could save us some time by not doing some lookups.
Do you think it make sense to explore that option?

subs/pantry/package.yaml
```diff
@@ -62,6 +62,7 @@ dependencies:
 - resourcet
 - rio-prettyprint
 - mtl
+- extra
```
snoyberg: I think we've been trying to avoid adding dependencies on these kinds of packages.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -134,8 +140,17 @@ HackageCabal
...
     tree TreeId Maybe
+    -- If available should be equal to True meaning that
+    -- exposed modules were saved in HackageExposedModules
+    hasExposedModules Bool Maybe
     UniqueHackage name version revision
 
+-- Cache of modules exposed by a Hackage package
+HackageExposedModules
+    cabal BlobId
+    moduleName P.ModuleNameP
```
snoyberg: Since there are likely to be a lot of duplicated module names, this would probably benefit from normalization, the same way we normalize package names. How about:

```
    moduleName ModuleId

Module
    name P.ModuleName
    UniqueModule name
```

Or something like that?

PR #4625, opened 2019-03-14
Bring cabal-install executable into $PATH for integration test

psibi:
This will fix integration tests (like `stack-init-solver`)
which depends on the cabal executable.

I tested that this PR works on a separate CI job here:
https://dev.azure.com/psibi2000/stack-issue-debug/_build/results?buildId=737

### DIFF
--- a/.azure/azure-nightly-template-linux.yml
+++ b/.azure/azure-nightly-template-linux.yml
@@ -34,6 +34,9 @@ jobs:
       export PATH=$HOME/.local/bin:$PATH;
       set -ex
       stack install cabal-install
+      # FIXME: Note that cabal-install might not be required once
+      # https://github.com/commercialhaskell/stack/issues/4410 get's
+      # fixed.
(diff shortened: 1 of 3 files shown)


qrilka:
@psibi regarding the solver command we have #4410 - it's possible that we'll just remove this command

psibi:
@qrilka Thanks for the heads up. I think we can remove the associated tests if we agree on to remove the `solver` option. As of now, I think it should be okay to merge this PR for making the integration tests pass. What do you think ?

Also will the `stack init --solver` option be also removed as part of #4410 ?

qrilka:
@psibi I didn't look much into `solver` yet, I think it's just incompatible with the latest Cabal versions, I plan to look into it soon and will check `stack init --solver` as well. Can't tell you more than that at the moment.

psibi:
Merging this as the build failure seems to be a sporadic timeout issue.

PR #4632, opened 2019-03-18
Make dot and ls dependencies not require GHC

### LINKED ISSUE #4390: Simple commands should work without ghc installed

snoyberg:
The previous approach in #4405 would no longer work, due to @qrilka's explanation in https://github.com/commercialhaskell/stack/pull/4405#issuecomment-473820049. This implementation adds a new flag `--global-hints` which allows the user to bypass the need for an installation of GHC. We could arguably flip the default to not requiring GHC to be installed.

This will close #4390.

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -98,12 +98,6 @@ instance HasProcessContext Ctx where
     processContextL = configL.processContextL
 instance HasBuildConfig Ctx where
     buildConfigL = lens ctxBuildConfig (\x y -> x { ctxBuildConfig = y })
-    {-
-instance HasSourceMap Ctx where
-    sourceMapL = envConfigL.sourceMapL
-instance HasEnvConfig Ctx where
-    envConfigL = lens ctxEnvConfig (\x y -> x { ctxEnvConfig = y })
-    -}
 
 -- | Read @<package>.buildinfo@ ancillary files produced by some Setup.hs hooks.
 -- The file includes Cabal file syntax to be merged into the package description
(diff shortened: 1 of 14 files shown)


src/Stack/Package.hs
```diff
@@ -96,9 +96,14 @@ instance HasPantryConfig Ctx where
...
     processContextL = configL.processContextL
-instance HasBuildConfig Ctx
+instance HasBuildConfig Ctx where
+    buildConfigL = lens ctxBuildConfig (\x y -> x { ctxBuildConfig = y })
+    {-
+instance HasSourceMap Ctx where
+    sourceMapL = envConfigL.sourceMapL
 instance HasEnvConfig Ctx where
     envConfigL = lens ctxEnvConfig (\x y -> x { ctxEnvConfig = y })
+    -}
```
qrilka: Did you want to remove this commented out code?

snoyberg: Good catch!

PR #4640, opened 2019-03-19
Fix integration test for #4101

snoyberg:
Running the integration tests on Azure at:

https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=788


snoyberg:
This fixed the integration test that was failing before, although some others are still failing on Windows and Linux. Merging, we can address those separately.

PR #4652, opened 2019-03-25
Use GHC_ENVIRONMENT to resolve module name ambiguity in doctests

qrilka:

Tested manually on an intermediate Stackage snapshot (`lens` vs `lenz` ambiguity in `language-nix` tests)
This overrides #4533 and makes  https://github.com/sol/doctest/pull/218 unnecessary

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE TupleSections         #-}
 -- | Perform a build
 module Stack.Build.Execute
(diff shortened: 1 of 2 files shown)


src/Stack/Build/Execute.hs
```diff
@@ -1809,22 +1810,39 @@ singleTest topts testsToRun ac ee task installedMap = do
...
-                let usePackageIds pc = modifyEnvVars pc $ \envVars ->
-                      Map.insert "HASKELL_PACKAGE_IDS" (T.unwords packageIds) envVars
-                menv <- liftIO $ usePackageIds =<< configProcessContextSettings config EnvSettings
+                let setEnv f pc = modifyEnvVars pc $ \envVars ->
+                      Map.insert "GHC_ENVIRONMENT" (T.pack f) envVars
+                    fp = toFilePath $ eeTempDir ee </> $(mkRelFile "test-ghc-env")
+                    snapDBPath = toFilePathNoTrailingSep (bcoSnapDB $ eeBaseConfigOpts ee)
+                    localDBPath = toFilePathNoTrailingSep (bcoLocalDB $ eeBaseConfigOpts ee)
+                    ghcEnv = mconcat $
+                        "clear-package-db\n" :
```
snoyberg: It's slightly more efficient to bypass the `mconcat` and do `"foo" <> "bar" <> ... <> foldMap f rest`, but that's not terribly important here.

qrilka: I'll fix it anyway

qrilka:
@snoyberg I think `HASKELL_PACKAGE_SANDBOX`(`ES`) is also aimed at `doctest` and probably could be removed. Also with that change the code in https://github.com/sol/doctest/blob/master/src/PackageDBs.hs could be needed only for backward compatibilty. But let's do it in a separate PR (or maybe good cleanup ticket for a newcomer?).

src/Stack/Build/Execute.hs
qrilka: Done in 99059212

Issue #4567, opened 2019-02-05
Custom GHC build


hsyl20:
I have some code that depends on a custom patched GHC. Currently my options to allow other people to build this code are:

1) Create a GHC bindist. Make it available somewhere online. Add this kind of configuration into stack.yaml:

```yaml
setup-info:
  ghc:
     linux64-custom-bugfix-tinfo6:
        8.6.3:
           url: "http://myserver/downloads/ghc-8.6.3-x86_64-bugfix-linux.tar.xz"

compiler: ghc-8.6.3
ghc-variant: bugfix
allow-newer: true
```

2) Give instructions to user on how to build the custom GHC and tell them to use the following Stack configuration updated for their system:

```yaml
extra-path:
- /home/login/repo/ghc/_build/stage1/bin

system-ghc: true
compiler: ghc-8.7.20190131
allow-newer: true
```

Both options are not satisfying.

What I would like instead is to be able to specify in stack configuration:
```yaml
compiler:
   git: https://github.com/me/ghc.git
   commit: 4b00a96cee23ba0bf658834709703368b02f7e30
   flavour: devel2
```

This would:
1) download GHC from git (similarly to an extra-dep).
2) build it using: ./hadrian/build.stack.sh -c -j --flavour=devel2 binary-dist
3) install the resulting bindist into: STACK_ROOT/programs/ARCH/ghc-4b00a96cee23ba0bf658834709703368b02f7e30-devel2/
4) use this GHC as a compiler

dbaynard:
Hi @hsyl20, thanks for the suggestion.

> Both options are not satisfying.

I'd like to know the pros and cons of each of the options — they aren't obvious to me.

Would you be interesting in contributing and maintaining a solution?

borsboom:
One thing to keep in mind is that building GHC from source requires a lot extra stuff installed on the developer's machine, so it wouldn't work "out of the box".  We've considered supporting building GHC from source in the past, but decided against it for this reason.

hsyl20:
@dbaynard 

Solution 1:
* Pro: users of my package only have to type "stack build"
* Con: I need to build and to provide bindists for different architectures/systems which requires *a lot* of time and access to various platforms

Solution 2:
* Pro: I don't have to provide any GHC bindist
* Con: users of my package have to build GHC themselves and to modify stack.yaml accordingly. We could script it but it is still worst than just typing "stack build"

[...]

Issue #4637, opened 2019-03-18
build GHC from source if bindist not available


notestaff:
For a platform for which only older bindists exist, if a project's resolver requires a newer ghc, it would be good if stack could (have the option to) download ghc sources and build ghc from source, using the older available bindist for bootstrapping.

mihaimaruseac:
That would be an improvement, but I would also want a warning be presented to the user and then have the user choose to do the download,compile&install.

notestaff:
Right, so where right now it just fails for lack of needed bindist, it would print “Rerun with —build-ghc-from-source if you want to do that”.

notestaff:
Maybe, eventually there could also be an option to upload the built bindist to some shared repository, from which others could download it?

mihaimaruseac:
That would be nice, but will open a can of worms labeled security and trust.

snoyberg:
Issue #4567 also asks for the ability to build GHC from source. I'm leaning towards wishlisting both of these issues until someone steps up to contribute code to make this happen.

snoyberg:
There's a PR too build from source at #4655. If and when that lands, I could imagine providing an argument to stack setup which would perform a source build.

snoyberg:
#4655 has been merged to master, though it doesn't completely address this issue. I'm moving this request to [this wishlist](https://github.com/commercialhaskell/stack/wiki/Wishlist), if you're interested in working on this (or anyone else for that matter), please add a comment.

barracuda156:
@snoyberg How would you bypass a need for existing GHC? AFAIK it does not build with any compilers bootstrappable from C (like nhc98).

mpilgrem:
@barracuda156, based on this [GHC wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/building/preparation/tools) (under 'ghc'), I understand your question is equivalent to: how to cross-compile GHC. This other [GHC wiki page](https://gitlab.haskell.org/ghc/ghc/-/wikis/building/cross-compiling) is devoted to that subject. I think cross-compiling GHC is (at least, currently) outside of the scope of the Stack project.

hasufell:
GHCup tries to support cross compilation. But the GHC build system is a constantly moving target and far from convenient for the casual end user.

barracuda156:
[...]

PR #4655, opened 2019-03-25
Build GHC from source (#4567)

hsyl20:
This patch adds the experimental support for building GHC from source. We can now specify a specific GHC to build from source in `stack.yaml` with the following syntax:

```
compiler-build:
   git: GHC_GIT_REPO
   commit: COMMIT
   flavour: FLAVOUR
```

In the setup phase Stack uses Hadrian to build a GHC binary distribution
with the specified flavour and installs it into
STACK_ROOT/programs/ARCH/ghc-git-COMMIT-FLAVOUR.

The built compiler can be used with the following configuration:

```
compiler: ghc-git-COMMIT-FLAVOUR
```

As GHC may rely on unreleased version of the global packages (e.g.
Cabal), you may have to add some extra-deps as follows:

```
extra-deps:
- git: GHC_GIT_REPO
  commit: COMMIT
  subdirs:
    - libraries/Cabal/Cabal
    - libraries/...
```

* [X] A test has been added but it takes a long time to run as it builds GHC. [Also note that currently GHC head builds non-installable bindists, cf [GHC#16498](https://gitlab.haskell.org/ghc/ghc/issues/16498)]

### DIFF
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -450,6 +450,80 @@ compiler: ghcjs-0.1.0.20150924_ghc-7.10.2
 compiler-check: match-exact
 ```
 
+#### Building GHC from source (experimental)
+
+(Since 2.0)
+
+Stack supports building the GHC compiler from source. The version to build and
+to use is defined by a a Git commit ID and an Hadrian "flavour" (Hadrian is the
+build system of GHC) with the following syntax:
+
+```yaml
+compiler: ghc-git-COMMIT-FLAVOUR
+```
+
+In the following example the commit ID is "5be7ad..." and the flavour is
+"quick":
+
+```yaml
+compiler: ghc-git-5be7ad7861c8d39f60b7101fd8d8e816ff50353a-quick
+```
+
+By default the code is retrieved from the main GHC repository. If you want to
+select another repository, set the "compiler-repository" option:
+
+```yaml
+compiler-repository: git://my/ghc/repository
+# default
+# compiler-repository: https://gitlab.haskell.org/ghc/ghc.git
+```
+
+Note that Stack doesn't check the compiler version when it uses a compiler built
+from source. Moreover it is assumed that the built compiler is recent enough as
+Stack doesn't enable any known workaround to make older compilers work.
+
+Building the compiler can take a very long time (more than one hour). Hint: for
+faster build times, use Hadrian flavours that disable documentation generation.
+
+#### Global packages
+
+The GHC compiler you build from sources may depend on unreleased versions of
+some global packages (e.g. Cabal). It may be an issue if a package you try to
+build with this compiler depends on such global packages because Stack may not
+be able to find versions of those packages (on Hackage, etc.) that are
+compatible with the compiler.
+
+The easiest way to deal with this issue is to drop the offending packages as
+follows. Instead of using the packages specified in the resolver, the global
+packages bundled with GHC will be used.
+
+```yaml
+drop-packages:
+- Cabal
+- ...
+```
+
+Another way to deal with this issue is to add the relevant packages as
+`extra-deps` built from source. To avoid mismatching versions, you can use
+exactly the same commit id you used to build GHC as follows:
+
+```
+extra-deps:
+- git: https://gitlab.haskell.org/ghc/ghc.git
+  commit: 5be7ad7861c8d39f60b7101fd8d8e816ff50353a
+  subdirs:
+    - libraries/Cabal/Cabal
+    - libraries/...
+```
+
+#### Bootstrapping compiler
+
+Building GHC from source requires a working GHC (known as the bootstrap
+compiler). As we use a Stack based version of Hadrian (`hadrian/build.stack.sh` in
+GHC sources), the bootstrap compiler is configured into `hadrian/stack.yaml` and
+fully managed by Stack.
+
+
 ### ghc-options
 
 (Since 0.1.4)
--- a/src/Stack/Script.hs
+++ b/src/Stack/Script.hs
@@ -254,6 +254,8 @@ allExposedModules gpd = do
       checkCond (PD.Impl compiler range) = case curCompiler of
         ACGhc version ->
           pure $ compiler == GHC && version `withinRange` range
+        ACGhcGit {} ->
+          pure $ compiler == GHC
         ACGhcjs version _ghcVersion ->
           pure $ compiler == GHCJS && version `withinRange` range
       -- currently we don't do flag checking here
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -728,12 +728,6 @@ buildGhcFromSource getSetupInfo' installed (CompilerRepository url) commitId fla
          -- RIO.Process doesn't wrap process' "shell".
          -- Instead we use "proc" with the "sh" command
          proc "sh" hadrianArgs runProcess_
-           `catch` \(e :: ExitCodeException) -> do
-              let dispLBS = displayBytesUtf8 . BL8.toStrict
-              logDebug ("Hadrian failed with " <> displayShow (eceExitCode e))
-              logDebug ("stdout: " <> dispLBS (eceStdout e))
-              logDebug ("stderr: " <> dispLBS (eceStderr e))
-              throwM e
 
          -- find the bindist and install it
          bindistPath <- parseRelDir "_build/bindist"
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -78,6 +78,7 @@ import              Path.CheckInstall (warnInstallSearchPathIssues)
 import              Path.Extra (toFilePathNoTrailingSep)
 import              Path.IO hiding (findExecutable, withSystemTempDir)
 import              Prelude (until)
+import qualified    Pantry
 import qualified    RIO
 import              RIO.List
 import              RIO.PrettyPrint
@@ -512,29 +513,15 @@ ensureCompiler sopts = do
                                         Just x -> return x
                                         Nothing -> throwString $ "MSYS2 not found for " ++ T.unpack osKey
                                 let tool = Tool (PackageIdentifier (mkPackageName "msys2") version)
-                                Just <$> downloadAndInstallTool (configLocalPrograms config) si info tool (installMsys2Windows osKey)
+                                Just <$> downloadAndInstallTool (configLocalPrograms config) info tool (installMsys2Windows osKey si)
                             | otherwise -> do
                                 logWarn "Continuing despite missing tool: msys2"
                                 return Nothing
                 _ -> return Nothing
 
-
-        -- If we need to install a GHC or MSYS, try to do so
-        -- Return the additional directory paths of GHC & MSYS.
-    (compilerTool, mmsys2Tool, compilerBuild) <-
-      case msystem of
-        Just system | canUseCompiler system -> do
-            -- Have the right ghc, may still need msys
-            mmsys2Tool <- getMmsys2Tool
-            return (Left system, mmsys2Tool, CompilerBuildStandard)
-        _ -> do
-
-            -- Install GHC
-            ghcVariant <- view ghcVariantL
+    let installGhcBindist installed = do
             config <- view configL
-            let localPrograms = configLocalPrograms config
-            installed <- listInstalled localPrograms
-
+            ghcVariant <- view ghcVariantL
             possibleCompilers <-
                     case whichCompiler $ wantedToActual wanted of
                         Ghc -> do
@@ -591,9 +578,27 @@ ensureCompiler sopts = do
                             (soptsStackYaml sopts)
                             suggestion
 
-            -- Install msys2 on windows, if necessary
-            mmsys2Tool <- getMmsys2Tool
-            return (Right compilerTool, mmsys2Tool, compilerBuild)
+            return (Right compilerTool, compilerBuild)
+
+    -- If we need to install a GHC, try to do so
+    -- Return the additional directory paths of GHC.
+    (compilerTool, compilerBuild) <-
+      case msystem of
+        Just system | canUseCompiler system -> return (Left system, CompilerBuildStandard)
+        _ -> do
+            -- List installed tools
+            config <- view configL
+            let localPrograms = configLocalPrograms config
+            installed <- listInstalled localPrograms
+            logDebug $ "Installed tools: \n - " <> mconcat (intersperse "\n - " (map (fromString . toolString) installed))
+            case soptsWantedCompiler sopts of
+               -- shall we build GHC from source?
+               WCGhcGit commitId flavour -> buildGhcFromSource getSetupInfo' installed  (configCompilerRepository config) commitId flavour
+               _ -> installGhcBindist installed
+
+
+    -- Install msys2 on windows, if necessary
+    mmsys2Tool <- getMmsys2Tool
 
     paths <- do
       -- Add GHC's and MSYS's paths to the config.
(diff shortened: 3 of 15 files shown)


snoyberg:
Thanks! A few initial comments

- I see that there's still a CI failure, can you address that?
- How do you specify which version of GHC is used for bootstrapping?
- Can you clarify a bit more about the global packages? Stack tries to assume that global packages are not the same as something on Hackage

hsyl20:
>     * I see that there's still a CI failure, can you address that?

I'm trying to fix it now. However I'm not sure I've done the correct thing in src/Stack/Types/BuildPlan.hs. I've blindly updated the version string but I'm not sure about it.

>     * How do you specify which version of GHC is used for bootstrapping?

I have updated the documentation. But to answer here: the bootstrapping compiler is configured via `hadrian/stack.yaml` in GHC source tree. Hence the commit id also "hashes" the bootstrap compiler used, which is very good for reproducible builds.

>     * Can you clarify a bit more about the global packages? Stack tries to assume that global packages are not the same as something on Hackage

I have tried to improve the documentation on this topic but I don't really understand " Stack tries to assume that global packages are not the same as something on Hackage". The issue we had in #4567 (cf https://github.com/commercialhaskell/stack/issues/4567#issuecomment-461811677) is that Stack was trying to build an older Cabal against a newer GHC while the only working Cabal was still unreleased.

snoyberg:
> I'm trying to fix it now. However I'm not sure I've done the correct thing in src/Stack/Types/BuildPlan.hs. I've blindly updated the version string but I'm not sure about it.

Thanks! I'll review the code in more detail soon.

> I have updated the documentation. But to answer here: the bootstrapping compiler is configured via hadrian/stack.yaml in GHC source tree. Hence the commit id also "hashes" the bootstrap compiler used, which is very good for reproducible builds.

That's even better than I was hoping for!

> I have tried to improve the documentation on this topic but I don't really understand " Stack tries to assume that global packages are not the same as something on Hackage". The issue we had in #4567 (cf #4567 (comment)) is that Stack was trying to build an older Cabal against a newer GHC while the only working Cabal was still unreleased.

I think the issue here is that the Stackage snapshots all specify an override of the Cabal library, which forces Stack to _ignore_ the global Cabal version for custom Setup.hs files. We may be moving away for that for other reasons. There are more details at https://github.com/commercialhaskell/stackage/issues/4425.

In #4671 I've added support for a `drop-packages` configuration, which will drop the Cabal library override and hopefully make this use case work more easily.

src/Stack/Setup.hs
```diff
@@ -669,6 +678,84 @@ ensureCompiler sopts = do
...
+             | osIsWindows = "./hadrian/build.stack.bat"
+             | otherwise   = "./hadrian/build.stack.sh"
+
+         logSticky $ "Building GHC from source with `"
+            <> RIO.display flavour
+            <> "` flavour. It can take a long time (more than one hour)..."
+
+         -- RIO.Process doesn't wrap process' "shell".
+         -- Instead we use "proc" with the "sh" command
+         void (proc "sh" hadrianArgs readProcess_)
```
snoyberg: `readProcess_` here will cause no output to be displayed to the user during the installation, and will take up a large amount of memory. `runProcess_` may be a better choice, perhaps with redirecting `stdout` to `stderr`.

src/Stack/SetupCmd.hs
```diff
@@ -102,9 +102,11 @@ setup SetupCmdOpts{..} wantedCompiler compilerCheck mstack = do
         , soptsSetupInfoYaml = scoSetupInfoYaml
         , soptsGHCBindistURL = scoGHCBindistURL
         , soptsGHCJSBootOpts = scoGHCJSBootOpts ++ ["--clean" | scoGHCJSBootClean]
+        , soptsBuildGHC = configCompilerGitBuild
         }
     let compiler = case wantedCompiler of
             WCGhc _ -> "GHC"
+            WCGhcGit{} -> "GHC"
```
snoyberg: I think it's worth saying `"GHC (built from Git)"` or similar

src/Stack/Types/BuildPlan.hs
```diff
@@ -137,9 +137,9 @@ configuration. Otherwise, we don't cache.
 
 loadedSnapshotVC :: VersionConfig LoadedSnapshot
 #if MIN_VERSION_template_haskell(2,14,0)
-loadedSnapshotVC = storeVersionConfig "ls-v6" "RV5MidVhkF2v-PlFB5-tfpcX5xQ="
+loadedSnapshotVC = storeVersionConfig "ls-v7" "LVOy5Emee44pWXVQYpWbT6TaNy4="
```
snoyberg: This is perfect, thanks. This code will actually be going away pretty soon with some in-flight changes, but this is absolutely the right way to handle it today.

src/Stack/Types/CompilerGitBuild.hs
```diff
@@ -0,0 +1,35 @@
...
+  ( CompilerGitBuild(..)
+  )
+where
+
+import           Stack.Prelude
+import           Data.Aeson.Extended
+import           Data.Text as T
+import           Pantry (Repo)
+
+data CompilerGitBuild = CompilerGitBuild
```
snoyberg: See my comment on the docs above, I'm not following why we need more than the URL and presumably repo type (though I'd be OK with hard-coding Git for now given GHC development practices).

subs/pantry/src/Pantry/Types.hs
```diff
@@ -1638,6 +1638,7 @@ data HpackExecutable
 -- @since 0.1.0.0
 data WantedCompiler
   = WCGhc !Version
+  | WCGhcGit !Text !Text
```
snoyberg: This is honestly the most controversial thing here IMO. I'm not sure if snapshots should have the ability to specify arbitrary GHC commits like this, though I understand the motivation. I'd also recommend replacing `Text` with more type-safe values. In particular, `data Flavour = FlavourQuick` would seem reasonable.

test/integration/tests/build-ghc/Main.hs
```diff
@@ -0,0 +1,29 @@
...
+         [ "compiler-build:"
+         , "  git: https://gitlab.haskell.org/ghc/ghc.git"
+         , "  commit: " ++ commitId
+         , "  flavour: " ++ flavour
+         , ""
+         , "compiler: ghc-git-" ++ commitId ++ "-" ++ flavour
+         ]
+
+       -- Setup the package
+       stack ["setup"]
```
qrilka: I wonder how long does it take to run this test? Probably a couple of hours? Not sure if it will be OK to run on every run of integration tests...

doc/yaml_configuration.md
```diff
@@ -427,6 +427,57 @@ compiler: ghcjs-0.1.0.20150924_ghc-7.10.2
...
+### compiler-build (experimental)
+
+(Since 1.10.0)
+
+Build a GHC compiler from source during the "setup" phase. The version to build
+is defined by a Git commit id and an Hadrian flavour (Hadrian is the build
+system of GHC). For instance:
+
+```yaml
+compiler-build:
```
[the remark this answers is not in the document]
hsyl20: Good idea! Fixed. The only new configuration option is now "compiler-repository" which defaults to the official GHC repo (via HTTPS).

doc/yaml_configuration.md
```diff
@@ -427,6 +427,57 @@ compiler: ghcjs-0.1.0.20150924_ghc-7.10.2
...
+  git: https://gitlab.haskell.org/ghc/ghc.git
+  commit: 5be7ad7861c8d39f60b7101fd8d8e816ff50353a
+  flavour: quick
```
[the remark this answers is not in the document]
hsyl20: That's what is used by Hadrian as you've seen below. But after the recent changes the word only appears in the doc and in Stack source, so it is shouldn't be an issue.

doc/yaml_configuration.md
```diff
@@ -427,6 +427,57 @@ compiler: ghcjs-0.1.0.20150924_ghc-7.10.2
...
+
+#### Global packages
+
+The GHC compiler you build from sources may depend on unreleased versions of
+some global packages (e.g. Cabal). It may be an issue if a package you try to
+build with this compiler depends on such global packages because Stack may not
+be able to find versions of those packages (on Hackage, etc.) that are
+compatible with the compiler.
+
+The easiest way to deal with this issue is to add the relevant packages as
```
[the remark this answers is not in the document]
hsyl20: Done. Not tested yet though.

I was thinking, maybe it would be clearer to have something like this to mean the same thing:

```yaml
extra-deps:
- global: Cabal
```

Because we are not really dropping a package in this case: we drop it from the resolver to have access to the "global" one (i don't like "global" either; maybe "bundled" or something else could be better).

src/Stack/Script.hs
```diff
@@ -254,6 +254,8 @@ allExposedModules gpd = do
       checkCond (PD.Impl compiler range) = case curCompiler of
         ACGhc version ->
           pure $ compiler == GHC && version `withinRange` range
+        ACGhcGit {} ->
+          pure $ compiler == GHC
```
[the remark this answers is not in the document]
hsyl20: I have added a note in the documentation about this.

src/Stack/Setup.hs
```diff
@@ -144,6 +146,8 @@ data SetupOpts = SetupOpts
     -- ^ Alternate GHC binary distribution (requires custom GHCVariant)
     , soptsGHCJSBootOpts :: [String]
     -- ^ Additional ghcjs-boot options, the default is "--clean"
+    , soptsBuildGHC :: !(Maybe CompilerGitBuild)
```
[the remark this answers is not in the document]
hsyl20: Done.

src/Stack/Setup.hs
```diff
@@ -239,6 +243,7 @@ setupEnv needTargets boptsCLI mResolveMissingGHC = do
             , soptsSetupInfoYaml = defaultSetupInfoYaml
             , soptsGHCBindistURL = Nothing
             , soptsGHCJSBootOpts = ["--clean"]
+            , soptsBuildGHC = configCompilerGitBuild config
```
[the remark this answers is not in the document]
hsyl20: Done.

src/Stack/Setup.hs
```diff
@@ -669,6 +678,84 @@ ensureCompiler sopts = do
...
+   , HasProcessContext env
+   , HasConfig env
+   ) => Memoized SetupInfo -> [Tool] -> CompilerGitBuild
+   -> RIO env (Either (ActualCompiler, Arch, Path Abs Dir) Tool, CompilerBuild)
+buildGhcFromSource getSetupInfo' installed (CompilerGitBuild repo flavour) = do
+   config <- view configL
+   let compilerTool = ToolGhcGit (repoCommit repo) flavour
+
+   -- detect when the correct GHC is already installed
+   if compilerTool `elem` installed
```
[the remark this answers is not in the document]
hsyl20: Yes the feature was quite easy to implement given what was already there!

src/Stack/Setup.hs
hsyl20: I use ``runprocess_`` now. It is more verbose given how Hadrian is but I'm quite used to it so I don't now what the users of this feature would expect.

src/Stack/SetupCmd.hs
hsyl20: Done.

src/Stack/Types/BuildPlan.hs
hsyl20: Thanks. It seems to be gone now ;)

src/Stack/Types/CompilerGitBuild.hs
hsyl20: Done. I have hard-coded Git for now.

test/integration/tests/build-ghc/Main.hs
hsyl20: Indeed. I am not sure we should enable this test by default. A GHC build with the "quick" flavour takes ~28min on my Intel 9700k (it took ~2h39 with my previous Intel i7-970).

----

For the record: a lot of time is wasted generating the documentation, so I've proposed a patch (https://gitlab.haskell.org/ghc/ghc/merge_requests/354) that:
1) allows flavours to disable doc generation
2) uses this feature to disable doc generation with the "quickest" flavour

As you can see in the discussion, it wasn't welcome. But a few days after point (1) has been reimplemented by someone else and merged without any discussion (https://gitlab.haskell.org/ghc/ghc/merge_requests/461)...

So now we could define and use a faster flavour for this test. We can host a GHC repository with the additional faster flavour somewhere or alternatively we could try to submit a GHC MR to implement point (2) now that the counter arguments are moot (or alternatively we could propose a new "quickestest" flavour).

(Just as I'm writing this, I've found a bug in Hadrian: when we disable doc generation the binary-dist target fails...)

----

Note that currently Hadrian generates invalid bindists so I'm waiting for https://gitlab.haskell.org/ghc/ghc/merge_requests/651 to be merged to get the test passing... Currently without this patch the installed GHC crashes.

doc/yaml_configuration.md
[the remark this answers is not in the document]
snoyberg: I don't want to add additional syntax for this case.

test/integration/tests/build-ghc/Main.hs
snoyberg: I'll add a feature to `master` to have a "superslow" flag to the integration test suites, and this test can be part of that. Sound reasonable?

subs/pantry/src/Pantry/Types.hs
snoyberg: Controversial in the sense of "this isn't obviously the right place for it," nothing more. It might make more sense to treat these Git-built GHCs like any other GHC instead of treating it as a completely separate compiler. For example, if I build `ghc-8.6.4.23` from some branch, and a Stackage snapshots wants `ghc-8.6.4`, should the Git-built one work for that or not? As you've set it up here, we're treating these things as completely separate compilers.

See #4637 for a related issue which would be impacted by the decision here

hsyl20: Ok I see. There are several issues:

1) The Version type isn't adapted and has deep roots into Cabal etc.: i've thought about unifying ``WCGhc`` and ``WCGhcGit`` by using something like [versionTags](https://www.stackage.org/haddock/lts-13.16/base-4.12.0.0/Data-Version.html#v:versionTags) to indicate the commit and the flavour, but since Cabal 2.0 [we don't have these tags anymore](https://www.stackage.org/haddock/lts-13.16/Cabal-2.4.1.0/Distribution-Make.html#t:Version).

2) Retrieving a meaningful GHC version is hard: only release branches (e.g. "ghc-8.6" branch) ensure that the ABI stays the same and it's just discipline by GHC HQ (we can imagine a custom GHC repo where it is not true). GHCs built from other branches such as master may return meaningless versions (e.g. ghc from "master" returns version "8.9.20190410" where the date is the date when ``configure`` was run...) and the ABI may be broken anytime.

So I don't think it is worth the effort to try to detect when there is a compatibility between the Git built compiler and the Stackage snapshot. Except maybe when there is a perfect match between Version and CommitID+Flavour.

---

The other way around is simpler (from Version to Commit/Flavour) but not trivial either. To implement #4637 we can imagine retrieving the Git tags from the ``compiler-repository``: e.g. to build 8.6.4 from source, we infer that it correspond to tag "ghc-8.6.4-release" (i.e. commit aac18e9..) built with "perf" flavour.

It's not always true though: e.g. 8.6.4 bindist on Windows has been released by mistake while built with the "quick" flavour; we would need to handle this quirk to get the same behavior between bindist/source versions :/ or FP Complete might distribute its own bindists ensured to be built with the "perf" flavour, but this is another can of worms...

So for now I think keeping both compilers separate is better.

snoyberg: OK, that makes sense. To be clear, I definitely didn't have what I thought was a better solution to the problem. Let's stick with this for now. Thank you for clarifying :)

test/integration/tests/build-ghc/Main.hs
snoyberg: Feature added in 4b9f1b14a87f210fd87fc40dbaf1ea108a68aa43. If you merge/rebase with master, you can add `superslow $` in front of the test code.

src/Stack/Setup.hs
```diff
@@ -669,6 +674,91 @@ ensureCompiler sopts = do
...
+             | osIsWindows = "./hadrian/build.stack.bat"
+             | otherwise   = "./hadrian/build.stack.sh"
+
+         logSticky $ "Building GHC from source with `"
+            <> RIO.display flavour
+            <> "` flavour. It can take a long time (more than one hour)..."
+
+         -- RIO.Process doesn't wrap process' "shell".
+         -- Instead we use "proc" with the "sh" command
+         proc "sh" hadrianArgs runProcess_
```
snoyberg: I don't think this will work on Windows, since we're trying to run a batch file, not a shell script. What about instead:

```haskell
proc hadrianCmd hadrianArgs runProcess_
```

And removing the `hadrianCmd` from the args above.

src/Stack/Setup.hs
```diff
@@ -669,6 +674,91 @@ ensureCompiler sopts = do
...
+           `catch` \(e :: ExitCodeException) -> do
```
snoyberg: I think all of this is unnecessary now, since the stdout and stderr will be displayed to the user while the process is running.

src/Stack/Setup/Installed.hs
```diff
@@ -39,17 +39,22 @@ import           RIO.Process
 data Tool
     = Tool PackageIdentifier -- ^ e.g. ghc-7.8.4, msys2-20150512
     | ToolGhcjs ActualCompiler -- ^ e.g. ghcjs-0.1.0_ghc-7.10.2
+    | ToolGhcGit !Text !Text   -- ^ e.g. ghc-git-COMMIT_ID-FLAVOUR
```
snoyberg: Going in the direction of allowing Git-build GHC's to be used on platforms that don't have the relevant bindists: what would you think about including the `Version` of the built compiler here? Then we could theoretically have a setting in `stack.yaml` which says "try to use any Git-built GHC if it's available".

src/Stack/Setup.hs
hsyl20: IIRC it wasn't working this way on Linux: we would need [shell](https://www.stackage.org/haddock/lts-13.16/process-1.6.3.0/System-Process.html#v:shell) but RIO only has [proc](https://www.stackage.org/haddock/lts-13.16/rio-0.1.8.0/RIO-Process.html#v:proc) AFAIK.

snoyberg: Is the shell script in the repo set to be executable?

hsyl20: Yes it is. It fails with "Did not find executable at specified path: ./hadrian/build.stack.sh"

src/Stack/Setup/Installed.hs
hsyl20: > what would you think about including the Version of the built compiler here? 
> Then we could theoretically have a setting in stack.yaml which says "try to use any Git-built GHC if it's available".

I would prefer if we start dealing with reported version in another PR (in the context of #4637 maybe). I like the fact that a ghc-git "version" is almost fully determined by commitID+flavour (it's not nix so it doesn't determine external tools such as GCC but we're close as we even determine the bootstrap compiler). If we start to infer a SemVer version from it, then maybe we will have to infer the integer-library used, the configure flags (e.g. RELEASE=YES/NO), etc. and try to match all this against stack passed options (ghc-variant, ghc-build, compiler version...).

I really think it deserves more thoughts and its own PR ;)

snoyberg: Fair enough, you make a good point 👍

snoyberg:
OK, the only thing remaining is the execution of the script on POSIX vs Windows. Everything else looks great. I'm going to futz with it a bit on my OS X and Windows machines and then merge to master. Thanks for enduring a thorough review process @hsyl20 and for the new feature!

test/integration/tests/build-ghc/Main.hs
hsyl20: Done. I have modified it a little to make ``run-single-test.sh`` work (it doesn't specify a speed)

src/Stack/Setup.hs
snoyberg: Please see 2768bb9b1fa3a73bf2a4903d583b526d64e2e78c. This is a really confusing aspect of the process package, I only figured this out since I've been bitten by it before.

hsyl20:
Thanks Michael for the review! My patch fixing GHC bindists has also been merged into GHC so the test passes (on Linux)!

PR #4660, opened 2019-03-26
Remove redundant word

dadepo:
Removed the redundant word "above" in the sentence: 

> Then run the above stack init command above again and it will succeed.


mattaudesse:
Thanks @dadepo

PR #4665, opened 2019-03-27
Hackage creds are read-only (fixes #2159)

### LINKED ISSUE #2159: Hackage credentials file is world-readable when saved
`saveCreds` (https://github.com/commercialhaskell/stack/blob/bcf73ec04a6d6701f7f2090bc8569e8d680dc72a/src/Stack/Upload.hs#L112) creates `~/.stack/upload/credentials.json` (containing Hackage username and password) without explicit file permissions, which results in the default `umask` being used, which is typically `rw-rw-r`, e.g. on Ubuntu.

This means that when your home directory is also world-readable (on Ubuntu it is, `rwxr-xr-x`), other users can grab your Hackage credentials.

For single user systems this is less problematic but not ideal for e.g. university setups where all users' homes are mounted over NFS.

The fix would be for stack to create this file with `rw-------` permissions (and ideally check this when reading so that upgrades from old versions of stack can notice the problem).
## 

Independently, as a person who generally dislikes any on-disk plaintext passwords, I'd prefer if the default behaviour was not to save the credentials on disk, but use a flag for that.

snoyberg:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -196,6 +196,8 @@ Bug fixes:
 - Fix detection of aarch64 platform (this broke when we upgraded to a newer Cabal version).
 - Docker: fix detecting and pulling missing images with `--docker-auto-pull`, see
   [#4598](https://github.com/commercialhaskell/stack/issues/4598)
+* Hackage credentials are not world-readable. See
+  [#2159](https://github.com/commercialhaskell/stack/issues/2159).
 
 ## v1.9.3
 
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
 -- | Provide ability to upload tarballs to Hackage.
 module Stack.Upload
     ( -- * Upload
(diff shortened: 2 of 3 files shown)


src/Stack/Upload.hs
```diff
@@ -90,12 +98,33 @@ loadCreds config = do
...
+
+  -- Disable writing
+  setFileMode fpTmp 0o400
+
+  -- Make sure the destination file, if present, is writeable
+  void $ tryIO $ setFileMode fp 0o600
+
+  -- And atomically move
+  renameFile fpTmp fp
+
```
nh2: Is there a benefit in `-- Disable writing` on the temp file?

Also, it seems `setFileMode fpTmp 0o600` is unnecessary, as [`openTempFile`](https://www.stackage.org/haddock/lts-13.14/base-4.12.0.0/System-IO.html#v:openTempFile), which `withTempFile` uses, has in its docs:

> The file is created with permissions such that only the current user can read/write it.

Correspondingly, `hClose h0` should also be unnecessary.

ChangeLog.md
```diff
@@ -192,6 +192,8 @@ Bug fixes:
 - Fix detection of aarch64 platform (this broke when we upgraded to a newer Cabal version).
 - Docker: fix detecting and pulling missing images with `--docker-auto-pull`, see
   [#4598](https://github.com/commercialhaskell/stack/issues/4598)
+* Hackage credentials are not world-writable. See
```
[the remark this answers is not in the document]
snoyberg: Yes, fixed, thanks!

src/Stack/Upload.hs
snoyberg: Oh, nice! I didn't notice that. Let me update.

snoyberg: > Is there a benefit in -- Disable writing on the temp file?

I suppose not, I think I saw discussion of permissions 400 being set on the file, but I can't really think of a reason to use that. I'll remove that line too.

snoyberg:
Comments added, thanks for the review @nh2!

PR #4666, opened 2019-03-27
package ids output should be debug level

snoyberg:


src/Stack/Build/Execute.hs
```diff
@@ -1850,7 +1850,7 @@ singleTest topts testsToRun ac ee task installedMap = do
                         "package-db " <> fromString localDBPath <> "\n" <>
                         foldMap (\ghcId -> "package-id " <> RIO.display (unGhcPkgId ghcId) <> "\n")
                             (thGhcId:packageIds)
-                logInfo $  "package ids: " <> RIO.displayShow packageIds
+                logDebug $  "package ids: " <> RIO.displayShow packageIds
```
qrilka:
```suggestion
```
I should have removed it actually before committing

snoyberg: Fixed

PR #4668, opened 2019-03-27
Find GHC on PATH via version suffixes (fixes #2433)

### LINKED ISSUE #2433: Find system ghc via binary names like "ghc-7.10.3"?
In https://github.com/commercialhaskell/stack/issues/2221, we're planning to switch away from using system ghc by default.  We could consider trusting that the correct version is installed and skipping the version check.

It turns out that for non system ghcs it is also potentially possible to skip the version check.  Someone on IRC mentioned that binaries named things like `ghc-7.10.3` exist, and that stack should use these.  Skipping the check should save a bit of execution time, potentially something like a tenth of a second.

This would mean not gracefully handling cases like https://github.com/commercialhaskell/stack/issues/2430 .  One option would be to perform such sanity checks after a failing build, if the Cabal errors for these cases are insufficient.  This way, the user gets feedback quickly.  For failing builds, it will wait a bit longer before exiting, to run the sanity checks.

snoyberg:
If you enable system-ghc, and have multiple versions of GHC available,
Stack will now find the appropriate one based on the version-suffixed
name.

Note: the code in Stack.Setup is _really_ difficult to follow at this
point. This PR makes the situation a bit worse, to avoid making this
diff larger than it has to be. A later PR to clean up Stack.Setup would
be a great idea.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -149,6 +149,10 @@ Other enhancements:
 * Warn when a Docker image does not include a `PATH` environment
   variable. See
   [#2472](https://github.com/commercialhaskell/stack/issues/2742)
+* When using `system-ghc: true`, Stack will now find the appropriate GHC
+  installation based on the version suffix, allowing you to more easily switch
+  between various system-installed GHCs. See
+  [#2433](https://github.com/commercialhaskell/stack/issues/2433).
 
 Bug fixes:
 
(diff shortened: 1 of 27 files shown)


fommil:
Excellent! What is the failure mode if no such `ghc-<version>` binary is discovered?

snoyberg:
Same behavior as today essentially:

* If `install-ghc` is enabled, it will download/setup GHC, failing if that's not possible
* If it's disabled, it will give an error message about GHC not being available

PR #4671, opened 2019-03-30
Add drop-packages to stack.yaml #4567

snoyberg:


src/Stack/Config.hs
```diff
@@ -535,8 +535,10 @@ loadBuildConfig = do
       map (second dpLocation) deps0
 
     let packages1 = Map.fromList packages0
-        snPackages = snapshotPackages snapshot `Map.difference` packages1
+        snPackages = snapshotPackages snapshot
+          `Map.difference` packages1
           `Map.difference` Map.fromList deps0
+          `Map.difference` Map.fromSet (const ()) (projectDropPackages project)
```
qrilka: probably it will make code a bit less uniform but what about `Map.withoutKeys`? using it will not require an extra map creation step

snoyberg: I wasn't aware of that function, thanks!

PR #4670, opened 2019-03-29
Snapshot to sourcemap

qrilka:
Please include the following checklist in your PR:

* [ ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [ ] The documentation has been updated, if necessary.

2 integration tests fail. That was filed as #4669 

Fixes #4409 
Fixes #4410 

I didn't remove `solver` from documentation

PR #4674, opened 2019-03-31
Remove use of 'store' and replace with 'persistent' for caches (closes #4214)

### LINKED ISSUE #4214: Drop store usage
We're moving over to SQLite for caching storage. This is more efficient, and will bypass issues of cache invalidation each time we make minor changes to the codebase. It will also make it easier to build stack on ARM, where store's memory usage is a problem.

At the end of this: we should be able to drop all orphan instances in Stack (search for `no-warn-orphans`).

borsboom:
A few questions/concerns:

- [x] How forwards compatible should we be?  Currently, if the way values like cache keys are persisted changes in the future, old Stack versions might noisily fail to load them rather than just ignoring them.  This is mitigated by the fact that PersistField instances are hand-written so the serialization won't change unintentionally, but we may find we need to change them in the future.  This is pretty easy to fix (just means an additional constructor for "unknown" values in a few sum types).

- [x] Currently this uses a file named `caches.sqlite3`.  Might this be used to store data other than caches in the future?  If so, we might want to change that.

- [x] This should be merged after https://github.com/commercialhaskell/stack/pull/4670, since that removes the use of `LoadedSnapshot`.  This code just replaces the the code that loads the snapshot from the cache with recreating the data every time, so Stack still works but startup is slower.

- [x] This does not try to clean up cache entries that are no longer relevant (e.g. because the package's directory isn't there). However I've left a path open to implementing that later by ensuring that cache keys can all be "decoded" back to their constituent components, so it would be relatively easy to run through the tables and check for existence of the directories they reference if needed.

- [x] There's some duplication of database initialization code and exception text from Pantry (e.g. `Stack.PersistentTH.initCacheStorage` is very similar to `Pantry.Storage.initStorage`).  We could generalize this and have the same code used in both places.  Would it make sense to have the general version in `pantry`, or extract that to a separate package that's depended on by both `stack` and `pantry`?


snoyberg:
> How forwards compatible should we be? Currently, if the way values like cache keys are persisted changes in the future, old Stack versions might noisily fail to load them rather than just ignoring them. This is mitigated by the fact that PersistField instances are hand-written so the serialization won't change unintentionally, but we may find we need to change them in the future. This is pretty easy to fix (just means an additional constructor for "unknown" values in a few sum types).

More forward compat sounds like a nice touch here.

> Currently this uses a file named caches.sqlite3. Might this be used to store data other than caches in the future? If so, we might want to change that.

The other data I currently have planned for storage here could arguably be called cache also, but I can imagine putting other things there in the future too. Let's call it something more generic.

>  There's some duplication of database initialization code and exception text from Pantry (e.g. Stack.PersistentTH.initCacheStorage is very similar to Pantry.Storage.initStorage). We could generalize this and have the same code used in both places. Would it make sense to have the general version in pantry, or extract that to a separate package that's depended on by both stack and pantry?

Let's keep it in pantry for now

src/Stack/PersistentTH.hs
```diff
@@ -0,0 +1,369 @@
...
+            (configCacheDeps new)
+        updateSet
+            ConfigCacheComponent
+            ConfigCacheComponentParent
+            parentId
+            ConfigCacheComponentValue
+            (maybe Set.empty configCacheComponents mold)
+            (configCacheComponents new)
+
+-- | Mark 'ConfigCache' as inactive in the database.
```
snoyberg: Perhaps a short comment explaining why we deactivate instead of delete?

snoyberg:
#4670 is now merged to master

borsboom:
> More forward compat sounds like a nice touch here.

Turns out this wasn't an issue, since the code only ever loads records for which it can construct a key, so by definition that key can be read back.  It is something to keep in mind for any future code that uses the database, though (for example, if we implement anything that selects all records from a table, we should make sure it handles failure to load gracefully).

> The other data I currently have planned for storage here could arguably be called cache also, but I can imagine putting other things there in the future too. Let's call it something more generic.

I've renamed to to `stack.sqlite3`.  This way it's clear which package the data belongs to (since Pantry uses `pantry.sqlite3`).

> > Would it make sense to have the general version in pantry, or extract that to a separate package that's depended on by both stack and pantry?
>
> Let's keep it in pantry for now

Alright, I've created a new `Pantry.SQLite` module that exports the shared bits.

> #4670 is now merged to master

I've merged master and fixed the conflicts.  

However, now I'm sure there will be conflicts with #4688 (since I moved most of the code that PR changes to a different module).

snoyberg:
> However, now I'm sure there will be conflicts with #4688 (since I moved most of the code that PR changes to a different module).

I agree, but I'll deal with those. Let's get this merged.

PR #4678, opened 2019-04-01
Always exact configuration

snoyberg:

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -76,6 +76,9 @@ Behavior changes:
   never been used yet, and there are no plans to implement signature
   verification.
 
+* Always use `--exact-configuration` (should mostly be a non-user-visible
+  enhancement).
+
 Other enhancements:
 
 * Defer loading up of files for local packages. This allows us to get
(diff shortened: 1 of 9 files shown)


qrilka:
@snoyberg What is this option? I don't see it documented and I'm not sure that I understand it completely. How does it relate to bounds checking?

snoyberg:
This flag means that we tell Cabal exactly which package identifiers should be used for each dependency, bypassing any bounds checking Cabal itself might do. The failure you noted in the previous PR was due to Cabal refusing to ignore bounds.

qrilka:
Oh, it's from Cabal. I see now. Probably we should add a hint about it in the changelog entry?

snoyberg:
I included a ChangeLog entry here: https://github.com/commercialhaskell/stack/pull/4678/commits/e06a3e3e3ab8a85e7660aa1a44237d2756ecb16e#diff-e705c8fadf1193ab59443a5e6c8cbe8bR79

qrilka:
I was referring to exactly that entry, my suggestion is to mention that it's a Cabal flag

snoyberg:
Oh, got it! Sure, I'll add that and merge to master

Issue #4471, opened 2018-12-28
SqliteBusy Error appears once in a while


psibi:
### Sqlite Busy error still appears

Related issue: https://github.com/commercialhaskell/stack/issues/4247

### Steps to reproduce

Unfortunately, there are no steps to reproduce it reliable. The Azure pipeline has been able to replicate it periodically:

* https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=198 (Windows)
* https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=185 (Windows)
* https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=171 (Linux)

I was sparsely able to reproduce it twice by running Stack build simultaneously on multiple Haskell projects (on a Linux machine).

### Expected

Not have these errors:

```
SQLite3 returned ErrorBusy while attempting to perform prepare "PRAGMA journal_mode=WAL;": database is locked
```

### Actual

Hitting up with the `ErrorBusy` error.

### Stack version

```
$ stack --version
Version 1.10.0, Git revision f99a25395af25c964f1a517f49f4c31cab2b82a9 (6747 commits) x86_64 hpack-0.31.1
```

### Method of installation

From the source

mihaimaruseac:
Do we have multiple threads accessing SQLite at the same time?

psibi:
Yes, I do think so. There is also this function inside the module `Pantry`: 

```
traverseConcurrently_
traverseConcurrentlyWith_
```

mihaimaruseac:
Then that's the issue. SQLite doesn't support multithreaded access to the database.

qrilka:
@mihaimaruseac why do you think it does not? See e.g. https://www.sqlite.org/threadsafe.html
From a quick glance it looks like `persistent-sqlite` doesn't set threading explicitly and thus serialized mode should be used.

mihaimaruseac:
I guess that's a new development in slqite. In the past it used to not be possible to properly multithread access the database.

nh2:
> Do we have multiple threads accessing SQLite at the same time?

Another important question to ask:

Do we have multiple **processes** accessing SQLite at the same time?

While multi-_threaded_ access isn't a problem for sqlite as linked above, only 1 _process_ can write to a DB at any given time.

From https://www.sqlite.org/faq.html#q5:

[...]

PR #4688, opened 2019-04-02
Take a write lock before using Pantry database #4471

snoyberg:
This makes two basic changes to the code:

* Avoids a connection pool, which forces the SQLite connection to be
closed at the end of each call to withStorage. This seems unfortunate;
we would like to be able to maintain connections. However, SQLite seems
to view such connections as holding a write lock even when not in the
middle of a database transaction.

* Uses an explicit file lock on the two longest-running write
operations: migration and Hackage population. Ideally the locking
mechanism in SQLite would be sufficient for this, but (1) Hackage
population can take a long time, and (2) evidence has shown that it
doesn't behave as we expect it to.

This may be insufficient for preventing the bug described in #4471, in
which case we may need to apply withWriteLock to more calls. However,
I'd like to start off small to avoid the additional overhead.

### DIFF
--- a/subs/pantry/src/Pantry/SQLite.hs
+++ b/subs/pantry/src/Pantry/SQLite.hs
@@ -1,47 +1,93 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Pantry.SQLite
-  ( P.Storage
+  ( Storage (..)
   , initStorage
-  , withStorage
   ) where
 
 import RIO hiding (FilePath)
-import qualified Pantry.Types as P
 import Database.Persist.Sqlite
 import RIO.Orphans ()
 import Path (Path, Abs, File, toFilePath, parent)
 import Path.IO (ensureDir)
-import Data.Pool (destroyAllResources)
-import Pantry.Types (PantryException (MigrationFailure))
+import Pantry.Types (PantryException (MigrationFailure), Storage (..))
+import System.FileLock (withFileLock, withTryFileLock, SharedExclusive (..))
 
 initStorage
   :: HasLogFunc env
   => Text
   -> Migration
   -> Path Abs File -- ^ storage file
-  -> (P.Storage -> RIO env a)
+  -> (Storage -> RIO env a)
   -> RIO env a
 initStorage description migration fp inner = do
   ensureDir $ parent fp
-  bracket
-    (createSqlitePoolFromInfo (sqinfo False) 1)
-    (liftIO . destroyAllResources) $ \pool -> do
-    migrates <- wrapMigrationFailure $ runSqlPool (runMigrationSilent migration) pool
-    forM_ migrates $ \mig -> logDebug $ "Migration executed: " <> display mig
-  bracket
-    (createSqlitePoolFromInfo (sqinfo True) 1)
-    (liftIO . destroyAllResources) $ \pool -> inner (P.Storage pool)
+
+  migrates <- withWriteLock fp $ wrapMigrationFailure $
+    withSqliteConnInfo (sqinfo True) $ runReaderT $
+    runMigrationSilent migration
+  forM_ migrates $ \mig -> logDebug $ "Migration executed: " <> display mig
+
+  inner $ Storage
+    { withStorage_ = withSqliteConnInfo (sqinfo False) . runSqlConn
+    , withWriteLock_ = withWriteLock fp
+    }
   where
     wrapMigrationFailure = handleAny (throwIO . MigrationFailure description fp)
-    sqinfo fk = set extraPragmas ["PRAGMA busy_timeout=2000;"]
-           $ set fkEnabled fk
+
+    sqinfo isMigration
+           = set extraPragmas ["PRAGMA busy_timeout=2000;"]
+
+           -- When doing a migration, we want to disable foreign key
+           -- checking, since the order in which tables are created by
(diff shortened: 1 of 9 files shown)


snoyberg:
I've kicked off integration tests on this PR, @psibi when they've completed do you think you can see if it seems to solve the problem from #4471?

psibi:
Sure, will give it a look once the tests complete.

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -230,32 +233,76 @@ ModuleName
...
+              shouldComplain <- readIORef shouldComplainRef
+              when shouldComplain $ do
+                run $ logWarn "Still waiting on the Pantry database write lock..."
+                loop
+            stopComplaining = writeIORef shouldComplainRef False
+            worker = withFileLock lockFile Exclusive $ const $ do
+              run $ logInfo "Acquired the Pantry database write lock"
+              stopComplaining
+              run inner
+        runConcurrently $ Concurrently complainer
```
lehins: I might be wrong here, but it looks like by using applicative here it can tie up the calling thread unnecessarily for a whole minute. What I mean is whenever `inner` is very quick and finishes early, `complainer` will still be blocked in `threadDelay`. Wouldn't it be better to use Alternative's `<|>` or `race` and make sure `complainer` always calls `loop`, so it gets killed by an exception instead of exiting the loop, i.e.:
```haskell
              when shouldComplain $ do
                run $ logWarn "Still waiting on the Pantry database write lock..."
              loop
```

snoyberg: You're right, though I'd prefer not to use a `race` here to avoid async exceptions. Let me take a stab at another approach...

borsboom:
This can come later, but I think it could be helpful to extract this SQLite database setup code from Pantry and move it to an independent module.  Stack itself will be using this for its own database as well (#4674), and this is a fairly common use case for SQLite databases. I can think of at least one other project I'm working on now that could make use of it right now.

snoyberg:
It looks like the code I've written (and probably the most recent commit on `registerDelay`) is causing a deadlock on Windows. Investigating on my Windows machine now.

lehins:
@snoyberg Just a suggestion, since I don't know if it will make difference, but maybe worth a try avoiding `checkSTM` together with `Alternative`:

```
atomically $ orElse (readTVar delay) (not <$> readTVar shouldStopComplainingVar)
```

snoyberg:
@lehins That version will always return the current value for `delay`, which will always be `False` initially.

lehins:
@snoyberg My bad, I confused it with `TMVar`.

snoyberg:
For some reason, the CI pipeline on Windows is taking a _really_ long time at processing the cabal files from Hackage for running the Pantry test suite. This is surprising, because:

1. The code I've added shouldn't, in principle, affect that. The work is all inside a single large transaction
2. I can't reproduce that problem on my local machine
3. I can't see a slowdown on other OSes

This definitely blocks merging this PR, but I may have to fiddle with things a bit randomly to figure out what exactly is happening here.

snoyberg:
I think I finally figured this out! I was using `runReaderT` instead of `runSqlConn` to run the actions. The latter will begin and commit a transaction, which _massively_ increases the throughput of high I/O actions. The former does not. I don't know why the performance on Windows is so much worse than elsewhere, or why my Windows laptop was so much faster than the Azure Windows machine. Nonetheless, this wasn't only a performance issue, but a correctness issue, since we rely on proper transactions. So hurrah that this got caught.

Once CI passes, I'll kick off integration tests as well.

subs/pantry/src/Pantry/SQLite.hs
```diff
@@ -1,47 +1,93 @@
...
+    withSqliteConnInfo (sqinfo True) $ runReaderT $
+    runMigrationSilent migration
+  forM_ migrates $ \mig -> logDebug $ "Migration executed: " <> display mig
+
+  inner $ Storage
+    { withStorage_ = withWriteLock fp . withSqliteConnInfo (sqinfo False) . runSqlConn
+    , withWriteLock_ = id
+    }
   where
     wrapMigrationFailure = handle (throwIO . MigrationFailure description fp)
```
borsboom: This only catches `SqliteException`, but I noticed that in some cases persistent calls `error`.  For example [if there's an unsafe migration](https://github.com/yesodweb/persistent/blob/master/persistent/Database/Persist/Sql/Migration.hs#L103) that doesn't get caught here and so the error doesn't give any useful advice for an end-user.

snoyberg:
Integration tests are still giving an ErrorBusy... even when I take a lock before every database action! I can only think of one last thing right now, from the WAL docs:

>  When the last connection to a particular database is closing, that connection will acquire an exclusive lock for a short time while it cleans up the WAL and shared-memory files. If a second database tries to open and query the database while the first connection is still in the middle of its cleanup process, the second connection might get an SQLITE_BUSY error. 

I think I'll try adding some logic to persistent-sqlite that, when it detects an ErrorBusy during a `PRAGMA journal_mode=WAL`, it retries a few times before giving up.

snoyberg:
I've attempted to implement this logic in persistent-sqlite:

https://github.com/yesodweb/persistent/commit/efbeabf77124085f3a58b58f73ee1bdbc88502a7

And I'm trying out a commit that uses this:

a85576ec99a90ac6db45a799bebd29ca2777f4ee

Note that I'm going back to using a connection pool and only taking a write lock for migrations and Hackage updates.

PR #4693, opened 2019-04-03
Fix --docker throwing exceptions on expected subcommand failure

borsboom:


borsboom:
It would seem that, at some point, the exception thrown by `sinkProcessStderrStdout` changed without the documentation or downstream code having been updated.  It currently uses `runProcess_`, which uses `checkExitCode`, which throws `ExitCodeException` rather than `ProcessExitedUnsuccessfully`.

The only downstream code that appears to be effected is the Docker support, so I just updated it to use the `ExitCodeException` instead.

PR #4695, opened 2019-04-03
More idiomatic database table keys

borsboom:
This comes a bit late as I hadn't noticed that #4674 was already merged, but this splits the keys into multiple fields as is more idiomatic in a database schema and easier to query, and ends up simplifying the code a bit too.


borsboom:
As expected, this breaks database migration.  Surprisingly, that also seems to cause test failure.  I guess the STACK_ROOT is being cached.

borsboom:
> As expected, this breaks database migration. Surprisingly, that also seems to cause test failure. I guess the STACK_ROOT is being cached.

Added a commit that lets the migrations run (and means any existing records will end up being ignored).

PR #4697, opened 2019-04-04
Inform the user why sudo is necessary in get-stack (fixes #4672)

### LINKED ISSUE #4672: Inform the user why sudo is necessary in get-stack
Currently it looks like `get-stack` prompts the user for `sudo` permissions in three broad cases:

- when attempting to perform filesystem operations related to installing `stack` to a target directory
  - e.g. https://github.com/commercialhaskell/stack/blob/master/etc/scripts/get-stack.sh#L529
- when trying to install some generic, required OS dependencies via `try_install`
  - e.g. https://github.com/commercialhaskell/stack/blob/master/etc/scripts/get-stack.sh#L507
- when trying to install some specifically required OS dependencies
  - e.g. (for apt-get) https://github.com/commercialhaskell/stack/blob/master/etc/scripts/get-stack.sh#L158

In each of these cases it would probably be helpful if the script could inform the user why exactly `sudo` is being requested so that they may better understand the implication of supplying these permissions.

borsboom:
This also adds some extra text to the docs pointing people to the manual installation instructions if they prefer not to use the script.  (ping @jkachmar)

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -18,6 +18,12 @@
 # Make pull requests at:
 # https://github.com/commercialhaskell/stack/blob/master/etc/scripts/get-stack.sh
 #
+# Note that this script will ask for root access using `sudo` in order to use
+# your platform's package manager to install dependencies and to install to
+# `/usr/local/bin`.  If you prefer more control, follow the manual
+# installation instructions for your platform at:
+# https://docs.haskellstack.org/en/stable/install_and_upgrade/
+#
 
 STACK_VERSION="1.9.3"
 HOME_LOCAL_BIN="$HOME/.local/bin"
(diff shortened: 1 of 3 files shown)


etc/scripts/get-stack.sh
```diff
@@ -18,6 +18,12 @@
 # Make pull requests at:
 # https://github.com/commercialhaskell/stack/blob/master/etc/scripts/get-stack.sh
 #
+# Note that this script will ask for root access using `sudo` in order to use
+# your platform's package manager to install dependencies and to install to
+# `/usr/local/bin`.  If you prefer more control, follow the manual
```
phadej: Where `stack upgrade` (which is recommended way to upgrade `stack`) will put new `stack` executable?

> You can install stack by copying it anywhere on your PATH environment variable. We recommend installing in the same directory where stack itself will install executables (that way stack is able to upgrade itself!). 

https://docs.haskellstack.org/en/stable/install_and_upgrade/#path

borsboom: Looks like the docs you quoted are out of date.  `stack upgrade` will overwrite its own executable, wherever it is, with the new version (prompting the user whether it's ok to use `sudo` if that's necessary).

borsboom: I've pushed a commit that removes the out-of-date info.

PR #4701, opened 2019-04-05
Load up user config when using scripts

### LINKED ISSUE #3705: `stack script` ignores `nix:  enable: true` in `~/.stack/config.yaml`
I want to make scripts that work on both nix and non-nix OS without changes

### Steps to reproduce

example-script.hs
```hs
#!/usr/bin/env stack
{- stack
  script
  --resolver lts-9.18
-}

import           System.Environment    (getArgs)

main :: IO ()
main = do
  args <- getArgs
  print args
```

~/.stack/config.yaml
```
nix:
  enable: true
  packages:
    - libcxx
    - icu
    - gcc
    - ncurses
    - zlib
```

```sh
 ~/projects/nuuz/nuuz-chrome-extension/packages/metalib   master  update-index-js.hs index.js
Already downloaded.
Running /home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2.temp/ghc-8.0.2/configure --prefix=/home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2/ in directory /home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2.temp/ghc-8.0.2/ exited with ExitFailure 1

checking for path to top of build tree...
/home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2.temp/ghc-8.0.2/configure: utils/ghc-pwd/dist-install/build/tmp/ghc-pwd-bindist: /bin/bash: bad interpreter: No such file or directory
configure: error: cannot determine current directory


Error: Error encountered while configuring GHC with
         /home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2.temp/ghc-8.0.2/configure --prefix=/home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2/
         run in /home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2.temp/ghc-8.0.2/

The following directories may now contain files, but won't be used by stack:
  - /home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2.temp/
  - /home/bjorn/.stack/programs/x86_64-linux/ghc-8.0.2/

Configuring GHC ...% 
```

But I can resolve this issue by adding `--nix` option
```hs
#!/usr/bin/env stack
{- stack
  script
  --nix
  --resolver lts-9.18
-}
```

### Expected
To work without `--nix`

### Stack version
```
 ✘  ~  stack --version
Version 1.5.1 x86_64
```

Sorry, can't upgrade to latest stack-1.6.3

### Method of installation

* nix

P.S. [without --nix, but with --verbose](https://pastebin.com/rbNnrjZe)

### LINKED ISSUE #3887: `stack script` ignores `local-programs-path` in ~/.stack/config.yaml
### General summary/comments

Also found #3705, so I assume `stack script` ignores ~/.stack/config.yaml completely.

### Steps to reproduce

For example:

1. Edit ~/.stack/config.yaml with a setting for `local-programs-path` to point somewhere else than ~/.stack/programs
2. Run `stack stream-fusion.hs`, where stream-fusion.hs is the attached file

### Expected

Just run the thing after initial setup

### Actual

Re-downloads GHC 8.2.2 and errors (might be machine-specific) while unpacking it into ~/.stack/programs. Also complains with `The following directories may now contain files, but won't be used by stack <some ~/.stack/programs paths>`, which it wouldn't in the first place.

```
$ stack stream-fusion.hs --verbose
Already downloaded.
Unpacking GHC into /$HOME/.stack/programs/x86_64-linux/ghc-nopie-8.2.2.temp/ ...
```

### Stack version

```
$ stack --version
Version 1.6.5, Git revision 24ab0d6ff07f28276e082c3ce74dfdeb1a2ca9e9 (5514 commits) x86_64 hpack-0.20.0
```

### Method of installation

Installed 1.6.3 through `nix-env`, but then did `stack upgrade` and used the downloaded exe for this reproduction (didn't want to wait for the upgrade of my whole distribution).

### LINKED ISSUE #4699: Next set of integration test improvements
Once #4698 is merged, I believe we'll only have two remaining test failures. Both of those are due to the fact that `stack script` is ignoring the global `system-ghc: true` setting. Both of those should be addressed by #3887, which I'll hit soon I hope.

The other minor improvement I want to make is to respect a `PANTRY_CONFIG` environment variable, which overrides the default of `$STACK_ROOT/pantry`. This would allow the cleanup code in `IntegrationSpec` to be _much_ nicer.

snoyberg:
* Fixes #3705 
* Fixes #3887 
* Fixes #4699

### DIFF
--- a/.azure/azure-nightly-template-linux.yml
+++ b/.azure/azure-nightly-template-linux.yml
@@ -37,7 +37,7 @@ jobs:
       # FIXME: Note that cabal-install might not be required once
       # https://github.com/commercialhaskell/stack/issues/4410 get's
       # fixed.
-      stack test --flag stack:integration-tests stack:test:stack-integration-test --interleaved-output
+      stack build --flag stack:integration-tests stack --interleaved-output --exec stack-integration-test
       set +ex
     displayName: Integration Test
   - script: |
--- a/.azure/azure-nightly-template-linux.yml
+++ b/.azure/azure-nightly-template-linux.yml
@@ -37,7 +37,7 @@ jobs:
       # FIXME: Note that cabal-install might not be required once
       # https://github.com/commercialhaskell/stack/issues/4410 get's
       # fixed.
-      stack build --flag stack:integration-tests stack --interleaved-output --exec stack-integration-test
+      etc/scripts/integration-tests.sh
       set +ex
     displayName: Integration Test
   - script: |
--- a/.azure/azure-nightly-template-osx.yml
+++ b/.azure/azure-nightly-template-osx.yml
@@ -39,7 +39,7 @@ jobs:
       # FIXME: Note that cabal-install might not be required once
       # https://github.com/commercialhaskell/stack/issues/4410 get's
       # fixed.
-      stack test --flag stack:integration-tests stack:test:stack-integration-test --interleaved-output
+      stack build --flag stack:integration-tests stack --interleaved-output --exec stack-integration-test
       set +ex
     displayName: Integration Test
   - script: |
--- a/.azure/azure-nightly-template-osx.yml
+++ b/.azure/azure-nightly-template-osx.yml
@@ -39,7 +39,7 @@ jobs:
       # FIXME: Note that cabal-install might not be required once
       # https://github.com/commercialhaskell/stack/issues/4410 get's
       # fixed.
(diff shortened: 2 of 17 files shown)


snoyberg:
And with that, the integration tests finally pass! W00t!

PR #4714, opened 2019-04-09
Prepare to cache GHC info

snoyberg:
This is a refactoring of a bunch of code in Stack.Setup and related modules to make it possible to begin caching GHC information. These changes already improve the behavior of Stack slightly, such as

* Warning on untested or unsupported GHC/Cabal versions
* Avoiding redundant global package checks
* Checking for multiple system GHCs in a sensible order

I'm opening up this PR first for review, and once this is merged, I'll work on the caching logic, which will allow us to bypass a few expensive external process calls on each Stack invocation.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -86,15 +86,15 @@ Behavior changes:
 * Always use the `--exact-configuration` Cabal configuration option when
   building (should mostly be a non-user-visible enhancement).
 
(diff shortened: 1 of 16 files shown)


qrilka:
@snoyberg is it expected to fail integration tests on Windows?

snoyberg:
No, I'll need to look into that

Issue #4476, opened 2018-12-30
Windows terminal 64 bits?


flip111:
After the stack installation on windows is done it's not really clear what terminal should be used.

I'm on windows 10 64 bits.

When i start my git bash (i had git installed) or windows command prompt and run stack it's using mingw32

When i start mingw64.exe from `%APPDATA%\Local\Programs\stack\x86_64-windows` stack can not be found.

It wouldn't hurt to add some explanation here to the manual. In particular i'm confused if everything is actually setup the right way to compile 64 bits programs.

EDIT:
related question. How to get the path to the gcc compiler? i like to compile some external C code. It's better to use GCC shipped here than to ask the users of my library to install gcc separately or visual studio. Also need the make tool and possible some other tools.

mattaudesse:
Hey @flip111 - this is more of a Windows CLI question than something directly specific to `stack`, and the answer will depend on an individual's personal environment + how it's set up (so it's hard to provide any set of steps that'll apply equally well to everyone, although we'd happily welcome a docs PR if you'd like to volunteer).

> After the stack installation on windows is done it's not really clear what terminal should be used.

Whichever one you'd like! I'm not a Windows user so I can't replicate the problem, but it sounds like you just need to update your `PATH` environment variable. Did you follow [these steps?](https://docs.haskellstack.org/en/stable/install_and_upgrade/#path)

[This](https://www.howtogeek.com/118594/how-to-edit-your-system-path-for-easy-command-line-access/) may be helpful for using Window's default terminal.

`bash` terminals generally let you modify your `PATH` via a "dotfile" which runs when you enter an interactive shell session. Some common places to look:
- `/home/user/.bashrc`
- `/home/user/.bash_profile`
- `/home/user/.profile`

There are lots of ways to run `bash` on Windows these days, and they can each behave differently, so I can't say for sure which will apply to your system. Googling how to update the path for your `bash` distro should fetch plenty of hits, but it generally looks something like this in practice:

```bash
# You'd append this to ~/.bashrc:
export PATH="/some/additional/bin:$PATH"
```

I'm going to close since this doesn't seem like a `stack` problem but please feel free to continue the thread until everything's running smoothly. Perhaps someone who's more of a Windows person could weigh in too.

flip111:
[...]

PR #4717, opened 2019-04-09
Create developing_on_windows.md

flip111:
Some things i'm not sure of:

* Should i put something in the changelog ? probably not
* Should i go more into why stack is shipping with msys / what is stack using msys for?
* Is stack actually using msys2 or is it msys? I would have to get back to my windows machine later to check.
* Tools not being able to find itself on path seems more like an issue that needs to be fixed than to be put in documentation. But as of now this seems to be the state it is ... so better document it i guess

Also i wanted to open this section for some more general tips and attention points for using stack on windows and not make it just about pacman. Too avoid scattering later information and make the section more future-proof sorts of.

PR created for https://github.com/commercialhaskell/stack/issues/4476#issuecomment-478665757

### DIFF
--- a/doc/developing_on_windows.md
+++ b/doc/developing_on_windows.md
@@ -1,0 +1,22 @@
+# Developing on windows #
+On windows stack comes with an installation of [msys2](https://www.msys2.org/). Msys2 will be used by stack to provide a unix-like shell for stack so it can do it's work <or something more specific here what stack does with msys ?>. No matter which terminal you choose (cmd.exe, powershell, git bash or any other) you can use this environment too by executing all programs through `stack exec -- program`. This is especially useful if your project needs some additional tools during the build phase. Build tools and other tools can be installed with Pacman. All tools can be found [in the package list](https://github.com/msys2/msys2/wiki/Packages). A [list of commands](https://github.com/msys2/msys2/wiki/Using-packages) that work with Pacman is also available. Just remember that pacman like all other tools should be started with `stack exec -- pacman`.
+
+## Setup.hs ##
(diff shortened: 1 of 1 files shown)


snoyberg:
> Should i put something in the changelog ? probably not

Agreed, not necessary

> Should i go more into why stack is shipping with msys / what is stack using msys for?

Maybe a short mention of the largest original motivator: running `configure` scripts for Haskell packages which use them.

> Is stack actually using msys2 or is it msys? I would have to get back to my windows machine later to check.

`msys2`. I recently was playing with that part of the code base :)

> Also i wanted to open this section for some more general tips and attention points for using stack on windows and not make it just about pacman.

I like that idea!

* * *

Recommendation: how about including a section for "Pacman packages to install for common Haskell packages." This could include things like dependencies of the `text-icu` package, which I've covered previously:

https://stackoverflow.com/a/34542892/369198

flip111:
> Recommendation: how about including a section for "Pacman packages to install for common Haskell packages." This could include things like dependencies of the `text-icu` package, which I've covered previously:

I don't like this too much by principle because i think it's better for packages to declare their own dependencies they need get from the operating system. But i'm not sure if users will visit the website of the package itself and a bit of redundancy doesn't hurt. Since i didn't see it mentioned with `text-icu` what the requirements are for windows (other than the SO answer) i'll just put it in for now.

On a side note i would like a checklist what a package should and could contain. I know there is this https://hackage.haskell.org/upload but i think it leaves out many topics.

snoyberg:
Thanks!

PR #4721, opened 2019-04-12
Cache GHC information in the SQLite database

snoyberg:
This bypasses a few relatively expensive external process executions on
each Stack invocation, namely `ghc --info` and `ghc-pkg dump`.

### DIFF
--- a/src/Stack/Storage.hs
+++ b/src/Stack/Storage.hs
@@ -26,23 +26,33 @@ module Stack.Storage
     , savePrecompiledCache
     , loadDockerImageExeCache
     , saveDockerImageExeCache
+    , loadCompilerPaths
+    , saveCompilerPaths
     ) where
 
 import qualified Data.ByteString as S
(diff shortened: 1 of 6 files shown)


src/Stack/Storage.hs
```diff
@@ -114,6 +123,33 @@ DockerImageExeCache
...
+  globalDbCacheModified Int64
+  info ByteString
+
+  -- This is the ugliest part of this table, simply storing a Show/Read version of the
+  -- data. We could do a better job with normalized data and proper table structure.
+  -- However, recomputing this value in the future if the data representation changes
+  -- is very cheap, so we'll take the easy way out for now.
+  globalDump Text
+
+  UniqueGhcInfo ghcPath
```
borsboom: Maybe call this `UniqueCompilerInfo` to match the table name?

snoyberg: Oh, good point, I'll update and then merge.

Issue #4702, opened 2019-04-05
DISCUSS: Make --interleaved-output the default


snoyberg:
A few different reasons:

1. People have been reporting issues about "frozen" compiles, when in fact it's just very large dependencies that take a long time to build. This output will make it clear that work is happening, just slowly.
2. I've had CI timeouts in the past due to lack of output for 10 minutes. Basically, same as 1, but impatient machines instead of impatient humans
3. It can be informative to see the output as it happens

Downsides:

1. Subjectively more confusing
2. Arguably people don't want to see the individual modules compiling for each dependency

CC @nh2

Wildsoft:
Maybe indicate some progress in some other way to avoid confusing output

NorfairKing:
I'd like `--interleaved-output` for local packages but not for dependencies, by default.
Or that `--test` implies `--interleaved-output` if there are multiple packages.

snoyberg:
> Maybe indicate some progress in some other way to avoid confusing output

Two problems with this:

1. We'd need to know what "some other way" is. In other words: can you be a bit more specific?
2. It doesn't exist yet, so would have to be implemented

> I'd like --interleaved-output for local packages but not for dependencies, by default.

I get the motivation here. However, it wouldn't address my "frozen compiles" concern above. Not eliminating the possibility, just pointing out a limitation.

snoyberg:
We can continue to refine the behavior up until the next release, but I've opened up a PR to change the default. This may also flush out some additional problems during testing. #4718

PR #4722, opened 2019-04-12
Set empty input to sinkProcessStderrStdout

snoyberg:
This is intended to address intermittent integration test failures,
which started as a result of the new integration test runner. This is a
weird corner case, but here goes:

* ./configure scripts will fail if run with `stdin` closed. We already
work around this in Stack.Build.Execute in the OTLogFile case

* However, in the OTConsole case, we use the sinkProcessStderrStdout
functions, which inherits stdin

* The current behavior can be classified as a bug: we don't want
subprocesses to be exposed to stdin in general.

    * Reviewer request: I've checked the other use cases for
      sinkProcessStderrStdout, and I believe in _all_ cases we do not want the
      subprocess to have such stdin access. Please confirm I haven't made a
      mistake.

* The new integration test runner explicit uses `setStdin closed`...

* And in some cases, the runner will try to build the `network` package,
which has a ./configure script. I don't know why this package isn't
built every time, but when it is built, it fails with a file descriptor
0 error

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -1193,6 +1193,9 @@ withSingleContext ActionContext {..} ExecuteEnv {..} task@Task {..} mdeps msuffi
                         OTLogFile _ h ->
                             proc (toFilePath exeName) fullArgs
                           $ runProcess_
+                            -- Don't use closed, since that can break
+                            -- ./configure scripts. See:
+                            -- https://github.com/commercialhaskell/stack/pull/4722
                           . setStdin (byteStringInput "")
                           . setStdout (useHandleOpen h)
                           . setStderr (useHandleOpen h)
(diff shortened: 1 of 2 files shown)


snoyberg:
Clarification: I believe the integration test failures started with the change for #4702 (interleaved output), which caused the `OTConsole` code path to trigger and the `network` build to fail. Yay integration tests catching a really obscure corner case!

snoyberg:
I'm going to merge to unblock integration tests on master. @borsboom if you see a problem with the usage of `setStdin (byteStringInput "")` in the case of Docker, please speak up.

borsboom:
> if you see a problem with the usage of setStdin (byteStringInput "") in the case of Docker, please speak up.

Nothing would surprise me, but I'll test it and let you know if anything goes wrong.

PR #4723, opened 2019-04-12
Additions necessary for integration of pantry into stackage-server:

lehins:
* Exposed few modules needed for stackage-server
* Addition of few class instances and exports needed for stackage-server
* Made pantry compatible with PostgreSQL. Improved speed and safety by added database aware queries. 
* Fixed concurrent blob and file name writes in multi-connection sql pool setting. Added few more exports needed for stackage-server
* Removed redundant constraints and switched many queries to `MonadIO`
* Added .cabal files to gitignore for all subs
* Added .hindent file for pantry with width set to `2`, since that is what majority of code has.

### DIFF
--- a/subs/pantry/package.yaml
+++ b/subs/pantry/package.yaml
@@ -108,9 +108,7 @@ library:
   - Pantry.Internal
   - Pantry.Internal.StaticBytes
   # For stackage-server
-  - Pantry.Storage
-  - Pantry.Types
-  - Pantry.Hackage
+  - Pantry.Internal.Stackage
 
   # FIXME must be removed from pantry!
   - Data.Aeson.Extended
--- a/subs/pantry/package.yaml
+++ b/subs/pantry/package.yaml
@@ -86,6 +86,9 @@ dependencies:
 - directory
 - filepath
 
+ghc-options:
+  - -Wall
+
 library:
   source-dirs: src/
   when:
@@ -104,6 +107,10 @@ library:
   # For testing
   - Pantry.Internal
   - Pantry.Internal.StaticBytes
+  # For stackage-server
+  - Pantry.Storage
+  - Pantry.Types
+  - Pantry.Hackage
 
   # FIXME must be removed from pantry!
   - Data.Aeson.Extended
--- a/subs/pantry/src/Pantry/Hackage.hs
+++ b/subs/pantry/src/Pantry/Hackage.hs
@@ -4,10 +4,12 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 module Pantry.Hackage
   ( updateHackageIndex
+  , forceUpdateHackageIndex
   , DidUpdateOccur (..)
   , RequireHackageIndex (..)
   , hackageIndexTarballL
   , getHackageTarball
+  , getHackageTarballOnGPD
   , getHackageTarballKey
   , getHackageCabalFile
   , getHackagePackageVersions
@@ -28,7 +30,7 @@ import qualified RIO.ByteString as B
 import qualified RIO.ByteString.Lazy as BL
 import Pantry.Archive
 import Pantry.Types hiding (FileType (..))
-import Pantry.Storage
+import Pantry.Storage hiding (TreeEntry, PackageName, Version)
 import Pantry.Tree
 import qualified Pantry.SHA256 as SHA256
 import Network.URI (parseURI)
@@ -39,6 +41,7 @@ import qualified Distribution.PackageDescription as Cabal
 import System.IO (SeekMode (..))
 import qualified Data.List.NonEmpty as NE
 import Data.Text.Metrics (damerauLevenshtein)
+import Distribution.PackageDescription (GenericPackageDescription)
 import Distribution.Types.Version (versionNumbers)
 import Distribution.Types.VersionRange (withinRange)
 
@@ -80,7 +83,26 @@ updateHackageIndex
   :: (HasPantryConfig env, HasLogFunc env)
   => Maybe Utf8Builder -- ^ reason for updating, if any
   -> RIO env DidUpdateOccur
-updateHackageIndex mreason = do
+updateHackageIndex = updateHackageIndexInternal False
+
+-- | Same as `updateHackageIndex`, but force the database update even if hackage
+-- security tells that there is no change.  This can be useful in order to make
+-- sure the database is in sync with the locally downloaded tarball
+--
+-- @since 0.1.0.0
+forceUpdateHackageIndex
+  :: (HasPantryConfig env, HasLogFunc env)
+  => Maybe Utf8Builder
+  -> RIO env DidUpdateOccur
+forceUpdateHackageIndex = updateHackageIndexInternal True
+
+
+updateHackageIndexInternal
+  :: (HasPantryConfig env, HasLogFunc env)
+  => Bool -- ^ Force the database update.
+  -> Maybe Utf8Builder -- ^ reason for updating, if any
+  -> RIO env DidUpdateOccur
+updateHackageIndexInternal forceUpdate mreason = do
   storage <- view $ pantryConfigL.to pcStorage
   gateUpdate $ withWriteLock_ storage $ do
     for_ mreason logInfo
@@ -118,6 +140,9 @@ updateHackageIndex mreason = do
         HS.checkForUpdates repo maybeNow
 
     case didUpdate of
+      _ | forceUpdate -> do
+            logInfo "Forced package update is initialized"
+            updateCache tarball
       HS.NoUpdates -> do
         x <- needsCacheUpdate tarball
         if x
@@ -200,11 +225,13 @@ updateHackageIndex mreason = do
               if oldHash == oldHashCheck
                 then oldSize <$ logInfo "Updating preexisting cache, should be quick"
                 else 0 <$ do
-                  logInfo "Package index change detected, that's pretty unusual"
-                  logInfo $ "Old size: " <> display oldSize
-                  logInfo $ "Old hash (orig) : " <> display oldHash
-                  logInfo $ "New hash (check): " <> display oldHashCheck
-                  logInfo "Forcing a recache"
+                  logWarn $ mconcat [
+                    "Package index change detected, that's pretty unusual: "
+                    , "\n    Old size: " <> display oldSize
+                    , "\n    Old hash (orig) : " <> display oldHash
+                    , "\n    New hash (check): " <> display oldHashCheck
+                    , "\n    Forcing a recache"
+                    ]
             pure (offset, newHash, newSize)
 
       lift $ logInfo $ "Populating cache from file size " <> display newSize <> ", hash " <> display newHash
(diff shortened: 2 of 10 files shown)


subs/pantry/package.yaml
```diff
@@ -104,6 +107,10 @@ library:
   # For testing
   - Pantry.Internal
   - Pantry.Internal.StaticBytes
+  # For stackage-server
```
snoyberg: I'd rather see these in some kind of `.Internal` namespace, I don't want these to be part of the publicly accessed API. Better yet, if there's some way to do it, I'd prefer if there was a minimal subset of these modules that could reasonably be exported from `Pantry` itself.

subs/pantry/src/Pantry/Hackage.hs
```diff
@@ -178,11 +200,13 @@ updateHackageIndex mreason = do
               if oldHash == oldHashCheck
                 then oldSize <$ logInfo "Updating preexisting cache, should be quick"
                 else 0 <$ do
-                  logInfo "Package index change detected, that's pretty unusual"
-                  logInfo $ "Old size: " <> display oldSize
-                  logInfo $ "Old hash (orig) : " <> display oldHash
-                  logInfo $ "New hash (check): " <> display oldHashCheck
-                  logInfo "Forcing a recache"
+                  logWarn $ mconcat [
```
snoyberg: Not a big deal, but `mconcat` has more overhead than using a bunch of `<>`s, since it forces an intermediate list representation.

subs/pantry/src/Pantry/Hackage.hs
```diff
@@ -463,11 +487,22 @@ getHackageTarball
...
   -> Maybe TreeKey
   -> RIO env Package
-getHackageTarball pir@(PackageIdentifierRevision name ver _cfi) mtreeKey = do
+getHackageTarball = getHackageTarballOnGPD (\ _ _ -> pure ())
+
+-- | Same as `getHackageTarball`, but allows an extra action to be performed on the parsed
+-- `GenericPackageDescription` and newly created `TreeId`.
+getHackageTarballOnGPD
+  :: (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+  => (TreeId -> GenericPackageDescription -> RIO env ())
```
snoyberg: Alternative approach: what about having a `HackageTarballResult` datatype that contains a `Package`, `TreeId`, and `GenericPackageDescription` and return that? Any specific reason why this callback should be called inside this function call?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -243,39 +258,61 @@ withStorage
 withStorage action =
   flip SQLite.withStorage_ action =<< view (P.pantryConfigL.to P.pcStorage)
 
+
+rdbmsAwareQuery
```
snoyberg: Please add some docs to explain what the different arguments are here. I'm tempted to say there should be a datatype helper to avoid positional arguments, e.g.:

```haskell
data RdmbsActions m a = RdbmsActions
   { raSqlite :: !(ReaderT SqlBackend m a)
   , raPostgres :: !(ReaderT SqlBackend m a)
   }
```

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -243,39 +258,61 @@ withStorage
...
-  => SafeFilePath
-  -> ReaderT SqlBackend (RIO env) FilePathId
-getFilePathId = fmap (either entityKey id) . insertBy . FilePath
-
 storeBlob
-  :: (HasPantryConfig env, HasLogFunc env)
+  :: MonadIO m
   => ByteString
-  -> ReaderT SqlBackend (RIO env) (BlobId, BlobKey)
+  -> ReaderT SqlBackend m (BlobId, BlobKey)
```
snoyberg: Are these `RIO`-removing changes necessary for Stackage Server, or a cleanup?

subs/pantry/src/Pantry/Storage.hs
```diff
@@ -700,7 +732,7 @@ loadPackageById rpli tid = do
         Package
             { packageTreeKey = P.TreeKey blobKey
             , packageTree = mtree
-            , packageCabalEntry = pentry
+            , packageCabalEntry = pantry
```
snoyberg: `pentry` was intended: it was the `package entry`.

subs/pantry/src/Pantry/Types.hs
```diff
@@ -205,8 +205,8 @@ newtype Revision = Revision Word
 -- whether a pool is used, and the default implementation in
 -- "Pantry.Storage" does not use a pool.
 data Storage = Storage
-  { withStorage_ :: (forall env a. HasLogFunc env => ReaderT SqlBackend (RIO env) a -> RIO env a)
-  , withWriteLock_ :: (forall env a. HasLogFunc env => RIO env a -> RIO env a)
+  { withStorage_ :: forall m a. MonadUnliftIO m => ReaderT SqlBackend m a -> m a
```
snoyberg: I think overall I'd rather push the code in stackage-server to use `RIO` rather than have to avoid its usage here. But I'm open to hearing that that would be too big a change.

subs/pantry/src/Pantry/Hackage.hs
lehins: Just because of this :
> that's pretty unusual
it doesn't seem like performance here is the main concern here :) Moreover I really hope hackage folks are not gonna start messing with the index file and cause this warning to appear.

Using `mconcat` here is more readable in my opinion. But as always, if you'd like me to change it to bunch of `<>`, I'll do that.

subs/pantry/package.yaml
lehins: The only thing I can think of is I can compile a list of all functions and types that are needed by stackage-server, and reexport them through a separate module. There are quite a few of them, since I am using pantry's tables directly and esqueleto for queries as well as reuse a bunch of types. Quick compilation of pantry related imports inside stackage-server:

```haskell
import Pantry (defaultHackageSecurityConfig)
import Pantry.Hackage (DidUpdateOccur(..), forceUpdateHackageIndex,
                       getHackageTarballOnGPD)
import Pantry.Storage (HackageCabalId, getTreeForKey, loadBlobById, treeCabal)
import Pantry.Types (CabalFileInfo(..), HpackExecutable(HpackBundled),
                     PackageIdentifierRevision(..), PantryConfig(..),
                     packageTreeKey)
```
```haskell
import Pantry.SHA256 (fromHexText)
import Pantry.Storage (BlobId, HackageCabalId)
import Pantry.Types (BlobKey(..), CabalFileInfo(..), FileSize(..),
                     HasPantryConfig(..), PackageIdentifierRevision(..),
                     PackageNameP(..), PantryConfig, TreeKey(..), VersionP(..))
```

```haskell
import Pantry.Storage as PS (BlobId, HackageCabalId, ModuleNameId,
                             PackageNameId, Tree(..), TreeEntry(..),
                             TreeEntryId, TreeId, Unique(..), VersionId,
                             unBlobKey)
import qualified Pantry.Storage as Pantry (migrateAll)
import Pantry.Types (HasPantryConfig(..), PantryConfig(..), Revision,
                     Storage(..))
```
```haskell
import Pantry.Storage (EntityField(..), PackageName, Unique(..), Version,
                       getBlobKey, getPackageNameById, getPackageNameId,
                       getTreeForKey, getVersionId, loadBlobById, treeCabal)
import Pantry.Types (mkSafeFilePath)
```

subs/pantry/src/Pantry/Hackage.hs
lehins: Correct me if I am wrong, but from my understanding `getHackageTarball` does not need to retrieve `GenericPackageDescription`, it only needs the `Package`.  `withCachedTree` allows for the direct  look up of the `Package` from the database if such is present, and only falling back onto building the `Package` by pulling it from Hackage. In other words that function `onGPD`, will only be called when this `Package` is new to Pantry.

To answer your question. We could add new datatype `HackageTarballResult` that would get all three: `Package`, `TreeId`, and `GenericPackageDescription`, but they will be used only in `stackage-server`, for Pantry pulling the cabal blob from database and parsing it will be an unnecessary overhead.

subs/pantry/src/Pantry/Storage.hs
lehins: Some of theses are necessary for Stackage Server others are just cleanup for consistency. RIO doesn't compose very well with Yesod, at least for now, right? So I use RIO inside stackage-server-cron, but everywhere else functions that query database still have to be `ReaderT SqlBackend m`.

I think it is an improvement, since `RIO` wrapper is really unnecessary, especially with `HasPantryConfig env` and `HasLogFunc env` constraints. Also if I remember correctly your suggestion was to try to keep everything in a most general form. Which make this suggestion to work nicely in stackage-server

subs/pantry/src/Pantry/Storage.hs
lehins: Lol. I did not realize that. I can see it now though. Can I rename it to `pEntry` or `packageEntry`, so I doesn't look like misspelled pantry? :)

subs/pantry/src/Pantry/Types.hs
lehins: I like that idea, but I wanna push back on it until we can use RIO with Yesod easily.

subs/pantry/src/Pantry/Storage.hs
lehins: This is only an internal function that is used in two places, an extra data type seems like an overkill. But if you feel strongly about it, I can add `RdmbsActions`.

I will add some documentation explaining this function.

subs/pantry/package.yaml
snoyberg: Alright, let's do reexporting from a `.Internal` module with a comment explaining that this is for advanced usages.

subs/pantry/src/Pantry/Hackage.hs
snoyberg: Nah, not worth changing, it's a common comment I add on PRs just to make the issue known.

subs/pantry/src/Pantry/Storage.hs
snoyberg: I don't think it's overkill at all, we could very easily write incorrect code without the types there to help us.

subs/pantry/src/Pantry/Storage.hs
snoyberg: > RIO doesn't compose very well with Yesod, at least for now, right?

Well, sort of. But it seems like all of this can be addressed with a single helper wrapper function in stackage-server. And since we already have such a wrapper function to calling into the database, I would have thought it wouldn't have a big overhead.

> Also if I remember correctly your suggestion was to try to keep everything in a most general form.

Not in general, no. Where there's a strong reason to support non-RIO monads I go in that direction (like the `unliftio` package). But sticking to `RIO` makes error messages much easier to follow, so I prefer it when it won't cause trouble.

subs/pantry/src/Pantry/Storage.hs
snoyberg: Either of those spellings is great.

subs/pantry/src/Pantry/Hackage.hs
snoyberg: If `getHackageTarball` doesn't need the `GenericPackageDescription`, and `getHackageTarball` is now implemented in terms of `getHackageTarballOnGPD`, that would mean that we are unnecessarily parsing the cabal file.

> To answer your question. We could add new datatype HackageTarballResult that would get all three: Package, TreeId, and GenericPackageDescription, but they will be used only in stackage-server, for Pantry pulling the cabal blob from database and parsing it will be an unnecessary overhead.

I think I'd prefer that over

1. Adding a new function to the API and
2. Using callback style where it isn't necessary

subs/pantry/package.yaml
lehins: I created an extra module `Pantry.Internal.Stackage` which exports only the stuff needed for stackage server.

subs/pantry/src/Pantry/Hackage.hs
lehins: I did add a new `HackageTarballResult` data type, which it is not part of a public api and is exported through an internal module to stackage server. I think this is a win win solution, since it avoids all three: unnecessary overhead of parsing cabal file when it is not needed, the callback style and addition of another function to the api.

subs/pantry/src/Pantry/Storage.hs
lehins: Done. Added `RdbmsActions` data type and some documentation.

subs/pantry/src/Pantry/Storage.hs
lehins: Switched to `packageEntry`

subs/pantry/src/Pantry/Types.hs
lehins: Ok, I switched it back to the original signature. 
```haskell
withStorage_ :: (forall env a. HasLogFunc env => ReaderT SqlBackend (RIO env) a -> RIO env a)
```
Also adjusted stackage server to use RIO for connecting to db as well.

subs/pantry/src/Pantry/Storage.hs
lehins: All switched `ReaderT SqlBackend m` -> `ReaderT SqlBackend (RIO env)`

PR #4725, opened 2019-04-13
Add compatibility with rio-0.1.9.0

### LINKED ISSUE #4724: 1.9.3 fails to build with rio 0.1.9.0
Version bounds are currently >= 0.1.5.0 && < 0.2, but stack 1.9.3 fails to build with rio 0.1.9.0. Might need to adjust the bounds.

```
[ 55 of 127] Compiling Text.PrettyPrint.Leijen.Extended ( src/Text/PrettyPrint/Leijen/Extended.hs, dist/build/Text/PrettyPrint/Leijen/Extended.dyn_o )

src/Text/PrettyPrint/Leijen/Extended.hs:144:14: error:
    Ambiguous occurrence ‘empty’
    It could refer to either ‘Text.PrettyPrint.Annotated.Leijen.empty’,
                             imported from ‘Text.PrettyPrint.Annotated.Leijen’ at src/Text/PrettyPrint/Leijen/Extended.hs:130:1-63
                          or ‘Stack.Prelude.empty’,
                             imported from ‘Stack.Prelude’ at src/Text/PrettyPrint/Leijen/Extended.hs:127:1-42
                             (and originally defined in ‘GHC.Base’)
    |
144 |     mempty = empty
    |              ^^^^^
```

qrilka:

Fixes #4724

### DIFF
--- a/src/Stack/Clean.hs
+++ b/src/Stack/Clean.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -19,7 +20,9 @@ import           Stack.Config (getLocalPackages)
 import           Stack.Constants.Config (distDirFromDir, workDirFromDir)
 import           Stack.Types.PackageName
 import           Stack.Types.Config
+#if !MIN_VERSION_rio(0,1,9)
 import           System.Exit (exitFailure)
+#endif
 
 -- | Deletes build artifacts in the current project.
 --
--- a/src/Stack/Clean.hs
+++ b/src/Stack/Clean.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ConstraintKinds #-}
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -20,9 +19,7 @@ import           Stack.Config (getLocalPackages)
 import           Stack.Constants.Config (distDirFromDir, workDirFromDir)
 import           Stack.Types.PackageName
 import           Stack.Types.Config
-#if !MIN_VERSION_rio(0,1,9)
 import           System.Exit (exitFailure)
-#endif
 
 -- | Deletes build artifacts in the current project.
 --
@@ -30,7 +27,7 @@ import           System.Exit (exitFailure)
 clean :: HasEnvConfig env => CleanOpts -> RIO env ()
 clean cleanOpts = do
     failures <- mapM cleanDir =<< dirsToDelete cleanOpts
-    when (or failures) $ liftIO exitFailure
+    when (or failures) $ liftIO System.Exit.exitFailure
   where
     cleanDir dir =
       liftIO (ignoringAbsence (removeDirRecur dir) >> return False) `catchAny` \ex -> do
--- a/src/Stack/Coverage.hs
+++ b/src/Stack/Coverage.hs
@@ -244,7 +244,7 @@ generateHpcReportForTargets opts = do
                              case nc of
                                  CTest testName ->
                                      liftM (pkgPath </>) $ parseRelFile (T.unpack testName ++ "/" ++ T.unpack testName ++ ".tix")
-                                 _ -> fail $
+                                 _ -> liftIO . fail $
                                      "Can't specify anything except test-suites as hpc report targets (" ++
                                      packageNameString name ++
                                      " is used with a non test-suite target)"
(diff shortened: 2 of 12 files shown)


src/Stack/Clean.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
```
snoyberg: If possible, it's best to avoid adding CPP, since it can cause unnecessary recompilation. Can you instead use qualified usage of the identifier?

Alternatively: We can bump the minimum version of `rio` in the cabal file.

Alternative alternatively: we simply don't care about this, since Stack should always be built from the relevant `stack.yaml` file.

src/Stack/Coverage.hs
```diff
@@ -244,7 +244,7 @@ generateHpcReportForTargets opts = do
                              case nc of
                                  CTest testName ->
                                      liftM (pkgPath </>) $ parseRelFile (T.unpack testName ++ "/" ++ T.unpack testName ++ ".tix")
-                                 _ -> fail $
+                                 _ -> liftIO . fail $
```
snoyberg: It wouldn't likely be better to use `throwString` anyway. But is this caused by a change in `rio`, or a change in GHC?

src/Stack/SDist.hs
```diff
@@ -24,7 +24,7 @@ import qualified Data.ByteString as S
 import qualified Data.ByteString.Char8 as S8
 import qualified Data.ByteString.Lazy as L
 import           Data.Char (toLower)
-import           Data.Data (cast)
+import           Data.Data (cast, gmapT)
```
snoyberg: This is also fixed in the 0.1.9.1 bump to rio.

src/Text/PrettyPrint/Leijen/Extended.hs
```diff
@@ -124,7 +125,11 @@ import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import qualified Data.Text.Lazy as LT
 import qualified Data.Text.Lazy.Builder as LTB
+#if !MIN_VERSION_rio(0,1,9)
 import Stack.Prelude hiding (Display (..))
+#else
+import Stack.Prelude hiding (Display (..), empty)
```
snoyberg: Should no longer be necessary with rio 0.1.9.1.

src/Stack/Coverage.hs
qrilka: that's now exported `MonadFail` but `throwString` probably makes more sense, thanks

src/Stack/Clean.hs
qrilka: I've caught this incompatibility when trying to build Stackage, if we just pin dependencies then we'll need to drop Stack from Stackage and this PR was thought to prevent that from happening

src/Stack/SDist.hs
qrilka: Sure, but that release came a bit later in time :)

src/Stack/Clean.hs
snoyberg: OK, fair enough, I rescind that idea. But still, finding ways that don't introduce `CPP` would be ideal.

src/Stack/SDist.hs
snoyberg: Yes, you're right. It just wasn't immediately clear that we were going to add these back in 0.1.9.1, so I wanted to mention it.

qrilka:
@snoyberg with 0.1.9.0 and 0.1.9.1 deprecated what's the current plan? Should we just close this PR and think about the compatibility with the next version?

snoyberg:
May as well leave it open. I'm going to try to carefully check the next patch release to make sure it doesn't include any breaking changes. Thanks to the `System.Exit` stuff, some changes will still be needed to Stack, but hopefully most of the changes here will be unnecessary.

snoyberg:
@borsboom Can you look this over when you have a chance? Also, what's the procedure for releasing a patch to Hackage?

borsboom:
Looks like the only change left is qualifying all the `exitWith` references, so I guess this will remain compatible with rio (>=0.1.5.0 && <0.1.9.0), and also compatible with rio (>=0.1.9.2 && <0.2)? 
 Perhaps a `package.yaml` constraint should make this explicit.

Doesn't look like the I've documented the procedure to release a patch to hackage is documented, but it would go something like this:

1. Cherry-pick the change to the `release` branch
2. Add a fourth version component in package.yaml (e.g. 1.9.3.1)
3. Add an entry to the changelog for 1.9.3.1, making it clear that this is a Hackage-only compatibility release with no changes to functionality.
4. Follow the steps in https://docs.haskellstack.org/en/stable/maintainers/releases/ create an sdist (using `etc/scripts/sdist-with-bounds.sh`), upload it to github, upload it to hackage, and update the cabal file metadata with version constraints.

snoyberg:
Yes, this should remain compatible. I personally don't think an explicit bound to ignore 0.1.9.0 and 0.1.9.1 is necessary, but I'll add it if you think it's prudent.

borsboom:
One more thing for my release steps above: over time I've made cabal metadata edits for stack-1.9.3 that loosen some maxbounds, and we should make sure the new release has those too.  One way that should cover most (if not all) of those is be to upgrade `stack-nightly.yaml` to the latest nightly snapshot.  The script that generates bounds will then pick those up.  This is a something I always do before real releases.

Also, the sdist should be uploaded to the existing stack-1.9.3 Github release, rather than making a new stack-1.9.3.1 github release.  If we did the latter, we'd basically have to do a full minor release (building new bindists etc.), which which case we might as well release the `stable` branch as stack-1.9.5.

PR #4729, opened 2019-04-14
Recommend Stack upgrade when appropriate (fixes #1681)

### LINKED ISSUE #1681: Check for recent version during calls?
The [cocoa pods](https://cocoapods.org) `pod` command checks for the latest version upon each invocation, and lets one know if there is a newer version available. I would like to see such a feature added to `stack`.

snoyberg:

### DIFF
--- a/src/Stack/Runners.hs
+++ b/src/Stack/Runners.hs
@@ -187,7 +187,7 @@ shouldUpgradeCheck = do
     let yesterday = addUTCTime (-24 * 60 * 60) now
     checks <- upgradeChecksSince yesterday
     when (checks == 0) $ do
-      mversion <- getLatestHackageVersion "stack" UsePreferredVersions -- FIXME ensure it doesn't force an update ever
+      mversion <- getLatestHackageVersion NoRequireHackageIndex "stack" UsePreferredVersions
       case mversion of
         Just (PackageIdentifierRevision _ version _) | version > mkVersion' Paths_stack.version -> do
           logWarn "<<<<<<<<<<<<<<<<<<"
--- a/src/Stack/Runners.hs
+++ b/src/Stack/Runners.hs
@@ -20,7 +20,10 @@ module Stack.Runners
     ) where
 
 import           Stack.Prelude
+import           Distribution.Version (mkVersion')
+import qualified Paths_stack
 import           RIO.Process (mkDefaultProcessContext)
+import           RIO.Time (addUTCTime, getCurrentTime)
 import           Stack.Build.Target(NeedTargets(..))
 import           Stack.Config
 import           Stack.Constants
@@ -28,6 +31,7 @@ import           Stack.DefaultColorWhen (defaultColorWhen)
 import qualified Stack.Docker as Docker
 import qualified Stack.Nix as Nix
 import           Stack.Setup
+import           Stack.Storage (upgradeChecksSince, logUpgradeCheck)
 import           Stack.Types.Config
 import           Stack.Types.Docker (dockerEnable)
 import           Stack.Types.Nix (nixEnable)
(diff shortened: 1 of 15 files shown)


snoyberg:
Sorry, I forgot one detail on this, please don't review/merge until I update this... there's a FIXME if anyone's curious.

snoyberg:
OK, second commit addresses the FIXME, this is ready for review.

src/Stack/Runners.hs
```diff
@@ -169,3 +177,32 @@ withRunnerGlobal go inner = do
...
+shouldUpgradeCheck = do
+  config <- ask
+  when (configRecommendUpgrade config) $ do
+    now <- getCurrentTime
+    let yesterday = addUTCTime (-24 * 60 * 60) now
+    checks <- upgradeChecksSince yesterday
+    when (checks == 0) $ do
+      mversion <- getLatestHackageVersion NoRequireHackageIndex "stack" UsePreferredVersions
+      case mversion of
+        Just (PackageIdentifierRevision _ version _) | version > mkVersion' Paths_stack.version -> do
```
borsboom: `Stack.Types.Version.stackVersion` can be used instead of `mkVersion' Paths_stack.version`.

snoyberg:
Apologies for merging master into this branch during a review, but due to how the caching system on CI is working, the tests will fail with a migration error unless I include the schema change on master. The relevant change to address your comments is 191301a34aa0cc70ff7136a8c4708426495719f2. Thanks for the review 👍

src/Stack/Storage.hs
```diff
@@ -151,6 +153,10 @@ CompilerCache
   globalDump Text
 
   UniqueCompilerInfo ghcPath
+
+-- History of checks for whether we should upgrade Stack
+UpgradeCheck
```
borsboom: If I understand the code correctly, this is being used to determine whether there have been any version checks in the previous 24 hours (before the current time).  It looks like this could grow indefinitely, and since there's no index on the `timestamp` field each of these checks would require a table scan, so over time this would get slower and slower.

I think you could add a unique constraint on `timestamp` (and change `logUpgradeCheck` to use `upsert`) to avoid the table scans, and I think records older than 24 hours could be purged since they're not being used anymore.

Alternatively, perhaps the approach could be simplified into doing just one version check on a given calendar date (rather than the last 24 hour period).  Then all you need to do is store a single value: the date of the last check.

snoyberg: I'd like to skip the calendar day approach, since (1) it allows for users getting bugged twice in quick succession around midnight, and (2) it can be wonky with timezones. I'll update the PR to delete all of the old records before inserting. I left the old records initially because I thought it could be nice to see when checks actually happen, but there's 0 obvious reason to do it, just a bad data hording tendency on my part.

src/Stack/Storage.hs
```diff
@@ -557,4 +557,6 @@ upgradeChecksSince since = withStorage $ count [UpgradeCheckTimestamp >=. since]
 
 -- | Log in the database that an upgrade check occurred at the given time.
 logUpgradeCheck :: HasConfig env => UTCTime -> RIO env ()
-logUpgradeCheck = withStorage . insert_ . UpgradeCheck
+logUpgradeCheck time = withStorage $ do
+  deleteWhere ([] :: [Filter UpgradeCheck])
```
borsboom: It looks like you'll only ever have a single record in the database, which I agree will work.  What about just updating the existing record if it already exists?

snoyberg: It seems like that introduces more complex logic to me: check number of entries, switch between updating and inserting, etc. What's the advantage to updating versus deleting-and-inserting?

borsboom: `upsert` has that logic (although you'd need to add a unique key that you always set to the same value to enforce that there's only one record).  In this case the difference is probably negligible, just avoids the overhead of deleting and creating new records unnecessarily.  In theory, there could be a need for another field in this table someday in the future, in which case deleting the whole record might delete more information than intended.

There may be the case for having one "singleton" table that will only ever contain one record and that we add fields to for any single values that need to be saved, rather than ending up with multiple single-record tables.

snoyberg: I've tried out the singleton table approach, sort of. It's a table with two columns: the action and timestamp. In the future, we can add additional actions that should be tracked. One that comes to mind is garbage collection. Maybe another would be Hackage index updating, if we ever care about that.

I used a integer representation of the action, but I can see an argument for a `String` field to allow for greater flexibility in the future.

borsboom: I like this approach.  No particular opinion or integer vs. string.

PR #4740, opened 2019-04-15
Reconfigure on new PATH env var (fixes #3138)

### LINKED ISSUE #3138: Consider reconfiguring if $PATH changes
As per https://github.com/haskell/cabal/issues/2015, it seems Cabal caches the `$PATH` when running `configure` and reuses it later. As a result, if a package needs a particular program on `$PATH` at compile time, a user who modifies their `$PATH` after a failure will continue to remain in a failure state until running a `stack clean`. Perhaps the `$PATH` should be considered an input dependency of configure steps?

snoyberg:


snoyberg:
I'm going to move ahead with merging this. It appears that currently CI is broken because a version of this database got cached and now `master` is failing at migration.

PR #4742, opened 2019-04-15
Atomic file writes

snoyberg:


snoyberg:
I've switched this over to use an atomic file write. When that functionality is merged into rio, we can switch to using it.

PR #4746, opened 2019-04-16
Lock files attempt 2

### LINKED ISSUE #4288: Write pantry hashes into a .lock file instead of emitting warnings
Currently `stack build` will emit warnings if I do not specify `size` and `sha256` for `github` dependencies in `stack.yaml`.

Running `stack freeze` will output the relevant information that I then can add to `stack.yaml` manually.

I propose to write the information that is currently produced by `stack freeze` to a `.lock` file after every successful build (e.g. `pantry.lock`):

- when the `.lock` file exists then for each listed dependency:
  - `stack` uses the information as if it were directly listed in `stack.yaml`; specifically it errors out on any hash mismatches
  - when `stack` detects that a dependency was updated in `stack.yaml` it silently ignores the information from the `.lock` file
- after a successful build `stack` creates the `.lock` file, overwriting any existing versions (which in effect means that we remove unused entries and add missing entries)

Regarding revision control, we can [follow conventions that have been developed elsewhere](https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/):
When developing an application, the `.lock` file should be checked into revision control.  When developing a library, the `.lock` file should be added to `.gitignore`.

A word on naming: If we expect the `.lock` file to be specific to `stack` we may want to name it `stack.lock` (or possibly `.stack.lock` to not interfere with file name completion); if on the other hand we think that it will be relevant for other build tools that implement support for `pantry` then we could name it `pantry.lock`.

Assuming we adopt this approach we can:
- remove any warnings related to pantry hashes
- remove the `stack freeze` command

CC @snoyberg

qrilka:

Tested manually and with local run of integration tests
Has some code and resolves #4550 
Resolves #4288

### DIFF
--- a/doc/lock_files.md
+++ b/doc/lock_files.md
@@ -211,7 +211,7 @@ If the lock file does not exist, it will be created by:
 If the `stack.yaml.lock` file exists, its last modification time is
 compared against the last modification time of the `stack.yaml` file
 and any local snapshot files. If any of those files is more recent
-than the `stack.yaml` file, and the file hashes in the lock file
+than the `stack.yaml.lock` file, and the file hashes in the lock file
 do not match the files on the filesystem, then the update procedure is
 triggered. Otherwise, the `stack.yaml.lock` file can be used as the
 definition of the snapshot.
--- a/doc/lock_files.md
+++ b/doc/lock_files.md
@@ -181,4 +181,5 @@ If the lock file does not exist, it will be created by:
 
 When loading a Stack project all completed package or snapshot locations
 (even when they were completed using information from a lock file) get
-collected to form a new lock file.
+collected to form a new lock file in memory and compare against the one
+on disk, writing if there are any differences.
--- a/doc/lock_files.md
+++ b/doc/lock_files.md
@@ -18,11 +18,12 @@ set of input files. There are a few problems with making this work:
 
 To address this, we follow the (fairly standard) approach of having a
 _lock file_. The goal of the lock file is to cache completed
-information about all packages and snapshot files so that:
+locations of project, snapshot packages and snapshots themselves so that:
 
 * These files can be stored in source control
 * Users on other machines can reuse these lock files and get identical
-  build plans
+  build plans given that the used local packages and local snapshots are
+  the same on those machines
 * Rerunning `stack build` in the future is deterministic in the build
   plan, not depending on mutable state in the world like Hackage
   revisions
@@ -31,8 +32,6 @@ information about all packages and snapshot files so that:
       to perform the build. However, by deterministic, we mean it
       either performs the same build or fails, never accidentally
       doing something different.
-* Stack can quickly determine the build plan in the common case of no
-  changes to `stack.yaml` or snapshot files
 
 This document explains the contents of a lock file, how they are used,
 and how they are created and updated.
@@ -42,11 +41,7 @@ and how they are created and updated.
 Relevant to this discussion, the `stack.yaml` file specifies:
 
 * Resolver (the parent snapshot)
-* Compiler override
 * `extra-deps`
-* Flags
-* GHC options
-* Hidden packages
 
 The resolver can either specify a compiler version or another snapshot
 file. This snapshot file can contain the same information referenced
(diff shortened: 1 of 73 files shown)


qrilka:
One note about the implementation: it persist all dependency locations and that means that `stack.yaml.lock` is quite large because it contains all snapshot packages so loading it takes noticeable time.

snoyberg:
I haven't looked at the code yet, but based on your comment @qrilka: do you think it makes more sense to take a different approach here? Two things I can think of:

* Don't include all packages. This seems like it will make it quite difficult to come up with a `SourceMapHash`
* Use a more efficient format, either as the primary format or as an optimized format for lookup. In the latter, I mean keeping the .lock file as a YAML file, but keeping a .lock.bin file next to it (or elsewhere) in a more efficient binary format, or maybe even storing in the SQLite database

qrilka:
@snoyberg I think I need to check the current timings to get a bit more clear understanding of the problem. But in any case it looks odd to add an optimization which makes Stack slower.

qrilka:
@snoyberg on Monday I've tried to optimize sourcemap-based lockfiles and I had some improvements but still if we parse almost the same what we have in a snapshot it takes some time. So in 40bf6f0 I've simplified everything and in that commit lock file contains only non-trivial location completions (e.g. snapshot locations won't get stored there) - it simplifies logic quite a lot and also it gives some speed improvements (up to 20% compared to optimized sourcemap-based locks).
I think we should go this simpler and a bit faster way.
One visible difference with the previous version is that we don't check snapshots but I don't see how that could cause any troubles.
If we choose this simple variant then `lock_files.md` needs to be rewritten.

qrilka:
BTW regarding timings I have this gist - https://gist.github.com/qrilka/d5e220f3955c8d8ea70ad40f720f5f12

qrilka:
@snoyberg 1 of integration test failed, so I need to fix that before it will be ready for a proper review

qrilka:
the problem with `1438-configure-options` looks to be unrelated to this change so this PR is ready to be reviewed

src/test/Stack/LockSpec.hs
```diff
@@ -0,0 +1,182 @@
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Stack.LockSpec where
```
snoyberg: Very happy to see this

subs/pantry/src/Pantry.hs
```diff
@@ -396,7 +399,7 @@ loadCabalFileImmutable
 loadCabalFileImmutable loc = withCache $ do
   logDebug $ "Parsing cabal file for " <> display loc
   bs <- loadCabalFileBytes loc
-  let foundCabalKey = BlobKey (SHA256.hashBytes bs) (FileSize (fromIntegral (B.length bs)))
+  let foundCabalKey = bsToBlobKey bs
```
snoyberg: Good helper function

src/test/Stack/LockSpec.hs
qrilka: Thanks to @psibi :)

qrilka:
@snoyberg please take another look, especially into the doc

doc/lock_files.md
```diff
@@ -206,36 +177,8 @@ If the lock file does not exist, it will be created by:
...
-3. For each incomplete package location:
-    * Lookup the value in the map created in (1)
-    * If present: use that completed information
-    * Otherwise: complete the information using the same completion
-      procedure from Pantry as in "creation"
-
-This should minimize the number of changes to packages incurred.
+When loading a Stack project all completed package or snapshot locations
+(even when they were completed using information from a lock file) get
+collected to form a new lock file.
```
snoyberg: > to form a new lock file

Can you clarify this a bit? I'm guessing the logic is "form a new lock file in memory and compare against the one on disk, writing if there are any differences." Is that the case?

qrilka: Yes, I'll add this

qrilka: done

PR #4750, opened 2019-04-16
Add configure-options settings (fixes #1438)

### LINKED ISSUE #1438: Add `configure-options` option
I'm opening this issue discuss the possibility of adding a `configure-options` option, to add arguments passed to `Setup.hs configure`, as mentioned in #846 (among others).

I'm broadly in favour of the idea, with the following misgivings and design considerations.

Stack may not be using the Cabal library forever, in which case these options would no longer work.  But moving away from Cabal is a ways off (if ever), and I think in that case we'd just bite the bullet and remove this option.

Some configure options must apply to all packages in a snapshot (e.g. `--enable-library-profiling` needs all packages to be built with that flag, and likely similar for `--enable-split-objs`), while others will work when only only to local packages or to targets.  `ghc-options` can already suffer from similar problems and they're not handled very well right now, but it's more acute with `configure-options`.   We'd need to be able to indicate which packages the configure-options apply to (e.g., all packages including snapshots, all local packages, or specific packages).

The alternative is that, when need for a new configure option is identified (e.g. in #1284 and #593), we add it to Stack explicitly.  Since configure arguments will normally fall into two or three categories, it should be possible to modify the current design to make it quite easy to add these whenever the need comes up.

For either approach, Stack's current design doesn't handle the snapshots case very well, but extensible snapshots (#863, #1265) should fix that.

ping @mgsloan and @snoyberg

snoyberg:
Addresses #1438

Two points for discussion on this PR:

* Should we also be adding `configure-options` to Pantry snapshots, like we have support for `ghc-options`?
* Should we attempt to fold the `ghc-options` values into this field? I think not, since some commands use `ghc-options` but don't use `configure-options` (like `ghci`)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -60,6 +60,9 @@ Major changes:
       is uniquely identified by a commit id and an Hadrian "flavour" (Hadrian is
       the newer GHC build system), hence `compiler` can be set to use a GHC
       built from source with `ghc-git-COMMIT-FLAVOUR`
+* `stack.yaml` now supports a `configure-options`, which are passed directly to
+  the `configure` step in the Cabal build process. See
+  [#1438](https://github.com/commercialhaskell/stack/issues/1438)
 
 Behavior changes:
 * `stack.yaml` now supports `snapshot`: a synonym for `resolver`. See [#4256](https://github.com/commercialhaskell/stack/issues/4256)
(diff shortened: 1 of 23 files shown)


qrilka:
Regarding that these options are quite similar to `ghc-options` it looks to make sense to have them supported in snapshots as well. The other question is that they probably won't get used much.
As for merging `ghc-options` into this field I agree that probably we should keep them separate and it's user's responsibility to check for any conflicts.
The PR looks good to me but I guess it's better to add support to snapshots as a part of it?

snoyberg:
@qrilka and I followed up in a private chat about this, and decided that we don't want to require tools that process snapshot files to support arbitrary configure options right now. So we're going to leave this PR as-is.

PR #4762, opened 2019-04-22
Add timestamp during conversion

snoyberg:
Pairs up with #4761


snoyberg:
It's based on a much older version of the code that performed some weird newline conversions on hashes. It's overhead that was only necessary for the conversion tool, so I haven't kept that code in the master branch.

PR #4770, opened 2019-04-23
MX linux support

rkaippully:

Tested this by running the script against MX Linux 18.2 and Debian Stretch.

During testing I noticed that the dpkg-query check in apt_get_install_pkgs always assumes
that the packages are installed even when they aren't. This happens only on MX and not on
Debian and is probably a subtle behavior change between the /bin/sh versions in MX and Debian testing (I am not entirely sure though). This check was introduced in commit b46e5a7 a few days
back, reverting that to the previous version works fine in MX and Debian stretch. Please let me know if this is a problem, I'll try to find an alternative fix.

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -611,7 +611,8 @@ try_install_pkgs() {
 apt_get_install_pkgs() {
   # Grepping the output of dpkg-query and checking the overall status of the pipeline
   # causes differing behaviors across various shell versions. So it is more portable to
-  # just check the staus of dpkg-query
+  # just check the staus of dpkg-query.
+  # See https://github.com/commercialhaskell/stack/pull/4770#pullrequestreview-229732166
   if dpkg-query -W "$@" > /dev/null 2>&1; then
     info "Already installed!"
   elif ! sudocmd "install required system dependencies" apt-get install -y ${QUIET:+-qq} "$@"; then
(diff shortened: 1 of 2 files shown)


etc/scripts/get-stack.sh
```diff
@@ -609,7 +609,7 @@ try_install_pkgs() {
 
 # Install packages using apt-get
 apt_get_install_pkgs() {
-  if ! dpkg-query -W "$@"|grep -v '^\S\+\s\+.\+$' > /dev/null; then
+  if dpkg-query -W "$@" > /dev/null; then
```
borsboom: Any particular reason you changed this?  I believe the reason for the `grep` was to make it detect if only _some_ of the required packages are missing.  That said, it could be more intelligent by only trying to install the actually missing dependencies (rather than all of them, even if only some are missing).

rkaippully: See my comment above. This is somehow broken in the version of `/bin/sh` that MX uses (which is based on Debian stable). I can try to figure out a better way to handle it.

borsboom: Ok, sorry I didn't read the description well enough.  I'm OK with the revert, but how about adding a comment with what you reverted and why (and a link to this discussion) so that if we change it again we won't accidentally break MX.

rkaippully: I added a comment and verified this in both Debian and MX.

borsboom:
Looks good, thanks!

PR #4771, opened 2019-04-24
Nicer DB locked messages

snoyberg:
* Include indication of whether it's Pantry or Stack
* Don't mention anything for the first 5 seconds to avoid spamming the
user

Since there are now two different delays involved, I've extracted the
"talkUntil" logic to its own function. In theory, this could be added as
a library function with some more cleanup.

### DIFF
--- a/subs/pantry/src/Pantry/SQLite.hs
+++ b/subs/pantry/src/Pantry/SQLite.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE RankNTypes #-}
 module Pantry.SQLite
   ( Storage (..)
   , initStorage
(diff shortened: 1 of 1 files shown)


subs/pantry/src/Pantry/SQLite.hs
```diff
@@ -69,33 +71,84 @@ initStorage description migration fp inner = do
...
+            complainer delay = run $ do
+              -- Wait five seconds before giving the first message to
+              -- avoid spamming the user for uninteresting file locks
+              delay $ 5 * 1000 * 1000 -- 5 seconds
+              logInfo $ "Unable to get a write lock on the " <> desc <> " database, waiting..."
+
+              -- Now loop printing a message every 1 minute
+              forever $ do
+                delay (60 * 1000 * 1000) -- 1 minute
+                  `onException` logInfo ("Acquired the " <> desc <> " database write lock")
```
lehins: This is minor, and not sure if it's worth the trouble, but it might be a good idea to do `withException` and only print the message about acquired lock whenever it is a `DoneTalking` exception. Otherwise upon async exception it will also print that it has acquired the database lock, which can be misleading.

Issue #3588, opened 2017-11-19
Terminal width detection isn't done on Windows


kadoban:
The terminal width detection (that is used for pretty-printing log/warning/error output) is not done on Windows.

If anyone is interested in contributing that, it should be fairly trivial for the right person. It should go in ```windows/System/Terminal.hs``` and do something similar in function to what is done in ```unix/System/Terminal.hsc```

mgsloan:
@eye942 Approach (1) seems good to me.  I don't think anyone is working on this, feel free to have at it

ndmitchell:
I recommend the terminal-size library which does this in a cross platform manner. It's used in ghcid with much success.

snoyberg:
Fixed by #4797

PR #4779, opened 2019-04-28
Implement getTerminalWidth for Windows

NeonGraal:

## Tests
I temporarily added a `putStrLn "!!!"` just before the `getWidth` call in `getTerminalWidth` to be sure that the function was being called.
I then ran the modified stack with `stack run stack -- ls stack-color` and got the "!!!"
To check the non-console case I ran the above command with the stdout and stderr redirected to files


NeonGraal:
Resolve [3588](https://github.com/commercialhaskell/stack/issues/3588)

snoyberg:
The last comment in the referenced issue (#3588) mentions the terminal-size library. Is there a reason you didn't go with that approach? My concern here is that this change introduces lower bounds on a few packages that ship with GHC, which we typically try to avoid doing.

NeonGraal:
Ah. I went the way of upgrading an existing dependency rather than including a new one. Didn't realise the lower bounds issue.

At first look using the `terminal-size` package may actually allow me to delete some code.

Which would you prefer, I commit a revert here and then another with the new change or cancel this PR and raise a new one.

snoyberg:
No strong preference to be honest, take your pick. If you pushed me for an answer, I'd probably lean towards new PR.

NeonGraal:
Implemented better in [https://github.com/commercialhaskell/stack/pull/4797]

PR #4781, opened 2019-04-28
Fix #4778: autorun `autoreconf` on Windows too

### LINKED ISSUE #4778: On Windows: Automatically run autoreconf -i when needed
I think this is a stack feature request (and not a bug). It may be a bug if https://github.com/commercialhaskell/stack/pull/3598 was intended to work on Windows as it does on Unix-like operating systems.

In a clean clone of the `process` library from GitHub, on Windows 10 with `stack --version` `Version 2.0.0, Git revision dfbf85ad7e8af5b01cf7b51367290870ffc2c90e (7456 commits) x86_64 hpack-0.31.1` , `stack build` fails with (extracts):
~~~~
Trying to generate configure with autoreconf in C:\Users\mikep\Documents\Code\GitHub\process\
Unable to run autoreconf: Executable named autoreconf not found on path: ...

Configuring process-1.6.5.0...
Warning: The 'build-type' is 'Configure' but there is no 'configure' script.
You probably need to run 'autoreconf -i' to generate it.
setup.exe: configure script not found.
~~~~
Following up, I discovered https://github.com/commercialhaskell/stack/pull/3598. I think that feature does not work on Windows because `autoreconf` is a shell script file, not an executable (I get the same error message even if the `autoreconf` file is on the path).

My feature request is to make https://github.com/commercialhaskell/stack/pull/3598 work for Windows users of stack. That is, automatically run the `autoreconf -i` script when needed. That would require the stack MSYS2 to come with `autoconf` and `perl` (see below).

I could build `process` on Windows using stack if I did the following:

1. Add both `autoconf` and `perl` to stack's MSYS2, using `stack exec pacman -- -S`. stack's version of MSYS2 does not come with them as standard.
2. Execute `stack exec sh -- autoreconf -i`.
3. Execute `stack build`.

mpilgrem:
In module `Stack.Setup`, `pacman` is used to add the `perl` and `autoconf` packages to stack's MSYS2.

In module `Stack.Build.Execute`, the code uses command `sh autoreconf -i` for Windows, instead of just `autoreconf -i`.

The implementation involves two hacks:

1. `pacman` and `sh autoreconf -i` seem to clear and not restore ANSI capability on native Windows terminals. A hack restores the ANSI capability (using the same hack that was once used for `git clone` - fix #4121).
2. MSYS2's current `perl` package (5.28.1-2) is broken. So the `perl` installed by stack is downgraded to 5.28.0-1, which works. This will need to be revisited once `perl` is fixed on MSYS2.

This has been tested on Windows 10 (only) by deleting stack's MSYS2 (so that stack reinstalls it) and by then building the `process` package (which makes use of `configure`) using stack.

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md None, I think.
* [x] The documentation has been updated, if necessary. Not necessary, I think.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -71,6 +71,7 @@ import           Stack.Config
 import           Stack.Constants
 import           Stack.Constants.Config
 import           Stack.Coverage
+import           Stack.DefaultColorWhen (defaultColorWhen)
 import           Stack.GhcPkg
 import           Stack.Package
 import           Stack.PackageDump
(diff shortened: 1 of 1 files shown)


src/Stack/Setup.hs
```diff
@@ -1483,6 +1484,28 @@ installMsys2Windows osKey si archiveFile archiveType _tempDir destDir = do
     menv <- mkProcessContext newEnv
     withWorkingDir (toFilePath destDir) $ withProcessContext menv
       $ proc "sh" ["--login", "-c", "true"] runProcess_
+    -- Install MSYS2 package perl. Used by the autoreconf shell script file.
+    -- perl 5.28.1-2 is broken (see
+    -- https://github.com/msys2/MSYS2-packages/issues/1611). perl 5.28.0-1
+    -- works, so downgrading to that version, in the alternative.
+    withWorkingDir (toFilePath destDir) $ withProcessContext menv
+    --  $ proc "pacman" ["-Sy", "--noconfirm", "perl"] runProcess_
+      $ proc
```
snoyberg: We've had bad experience in the past with including automated installation of pacman packages, specifically with Git before. I'd rather address this with documentation and perhaps a warning from Stack when the `autoreconf` call fails on Windows.

mpilgrem: Ok. I’ll revise the pull request accordingly, in due course.

mpilgrem: Pull request revised.

src/Stack/Build/Execute.hs
```diff
@@ -911,8 +912,40 @@ ensureConfig newConfigCache pkgDir ExecuteEnv {..} announce cabal cabalfp task =
...
+            logInfo "  stack exec where -- autoreconf"
+            logInfo ""
+            logInfo $ "If perl or autoreconf is not on the path in the " <>
+              "required location, add them with command (note that the " <>
+              "relevant package name is 'autoconf' not 'autoreconf'):"
+            logInfo ""
+            logInfo "  stack exec pacman -- --sync --refresh autoconf"
+            logInfo ""
+            logInfo $ "Some versions of perl from MYSY2 are broken. See " <>
+              "https://github.com/msys2/MSYS2-packages/issues/1611. To test " <>
```
snoyberg: This all looks great. Only other slight tweak: can you also add a link to this PR so that people can get a bit more context from Stack itself?

mpilgrem: In the latest version of the pull request, I have added a link to the PR as you request.

PR #4782, opened 2019-04-29
Fix freeze test (publish-date field was added to snapshots)

qrilka:
`publish-date` was added to snapshots and thus hashes and file sizes are now different.
Currently we have "reverse" failure on Azure because of S3 cache - https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=1483
We need to clear caches before merging this


qrilka:
freeze test looks to be OK, the remaining failures appear to be about missing 13.19 in stackage-snapshots

Issue #4789, opened 2019-04-30
Warnings when parsing lock file


snoyberg:
Actions:

1. Using a `stack` executable from commit 595e9eade0511ca046215ee0b7593b5ca67c71ce
2. `rm -f stack.yaml.lock`
3. Run `stack build` in the Stack repo at that same commit
4. Get some warnings

```
$ stack build 
Using archive from https://github.com/snoyberg/filelock/archive/97e83ecc133cd60a99df8e1fa5a3c2739ad007dc.tar.gz without a specified cryptographic hash
Cached hash is 78332e0d964cb2f24fdbb6b07c2a6a84a029c4fe540a0435993c85ad58eab051, file size 9526
For security and reproducibility, please add a hash and file size to your configuration
Using archive from https://github.com/snoyberg/filelock/archive/97e83ecc133cd60a99df8e1fa5a3c2739ad007dc.tar.gz without a specified cryptographic hash
Cached hash is 78332e0d964cb2f24fdbb6b07c2a6a84a029c4fe540a0435993c85ad58eab051, file size 9526
For security and reproducibility, please add a hash and file size to your configuration
```

The missing cryptographic hash isn't true: the `stack.yaml.loc` does in fact include the `sha256` and filesize of the tarball. Running `stack build` a second time generates slightly different warnings:

```
$ stack build 
Warning: /Users/michael/Documents/stack/stack.yaml.lock: Unrecognized fields in UnresolvedPackageLocationImmutable.PLIArchive: cabal-file, name, pantry-tree, version
Using archive from https://github.com/snoyberg/filelock/archive/97e83ecc133cd60a99df8e1fa5a3c2739ad007dc.tar.gz without a specified cryptographic hash
Cached hash is 78332e0d964cb2f24fdbb6b07c2a6a84a029c4fe540a0435993c85ad58eab051, file size 9526
For security and reproducibility, please add a hash and file size to your configuration
```

Looks like this might be a bug in the parsing or rendering of a `PLIArchive`. CC @qrilka

snoyberg:
I've opened up a PR (#4790) for the warnings during the parsing. That addresses "unrecognized fields" bit above by using the `Data.Aeson.Extended` machinery better.

I've opened up a second PR (#4791) for the warnings about "without specified cryptographic hash." The issue there was that we were looking up the package name from the raw information, instead of using the completed information.

Once both PRs are merged we can close this issue.

snoyberg:
Both PRs and merged, closing

PR #4790, opened 2019-04-30
Fix some misplaced parse warnings #4789

snoyberg:
This fixes the warnings about extraneous fields, by having `PackageMetadata` content parsed within the `WarningParser` which tracks which fields are used. This does _not_ fully address #4789, please do not close that issue when merging this.

### DIFF
--- a/subs/pantry/src/Pantry/Types.hs
+++ b/subs/pantry/src/Pantry/Types.hs
@@ -506,15 +506,6 @@ instance Display Repo where
     (if T.null subdir
       then mempty
       else " in subdirectory " <> display subdir)
-instance FromJSON Repo where
-    parseJSON =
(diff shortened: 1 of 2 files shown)


snoyberg:
Looking through this, it appears to me that we can simply remove the `logWarn` calls in `Pantry.Archive` which are generating the rest of the warnings, since lock files fully address this. Optionally, we could take a parameter to indicate whether we should warn on these, for use cases outside of Stack which don't have lock files, but that seems excessive. @qrilka what do you think?

snoyberg:
I've added an additional commit which addresses a _different_ `FromJSON` instance parsing bug. I'm including it here since the test suite changes overlap with what is already present in this PR.

PR #4791, opened 2019-05-01
Get package name from completed information #4789

snoyberg:
The completed package information already contains the package name,
bypassing the need for a package completion call if the cache (lock
file) already contains that information. This bypasses a spurious
warning about lacking cryptographic hashes, and likely improves
performance, for the repo and archive cases.

This wasn't discovered initially because the Hackage use case never had
an overhead: the specification of a RPLIHackage already contains the
package name in all cases.

### DIFF
--- a/subs/pantry/src/Pantry/Types.hs
+++ b/subs/pantry/src/Pantry/Types.hs
@@ -107,6 +107,7 @@ module Pantry.Types
   , SnapshotCacheHash (..)
   , getGlobalHintsFile
   , bsToBlobKey
+  , pliIdent
   ) where
 
 import RIO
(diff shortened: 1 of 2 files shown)


subs/pantry/src/Pantry/Types.hs
```diff
@@ -2269,7 +2270,13 @@ getGlobalHintsFile = do
...
+-- @since 0.1.0.0
 bsToBlobKey :: ByteString -> BlobKey
 bsToBlobKey bs =
     BlobKey (SHA256.hashBytes bs) (FileSize (fromIntegral (B.length bs)))
+
+-- | Identifier from a 'PackageLocationImmutable'
+pliIdent :: PackageLocationImmutable -> PackageIdentifier
+pliIdent (PLIHackage ident _ _) = ident
+pliIdent (PLIArchive _ pm) = pmIdent pm
+pliIdent (PLIRepo _ pm) = pmIdent pm
```
qrilka: @snoyberg there's already `packageLocationIdent`  -https://github.com/commercialhaskell/stack/blob/38826df905752a9c708c085f021c3c3b6d592879/subs/pantry/src/Pantry.hs#L1337

snoyberg: Doh, thanks. I'll send a PR to fix that.

PR #4793, opened 2019-05-01
Support symlinks to directories (fixes #4776)

### LINKED ISSUE #4776: Unable to find symlink destination
### General summary/comments

`stack` on `master` seems to be having trouble locating a symlink target that definitely does exist.

```
colin@yumi ~/c/h/chainweb> stack test
Cloning 37ad6ebb25c975c3c0ab9d5554cd5fb6d4809da8 from https://github.com/kadena-io/pact.git
Unsupported tarball from /tmp/with-repo-archive16624/foo.tar: Symbolic link dest not found from docs/en/img to ../img, looking for docs/img.
```

`docs/img/` does exist (it's a directory). 

### Steps to reproduce

For example:

1. Clone [this branch](https://github.com/kadena-io/chainweb-node/pull/172)
2. `stack build`
3. See the error I pasted above. 

### Expected

The pinned dependency on Pact is pulled and built.

### Actual

`stack` fails somewhere between pulling and building.

### Stack version

```
Version 2.0.0, Git revision dfbf85ad7e8af5b01cf7b51367290870ffc2c90e x86_64 hpack-0.31.1
```

### Method of installation

A very recent `stack upgrade --git`.

snoyberg:
CC @fosskers. I believe this addresses the symlink issue. There's _also_ an issue with parsing lock files which seems to be getting triggered here, which I'm looking into as well.


fosskers:
I'll test this out.

fosskers:
Seems to work :) Thanks

snoyberg:
Awesome!

PR #4797, opened 2019-05-04
Win terminal size

NeonGraal:


NeonGraal:
Tests
I temporarily added a `putStrLn "!!!"` just before the getWidth call in getTerminalWidth to be sure that the function was being called.
I then ran the modified stack with `stack run stack -- ls stack-colors` and got the "!!!"
To check the non-console case I ran the above command with the stdout and stderr redirected to files

NeonGraal:
Resolve #3588

Issue #4426, opened 2018-12-02
Received ExitFailure 1 when running stack setup


AlexGoico:
### General summary/comments

I am on MacOS Mojave (10.14) and am unable to get past stack setup.

### Steps to reproduce

1. Create a project `stack new projectName`
2. Inside projectName, run `stack setup`

### Expected

ghc to be installed into the project

### Actual

```
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Already downloaded.                
Configuring GHC ...                                                             
Received ExitFailure 1 when running
Raw command: /Users/alexgoico/.stack/programs/x86_64-osx/ghc-8.4.4.temp/ghc-8.4.4/configure --prefix=/Users/alexgoico/.stack/programs/x86_64-osx/ghc-8.4.4/
Run from: /Users/alexgoico/.stack/programs/x86_64-osx/ghc-8.4.4.temp/ghc-8.4.4/
```

Here is `stack setup --verbose` [output](https://gist.github.com/AlexGoico/1e8cbffe3ad2c66780d028821078bc0f).

### Stack version

```
$ stack --version
Version 1.9.1, Git revision f9d0042c141660e1d38f797e1d426be4a99b2a3c (6168 commits) x86_64 hpack-0.31.0
```

### Method of installation

* Haskell Platform x64 (https://www.haskell.org/platform/mac.html)

mpereira:
I see the same on 1.9.3. The `config.log` shows a few errors:
```
This file contains any messages produced by compilers while
running configure, to aid debugging if configure makes a mistake.

It was created by The Glorious Glasgow Haskell Compilation System configure 8.6.3, which was
generated by GNU Autoconf 2.69.  Invocation command line was

  $ /Users/murilo/.stack/programs/x86_64-osx/ghc-8.6.3.temp/ghc-8.6.3/configure --prefix=/Users/murilo/.stack/programs/x86_64-osx/ghc-8.6.3/

## --------- ##
## Platform. ##
## --------- ##

hostname = balloon
uname -m = x86_64
uname -r = 18.2.0
uname -s = Darwin
uname -v = Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64

/usr/bin/uname -p = i386
/bin/uname -X     = unknown

/bin/arch              = unknown
/usr/bin/arch -k       = unknown
/usr/convex/getsysinfo = unknown
[...]

PR #4803, opened 2019-05-05
Add note about installing SDK headers on macOs >= 10.14

basti1302:
Adds a note about installing additional headers on more recent macOs versions, when `xcode-select --install` is not enough. Follow up from https://github.com/commercialhaskell/stack/issues/4426#issuecomment-488641809


basti1302:
Sorry, accidentally created a duplicate.

Issue #4795, opened 2019-05-02
DISCUSS: Policy on including lock files in repo


snoyberg:
Now that lock files have landed (yay!), we need to establish a policy for the Stack repo on how to deal with them. We should also consider whether we make this policy a general recommendation for Stack users. Possible options I see:

1. Don't check in the lock files at all, no other change. I don't like this: it means we lose out on reproducible build plans.
2. Obviate the need for lock files by insisting that all additional hashes be included in the snapshot files. I don't like this either: the new format is backwards incompatible with Stack <2 (at least producing warnings), and adding all of the hashes manually is a pain.
3. Check in the lock files in the repo. This is what I favor. However I'd like to take it one half-step further: add a step to the CI scripts to ensure that no new lock files have been generated during the build, and that no existing lock files have been modified. This ensures that the fully reproducible plan is checked into the repo.

CC @borsboom @qrilka @psibi

borsboom:
Our recommendation should probably be similar to that used by Rust's `cargo` tool: [check in lock files for binaries, but not for libraries](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

I agree it'd make sense to enforce no addition/modification of lock files in CI. Not sure if you already have thoughts on how to implement that, but here's what I'm coming up with: run something like `find . -name '*.lock' -print0|xargs -0 shasum -a 256` before and after the build, and compare the output.  Alternatively, could just check git for any dirty or untracked/unignored files after the build, which would also catch other mistakes (like forgetting to update `.gitignore`), but might also lead to false positives.

Might it make sense to add an option to `stack` that will cause it to fail if it would need to add or update lock files?  I could see this coming in handy for lots of projects, and the shell script logic above is a bit hairy.

snoyberg:
> Alternatively, could just check git for any dirty or untracked/unignored files after the build, which would also catch other mistakes (like forgetting to update .gitignore), but might also lead to false positives.

I was leaning in this direction. We may actually consider a forgotten .gitignore update to be a proper bug that blocks CI from passing.

> Might it make sense to add an option to stack that will cause it to fail if it would need to add or update lock files? I could see this coming in handy for lots of projects, and the shell script logic above is a bit hairy.

I thought about this too. It should be pretty easy to add such a feature, essentially checking a flag before writing the lock file. I don't see a downside to this, so we may as well add it. How about having a `--lock-file` flag with a bunch of different options:

[...]

PR #4811, opened 2019-05-07
Add new --lock-file flag #4795

snoyberg:


src/Stack/Options/GlobalParser.hs
```diff
@@ -90,6 +94,12 @@ globalOptsFromMonoid defaultTerminal GlobalOptsMonoid{..} = do
     , globalStylesUpdate = globalMonoidStyles
     , globalTermWidth = getFirst globalMonoidTermWidth
     , globalStackYaml = stackYaml
+    , globalLockFileBehavior =
+        let defLFB =
+              case (getFirst globalMonoidResolver, getFirst globalMonoidCompiler) of
```
qrilka: Does compiler override change anything in this case?
If we talk about getting different packages in a source map then extra targets on command line could more easily give you a different source map though neigther `LFBReadOnly` nor `LFBReadWrite` look meaningful in this case

snoyberg: Maybe I'm being overly cautious about the compiler. I think you're right: I can't think of anything that would cause lock files to behave differently there.

I was under the (perhaps mistaken) impression that extra targets on the command line would not affect the source map here. If they do, you're right, that's the bigger concern. Is there an easy way at this stage of processing to know whether there are extra targets?

qrilka: It appears that it was me who got confused by different source maps that we have :)
In `SMWanted` we don't have any targets yet so it looks like resolver is the only option that could cause problems here.

Issue #4788, opened 2019-04-30
`stack haddock` no longer rebuilds docless deps


fosskers:
### General summary/comments (optional)

In `1.9.3`, when moving to a new LTS, if one **hadn't** specifically built their project and all its deps via:
```
stack build --test --haddock
```
then the first time `stack haddock` is invoked after an initial vanilla `stack build`, all deps are rebuilt from scratch with their Haddocks included for linking. This used to cost a lot of time, so I've been very careful when bumping snapshots. 

New scenario: I'm on the `master` branch of `stack`, in a project that only ever had a naked `stack build` ran on it (so, no docs included). Now I notice that `stack haddock` will run just fine, but quickly warns me that since none of the deps have their docs built, it won't attempt to link to them (which is arguably the entire point of `stack haddock --open`, and `stack hoogle --server`, etc.). Nor does it rebuild them all, like it used to do. 

### Steps to reproduce

1. Have an LTS you haven't pulled the deps from yet.
2. Perform `stack build` in a project that uses that LTS.
3. Perform `stack haddock`, and notice it warn about missing dep docs.

### Expected

Perhaps `stack build` should always invoke Haddock for deps, just so the links are always there?

### Actual

My project docs don't link to any external dependencies. Is there a way for me to rebuild my deps, without nuking my whole `~/.stack/`?

### Stack version

```
> stack --version
Version 2.0.0 (7456 commits) x86_64 hpack-0.31.2
```

### Method of installation

Self-built from `master`.

snoyberg:
Potentially related: #4754

qrilka:
@fosskers is this the warning you're seeing:
```
Warning: The documentation for the following packages are not installed. No
links will be generated to these packages: async-2.2.1, exceptions-0.10.2,
hashable-1.2.7.0, microlens-0.4.10, primitive-0.6.4.0, rio-0.1.8.0,
transformers-compat-0.6.4, typed-process-0.2.4.0, unliftio-0.2.10,
unliftio-core-0.1.2.0, unordered-containers-0.2.9.0, vector-0.12.0.3
```
?

fosskers:
Yes

snoyberg:
@fosskers I've just merged #4813. I'm going to close this out, if the issue continues, please comment and I'll reopen

fosskers:
I'm on `master`, but not noticing any dep rebuild when I `stack haddock` in an established project.

fosskers:
[...]

Issue #4754, opened 2019-04-18
Stack master: does not build haddocks with --no-haddock-deps after successful build


snoyberg:
This is a regression on Stack master, tested with `ddca714c73701fcefd0d4493364f6661a543e73f`.

With Stack 1.9.3:

```
stack-1.9.3 new foo --resolver ghc-8.6.4 && cd foo && stack-1.9.3 build && stack-1.9.3 haddock --no-haddock-deps --open
```

results in the Haddocks for the library being generated. However, for Stack master:

```
stack new foo --resolver ghc-8.6.4 && cd foo && stack build && stack haddock --no-haddock-deps --open
```

results in the Haddocks not being generated. This is likely a relatively simple dirtiness checking bug, maybe related to configure caches, but I'm not certain.

Marking as P1, but not investigating myself immediately. The codepaths may change significantly for component-based builds, so I'd rather see that land first.

qrilka:
It is a dirtiness check bug and `--no-haddock-deps` doesn't play any role here, I remember somewhat how I deleted haddocks check but it looks like it wasn't a correct decision.

snoyberg:
Addressed by #4813, closing.

PR #4813, opened 2019-05-07
4754 rebuild haddocks

qrilka:
Properly checks haddock usage in config caches and precompiled caches


qrilka:
Fixes #4754 
Fixes #4788

Issue #4796, opened 2019-05-02
`--file-watch` no longer notices `stack.yaml` changes


fosskers:
### General summary/comments (optional)

One workflow of mine is that while I'm pinning a certain dep in my `stack.yaml`, I'll open up `stack build --file-watch` in case it warns me about subdeps that also need pinning. 

I've just hit a case on Stack's `master` branch, where after pinning a dep and saving my buffer, `--file-watch` spins once (apparently noticing the file change), but then warns to add the dep I just added. Turning off `--file-watch` and manually running `stack build` over and over each for each subdep works as expected.

### Steps to reproduce

1. Create a new project with the LTS set to `lts-12.26` (old on purpose).
2. Pin `deque-0.4.2.1` in `extra-deps`.
3. `stack build --file-watch`, and be warned of the missing `strict-list` dep.
4. Add the dep it indicates, and save `stack.yaml`
5. Notice no change in the output of `--file-watch`.
6. Leave `--file-watch` and do `stack build`
7. Notice the warning about `semigroupoids`.

### Expected

I'd be further prompted to pin a `semigroupoids` dep.

### Actual

I'm prompted again to pin `strict-list`.

### Stack version

```
> stack --version
Version 2.0.0, Git revision 91fc08023314431e7e56c22897790b6429767d90 (dirty) (7495 commits) x86_64 hpack-0.31.2
```

### Method of installation

Self-built from Stack's `master` branch.

snoyberg:
Thanks for the report @fosskers, please see https://github.com/commercialhaskell/stack/pull/4820.

fosskers:
Thanks @snoyberg !

snoyberg:
PR merged to the RC branch, closing

PR #4820, opened 2019-05-12
Reload config files with --file-watch

snoyberg:
@borsboom This addresses #4796. Since it's a bug fix, it should be included in this release, even though we're after code freeze. Can you confirm this is the right branch to merge to (v2.1), and review/merge if so?


borsboom:
Yes, that's the right branch.

PR #4828, opened 2019-05-15
Perform CI on the v2.1 branch

snoyberg:
CC @psibi @borsboom. It makes me wonder if we need a different release
branch naming scheme so that we can have the default be to check all PRs
against release branches.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,6 +5,7 @@ trigger:
     - ci/*
     - master
     - stable
+    - v2.1
 
 variables:
   BASE_BRANCH: "master"


psibi:
@snoyberg Maybe we can have release branches like this:

```
release/2.1
```

Then all we need is this rule in the azure config for it to work:

```
release/*
```

snoyberg:
It's a nice PR in theory, but the CI still didn't run on it 😛. Any idea why?

psibi:
@snoyberg I didn't notice that. After thinking some time - I think you have to merge this PR to make future PR to `v2.1` branch make CI run. The reason being the current state of branch `v2.1` states that CI will not be invoked. At least that's what my guess is.

borsboom:
@psibi I'm sure you are correct about why it's not running.

I also agree that having a convention for the name of release candidate branches that automatically have them trigger CI is a good idea.  I'd propose `rc/*`, rather than hard-coding `v2.1`.  I'm happy to move the RC branch over to `rc/v2.1` to make this work.

Also, it would be good to have the nightly integration tests run on these.  It's not clear to me how to configure that though (`azure-nightly-pipeline.yml` doesn't seem to mention the branches).

psibi:
@borsboom I have created #4833 to enable CI for both unit and integration tests on `rc/*` branches.

snoyberg:
Looks like #4833 supersedes this, closing.

PR #4833, opened 2019-05-18
Update Azure CI rules for RC release

psibi:
This PR enables following:
* Any PR or Push to `rc/*` branch will trigger a Integration test build.
* Any PR or Push to `rc/*` branch will trigger a Unit test build.

### DIFF
--- a/azure-nightly-pipeline.yml
+++ b/azure-nightly-pipeline.yml
@@ -1,8 +1,4 @@
-trigger:
-  batch: true
-  branches:
-    include:
-    - rc/*
+trigger: none
 pr:
   autoCancel: true
   branches:
--- a/azure-nightly-pipeline.yml
+++ b/azure-nightly-pipeline.yml
@@ -1,5 +1,13 @@
-trigger: none
-pr: none
+trigger:
+  batch: true
+  branches:
+    include:
+    - rc/*
+pr:
+  autoCancel: true
+  branches:
+    include:
+    - rc/*
 
 variables:
   BASE_BRANCH: "master"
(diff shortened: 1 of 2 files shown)


borsboom:
I don't think pushes to `rc/*` branches should *trigger* the nightly integration tests.  Like with the `master` and `stable` branches, they should just run nightly.

psibi:
@borsboom I have pushed a commit which modifies the behavior. Any pushes to `rc/*` branch will *not* trigger the integration build. But any PR to `rc/*` branch will trigger the integration build. I hope that's the behavior we wanted. Also, it will have the scheduled tests running as usual.

I also tested that this configurations works by creating a sample repository as it wasn't obvious to me.

borsboom:
I don't think we run integration tests on PRs against `master` and `stable` do we?  IIUC the integration tests are just run once per night on the branch itself (not PRs).  I think the `rc/*` branches should work just like `stable` and `master` when it comes to the azure-nightly-pipelines.

psibi:
@borsboom I think I misunderstood the use case.

> I don't think we run integration tests on PRs against master and stable do we? 

No.

>  I think the rc/* branches should work just like stable and master when it comes to the azure-nightly-pipelines.

I have enabled nightly integration tests for `rc/*` branch ( through their UI, it's not supported via yaml config). But for it to get properly tested, you have to created `rc/*` branch. I'm closing this PR.

borsboom:
I've moved the `v2.1` branch to `rc/v2.1`, so we should see it run the integration tests nightly now.

psibi:
@borsboom We seem to have two rc based branches now:

* rc/v2.1
* rc/1.9.3.1

Do we need `rc/1.9.3.1` ? Not sure if the tests will run there as it seems to be outdated and has no azure configuration files in that branch.

borsboom:
I've removed `rc/1.9.3.1`.

PR #4847, opened 2019-05-30
Compare package names to be pruned correctly

### LINKED ISSUE #4845: stack ls dependencies doesn't prune anymore
### General summary/comments

This is a bug in `stack 2.1.0.1 rc`. It is not a problem in `1.9.3`.
The `pruneGraph` function in `Stack.Dot` seems to have regressed. The [tests](https://github.com/commercialhaskell/stack/blob/aa3a5f2573ffe8791f63e621447c9084d5aa0ba1/src/test/Stack/DotSpec.hs#L54-L59) did not catch this because the properties were either wrong or `show` for `PackageName` is no longer being overridden. I could make the tests fail by replacing `show` with `packageNameString`.

### Steps to reproduce

1. Go to any stack project
1. Run command `stack ls dependencies --no-include-base`. The bug can also be reproduced using `--prune base` and it can also be seen in `stack dot`.

### Expected

`stdout` shows all dependencies except `base` and its dependencies

### Actual

All dependencies including `base` and its dependencies are printed.

### Stack version

```
$ stack --version
Version 2.1.0.1, Git revision 72c9f8c741be762a036ce16b6306d08a8be666ad (7614 commits) RELEASE-CANDIDATE x86_64 hpack-0.31.1
```

### Method of installation

* Official osx binary, downloaded from github releases.

akshaymankar:
Fixes #4845 

**Please also shortly describe how you tested your change. Bonus points for added tests!**

The tests and the implementation both were wrong, I fixed both. However I have found that the behaviour of `pruneGraph` function is a bit weird. But that is how it has been since the beginning, so I am not very sure if I should report a bug. Here are my observations:

1. The `dontPrune` parameter is not always respected. I wrote this property to verify what I thought should've been the behaviour and it does not pass:
    ```haskell
    prop "dontPrune packages are not pruned" $ do
      let resolvedGraph = runIdentity (resolveDependencies Nothing graph stubLoader)
          allPackages g = Set.map packageNameString (Map.keysSet g `Set.union`  fold (fmap fst g))
      forAll (sublistOf (Set.toList (allPackages resolvedGraph))) $ \toPrune ->
        let pruned = pruneGraph [pkgName "one", pkgName "two"] toPrune resolvedGraph
        in if "one" `elem` allPackages pruned
              then if "two" `elem` allPackages pruned
                      then succeeded
                      else failed {reason = "expected `two` to not be pruned but did not find it in pruned packages: " ++ show (allPackages pruned)}
              else failed {reason = "expected `one` to not be pruned did not find it in pruned packages: " ++ show (allPackages pruned)}
    ```
1. If package a depends only on package b, and we're running `stack dot --prune b`, package a will also get pruned. For e.g., Running `stack dot --no-include-base` in stack repository also prunes `array`.

I am not sure if this is expected, if not feel free to create bugs. I will try to fix them.

### DIFF
--- a/src/Stack/Dot.hs
+++ b/src/Stack/Dot.hs
@@ -53,7 +53,7 @@ data DotOpts = DotOpts
     -- ^ Include dependencies on base
     , dotDependencyDepth :: !(Maybe Int)
     -- ^ Limit the depth of dependency resolution to (Just n) or continue until fixpoint
-    , dotPrune :: !(Set String)
+    , dotPrune :: !(Set PackageName)
     -- ^ Package names to prune from the graph
     , dotTargets :: [Text]
     -- ^ stack TARGETs to trace dependencies for
@@ -215,14 +215,14 @@ payloadText opts payload =
 -- unless they are in @dontPrune@
 pruneGraph :: (F.Foldable f, F.Foldable g, Eq a)
            => f PackageName
-           -> g String
+           -> g PackageName
            -> Map PackageName (Set PackageName, a)
            -> Map PackageName (Set PackageName, a)
 pruneGraph dontPrune names =
   pruneUnreachable dontPrune . Map.mapMaybeWithKey (\pkg (pkgDeps,x) ->
-    if packageNameString pkg `F.elem` names
+    if pkg `F.elem` names
       then Nothing
-      else let filtered = Set.filter (\n -> packageNameString n `F.notElem` names) pkgDeps
+      else let filtered = Set.filter (\n -> n `F.notElem` names) pkgDeps
            in if Set.null filtered && not (Set.null pkgDeps)
                 then Nothing
                 else Just (filtered,x))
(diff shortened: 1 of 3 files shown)


src/Stack/Dot.hs
```diff
@@ -220,9 +220,9 @@ pruneGraph :: (F.Foldable f, F.Foldable g, Eq a)
            -> Map PackageName (Set PackageName, a)
 pruneGraph dontPrune names =
   pruneUnreachable dontPrune . Map.mapMaybeWithKey (\pkg (pkgDeps,x) ->
-    if show pkg `F.elem` names
+    if packageNameString pkg `F.elem` names
```
qrilka: What about going other way around? I.e. compare package names and not strings? Not a large difference but adds a bit of type safety

akshaymankar: Good idea!

borsboom:
Thank you!  Can you rebase this against `rc/v2.1` so we can include the fix in the upcoming stack-2.1 release?

akshaymankar:
@borsboom I have rebase the commits and also made this PR merge into `rc/v2.1` branch.

PR #4856, opened 2019-06-08
Documentation: Replace placeholder by internal link

penteract:
Direct readers to the part of the docs where custom snapshots are explained.


borsboom:
Thanks!

PR #4862, opened 2019-06-13
Don't change directories when cloning a repo

amalloy:
This was [making it impossible to pin to a specific commit when using relative file paths as repo URLs](https://stackoverflow.com/a/56573316/625403).

I tested this change by running `stack test` three times:

1. Before I made any changes, tests pass
1. After intentionally implementing my change incorrectly, tests fail
1. After a correct implementation of my change, tests pass again

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -14,6 +14,8 @@ Behavior changes:
 Other enhancements:
 
 Bug fixes:
+* Fix to allow dependencies on specific versions of local git repositories. See
+  [#4862](https://github.com/commercialhaskell/stack/pull/4862)
 
 
 ## v2.1.1.1
--- a/subs/pantry/src/Pantry/Repo.hs
+++ b/subs/pantry/src/Pantry/Repo.hs
@@ -159,12 +159,13 @@ withRepo
   -> RIO env a
   -> RIO env a
 withRepo repo@(Repo url commit repoType' _subdir) action =
-  withSystemTempDirectory "with-repo" $
-  \tmpdir -> withWorkingDir tmpdir $ do
-    let suffix = "cloned"
-        dir = tmpdir </> suffix
-
-    let (runCommand, resetArgs, submoduleArgs) =
+  withSystemTempDirectory "with-repo" $ \tmpDir -> do
+    -- Note we do not immediately change directories into the new temporary directory,
+    -- but instead wait until we have finished cloning the repo. This is because the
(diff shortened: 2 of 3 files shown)


amalloy:
I've included a Changelog update in what I think was the right place, and I believe no documentation update is necessary (already-documented behavior now works better).

ChangeLog.md
```diff
@@ -277,6 +277,8 @@ Bug fixes:
   [#4292](https://github.com/commercialhaskell/stack/issues/4292)
 * Fix for git packages to update submodules to the correct state. See
   [#4314](https://github.com/commercialhaskell/stack/pull/4314)
+* Fix to allow dependencies on specific versions of local git repositories. See
```
snoyberg: This should go in "unreleased changes" instead. It should also likely be mentioned in the `pantry` changelog.

subs/pantry/src/Pantry/Repo.hs
```diff
@@ -176,23 +173,21 @@ withRepo repo@(Repo url commit repoType' _subdir) action =
...
+          when osIsWindows $ void $ liftIO $ hSupportsANSIWithoutEmulation stdout
 
     logInfo $ "Cloning " <> display commit <> " from " <> display url
-    runCommand ("clone" : [T.unpack url, suffix])
-    -- On Windows 10, an upstream issue with the `git clone` command means that
-    -- command clears, but does not then restore, the
-    -- ENABLE_VIRTUAL_TERMINAL_PROCESSING flag for native terminals. The
-    -- folowing hack re-enables the lost ANSI-capability.
-    when osIsWindows $ void $ liftIO $ hSupportsANSIWithoutEmulation stdout
+    runCommand ["clone", T.unpack url, dir]
```
snoyberg: Please add a comment that this is intentionally _not_ run within the `withWorkingDir` below in order to allow relative paths to work.

amalloy:
The hlint failure was on unmodified code in a different file; I'm not inclined to touch that as part of this PR.

I've adjusted the target branch, and made the changes you requested, except: the `pantry` changelog appears to currently be empty, after a year's worth of commits to it. Do you want this to be the first change documented therein?

snoyberg:
> the pantry changelog appears to currently be empty, after a year's worth of commits to it

Right, that's intended: `pantry` has been in development for a year, but only last week had its first release. This is the first actual change versus a released version.

amalloy:
Done.

PR #4869, opened 2019-06-16
Appease hlint

snoyberg:
I still think the original was nicer, but it's easier this way :)


psibi:
I changed some settings in the Azure pipelines dashboard. Closing and re-opening the PR will trigger the test suite build ( Note that this will not trigger the integration test suite ).

PR #4898, opened 2019-06-23
Split up user and project storage (fixes #4893)

### LINKED ISSUE #4893: Move configure information for local packages back to .stack-work to improve caching
Hello!

In `stack` 2.X, it looks like local configuration information for packages was moved to `~/.stack/stack.sqlite3`. This breaks a common pattern for caching Haskell dependencies separate from local builds in CI. Our CI caches used to work like this:
1. Build all Haskell dependencies
2. Cache `~/.stack`
3. Build our local Haskell packages
4. Cache all `.stack-work` directories

Separate caches are really nice because we don't have to pay to re-cache `~/.stack` if deps didn't change, and it is easy to share the dependency caches across branches.

Now that local configuration information for packages is in `~/.stack`, we had to combine our dependency and local caches, which means we have to cache/uncache `~/.stack` (or some yet-undetermined subset of `~/.stack`, maybe just the SQLite file?) after building local packages. This means an extra couple minutes wasted in CI, even if dependencies didn't change.

It would be great to keep everything related to local packages _outside_ of `~/.stack` so we can regain this nice caching setup, and also maintain the conceptual simplicity of "global/dependency stuff lives in `~/.stack`, while everything local to a project lives in `.stack-work`".

After talking to @borsboom and @snoyberg, it appears this was done out of simplicity so `stack` only needed to handle one SQLite database, which certainly makes sense. However, they indicated that having local configuration stay in `.stack-work` and out of `~/.stack` would be a good reason to add a bit of complexity and have multiple databases, hence me opening this issue.

---

For some additional context, here is what happens when building a local package if `~/.stack/stack.sqlite3` is nuked, which replicates what happens if you don't cache `~/.stack` in CI:

```sh
backend $ stack build core --fast            
backend $ rm ~/.stack/stack.sqlite3*
backend $ stack build core --fast
core-0.0.0: unregistering (old configure information not found)
sendgrid-client-0.0.0: unregistering (Dependency being unregistered: core-0.0.0)
core> configure (lib)
```

Stack rebuilds local packages from scratch when this happens.

snoyberg:

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -29,7 +29,7 @@ module Stack.Config
   ,getInNixShell
   ,defaultConfigYaml
   ,getProjectConfig
-  ,loadBuildConfig
+  ,withBuildConfig
   ) where
 
 import           Control.Monad.Extra (firstJustM)
@@ -64,7 +64,8 @@ import           Stack.Config.Nix
 import           Stack.Constants
 import           Stack.Build.Haddock (shouldHaddockDeps)
 import           Stack.Lock (lockCachedWanted)
-import           Stack.Storage (initStorage)
+import           Stack.Storage.Project (initProjectStorage)
+import           Stack.Storage.User (initUserStorage)
 import           Stack.SourceMap
 import           Stack.Types.Build
 import           Stack.Types.Compiler
@@ -373,9 +374,9 @@ configFromConfigMonoid
        hsc
        (maybe HpackBundled HpackCommand $ getFirst configMonoidOverrideHpack)
        clConnectionCount
-       (\configPantryConfig -> initStorage
+       (\configPantryConfig -> initUserStorage
          (configStackRoot </> relFileStorage)
-         (\configStorage -> inner Config {..}))
+         (\configUserStorage -> inner Config {..}))
 
 -- | Get the default location of the local programs directory.
 getDefaultLocalProgramsBase :: MonadThrow m
@@ -448,8 +449,10 @@ loadConfig inner = do
 
 -- | Load the build configuration, adds build-specific values to config loaded by @loadConfig@.
 -- values.
-loadBuildConfig :: RIO Config BuildConfig
-loadBuildConfig = do
+withBuildConfig
+  :: RIO BuildConfig a
+  -> RIO Config a
+withBuildConfig inner = do
     config <- ask
 
     -- If provided, turn the AbstractResolver from the command line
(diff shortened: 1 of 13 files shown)


src/Stack/Storage/User.hs
```diff
@@ -30,9 +24,11 @@ module Stack.Storage
     , saveCompilerPaths
     , upgradeChecksSince
     , logUpgradeCheck
+    -- * Shared utils
+    , updateList
```
borsboom: Perhaps move `updateList` and `updateSet` to another module, since there're not really "owned" by the user database code and could be used for any database?

src/Stack/Config.hs
```diff
@@ -526,13 +529,20 @@ loadBuildConfig = do
...
-        , bcSMWanted = wanted
-        , bcExtraPackageDBs = extraPackageDBs
-        , bcStackYaml = stackYamlFP
-        , bcCurator = projectCurator project
-        }
+    -- Unfortunately redoes getProjectWorkDir, since we don't have a BuildConfig yet
+    workDir <- view workDirL
+    let projectStorageFile = parent stackYamlFP </> workDir </> relFileStorage
+
+    initProjectStorage projectStorageFile $ \projectStorage -> do
```
borsboom: I wonder how much overhead this adds?  `withBuildConfig` appears to be used in a fair number of places (sometimes indirectly), and not all of those actually need the database.  Is it worth having a way to load the build config _without_ making the database available?

src/Stack/Storage/User.hs
```diff
@@ -60,42 +55,6 @@ share [ mkPersist sqlSettings
       , mkMigrate "migrateAll"
     ]
     [persistLowerCase|
-ConfigCacheParent sql="config_cache"
```
borsboom: Does removing the tables lead to any migration errors due to unsafe migrations?

snoyberg: No, Persistent simply ignores tables it isn't told of.

src/Stack/Storage/User.hs
snoyberg: Addressed in 36f93646fe18008a23827288b8466878c41b5986

src/Stack/Config.hs
snoyberg: I'd rather leave that alone for now, and if we see a performance impact in the future we can lazily fill this in.

src/Stack/Storage/User.hs
borsboom: Great, LGTM (well, once hlint warnings resolved :))

src/Stack/Config.hs
borsboom: Sounds good.

Issue #4892, opened 2019-06-20
Stack local-doc-root is wrong for 2.1.1


dfithian:
### General summary/comments (optional)

After upgrading to 2.1.1 and trying to install docs, I found that `stack path --local-doc-root` is wrong.

### Steps to reproduce

```
stack build --haddock --no-haddock-deps
...
Updating Haddock index for local packages in
.../.stack-work/install/x86_64-osx/9e8c24a71c82f9b19c848e4aa52470a8c1ccf223f368a919481ba397b1f5de2b/8.6.5/doc/index.html
Updating Haddock index for local packages and dependencies in
...//.stack-work/install/x86_64-osx/9e8c24a71c82f9b19c848e4aa52470a8c1ccf223f368a919481ba397b1f5de2b/8.6.5/doc/all/index.html
Haddock index for snapshot packages already up to date at:
/Users/dan/.stack/snapshots/x86_64-osx/9e8c24a71c82f9b19c848e4aa52470a8c1ccf223f368a919481ba397b1f5de2b/8.6.5/doc/index.html
```

```
ls $(stack path --local-doc-root)
ls: .../.stack-work/install/x86_64-osx/6c9a92c41e7ce4083f1bd67c8cbacc2a6620c22b954d11bc44a07a6f2b8217d9/8.6.5/doc: No such file or directory
```

The docs are in the hash `9e8c24a71c82f9b19c848e4aa52470a8c1ccf223f368a919481ba397b1f5de2b` but Stack says they are in `6c9a92c41e7ce4083f1bd67c8cbacc2a6620c22b954d11bc44a07a6f2b8217d9`.

I did check `stack path --local-install-root` and it uses the `9e8c24a71c82f9b19c848e4aa52470a8c1ccf223f368a919481ba397b1f5de2b` hash.

Important to note that I am using a custom snapshot.

### Stack version

```
stack --version
Version 2.1.1, Git revision f612ea85316bbc327a64e4ad8d9f0b150dc12d4b (7648 commits) x86_64 hpack-0.31.2
```

### Method of installation

Homebrew

dfithian:
As a workaround, I am using `$(stack path --local-install-root)/doc`

qrilka:
That's a side effect of implicit snapshots in Stack 2 - enabling or disabling haddocks changes snapshot hash, and in your case you should see difference between `stack path --local-doc-root` and `stack path --no-haddock-deps --local-doc-root` with the latter pointing to the proper location

[...]

PR #4903, opened 2019-06-24
Suppress haddock-dependent paths

dfithian:
https://github.com/commercialhaskell/stack/issues/4892

I did my best, let me know any feedback!

Verified by hand:

```bash
➜  stack git:(master) stack exec stack -- path --local-doc-root          
➜  stack git:(master) stack exec stack -- --haddock path --local-doc-root
/Users/dan/tvision/git/stack/.stack-work/install/x86_64-osx/95706d9a7eb5a51fa4f678a8b9b1b45f1c109a287d472625bc347d888d9a84f7/8.2.2/doc
```

@snoyberg @qrilka

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -10,6 +10,10 @@ Release notes:
 Major changes:
 
 Behavior changes:
+* `stack path` no longer outputs haddock-dependent paths unless `--haddock` is
+  specified. [#4892](https://github.com/commercialhaskell/stack/issues/4892).
+  Current list of affected paths: `local-doc-root`, `local-hoogle-root`,
+  `dist-dir`.
 
 Other enhancements:
 
--- a/src/Stack/Path.hs
+++ b/src/Stack/Path.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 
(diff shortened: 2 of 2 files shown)


snoyberg:
I think this behavior could be very confusing: now `--local-doc-root` will fail to produce output without `--haddock` without any explanation for why. The original goal in the code was to rerun things with Haddock enabled for the relevant settings, but for some reason it didn't work. That IMO would be the ideal thing to fix here.

dfithian:
Okay, I thought this was what you were asking for [here](https://github.com/commercialhaskell/stack/issues/4892#issuecomment-504713382) but I am happy to do whatever, so long as the behavior can be specified. Can you be more specific as to the behavior you want so that I can test it appropriately? In the original issue description, `stack path --local-doc-root` was not the location where haddock ended up via `stack build --haddock --no-haddock-deps`, so are you saying that you want the paths to match? Are there more test cases I should be aware of?

snoyberg:
The problem is, when I was working on updating this specific code, I didn't really understand the `UseHaddocks` bit, or have solid test cases to work from. Which is why I ended up breaking it, and why I'm having trouble giving you clear guidance. Sorry about that.

I believe the requirements here should be one of the following:

* `stack path --local-doc-root` gives a meaningful error message, while `stack path --local-doc-root --haddock` succeeds
* `stack path --local-doc-root` gives the same output as `stack path --local-doc-root --haddock`, which is the same path specified with `stack build --haddock`

Does that make sense? I prefer the latter myself.

qrilka:
@snoyberg I don't quite understand what this PR is trying to solve and the way it works in the version 2.1.1 is exactly what you give in your 2nd option, e.g.:
```
qrilka@qdesktop ~/ws/h/hatrace $ stack-2.1.1 path --local-doc-root
/home/qrilka/ws/h/hatrace/.stack-work/install/x86_64-linux-tinfo6/eaf56c794f2da90e300a3577de4d3f6adad92b776bfc0e13009088580835816a/8.6.3/doc
qrilka@qdesktop ~/ws/h/hatrace $ stack-2.1.1 path --local-doc-root --haddock
/home/qrilka/ws/h/hatrace/.stack-work/install/x86_64-linux-tinfo6/eaf56c794f2da90e300a3577de4d3f6adad92b776bfc0e13009088580835816a/8.6.3/doc
```

snoyberg:
@qrilka Looks like you're right. @dfithian I was debugging too many different things at once, and started seeing things that weren't there. It looks like the original bug report had to do with `--no-haddock-deps` producing a different hash, not a lack of `--haddock` being passed to `stack path`. Despite my initial thoughts, it could be that I actually implemented `Stack.Path` correctly in the first place.

dfithian:
Oh, I see. I agree that it is working correctly currently and I can see that additionally passing `--no-haddock-deps` changes the hash. That seems appropriate given the straightforward use case of using `--haddock`, and it is apt that you would need to specify `--no-haddock-deps` for the correct path in that case. So I will close this PR. Thank you both for your explanation and your time!

PR #4914, opened 2019-06-27
Package-local dist dir for doctest

qrilka:

Fixes https://github.com/commercialhaskell/stackage/issues/4667
Tested manually.


src/Stack/Build/Execute.hs
```diff
@@ -1935,7 +1935,11 @@ singleTest topts testsToRun ac ee task installedMap = do
                 -- env variable GHC_ENVIRONMENT is set for doctest so module names for
                 -- packages with proper dependencies should no longer get ambiguous
                 -- see e.g. https://github.com/doctest/issues/119
+                -- also we set HASKELL_DIST_DIR to a package dist directory so
+                -- doctest will be able to load modules autogenerated by Cabal
+                distRelativeDir' <- distRelativeDir
                 let setEnv f pc = modifyEnvVars pc $ \envVars ->
+                      Map.insert "HASKELL_DIST_DIR" (T.pack $ toFilePath distRelativeDir') $
```
snoyberg: Perhaps we can use an absolute path instead of a relative path in case things get called from different subdirs instead, what do you think?

snoyberg: I've rebased this PR onto latest stable and pushed a commit that uses an absolute path here.

mihaimaruseac:
We tested and it works.

PR #4915, opened 2019-06-27
Disable WAL mode for SQLite databases (fixes #4876)

### LINKED ISSUE #4876: Can no longer compile two things at once with 2.1.1 from 1.9.3
It used to be fine compiling a project in different .stack-work directories or different projects at the same time.
After upgrading to stack 2.1.1 I can't even build _different_ projets at the same time let a long the same project.

Just get this error until the first build has finished. After that I can try build the second project.
This makes stack ~unusable on CI, and~ pretty awful to use locally with multiple projects.

```
Encountered error while migrating Pantry database:
    SQLite3 returned ErrorProtocol while attempting to perform prepare "PRAGMA journal_mode=WAL;": locking protocol
Please report this on https://github.com/commercialhaskell/stack/issues
As a workaround you may delete Pantry database in /home/david/.stack/pantry/pantry.sqlite3 triggering its recreation.
```

borsboom:
Tested both against existing databases that had WAL enabled before, and a new STACK_ROOT with new databases.  In the former case, WAL remained enabled on the existing databases, and in the latter case WAL was not enabled on the new databases.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,10 @@ Major changes:
 
 Behavior changes:
 
+* Disable WAL mode for SQLite3 databases, to improve compatibility with
+  some platforms and filesystems.  See
+  [#4876](https://github.com/commercialhaskell/stack/issues/4876).
+
 Other enhancements:
 
 * Do not rerun expected test failures. This is mostly a change that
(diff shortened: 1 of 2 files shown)


borsboom:
(CI failed on unrelated code due to new hlint version triggering new warnings -- fixed in separate PR https://github.com/commercialhaskell/stack/pull/4918)

PR #4921, opened 2019-06-29
Expose package library when running tests

### LINKED ISSUE #4896: Stack 2.1.1 doesn't set package environments like it used to

With previous versions of Stack, `stack test` set the package environment so the current package was loaded. This is useful for [IHaskell](https://github.com/gibiansky/IHaskell) which exercises the GHC API.

Stack 2.1.1 doesn't do this, and so I'm getting a lot of errors and a failed test suite.

### Steps to reproduce

1. `git clone https://github.com/gibiansky/IHaskell`
1. `cd IHaskell`
1. `stack test`

### Expected

`ihaskell-0.10.0.0: Test suite hspec passed`

This is with version `1.9.3 x86_64 hpack-0.31.2`.

### Actual

https://travis-ci.org/gibiansky/IHaskell/jobs/548734133

### Stack version

```
$ stack --version
```

### Method of installation

qrilka:

Fixes #4896 
@snoyberg this appears to be just a quickfix. I'm not yet sure that it's the correct way.
I didn't quite understand the motivation of excluding package library from dependencies in the commit 9ceb409a887bfcb1c0a09c3c783e409f9ecaed97  - does it make sense to try reverting that or it's just a workaround until we have proper component-based builds?


src/Stack/Build/Execute.hs
```diff
@@ -1949,7 +1961,7 @@ singleTest topts testsToRun ac ee task installedMap = do
                         "package-db " <> fromString snapDBPath <> "\n" <>
                         "package-db " <> fromString localDBPath <> "\n" <>
                         foldMap (\ghcId -> "package-id " <> RIO.display (unGhcPkgId ghcId) <> "\n")
-                            (thGhcId:M.elems allDepsMap)
```
snoyberg: Why is `thGhcId` now excluded?

qrilka: what do you mean by "excluded"? I see it in `pkgGhcIdList ++ thGhcId:M.elems allDepsMap`

qrilka:
Oh, sublibraries, yes, that was also one of my guesses
So then I just need to add a changelog and check integration tests as an extra safety measure, right?

src/Stack/Build/Execute.hs
snoyberg: Sorry, I'm stupid, ignore me. Totally misread the code.

snoyberg:
Correct. Let me know if you want me to kick off integration test runs on Azure.

qrilka:
Integration tests failed because of `tar` issue and a problem with `stack clean` on Windows (@psibi should create an issue about it).
Otherwise this PR should be OK to merge

gelisam:
Has this fix been later reverted? I get the correct behaviour when I checkout and build commit 61edf6b, but not when I install the latest version of stack via `stack upgrade --git`.

snoyberg:
I think the fix is only on stable, and hasn't yet been merged to master.

gelisam:
`stack upgrade --git --git-branch stable` did the trick, thanks!

PR #4922, opened 2019-06-30
Tell the user that a boot package was pruned

### LINKED ISSUE #4897: Tell the user that a boot package was pruned
I use Stack on CI and sometimes override dependencies intentionally.  Recently, they started failing, but the error message isn't particularly descriptive:

```
Error: While constructing the build plan, the following exceptions were
encountered:
In the dependencies for directory-1.3.3.2:
    time must match >=1.4 && <1.10, but the stack configuration has no specified
         version  (latest matching version is 1.9.3)
needed since directory is a build target.
```

`stack ls dependencies` shows `time 1.6.0.1`, and similarly on [Stackage](https://www.stackage.org/lts-7.24/package/time-1.6.0.1).  Therefore, I was left very confused why Stack suddenly forgot `time` existed.

It took me a while but I eventually figured I should check the release notes in case it was intentional.  I saw this:

> Overriding GHC boot packages results in any other GHC boot
packages depending on it being no longer available as a dependency,
such packages need to be added explicitly when needed. See
[#4510] (#4510).

So that explained a lot, but figuring this out wasted a lot of time.  Could the error message be made a little more detailed?  It at least state that a package was pruned behind the scenes, and preferably include a reason for doing so and/or link to #4510 for rationale.

Rufflewind:
Fixes #4897.

* [N/A] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [N/A] The documentation has been updated, if necessary.

---

Manual test:

```shell
cabal get process-1.6.5.0 &&

cd process-1.6.5.0 &&

cabal get unix-2.7.2.2 &&

cat >stack.yaml <<EOF &&
resolver: lts-13.26
packages: [.]
extra-deps: [./unix-2.7.2.2]
EOF

"${LOCALLY_BUILT_STACK}" build
```

```

In the dependencies for process-1.6.5.0:
    directory must match >=1.1 && <1.4, but this GHC boot package has been pruned (issue
              #4510); you need to add the package explicitly to extra-deps  (latest
              matching version is 1.3.3.2)
needed since process is a build target.

  * Recommended action: try adding the following to your extra-deps
    in /tmp/ppc/process-1.6.5.0/stack.yaml:

- directory-1.3.3.2@sha256:79e15a62f90cb9f20f46b2b56fe67d00d71ed54cfad07ea337695f9ebb74baa4,2829

Plan construction failed.
```

### DIFF
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -1109,7 +1109,7 @@ pprintExceptions exceptions stackYaml stackRoot parentMap wanted' prunedGlobalDe
             let prunedDeps = map (style Current . fromString . packageNameString) pruned
             in Just $ flow "Can't use GHC boot package" <+>
                       (style Current . fromString . packageNameString $ name) <+>
-                      flow "when it has an overriden dependency, " <+>
+                      flow "when it has an overridden dependency (issue #4510);" <+>
                       flow "you need to add the following as explicit dependencies to the project:" <+>
                       line <+> encloseSep "" "" ", " prunedDeps
         | otherwise = Just $ flow "Unknown package:" <+> (style Current . fromString . packageNameString $ name)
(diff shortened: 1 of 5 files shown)


dbaynard:
Thanks for the PR! I'll take a look, shortly.

dbaynard:
Thanks for this; the code itself looks good.

However, there's now some code duplication between the `DependencyPlanFailures` and `UnknownPackage` cases of `pprintException`. It may improve the clarity of the code to add an additional `BadDependency` constructor. However, this requires more substantial changes.

It it's worth it, I'll make the changes. If it's not, I'll look to merge.

An integration test would be good, too — would you be willing to put one together, @Rufflewind ?

Rufflewind:
Added integration test.

PR #4930, opened 2019-07-03
Fix links

leifmetcalf:
https://github.com/commericalhaskell/ghc/releases/new redirects to https://github.com for me. https://github.com/commericalhaskell/ghc/releases goes to the releases. I'm not sure about this one, maybe only maintainers can see that page. It seems odd that I get redirected to the home page though.

The newcomers page doesn't seem to have the information it used to, so I've changed that link to the building quickstart. Maybe we should just link to the [build guide](https://gitlab.haskell.org/ghc/ghc/wikis/building)?

EDIT: I forgot to add [skip ci] to my commit message, sorry!

### DIFF
--- a/doc/maintainers/ghc.md
+++ b/doc/maintainers/ghc.md
@@ -10,7 +10,7 @@
         git fetch upstream
         git push origin ghc-X.Y.Z-release
 
-  * [Publish a new Github release](https://github.com/commercialhaskell/ghc/releases)
+  * [Publish a new Github release](https://github.com/commercialhaskell/ghc/releases/new)
     with tag `ghc-X.Y.Z-release` and same name.
 
   * Download all the relevant GHC bindists from https://www.haskell.org/ghc/download_ghc_X_Y_Z and upload them to the just-created Github release (see
--- a/doc/maintainers/ghc.md
+++ b/doc/maintainers/ghc.md
@@ -10,7 +10,7 @@
         git fetch upstream
         git push origin ghc-X.Y.Z-release
 
-  * [Publish a new Github release](https://github.com/commercialhaskell/ghc/releases/new)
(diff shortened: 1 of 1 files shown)


mattaudesse:
Thanks @leifmetcalf - sorry this PR has been sitting unacknowledged for so long!

You're right that the `new` page in https://github.com/commericalhaskell/ghc/releases/new is only accessible by maintainers, so it's probably correct to leave it as-is (albeit confusing for anyone without maintainer credentials).

Perhaps the newcomers link should point [here](https://gitlab.haskell.org/ghc/ghc/wikis/contributing#newcomers-to-ghc) now? When I checked the old version it stated the content had been moved there.

This is an easy merge if you'd care to update the links. Thanks!

leifmetcalf:
I've reverted the link back to https://github.com/commericalhaskell/ghc/releases/new. At some point the code (which the docs says comes from the newcomers page)

```
git config --global url."git://github.com/ghc/packages-".insteadOf git://github.com/ghc/packages/ && \
git clone -b ghc-${GHC_VERSION}-release --recursive https://gitlab.haskell.org/ghc/ghc.git ghc-${GHC_VERSION} && \
cd ghc-${GHC_VERSION}/ && \
cp mk/build.mk.sample mk/build.mk && \
sed -i 's/^#BuildFlavour *= *perf$/BuildFlavour = perf/' mk/build.mk && \
./boot && \
./configure --enable-tarballs-autodownload && \
sed -i 's/^TAR_COMP *= *bzip2$/TAR_COMP = xz/' mk/config.mk && \
make -j$(cat /proc/cpuinfo|grep processor|wc -l) && \
make binary-dist
```

must have been removed, or moved elsewhere. That's why I changed the link to https://gitlab.haskell.org/ghc/ghc/wikis/building/quick-start, which has more relevant information (although it doesn't contain that snippet of code).

PR #4939, opened 2019-07-06
Fix support for non-ASCII module names

### LINKED ISSUE #4938: Stack 2.1.1 fails to build package with non-ascii module name

Stack 2.1.1 fails to build when module names contain ä. Stack 1.9.3 does NOT fail.

### Steps to reproduce

1.  Clone `https://github.com/bjornbm/umlaut-stack.git`
2.  `stack build --verbose`

### Expected

A successful build.

### Actual

The build fails with (last lines of output, for full output see [gist](https://gist.github.com/bjornbm/77d8a7baa9924e34ee412ad7a075ff6a)):

```
2019-07-04 22:13:21.591710: [info] Linking .stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/umlaut-exe/umlaut-exe ...
2019-07-04 22:13:23.467465: [debug] Process finished in 4757ms: /Users/bjorn/.stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5 --builddir=.stack-work/dist/x86_64-osx/Cabal-2.4.0.1 build lib:umlaut exe:umlaut-exe --ghc-options ""
2019-07-04 22:13:23.468057: [debug] Start: getPackageFiles /Users/bjorn/repos/umlaut/umlaut.cabal
2019-07-04 22:13:23.470598: [debug] Finished with exception in 2ms: getPackageFiles /Users/bjorn/repos/umlaut/umlaut.cabal
Exception thrown: ModuleName.fromComponents: invalid components ["T\195\164st"]
  error, called at ./Distribution/ModuleName.hs:105:46 in Cabal-2.4.0.1-3bZfb3D11r2LsqHGk5ybV3:Distribution.ModuleName
2019-07-04 22:13:23.473932: [error] ModuleName.fromComponents: invalid components ["T\195\164st"]
  error, called at ./Distribution/ModuleName.hs:105:46 in Cabal-2.4.0.1-3bZfb3D11r2LsqHGk5ybV3:Distribution.ModuleName
```

The error is reported by a cabal function `ModuleName.fromComponents`, but since the ChangeLog.md says that Cabal 2.4 has been used since stack-1.9.1 I assume there is a change in how stack passes the module names to cabal.

May be related to UTF normalisation? (ore the absence thereof?)

Interestingly if the `Täst` module is imported by `Main.hs` instead of `Lib.hs` the build completes just fine.

### Stack version

```
$ stack --version
```

### Method of installation

`stack upgrade --force-download`

qrilka:

Tested manually and added an integration test.
Fixes #4938

Another case when importing `Data.ByteString.Char8` is an error...


borsboom:
The new integration test seems to be failing on macOS on CI (see https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=1945)

However, I just tried locally on my MacBook Pro and it seems to pass here, so not sure what's going on.

neongreen:
It might depend on the file system. https://apple.stackexchange.com/questions/83935/unicode-normalization-for-filenames-and-copied-text-from-pdfs

`Täst` in the source code uses a precomposed `ä`, but Git on macOS might be creating a filename with a two-symbol `ä` (there's a setting in Git that controls that), or this might happen at the filesystem level (HFS+ and APFS have different rules).

neongreen:
An easy way to find out if that is the case would be to try with `ß` instead.

borsboom:
I did a trial of upgrading CI to Mojave/10.14 (it was on High Sierra/10.13), and that may have fixed it.  Doing the full test now with Mojave.  Aside from that, indications so far are that this is not triggered in Stack itself but more likely in GHC's package handling (or maybe Cabal).

borsboom:
https://github.com/commercialhaskell/stack/pull/4953 looks to fix this (see https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=1955).

@snoyberg did discover that the `.cabal` file generated by `hpack` is different when the test fails.  On Azure it wrote `Ta\204\136st` while on @snoyberg's machine it wrote `T\195\164st`, so it's possible the problem is introduced by `hpack` doing reading the modules in a different way than the other tools and getting tripped up by macOS's unicode normalization.

bjornbm:
I didn't quite catch if the change you made fixes the problem on both HFS+ and APFS… it sounded like it is likely still broken on HFS+?? (I assume the High Sierra builds were on HFS+ and Mojave on APFS.)

FWIW elsewhere I have used [unicode-transforms](http://hackage.haskell.org/package/unicode-transforms) to normalize filenames in macOS to be able to work with them in a sane fashion (e.g., `length filename` returning `4` instead of `5` for something like "Täst"). The gist of it being something along the lines of:

```
import Data.Text.Normalize (normalize, NormalizationMode (NFC))

filenameToText :: FilePath -> Text
filenameToText = normalize NFC . pack
```

(There is a `normalize` that operates on bytestrings instead of `Text` as well.)

If `hpack` needs fixing I guess it would be somewhere around https://github.com/sol/hpack/blob/master/src/Hpack/Util.hs#L89.

borsboom:
I think this needs some more investigation.  I actually thought my system was still on HFS+, but turns out upgrading to Mojave also changed my filesystem to APFS, so probably the same is true on Azure CI.  So what you say about it still being broken with HFS+ is very plausible.  

The original bug that was fixed was a problem on all operating systems and all filesystems, whereas what we're talking about is a different bug that only affects some macOS systems.  It's probably better to open a separate issue.

Issue #4936, opened 2019-07-04
stack clean --full fails on Windows


psibi:
Right now, the integration test build for Windows seem to be failing
because of `stack clean --full`.

On investigating the reason seems to be that, the stack process is
already opening the `stack.sqlite3` file and then trying to delete
it. While both OS X and Linux happily deletes the files, that doesn't
happen in Windows where an exception is thrown.

### Steps to reproduce

Only reproducible on Windows environment:

```
stack clean --full
```
### Expected

Successful delete of various files and exiting with a status code of 0.

### Actual

An exception is thrown.

### Method of installation

* Official binary, downloaded from stackage.org or fpcomplete's package repository

dbaynard:
Closing — please reopen, @psibi, if this isn't fixed

PR #4940, opened 2019-07-07
Fix Windows integration test suite

psibi:


psibi:
@snoyberg This fixes all the windows related integration tests: https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=1935

Also, I tried running `stack clean --full` on my friends Windows machine and this error didn't pop up on his machine. So this error may actally be related to how we run stack integration tests (I think we are using stack script to run it in the CI, but I haven't looked into it deeply). In that case, I think we can also close #4936 unless some real Windows user is hitting up with this issue.

PR #4966, opened 2019-07-16
mention mac-specific issues for stack scripts (fixes #4965)

### LINKED ISSUE #4965: stack script interpreter manual could mention these gotchas for mac users
I was really mystified why one of my hledger addon stack scripts would not compile like the others; it always complained about the shebang line. The fix was to stop using CPP in it. https://gitlab.haskell.org/ghc/ghc/issues/6132, Can't use both shebang line and #ifdef declarations in the same file, was fixed a few years back - except on mac. 

Another one that I read about: shebang lines don't work on mac unless they refer to a compiled executable, so eg `#!/usr/local/bin/runhaskell` has no effect; it should be `#!/usr/bin/env runhaskell` instead.

https://docs.haskellstack.org/en/stable/GUIDE/#script-interpreter could mention these.

simonmichael:


dbaynard:
I had to force push to change branch! Documentation PRs go to `stable`, not `master`. I'll merge once CI completes.

PR #4974, opened 2019-07-23
Haddock use relative paths in output (fixed #4971)

### LINKED ISSUE #4971: Links with absolute paths in --haddock

Prior to stack-2, `stack build --haddock` would produce HTML files with relative links. This made it simple to (e.g.) `aws s3 sync` to serve the documentation, and also worked when viewing the files locally.

Since stack-2, I get absolute filesystem paths in all the links:

```html
<a href="/home/patrick/code/project/.stack-work/.../doc/...">
```

This still works locally, but won't work when served.

I've looked through the documentation for Stack and Haddock (specifically `--source-base` and friends) but nothing I try will recover the old behavior.

My general question is, does anyone know how to get relative links back? For now, I'm resorting to `sed`ing them out before uploading.

~~**CAVEAT**: I'm only 90% sure this is a stack-2 thing. I don't refresh my documentation super frequently so it's possible it worked on an old resolver/GHC/haddock and is breaking on a newer one, but I *believe* it was the upgrade to stack-2 that did it. Apologies in advance if I'm incorrect.~~ This has been confirmed below.

### Steps to reproduce

1. Have an S3 bucket as a website to serve documentation from
1. Run something like:

   ```console
   stack build --haddock
   aws s3 sync --acl public-read --delete \
     "$(stack path --local-doc-root)"/ \
     s3://{BUCKET}/{PREFIX}/
   ```

### Expected

Links from the main S3 website should work (because they are relative).

### Actual

Links do not work because they include `/home/$USER/blah/blah` in them.

### Stack version

```
$ stack --version
Version 2.1.3, Git revision 636e3a759d51127df2b62f90772def126cdf6d1f (7735 commits) x86_64 hpack-0.31.2
```

### Method of installation

https://aur.archlinux.org/packages/stack-static/, which packages the official binary for Arch.

borsboom:
* [N/A] The documentation has been updated, if necessary.

Tested by generating haddocks for a package and confirming by`grep`ing all the output that no absolute links are generated.  Also spot checked that links are correct.

@qrilka It looks like the links changed to absolute in !4596.  I don't understand the original bug that fixed well enough to know whether my changes here will cause any problems, so please test it.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,9 @@ Other enhancements:
 
 Bug fixes:
 
+* Fix using relative links in haddocks output.  See
+  [#4971](https://github.com/commercialhaskell/stack/issues/4971).
+
 
 ## v2.1.3.1
 
--- a/src/Stack/Build/Haddock.hs
+++ b/src/Stack/Build/Haddock.hs
@@ -229,8 +229,10 @@ generateHaddockIndex descr bco dumpPackages docRelFP destDir = do
                         docRelFP FP.</>
                         packageIdentifierString dpPackageIdent FP.</>
                         (packageNameString name FP.<.> "haddock")
+                    docPathRelFP =
+                        fmap ((docRelFP FP.</>) . FP.takeFileName) dpHaddockHtml
                     interfaces = intercalate "," $
-                      maybeToList dpHaddockHtml ++ [srcInterfaceFP]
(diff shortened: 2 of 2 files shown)


jarcane:
Any idea when this will be merged?

Ailrun:
When will it be released?

PR #4990, opened 2019-08-07
Fix yaml config doc

sio-funmatsu:
# Fix document issue link

- The link display is broken on https://docs.haskellstack.org
  - [fixed part](https://docs.haskellstack.org/en/stable/yaml_configuration/#4928httpsgithubcomcommercialhaskellstackissues4928)


borsboom:
Thanks.  Right now this looks like you're requesting to merge a branch derived from `master` into `stable`, which means it wants to bring all the rest of `master` along with it.  Can you rebase against `stable`?

snoyberg:
@sio-funmatsu bumping

snoyberg:
I actually don't see the original problem described, and the Markdown looks fine. Also no response to bumping, so closing.

Issue #4991, opened 2019-08-09
missing package libtinfo-dev on ubuntu:bionic


charala1:
`stack build --only-snapshot` fails in an `ubuntu:bionic`- based docker image with

```
Foreign.Marshal.Compat.Repl
[0m[91mbase-compat                      > [ 94 of 112] Compiling Foreign.Compat
[0m[91mbase-compat                      > [ 95 of 112] Compiling Foreign.Compat.Repl
[0m[91mhaskeline                        > /usr/bin/ld.gold: error: cannot find -ltinfo
haskeline                        > collect2: error: ld returned 1 exit status
```

borsboom:
`libtinfo-dev` is a dependency of **haskeline**, not of Stack or GHC.  There are lots of packages out there with various dependencies on system libraries, and Stack's installer can't install all of them (to see the full list of system dependencies for all packages in Stackage, check out [this script](https://github.com/commercialhaskell/stackage/blob/master/debian-bootstrap.sh)), so all it installs is the minimum to be able to use all of Stack's functionality and compile a basic Haskell program that uses base libraries included with GHC.

PR #4992, opened 2019-08-09
Updated etc/scripts/get-stack.sh

charala1:
Fixed https://github.com/commercialhaskell/stack/issues/4991


borsboom:
Thanks, but I'm afraid installing dependencies of non-base libraries is out of scope for the installer script.  See https://github.com/commercialhaskell/stack/issues/4991#issuecomment-520255024.

PR #5002, opened 2019-08-15
Add build-output-timestamps flag

chrisdone:
This adds timestamps to Cabal's build output log files when specified in the config file. It's helpful primarily for debugging/analytics of timing of build steps.

You can test it simply by adding to your `stack.yaml`:

```yaml
build-output-timestamps: true
```

And then running with something like this on a *multi-package project*:

```
stack clean && stack build --no-interleaved-output
```

[Example output](https://gist.github.com/chrisdone/20a1e08f0fe18bd6748c87281930e698).

For non-multi-package projects, you can just use pre-existing tools for prefixing lines with timestamps. But on a multi-package project, we don't have that option. Hence this patch.


borsboom:
A couple questions:

- Would it make sense to add a CLI option?  It's something I could see wanting to enable on a one-off basis rather than enabling more permanently via a config file.

- Does this provide a concrete benefit over piping stack's output to an external tool that prepends timestamps to each line?

chrisdone:
> Would it make sense to add a CLI option? It's something I could see wanting to enable on a one-off basis rather than enabling more permanently via a config file.

Sure, I'd agree. I didn't want the change to be invasive, but I can go ahead and add that to the `stack build` argument parser. `--build-output-timestamps`? For my personal use, I might actually use the `.yaml` flag in my global stack config. But turning it on in an ad-hoc basis is also a perfectly natural use-case.

> Does this provide a concrete benefit over piping stack's output to an external tool that prepends timestamps to each line?

When I wrote the PR I thought we couldn't use the interleaved output for time diffing, but actually the fact that each line is prefixed with the package name probably means that we could. So actually, the equivalent command would be to use `ts` from `moreutils` like this:

```
$ stack build 2>/dev/stdout | ts '%Y-%m-%dT%H:%M:%.S'
2019-08-16T10:27:17.599245 prana-ghc-8.4: unregistering (missing dependencies: prana-types)
2019-08-16T10:27:17.599335 prana-interpreter-0: unregistering (missing dependencies: prana-primops, prana-types)
2019-08-16T10:27:17.599367 prana-primops-0: unregistering (local file changes: app/Main.hs prana-primops.cabal src/Lexer.x src/Parser.y src/ParserM.hs src/Prana/PrimOp.hs src/P...)
2019-08-16T10:27:17.599390 prana-types-0: unregistering (missing dependencies: prana-primops)
```

And then the parser of the log output would parse the package prefixes.

There are pros and cons for both:

1. Pros for this PR:
  * You can enable in a config to be on all the time, so if you just ran a big build and don't want to sit through it again to see time info, you can just open the files.
  * The log files go to a standard location (already under a `.gitignore`'d dir).
  * No extra shell tools need to be installed to accumulate the right data (e.g. in CI, or Windows).
  * You can inspect just one package vs all.
  * One possible difference; it *might* be faster to write to a log file than to stdout, especially when enabling `--ghc-options=-v`, if we'd like to parse the phases of each GHC pass per module (parsing, desugaring, checking, rule rewriting, codegen) because those spit out a _lot_ of information.

2. Pros for piping interleaved output:
   * No extra change needed to Stack; less maintenance and docs.
   * Packages like `ts` might have more tooling around it.

In the end, I'm going to write a parser anyway that will consume the various stages (`Configuring`, `Compiling X module`, `Linking`) and produce an overall report of how long each package took and within each package how much compiling each module took and the linking step. I'll probably write those to an sqlite database to be queried by the developer interested in the stats. From a parsing perspective, it doesn't matter a great deal whether the log format is split into packages or prefixed with a package. So this discussion doesn't block my work.

What're your thoughts?

borsboom:
Ok, that all makes sense and also makes it more clear why you'd want to enable it globally (since it just adds the timestamps to files, not to regular output).

chrisdone:
Here is the mentioned tool: https://github.com/fpco/build-profile

chrisdone:
Merging config option for now. May add commandline flag later.

PR #5003, opened 2019-08-15
Adds support for configuring version of stack installed.

mavenraven:

I tested this change on OS X by running:
`cat get-stack.sh | sh -s - -f`
`stack --version`
which outputs
`Version 2.1.3, Git revision 0fa51b9925decd937e4a993ad90cb686f88fa282 (7739 commits) x86_64 hpack-0.31.2`
then
`cat get-stack.sh | sh -s - -f -v 1.9.3`
which outputs
`Version 1.9.3, Git revision 40cf7b37526b86d1676da82167ea8758a854953b (6211 commits) x86_64 hpack-0.31.1`

The use case for this feature is to allow for using a deterministic version in our CI. It also helps provide a migration path from 1.9.x -> 2.x as we're currently using `stack container image`.  I hope you'll accept it. Thanks!

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,9 @@ Other enhancements:
 * Add `ls dependencies json` which will print dependencies as JSON.
   `ls dependencies --tree`  is now `ls dependencies tree`. See
   [#4424](https://github.com/commercialhaskell/stack/pull/4424)
+* Add ability to set which version of stack to be downloaded by
+  `get-stack.sh` with `--version` argument.
+
 
 Bug fixes:
 
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -25,7 +25,8 @@
 # https://docs.haskellstack.org/en/stable/install_and_upgrade/
 #
 
-STACK_VERSION="2.1.3"
+DEFAULT_STACK_VERSION="2.1.3"
+STACK_VERSION=""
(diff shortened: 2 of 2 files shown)


borsboom:
I'm definitely sympathetic to your needs, but I don't think this is the approach we should take.  We adjust the script fairly often and the risk of breaking compatibility with older versions of Stack is too great (as is the maintenance burden if having to test with older versions).

I've seen other projects take a different approach: have the installer script versioned along with the tool itself.  So we'd have a script to install v2.1.3, another script to install v1.9.3, etc.  We could set things up so https://get.haskellstack.org/ would continue to install the latest version, but provide scripts for specific version at URLs such as https://get.haskellstack.org/v1.9.3.  I think this would handle your use case just as well, but would be a lot easier to maintain.

For the moment, you can install a specific version of Stack on CI using a command like this:

    wget -qO- https://github.com/commercialhaskell/stack/releases/download/v1.9.3/stack-1.9.3-linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C /usr/local/bin '*/stack' 

(you'll also need to ensure any system dependencies like GCC, libgmp, etc. are installed).

mavenraven:
The versioned url would be great! I saw that workaround line in stackage Dockerfile, but I would love to be able to use the script to get the system packages.

mavenraven:
Going forward, we're just going to keep a checked in version of the script in our repo in the same vein as https://github.com/takari/maven-wrapper.

Issue #3394, opened 2017-08-29
Feature Request: allow setup-info urls to be relative file paths


duog:
Currently if the setup-info url is a relative path stack fails complaining it is not an absolute path.

I want to distribute a custom build of ghc with my project, and reference it from the setup-info url field, however I am prevented from doing so. I instead must use a shell script and call
```
stack setup --ghc-bindist $(pwd)/bindist/xxx
```

Issue #2983, opened 2017-02-05
setup-info field in stack.yaml not consulted


ndmitchell:
Using stack 1.3.2 on Windows 64bit, I have a `stack.yaml` file containing `setup-info: http://example.com/none.yaml`. I have disabled access to `raw.githubusercontent.com`. If I run either `stack setup` or `stack build` then I get an error about not being able to reach `raw.githubusercontent.com`, rather than a 404 failing to download `example.com/none.yaml`. If I use the flag `--setup-info-yaml` with `example.com` then it fails witha  404 as expected. The documentation suggests these two settings are equivalent, but they don't seem to be.

mgsloan:
Hmm, indeed it seems that `--setup-info-yaml` overrides the default url, whereas the `setup-info` field extends the list of configs to load.

I wonder how impactful it would be to change this semantics such that the default url doesn't get used when `setup-info` is manually specified.  Seems like a reasonable change to me

borsboom:
In the case of a inline setup-info, it probably should extend the default configs, since that's commonly used to add one custom GHC bindist (e.g. if you want to test the latest GHC release candidate).  For the cases where it references an external YAML URL, you're right that the more common use case would probably be to totally override the default URL.  Potentially bit confusing to have the same field behave a bit differently in these two cases though.

AleXoundOS:
Still confusing.
I have `setup-info: "http://my-local-server/stack-setup.yaml"` in both config.yaml and project's stack.yaml. However none of them help in case of running stack setup. I still have to manually give `--stack-setup-yaml=http://my-local-server/stack-setup.yaml` option to stack setup on the first run.

borsboom:
I think that field expects the setup info to be inline, not point to a separate file/URL.  The fact that's it's silently ignoring an entry is a bug though, and being able to point it to a separate file/URL would be a helpful feature.

borsboom:
Yes, I'd say (1) is required and (2) would be a nice feature but needs a bit of thought.  Inline `setup-info` _augments_ the downloaded setup-info, whereas the `stack --setup-info-yaml` _replaces_ it, so it would be confusing to have both behaviours in the same `setup-info` field.

afcowie:
[...]

Issue #2982, opened 2017-02-05
Combine stack-info-yaml argument and config.yaml


ndmitchell:
There are two config files that Stack uses for the global URL settings, namely config.yaml (typically in `$STACK_ROOT`) and the arguments passed to `stack-setup-yaml`. The typical reason to modify either seems to be that you want to make Stack always hit a local mirror (either for caching, reproducibility, or due to network blocks). If you want to modify one, you probably want to modify the other. They are both set in different ways, sourced from different places, and have different non-overlapping subsets of information. Combining them into one would simplify lives for people who need to modify these things.

ndmitchell:
Note that the stack-setup-yaml file is far more convenient (passed on the command line, able to be set in stack.yaml) so would be my preference of the model to follow.

lwm:
I'd love when Stack could run totally offline. I'm following the Stack list thread and would like to see this all being made easier. So, those two files are described as:

  * stack-setup.yaml ([default](https://raw.githubusercontent.com/fpco/stackage-content/master/stack/stack-setup-2.yaml )): Stack's dependencies.
  * config.yaml (no default): User's global configuration

Could we see an example of the options you need from both files without sensitive info?

ndmitchell:
For stack-setup.yaml I take the default, delete 90%+ of it, and then rewrite `http://...` to http://myinternalserver/...`.

For config.yaml I take https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md#package-indices and https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md#urls, slap them together, and do the same `http://` trick.

To make it easy, you'd really want to just set "for `http://` read `http://mylocalserver/` or `C:/mylocalcache/`, and once, rather than a few 100 times over two files.

dpacbach:
(+1)

This should make it easier to configure stack to work in environments where local servers are required.

ndmitchell:
I've set up a fully worked example at https://github.com/ndmitchell/offline-stack#readme which is tested to ensure it works. I suggest using that as a template.

philipcraig:
Great setup @ndmitchell

jonathankochems:
(+1) This would really help getting stack into work environments where local mirroring of packages are required.

Blaisorblade:
“Using Stack offline” would also probably help with bad connections (#3088), and with people behind firewalls that block S3 (I forget which is the issue).

Blaisorblade:
[...]

Issue #2913, opened 2017-01-10
`stack setup` should not fail if it couldn't download default stack-setup-yaml


vasiliyl:
Our build environment is isolated from the internet.  Recently we've put into service a new
build machine with latest stack (reproduced on 05acb2c) and discovered that `stack setup`
build step fails with an error (part of output from `stack setup -v`):

    ...
    2017-01-11 01:56:19.205442: [debug] Using standard GHC build
    @(Stack/Setup.hs:612:9)
    HostCannotConnect "raw.githubusercontent.com" [connect: does not exist (Network is unreachable)]

Other machines pass this build step because earlier versions of stack cached everything needed.

A small investigation showed that stack always tries to download default stack-setup.yaml
(https://github.com/commercialhaskell/stack/blob/05acb2c/src/Stack/Setup.hs#L735)
and errors if it couldnt.

I think, in case when stack-setup.yaml is unavailable, stack should
issue a warning and continue with processing setup-info from config.

### some details

relevant parts of `stack.yaml`:

    compiler: ghc-8.0.1
    setup-info:
        ghc:
            linux64:
                8.0.1:
                    url: "http://server.local/ghc-8.0.1.tar.xz"
                    sha1: 185b114913b1b63b323203c387bf3a5e1b64ff34

stack version

    $ stack --version
    Version 1.3.3, Git revision 05acb2c0b6f84fe6981e88bebdcd44561aefa8bc (4474 commits) x86_64 hpack-0.15.0

mgsloan:
Ideally, it wouldn't fetch the stack-setup.yaml if it can find the needed info locally.  This would be a bit tricky to implement, as it would require interminging the logic for fetching the setup-info with the logic using it.

The warning could be an ok alternative, as the setup will fail if the requisite info isn't available.

PR #5031, opened 2019-09-13
Setup-info configurability

NightRa:
This PR comes to fix various issues related to how `stack` decides where to resolves it's tooling locations, for the initial purpose of using `stack` offline or behind a firewall. 

* See the documentation changes and the changelog for details on the new behaviors.

Stack 2 broke the previous behavior of the `setup-info` field - which now does not overwrite the default location. [#2983](https://github.com/commercialhaskell/stack/issues/2983), [#2913](https://github.com/commercialhaskell/stack/issues/2913)

This existing breakage allowed for a backwards-compatible implementation of the wanted behavior - via the `setup-info-locations` yaml config for remote locations, and using `setup-info` for inline configuration only.

Also, as the `--stack-setup-yaml` CmdArg has been deprecated for 3 years, it is now removed, in favor of `--setup-info-yaml` [#2647](https://github.com/commercialhaskell/stack/issues/2647)

Design decisions:
1. Usability / Backwards compatability: The default `stack-setup-2.yaml` is added to the `setup-info-locations` only if no locations were specified (either in the cmd args or in the config yaml)
2. Extendability: Allow extending setup locations - inline configs extend the default, and locations allow extensions by the locations following them.

PR considerations:
* Documentation was updated
* Changelog was updated
* Manual tests performed:
  * Overriding precedence between `setup-info`, command line args, `setup-info-locations` and the default location, with missing / invalid locations to check that no unwanted accesses are made.

Relevant issues:
- Fixes [setup-info field in stack.yaml not consulted](https://github.com/commercialhaskell/stack/issues/2983)
- Fixes [Combine stack-info-yaml argument and config.yaml](https://github.com/commercialhaskell/stack/issues/2982)
- Fixes [`stack setup` should not fail if it couldn't download default stack-setup-yaml](https://github.com/commercialhaskell/stack/issues/2913)
- Fixes [Use setup-info in stack.yaml](https://github.com/ndmitchell/offline-stack/issues/2)
- Fixes [Feature Request: allow setup-info urls to be relative file paths](https://github.com/commercialhaskell/stack/issues/3394)
- Change Monoid ordering of `SetupInfo` [Consider if other order of combination is more appropriate for the monoids on config types](https://github.com/commercialhaskell/stack/issues/2078)
- [Improve UI for `stack setup --stack-setup-yaml` option](https://github.com/commercialhaskell/stack/issues/2647)
though.

### DIFF
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -707,8 +707,6 @@ ghc:
             url: "installs/ghc-8.2.2.tar.xz"
 ```
 
-Executing `stack setup` outside of a project will use the global `stack.yaml` to determine the relevant `setup-info` information and locations.
-
 ### setup-info
 
 (Since 0.1.5)
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -653,7 +653,8 @@ The first location which specifies the location of a tool `(Tool, Platform, Vers
 ```yaml
 setup-info-locations:
 - C:/stack-offline/my-stack-setup.yaml
-- \\smbServer\stack\my-stack-setup.yaml
+- relative/inside/my/project/setup-info.yaml
+- \\smbShare\stack\my-stack-setup.yaml
 - http://stack-mirror.com/stack-setup.yaml
 - https://github.com/commercialhaskell/stackage-content/raw/master/stack/stack-setup-2.yaml
 ```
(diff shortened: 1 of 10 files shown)


doc/yaml_configuration.md
```diff
@@ -646,18 +688,7 @@ setup-info:
...
-setup-info: "https://example.com/my-stack-setup-info.yaml"
-```
-
-This may be either URL or (since 1.2.0) absolute file path.
-
-Note that this **adds** the specified setup info metadata to the default.
-If you need to **replace** it, use the `stack --setup-info-yaml` command-line
-argument instead.  The default setup metadata is in
-[stack-setup-2.yaml](https://github.com/commercialhaskell/stackage-content/raw/master/stack/stack-setup-2.yaml).
+Note that specifying this config *does not* the default `stack-setup-2.yaml` from being consulted as a fallback.
```
snoyberg: I think there's a typo in this sentence, I'm not sure how it's supposed to read.

src/Stack/Options/ConfigParser.hs
```diff
@@ -128,6 +129,11 @@ configOptsParser currentDir hide0 =
             <> help "Install binaries to DIR"
             <> hide
              ))
+    <*> many (
+        strOption
+            ( long "setup-info-yaml"
+           <> help "Alternate URL or absolute path for stack dependencies"
```
snoyberg: I would think a relative path would be even more convenient and recommended, so that the YAML file can be included in a repo and the code will work on multiple machines. Any underlying reason for preferring absolute paths?

NightRa:
Fixed PR comments & Implemented relative-path setup-info yaml files & tool paths (ghc & 7z).

There are 2 outstanding issues which are out-of-scope for the current PR and should be handled seperately:
1. [#5095](https://github.com/commercialhaskell/stack/issues/5095) - when used outside of a project,  `global-project/stack.yaml` configs (which are not project configs) are ignored, which shouldn't be the case.
2. When specifying a relative `setup-info-locations` path in `%STACK_ROOT%/config.yaml`, it's relative to the current project - either the local project or the global project, which is counter-intuitive, and should be relative to the `config.yaml` file - so only urls or absolute paths should be used in `config.yaml` for now.

NightRa:
Also, feel free to re-arrange the change-log, I had no clue which sections the changes should be in.

snoyberg:
Thanks!

PR #5076, opened 2019-11-07
Revert bad merge on stable

snoyberg:
The merge that @chrisdone did was correct, but shouldn't have been merged to stable (it should have remained on master). We _could_ do a force-push to stable to undo that and restore the previous commit (eb9f921e86e31d5df97d1b4041ae5c5e737b8ec2). This PR instead reverts both Chris's merge and a small commit I added on top to fix integration tests.

Once this is merged in, we would need to also revert this revert on master to avoid losing Chris's changes on master later. I have a branch for that ready to go: https://github.com/commercialhaskell/stack/pull/new/undo-revert-on-master.

@borsboom would you prefer the force-push approach or the revert+double revert approach?

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 
-## v2.3.x (unreleased changes)
+## Unreleased changes
 
 **Changes since v2.1.3.1**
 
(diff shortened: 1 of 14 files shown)


borsboom:
> would you prefer the force-push approach or the revert+double revert approach?

Force push does seem simpler, but could cause problems for anyone else who may have pulled stable at the wrong time, so probably the revert+double revert is better.

snoyberg:
That was stupid of me, I had this merging into `master` instead of `stable`. @borsboom can you give another sanity check before I merge? Then I'll open the second PR for `master`.

PR #5088, opened 2019-11-18
Build windows installer

jeffhappily:

### DIFF
--- a/etc/scripts/release.hs
+++ b/etc/scripts/release.hs
@@ -343,12 +343,13 @@ rules global@Global{..} args = do
             Just _ -> [x, x <.> sha256Ext, x <.> ascExt]
     binaryPkgFileNames =
         case platformOS of
-            Windows -> [binaryExeFileName, binaryPkgZipFileName, binaryPkgTarGzFileName]
+            Windows -> [binaryExeFileName, binaryPkgZipFileName, binaryPkgTarGzFileName, binaryInstallerFileName]
             _ -> [binaryExeFileName, binaryPkgTarGzFileName]
     binaryPkgZipFileName = binaryName <.> zipExt
     binaryPkgTarGzFileName = binaryName <.> tarGzExt
     -- Adding '-bin' to name to work around https://github.com/commercialhaskell/stack/issues/4961
     binaryExeFileName = binaryName ++ "-bin" <.> exe
+    binaryInstallerFileName = binaryName ++ "-installer" <.> exe
     binaryName =
         concat
             [ stackProgName
--- a/etc/scripts/release.hs
+++ b/etc/scripts/release.hs
@@ -270,15 +270,15 @@ rules global@Global{..} args = do
 
     releaseDir </> binaryInstallerFileName %> \out -> do
         need [releaseDir </> binaryExeFileName]
-        need [releaseDir </> "stack-install" <.> nsiExt]
+        need [releaseDir </> binaryInstallerNSIFileName]
 
         actionOnException
             (command_ [Cwd releaseDir] "c:\\Program Files (x86)\\NSIS\\Unicode\\makensis.exe"
                 [ "-V3"
-                , "stack-install.nsi"])
+                , binaryInstallerNSIFileName])
             (removeFile out)
 
-    releaseDir </> "stack-install" <.> nsiExt %> \out -> do
+    releaseDir </> binaryInstallerNSIFileName %> \out -> do
         need ["etc" </> "scripts" </> "build-stack-installer" <.> "hs"]
         cmd "stack etc/scripts/build-stack-installer.hs" 
             [ binaryExeFileName
@@ -367,6 +367,7 @@ rules global@Global{..} args = do
     binaryPkgTarGzFileName = binaryName <.> tarGzExt
     -- Adding '-bin' to name to work around https://github.com/commercialhaskell/stack/issues/4961
     binaryExeFileName = binaryName ++ "-bin" <.> exe
+    binaryInstallerNSIFileName = binaryName ++ "-installer" <.> nsiExt
     binaryInstallerFileName = binaryName ++ "-installer" <.> exe
     binaryName =
         concat
--- a/etc/scripts/release.hs
+++ b/etc/scripts/release.hs
@@ -270,7 +270,7 @@ rules global@Global{..} args = do
 
     releaseDir </> binaryInstallerFileName %> \out -> do
         need [releaseDir </> binaryExeFileName]
-        cmd "stack etc/scripts/build-stack-installer.hs" :: Action ()
+        need ["stack-install" <.> nsiExt]
         copyFile' (releaseDir </> binaryExeFileName) stackExeFileName
         actionOnException
             (command_ [] "c:\\Program Files (x86)\\NSIS\\Unicode\\makensis.exe"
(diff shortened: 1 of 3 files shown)


etc/scripts/release.hs
```diff
@@ -268,6 +268,20 @@ rules global@Global{..} args = do
                 cmd "strip -o"
                     [out, releaseBinDir </> binaryName </> stackExeFileName]
 
+    releaseDir </> binaryInstallerFileName %> \out -> do
+        need [releaseDir </> binaryExeFileName]
+        cmd "stack etc/scripts/build-stack-installer.hs" :: Action ()
```
borsboom: I'd suggest moving building the `.nsi` file to a separate target that's just responsible for running `build-stack-instasller.hs`.  Also make sure you `need` build-stack-installer.hs, so that if that file is changed it triggers re-running the script and generating the NSI.

jeffhappily: You mean something like this? https://github.com/commercialhaskell/stack/pull/5088/commits/493186304831fa5f19a58283a6692ff1342141c1#diff-cf9425a8bb87aa8706fdee9e97173e34R285

borsboom:
I see this is still marked WIP.  Do you think it's at a point where I can test it on Windows?

jeffhappily:
> I see this is still marked WIP. Do you think it's at a point where I can test it on Windows?

Yup, I just removed the WIP

jeffhappily:
> I tested this in Windows, and ran into an issue where it was writing files and copying things into the top-level directory of the repo, while all build artifacts should go under `_release`. This also led to a permissions error when it tried to copy the stack executable overtop of a pre-existing `stack.exe` in the top-level directory and got a permissions error (since that's the `stack` that was already running, and Windows doesn't allow you to overwrite a running executable).
> 
> I think there should be a `_release/installer/stack-<X.Y.Z>-windows-<ARCH>` directory in which the NSI file gets written.
> 
> In investigating how to do the above, I noticed that the generated NSI file has the paths for the source `stack.exe` and destination installer EXE written in it. Rather than having to copy the source `stack.exe` from its original location, and the copy the installer to where it belongs, the correct paths should be written into the generated NSI. This might be easier if the NSI generation code is inlined into `release.hs` rather than being in a separate script (alternatively, the separate generator script could accept command-line arguments for the paths of the destination NSI file, source `stack.exe`, and destination installer EXE)

I just made the changes accordingly, the `build-stack-installer.hs` is now taking args for the file names

etc/scripts/release.hs
```diff
@@ -268,6 +268,24 @@ rules global@Global{..} args = do
...
+        need [releaseDir </> binaryExeFileName]
+        need [releaseDir </> binaryInstallerNSIFileName]
+
+        actionOnException
+            (command_ [Cwd releaseDir] "c:\\Program Files (x86)\\NSIS\\Unicode\\makensis.exe"
+                [ "-V3"
+                , binaryInstallerNSIFileName])
+            (removeFile out)
+
+    releaseDir </> binaryInstallerNSIFileName %> \out -> do
```
borsboom: I think it'd be best to put the NSI file in a different subdirectory.  Something like `_release/installer/stack-<X.Y.Z>-windows-<ARCH>/stack-installer.nsi` (or `releaseInstallerDir </> binaryName </> binaryInstallerNSIFileName`, after setting `releaseInstallerDir` to `releaseDir </> "installer"`)

borsboom: Oh now I noticed you added the version to `stack-installer.nsi` itself.  That is probably fine.  I'll test it now.

etc/scripts/release.hs
```diff
@@ -268,6 +268,24 @@ rules global@Global{..} args = do
...
+        need ["etc" </> "scripts" </> "build-stack-installer" <.> "hs"]
+        cmd "stack etc/scripts/build-stack-installer.hs" 
+            [ binaryExeFileName
```
borsboom: This ends up creating an installer that installs the executable named e.g. `stack-2.2.0-windows-x86_64-bin.exe`, which means you can't just run `stack` after installing it.  Maybe there's a way to tell the NSI file to rename it to `stack.exe`?  Or maybe you can point the NSI file to `bin\stack-2.2.0-windows-x86_64\stack.exe` instead, which already has the right filename.

jeffhappily: NSI did provide a oname args to overwrite the output name, I've added it in `build-stack-installer.hs`

jeffhappily:
I've added the script to install `makensis`

borsboom:
Just tried this out on Azure and did doesn't seem installing makensis worked:
```
hocolatey v0.10.15
Installing the following packages:
nsis-unicode
By installing you accept licenses for the packages.
Error retrieving packages from source 'https://chocolatey.org/api/v2/':
 Object reference not set to an instance of an object.
nsis-unicode not installed. The package was not found with the source(s) listed.
 Source(s): 'https://chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://chocolatey.org/docs/troubleshooting for more 
 assistance.

Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Failures
 - nsis-unicode - nsis-unicode not installed. The package was not found with the source(s) listed.
 Source(s): 'https://chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://chocolatey.org/docs/troubleshooting for more 
 assistance.
```
(see https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=2610&view=logs&j=b2bb9c25-af23-5e2c-3485-082d2f8ef4d1&t=3a6dce7b-8a39-5aeb-06ab-ae0a4c8f59eb)

borsboom:
I did a little experiment, and it seems that using `choco install` fails if run from Bash, but it does work from Powershell:
```
  - powershell: |
      choco install nsis-unicode -y
    displayName: 'Chocolatey installation ${{parameters.os}}'
```
(it doesn't seem that the `choco install hg` was necessary, since `hg` is already installed by default)

borsboom:
Strangely, the installer build by CI (after moving the `choco install nsis-unicode` to a powershell script) doesn't seem to work for me.  When I run it, I get a dialog box with this error:

> ---------------------------
> NSIS Error
> ---------------------------
> Error writing temporary file. Make sure your temp folder is valid.
> ---------------------------
> OK   
> ---------------------------

However, when I build in my Windows VM using `stack etc/scripts/release.hs build`, the installer works fine.

jeffhappily:
> Strangely, the installer build by CI (after moving the `choco install nsis-unicode` to a powershell script) doesn't seem to work for me. When I run it, I get a dialog box with this error:
> 
> > ## NSIS Error
> > ## Error writing temporary file. Make sure your temp folder is valid.
> > ## OK
> 
> However, when I build in my Windows VM using `stack etc/scripts/release.hs build`, the installer works fine.

Do you mean the installer in this build? https://dev.azure.com/commercialhaskell/stack/_build/results?buildId=2612&view=artifacts&type=publishedArtifacts
Because it seems to work on my Windows machine though

borsboom:
That is the one I tried before, but I just tried it again and this time it worked  🤷‍♂ .  Maybe my download was corrupt last time.

So last thing is to move the `choco install` into a separate Powershell script, and then I think we're done.

jeffhappily:
Should I keep `choco install hg -y`? Since you said it's not necessary

borsboom:
You can remove `choco install hg -y`.

borsboom:
Looks great!  Nice work altogether.

jeffhappily:
Thanks!

Issue #5103, opened 2019-11-26
stack script needs network to work normally?


ppzzppz:
The following file:
```haskell
#!/usr/bin/env stack
{- stack --resolver global runghc
 --package process
 -}
{-# LANGUAGE OverloadedStrings #-}

main = do
  putStrLn "hello"
```
run command : stack xx.hs

Error reported in case of  without network
```
HttpExceptionRequest Request {
  host                 = "mirrors.tuna.tsinghua.edu.cn"
  port                 = 80
  secure               = False
  requestHeaders       = [("Accept","application/json"),("User-Agent","The Haskell Stack")]
  path                 = "/stackage/snapshots.json"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 0, addrAddress = <assumed to be undefined>, addrCanonName = <assumed to be undefined>}, host name: Just "mirrors.tuna.tsinghua.edu.cn", service name: Just "80"): does not exist (nodename nor servname provided, or not known))
```
expect： to work without network connection

qrilka:
@ppzzppz this seems to be an bug with an easy fix if the point is about the `global` resolver, see https://github.com/commercialhaskell/stack/blob/stable/src/Stack/Config.hs#L153 - snapshots are not used for the `ARGlobal` value.
PR with a fix will be welcomed.

ppzzppz:
I have tried and find that only “resolver global” doesn’t meet our expectation. The problem doesn’t seem too difficult, and I will try to fix

qrilka:
Much appreciated, let me know if I could help with it

snoyberg:
Resolved by #5104, thanks!

PR #5104, opened 2019-11-28
fix "--resolver global " needs network 

ppzzppz:
issue: https://github.com/commercialhaskell/stack/issues/5103

* [ x ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [ x ] The documentation has been updated, if necessary.

Test case:
```
stack test   673 examples, 0 failures
stack --resolver global   is ok
stack --resolver nightly is ok
stack --resolver lts-14   is ok
stack --resolver lts-14.13 is ok
```

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -146,7 +146,6 @@ makeConcreteResolver
     -> RIO env RawSnapshotLocation
 makeConcreteResolver (ARResolver r) = pure r
 makeConcreteResolver ar = do
-    snapshots <- getSnapshots
     r <-
         case ar of
             ARResolver r -> assert False $ makeConcreteResolver (ARResolver r)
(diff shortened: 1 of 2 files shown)


src/Stack/Config.hs
```diff
@@ -146,7 +146,7 @@ makeConcreteResolver
     -> RIO env RawSnapshotLocation
 makeConcreteResolver (ARResolver r) = pure r
 makeConcreteResolver ar = do
-    snapshots <- getSnapshots
+    -- Close #5103
```
qrilka: This doesn't seem to be a meaningful comment here

src/Stack/Config.hs
```diff
@@ -157,16 +157,19 @@ makeConcreteResolver ar = do
...
                     Nothing -> throwString $ "No LTS release found with major version " ++ show x
                     Just y -> return $ ltsSnapshotLocation x y
-            ARLatestLTS
-                | IntMap.null $ snapshotsLts snapshots -> throwString "No LTS releases found"
-                | otherwise ->
-                    let (x, y) = IntMap.findMax $ snapshotsLts snapshots
-                     in return $ ltsSnapshotLocation x y
+            ARLatestLTS -> do
+                snapshots <- getSnapshots
+                let r | IntMap.null $ snapshotsLts snapshots = throwString "No LTS releases found"
```
qrilka: Minor nitpicking but will it not look simpler if `if` were used here? Then `let` will not be required and that trailing `r` as well...

ppzzppz:
Received. Thanks for review.

Could changeLog be written under "Changes since v2.1.3.1"?


Such as the following：

```
**Changes since v2.1.3.1**

Release notes:

Major changes:

Behavior changes:

Other enhancements:

* Remove warning for using Stack with GHC 8.8 and Cabal 3.0.
* `--resolver global` doesn't retrieve snapshots list from the internet beause doesn't need it

Bug fixes:

```

qrilka:
Surely it's since v2.1.3.1 the question is whether it's an enhancement or a bug fix, I'm a bit more inclined to see it as a bug fix

PR #5106, opened 2019-11-28
Sigmundv add redhatenterpriseworkstation support

sigmundv:
I have added 'redhatenterpriseworkstation' as an option in the do_distro function in the get-stack.sh script. This is a trivial change that allows the script to run on this platform.


mattaudesse:
Sweet, thanks @sigmundv.

The only thing is we might be better off targeting `stable` rather than `master` (see [here](https://github.com/commercialhaskell/stack/pull/4352#issuecomment-429128795) for why). Mind toggling the destination branch, please?

dbaynard:
@sigmundv Please rebase your branch on to `stable`, too!

mattaudesse:
@sigmundv just in case it wasn't clear (my bad if so) we're looking for the same one line `git diff` you originally submitted but targeting `stable` rather than `master`.

You may find it easier to start fresh and branch directly from `stable`, add the `redhatenterpriseworkstation` bit again and submit that as a new PR, or you can update your current feature branch as @dbaynard suggested. We'll be all set so long as we're not including the extra commits from `master` (since those are generally pushed through to `stable` more selectively by the core developers during releases).

Thanks again!

mattaudesse:
Closing since progress has stalled out.

If anyone would like to pick this back up please refer to [this note](https://github.com/commercialhaskell/stack/pull/5106#issuecomment-562660431).

PR #5145, opened 2020-01-14
Remove the TMP env var on Windows integration tests

snoyberg:
@jeffhappily and I were debugging some integration test failures on CI, and believe this fix should do the trick for failing Windows builds. Sibi: is there a reason why this may cause problems?


snoyberg:
We can ignore any Linux or macOS failures here. Those are known and we're working around them separately. This should focus on just the Windows integration tests.

PR #5152, opened 2020-01-28
travis-complex.yml: Address validator warnings and hints

curiousleo:
Based on this Travis configuration validation report:

Screenshot from 2020-01-28 14-48-05

- `root`: deprecated key `sudo` (The key `sudo` has no effect anymore.)
   => `sudo: false` removed
- `root`: missing `os`, using the default `linux`
   => `os: linux` added at top-level
- `root`: key `matrix` is an alias for `jobs`, using `jobs`
   => `matrix` renamed to `jobs` at top-level

With the changes in this PR, the warnings go away.

### DIFF
--- a/doc/travis-complex.yml
+++ b/doc/travis-complex.yml
@@ -8,8 +8,8 @@
 # Copy these contents into the root directory of your Github project in a file
 # named .travis.yml
 
-# Use new container infrastructure to enable caching
-sudo: false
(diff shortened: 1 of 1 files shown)


curiousleo:
The build failures appear to be caused by AWS access key issues, unrelated to the changes in this PR.

PR #5164, opened 2020-02-06
V2.1.5

JoseD92:
Bump stack version for new release.


etc/dockerfiles/stack-build/lts-14.0/Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL maintainer="manny@fpcomplete.com"
 ARG GHC_VERSION=8.6.5
 ARG LTS_SLUG=lts-14.0
 ARG PID1_VERSION=0.1.2.0
-ARG STACK_VERSION=2.1.3
+ARG STACK_VERSION=2.1.5
```
borsboom: This should wait until after the new release is done.  Also, it should be done by adding a _new_ Dockerfile rather than modifying an existing one, like this: https://github.com/commercialhaskell/stack/blob/7fdb9acea87f0b967424dc8b8f97bbb472dde90b/etc/dockerfiles/stack-build/lts-13.25/Dockerfile

borsboom:
Have you checked build compatiblity with the latest minor versions of all the supported LTS snapshots?  Note: no need to check compatibility with latest `nightly`, we _know_ that won't work and won't be fixable in a minor release (since latest nightlies use Cabal-3.x).

borsboom:
The `pantry >= 0.2` constraint in `package.yaml` is redundant and should be removed.

borsboom:
I believe you'll have to update the stack version in [.github/ISSUE_TEMPLATE/bug_report.md](https://github.com/commercialhaskell/stack/blob/7fdb9acea87f0b967424dc8b8f97bbb472dde90b/.github/ISSUE_TEMPLATE/bug_report.md) and [.github/ISSUE_TEMPLATE/feature_request.md](https://github.com/commercialhaskell/stack/blob/7fdb9acea87f0b967424dc8b8f97bbb472dde90b/.github/ISSUE_TEMPLATE/feature_request.md).

borsboom:
Just a note: make sure this isn't merged to `stable` until the release has actually been performed (since otherwise get.haskellstack.org script will point to a non-existent version, among other problems).

borsboom:
It might be good to move this branch directly into the `commercialhaskell/stack` repo, so that we can run the nightly pipeline on it to ensure integration tests work and build the bindists.

JoseD92:
> Have you checked build compatiblity with the latest minor versions of all the supported LTS snapshots? Note: no need to check compatibility with latest `nightly`, we _know_ that won't work and won't be fixable in a minor release (since latest nightlies use Cabal-3.x).

after I removed pantry dependency I got:
```
$ stack --stack-yaml=stack-lts-12.yaml test --pedantic
Building all executables for `stack' once. After a successful build of all of them, only specified executables will be rebuilt.
stack> build (lib + exe + test)
Preprocessing library for stack-2.1.5..
Building library for stack-2.1.5..
[  5 of 110] Compiling Stack.Prelude
            
/home/jose/Documents/fpco/stack2/src/Stack/Prelude.hs:32:1: error:
    Could not find module ‘Pantry’
   |        
32 | import           Pantry               as X hiding (Package (..), loadSnapshot)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            
            
--  While building package stack-2.1.5 using:
      /home/jose/Documents/fpco/stack2/.stack-work/dist/x86_64-linux-tinfo6/Cabal-2.2.0.1/setup/setup --builddir=.stack-work/dist/x86_64-linux-tinfo6/Cabal-2.2.0.1 build lib:stack exe:stack test:stack-test --ghc-options " -fdiagnostics-color=always"
Progress 1/2

```

borsboom:
> after I removed pantry dependency I got:

That's OK, since `pantry` isn't in the snapshot at all, Stack can "pull" it in along with itself.

snoyberg:
I think we'll be going straight to 2.3. Can this be closed @borsboom?

borsboom:
Yes, closing.

PR #5162, opened 2020-02-05
Bump the default LTS from 11 to 14

snoyberg:
I'm fairly certain that we would also need to modify the integration
tests to use a more recent GHC, but not sure on procedure there.

Note: Documentation fixes for https://docs.haskellstack.org/en/stable/ should target the "stable" branch, not master.

Please include the following checklist in your PR:

* [ ] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [ ] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests!

PR #5167, opened 2020-02-07
Newer nightly

jeffhappily:


snoyberg:
@jeffhappily can you rebase this into a single commit?

@borsboom any objection to this?

borsboom:
It looks like this leaves us without any GHC 8.6-based snapshot, which means we can't be guaranteed that Stack will be build-able with that version.  After #5162 is merged this would be fine (since it changes `stack.yaml` to use a GHC 8.6-based snapshot), but if not a `stack-lts-14.yaml` and `snapshot-lts-14.yaml` should be added and CI configuration adjusted to also build that.

snoyberg:
With the change to GHC version based Stack config files, this won't be necessary. I'll include a `stack-ghc-88.yaml` in the `newer-lts` branch instead.

snoyberg:
Just realized that the code changes in here need to be included, my mistake.

Issue #5147, opened 2020-01-16
Local package dependency `ghc-pkg: cannot find package`


lehins:
### General summary/comments (optional)

When one package `bar` has another local package `foo` as a dependency (either in `packages` or `extrs-deps`), rebuilding separately package `foo` will cause package `bar` erroring during build with:

> ghc-pkg: cannot find package foo

### Steps to reproduce

Here are two sample projects that replicate the problem: https://github.com/lehins/bugs/tree/master/haskell/stack-ghc-pkg-cannot-find-package

```
$ git clone https://github.com/lehins/bugs
$ cd bugs/haskell/stack-ghc-pkg-cannot-find-package
$ cd bar
$ stack build
$ cd ../foo
$ stack build
$ cd ../bar
$ stack build --force-dirty
...
Standard error:

ghc-pkg: cannot find package foo
```

Not 100% sure, but I think it is important that `foo` and `bar` are both on a different snapshots, but with the same ghc, eg. one is `lts-14.4` another `lts-14.20`.

### Expected

Regardless of the order packages build successfully.

### Actual

Here is also a sample log with `--verbose`: https://github.com/lehins/bugs/blob/master/haskell/stack-ghc-pkg-cannot-find-package/output.log

### Stack version

Stack version used was the latest, but I've been observing this issue since `2.0`, but was unable to reproduce it reliably on small set of packages
```
$ stack --version
Version 2.2.0, Git revision c3c33b61a4d5c286a6718550d285305270ba4bf0 PRE-RELEASE x86_64 hpack-0.31.2
```

### Method of installation

* `stack upgrade --git`

lehins:
Just ran into this bug again, but both of the packages had exactly the same resolver in them. All I did was compile `bar` with different ghc flags and `foo` would no longer build.

lehins:
Here is another depiction of the same bug, I believe. Similar setup, but the error is:
```
setup: Cannot process the test suite 'tests' because test suites are not
enabled. Run configure with the flag --enable-tests
```

snoyberg:
Fixed by #5169

PR #5169, opened 2020-02-10
Include setup-config modification time into reconfigure checks

qrilka:
This resolves issue #5147 appearing because a path for setup-config for a
package depend on OS and Cabal versions and is stored in .stack-work
inside package directory

Tested locally using repro provided by @lehins in #5147

### DIFF
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -11,6 +11,7 @@ module Stack.Build.Cache
     ( tryGetBuildCache
     , tryGetConfigCache
     , tryGetCabalMod
+    , tryGetSetupConfigMod
     , getInstalledExes
     , tryGetFlagCache
     , deleteCaches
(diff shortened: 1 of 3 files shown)


qrilka:
@snoyberg I've added label WIP because I have a feeling that in principle we could have a better way to address this - e.g. if there will be no per-package `.stack-work`. Is it possible to find the issue which caused Stack to have per-package `.stack-work`?
Also I have somewhat related idea that having per-snapshot dist directories could help us caching local builds but that will have a problem of cache invalidation and also has much higher scope than the addressed issue.
But in principle this PR looks to be good-enough short-term (?) solution. If there are no objections then I could a changelog entry marking #5147 as resolved.

lehins:
I'll install this version locally to make sure issue is gone. I will report if I run into any problems.

qrilka:
And hlint hints don't seem to be related to this PR

lehins:
I've been using this `5147-fix-shared-setup-config` for 11 days now and can report that I haven't seen the originally reported issue pop up in my workflow. Thank you @qrilka for fixing this!

qrilka:
Oops @snoyberg - I added WIP label because of missing changelog entry, I'll add it a bit later into `master` if there are no objections

qrilka:
Done in a9f274ab

PR #5171, opened 2020-02-10
Add http proxy env vars in the stack-created nix shell

carlosdagos:
Hello! Draft PR for now as I don't know if this has further 
implications. Help/Feedback would be really appreciated :slightly_smiling_face: 

This helps with an issue where http proxy environment variables get dropped
by a pure nix shell.

When stack can't find a nix shell file, it creates its own ad-hoc nix
shell. If the `nix.pure` config is set, this will use a command such
as `nix-shell --pure ...`, which drops http proxy environment
varibles.

For users who depend on http proxy env vars to access the
internet (such as in some corporate environments), this means that
`stack` will be unable talk to the internet and fail at downloading
package indices or any other supporting assets. This can be somewhat
confusing because `nix.pure` is set by default to `true`.

### DIFF
--- a/src/Stack/Nix.hs
+++ b/src/Stack/Nix.hs
@@ -48,6 +48,16 @@ runShellAndExit = do
              traverse (resolveFile projectRoot) $ nixInitFile (configNix config)
          Nothing -> pure Nothing
 
+     -- See: https://github.com/NixOS/nixpkgs/blob/master/lib/fetchers.nix
+     -- This looks up the proxy env vars and returns a list of tuples
+     -- containing the var name and its env value, for those that exist in the
+     -- environment
+     proxyImpureEnvVars <- do
+       let varNames = ["http_proxy", "https_proxy", "ftp_proxy", "all_proxy", "no_proxy"]
+       proxyVars <- forM varNames $ \n -> do
+          fmap (\v -> (T.pack n, T.pack v)) <$> liftIO (lookupEnv n)
+       pure $ catMaybes proxyVars
+
      -- This will never result in double loading the build config, since:
      --
      -- 1. This function explicitly takes a Config, not a HasConfig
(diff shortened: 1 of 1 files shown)


qrilka:
I'd vote against this PR as it actually works against pure Nix mode.
At the same time I agree that pure mode by default is sometimes confusing for users who don't spend much time reading the docs. Improving documentation is a better way to help with the problem described.

carlosdagos:
Thanks for your feedback @qrilka.

I can understand the uneasiness regarding nix-shell purity. The reason I linked to the [`lib/fetches.nix`](https://github.com/NixOS/nixpkgs/blob/master/lib/fetchers.nix) file is to make the argument that the nix builders borrow these env vars in order to fetch artefacts from the network. 

This patch is meant simply to help `stack` fetch artefacts from the network as well.

Without this patch there's obviously other solutions, for example adding the http proxy env vars to `~/.bashrc` (which a pure nix-shell will still source). I don't have one in my system, but obviously I'm not against adding one and just add the http env vars in there. Another solution clearly is to just set `nix.pure` to `false`, which honestly I'm not fussed about either.

With this patch however, `stack` passes through these env vars to the ad-hoc pure nix-shell transparently without modifying the default config.

balsoft:
(Review on request from @qrilka)

I'd rather add a statement in documentation regarding `nix-shell` purity and the need to set `nix.pure` to false when you need external variables to pass through. Alternatively, passing those environment variables should be pretty easy to do upstream -- just add them to https://github.com/NixOS/nix/blob/d8972317fc4314864619cadd5620ae780da657a3/src/nix-build/nix-build.cc#L100 and submit a PR describing what and why.

qrilka:
Custom `shell.nix` could be another solution also.
Also there is an option shared by @balsoft that `nix-shell` could be changed to pass those variables itself

carlosdagos:
Hi all. Referenced PR got merged to nix as per @balsoft’s input. Closing this one 😀 Thanks!

balsoft:
@carlosdagos Cool, glad I could be of help!

qrilka:
Thank you both @carlosdagos and @balsoft

PR #5177, opened 2020-02-14
WIP: Add new docker file for lts 15

jeffhappily:

### DIFF
--- a/etc/dockerfiles/stack-build/lts-15.0/Dockerfile
+++ b/etc/dockerfiles/stack-build/lts-15.0/Dockerfile
@@ -1,0 +1,89 @@
+FROM ubuntu:18.04
+
+LABEL maintainer="manny@fpcomplete.com"
+
+ARG GHC_VERSION=8.8.2
+ARG LTS_SLUG=nightly-2020-02-11
+ARG PID1_VERSION=0.1.2.0
+ARG STACK_VERSION=2.1.3
+ARG CUDA_VERSION=10.0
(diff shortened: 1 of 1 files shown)


etc/dockerfiles/stack-build/lts-15.0/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+
+LABEL maintainer="manny@fpcomplete.com"
+
+ARG GHC_VERSION=8.8.2
+ARG LTS_SLUG=nightly-2020-02-11
+ARG PID1_VERSION=0.1.2.0
+ARG STACK_VERSION=2.1.3
+ARG CUDA_VERSION=10.0
+ARG JVM_PATH=/usr/lib/jvm/java-8-openjdk-amd64
+ARG LLVM_VERSION=7.0
```
borsboom: In this case, I think it should be set to `7` since the directory is `/usr/lib/llvm-7/include`.  In fact I would recommend to switching to the approach used for the JVM, and making this an `LLVM_PATH` variable instead, set to `/usr/lib/llvm-7`, and then adjust the way it's used when setting the `CPATH` environment variable below.

borsboom:
Since the Dockerfiles are moving to the Stackage repo (https://github.com/commercialhaskell/stackage/pull/5173), I think this MR should be moved there.

In addition, LTS-15.0 has now been released, so the new MR can use the final `lts-15.0` snapshot slug.

PR #5178, opened 2020-02-14
4524 Always use verified downloads

### LINKED ISSUE #4524: (Code ?) improvement: always use verified downloads
While working on #4515 it was found that `Stack.New` uses `download` and not `verifiedDownload` and the main difference between those is that `download` doesn't tack download progress and thus doesn't check download length as `verifiedDownload` could (with correct callback specified as `progressSink` parameter). Also it looks that `redownload` doesn't include verifications from `verifiedDownload` - it makes sense to make any downloads properly verified.

v0d1ch:
fixes #4524

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -31,14 +31,12 @@ module Stack.Setup
 import qualified    Codec.Archive.Tar as Tar
 import              Conduit
 import              Control.Applicative (empty)
-import              Control.Monad.State (get, put, modify)
 import "cryptonite" Crypto.Hash (SHA1(..), SHA256(..))
 import              Pantry.Internal.AesonExtended
 import qualified    Data.ByteString as S
 import qualified    Data.ByteString.Lazy as LBS
 import qualified    Data.Conduit.Binary as CB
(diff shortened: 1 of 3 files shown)


src/Stack/Setup.hs
```diff
@@ -1619,42 +1620,52 @@ chattyDownload label downloadInfo path = do
             , drLengthCheck = mtotalSize
             , drRetryPolicy = drRetryPolicyDefault
             }
-    x <- verifiedDownload dReq path chattyDownloadProgress
+    x <- verifiedDownload dReq path (chattyDownloadProgress label mtotalSize)
```
qrilka: what about adding it as something like 
```
verifiedDownloadWithProgress req path lbl msize = verifiedDownload req path (chattyDownloadProgress lbl msize)
```
into `Network.HTTP.StackClient` with `chattyDownloadProgress` as a part of it. That way we e.g. won't need o export `chattyDownloadProgress` from `Stack.Setup` which looks a bit odd

v0d1ch: @qrilka Thanks for the review! I moved some of the relevant code from `Setup` to `StackClient` module so hopefully all is good now.

PR #5188, opened 2020-02-20
On newer-lts branch: switch Windows to 64-bit builds

snoyberg:
It looks to me like there may be a bug in 32-bit GHC 8.6.5 on Windows. I've been seeing:

```
pantry                > ghc.exe:  | D:\a\1\s\.stack-root\snapshots\d08f6e4c\lib\i386-windows-ghc-8.6.5\persistent-sqlite-2.9.3-HO1GqQrWtVL3PHyDCYo1QZ\HSpersistent-sqlite-2.9.3-HO1GqQrWtVL3PHyDCYo1QZ.o: unknown symbol `___udivmoddi4'
```


.azure/azure-nightly-template-windows.yml
```diff
@@ -43,7 +43,7 @@ jobs:
   - powershell: |
       $env:STACK_ROOT = "$(Build.SourcesDirectory)\.stack-root"
       $env:PATH += ";$env:HOME\.local\bin"
-      stack etc/scripts/release.hs --arch=i386 build
+      stack etc/scripts/release.hs --arch=x86_64 build
```
borsboom: This MR looks like it has the same issue as #5189 -- builds 64-bit release twice.

borsboom: This MR looks like it has the same issue as #5189 -- builds 64-bit release twice.

snoyberg: It's actually better to simply close this PR in favor of #5189.

PR #5189, opened 2020-02-20
CI: Build Windows 64-bit only

snoyberg:
Separate from #5188, this applies the change directly to master. If this is merged, then #5188 should be closed and #5162 should be rebased.

### DIFF
--- a/.azure/azure-nightly-template-windows.yml
+++ b/.azure/azure-nightly-template-windows.yml
@@ -44,8 +44,6 @@ jobs:
       $env:STACK_ROOT = "$(Build.SourcesDirectory)\.stack-root"
       $env:PATH += ";$env:HOME\.local\bin"
       stack etc/scripts/release.hs --arch=x86_64 build
-      git clean -f
-      stack etc/scripts/release.hs --arch=x86_64 build
       copy _release/stack-* $(Build.ArtifactStagingDirectory)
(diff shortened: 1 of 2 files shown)


.azure/azure-nightly-template-windows.yml
```diff
@@ -43,7 +43,7 @@ jobs:
   - powershell: |
       $env:STACK_ROOT = "$(Build.SourcesDirectory)\.stack-root"
       $env:PATH += ";$env:HOME\.local\bin"
-      stack etc/scripts/release.hs --arch=i386 build
+      stack etc/scripts/release.hs --arch=x86_64 build
```
borsboom: It looks like this is now building 64-bit _twice_ (look two lines down), instead of building 32-bit and then 64-bit.

snoyberg: Doh, good catch, pushing a fix.

borsboom: Does this mean we're going to stop building and distributing 32-bit `stack` bindists?

snoyberg: Yes. Recent CI failures when upgrading to GHC 8.6.5 indicate that there are GHC bugs which will cause 32 bit Stack to crash

PR #5191, opened 2020-02-21
Add snapshot for building with lts-14 series

afcowie:
There is a `nightly` snapshot (which is, incidentally, rather out of date) and a few snapshots building against older stable `lts`s. I was stable on GHC 8.6.5 and `lts-14.27` so I added a snapshot file for the most recent stable `lts-14.27`. 

(I tried `lts-15.0` but haven't got that working yet)


snoyberg:
See #5162 and #5189

Issue #3824, opened 2018-01-30
Remove uses of throwString / any other sources of CallStack in non-panic exceptions


mgsloan:
Currently `throwString` is used in 38 places.  However, it includes a callstack, which I think is ugly for non-panic situations - https://github.com/commercialhaskell/stack/issues/3823 .  Probably better to use `fail`, or potentially add or own utility that directly throws a string.

rajcspsg:
@mgsloan Any update on my previous comment. Can I take this issue to work?

mgsloan:
Hey, sure, that'd be great, thanks!

snoyberg:
@rajcspsg Bumping, were you still planning on working on this?

rajcspsg:
@snoyberg - I'm really sorry. I'm not able to work on this in near future.

snoyberg:
No problem, thanks for the update

tjlaxs:
What would be a panic case? I went through these and I think in every case we pretty much know the state and there is no panic in my understanding. So this would end up being either just a search and replace. And it compiles so I guess it works... just kidding 😄 

Also testing these would turn pretty long in many cases (and maybe pointless). I did a quick Hspec test to see what it would look like with Stack.Upload.uploadBytes:
```haskell
  it "uploadBytes fails without callstack" $ example $ do
    let hc = HackageCreds "a" "b" "c"
        up = uploadBytes "http://hackage.haskell.org/" hc "nonexistent.tar" ""
    catch up (\e -> displayException (e :: IOException) `shouldBe` "user error (Authentication failure uploading to server)")
```
(I did this purely to brush up my exception understanding and also exposed the data constructor for the HackageCreds while doing so for being a bit lazy.)

Would you like a PR without the testing part?

snoyberg:
PR sounds good, thanks!

PR #5192, opened 2020-02-21
3824 Replaced throwString with fail

tjlaxs:

Quick test described in #3824

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -290,7 +290,7 @@ queryBuildInfo selectors0 =
             _ -> err $ "Cannot apply selector to " ++ show value
       where
         cont = select (front . (sel:)) sels
-        err msg = throwString $ msg ++ ": " ++ show (front [sel])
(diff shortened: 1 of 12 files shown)


snoyberg:
Sorry, I must have misunderstood the original discussion. Moving to `fail` isn't a good idea, since (1) it's not really any better than `throwString` and (2) with GHC 8.8 it's not compatible with `RIO`. I would be in favor of moving towards concrete exception types.

tjlaxs:
I would be willing to try to fix this, but might need a bit of more information what you mean by "concrete exception types". With `RIO` I guess something like on the "Exceptions" section on rio packages documentation would still be relevant? Looking at the code base this pattern seems to show up in multiple places.

snoyberg:
Yes, that's the basic idea. Want to push a commit making a change in one or two locations and I can review if it's in the right direction?

snoyberg:
No update here, I'm going to close for now. If you're interested in pursuing, please reopen.

PR #5201, opened 2020-02-26
Fix integration test

jeffhappily:

### DIFF
--- a/snapshot-lts-11.yaml
+++ b/snapshot-lts-11.yaml
@@ -25,9 +25,7 @@ packages:
 - typed-process-0.2.6.0@rev:0 # for rio-0.1.10.0
 - rio-prettyprint-0.1.0.0@rev:0
 - hi-file-parser-0.1.0.0@rev:0
-# - http-download-0.1.0.0@rev:0
-- github: commercialhaskell/http-download
-  commit: c3b7eb3dcbb78491e91065a616a45cfa7ebc265c
+- http-download-0.2.0.0@rev:0
 - github: snoyberg/filelock
   commit: 97e83ecc133cd60a99df8e1fa5a3c2739ad007dc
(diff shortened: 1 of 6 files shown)


snapshot-lts-11.yaml
```diff
@@ -25,7 +25,9 @@ packages:
 - typed-process-0.2.6.0@rev:0 # for rio-0.1.10.0
 - rio-prettyprint-0.1.0.0@rev:0
 - hi-file-parser-0.1.0.0@rev:0
-- http-download-0.1.0.0@rev:0
+# - http-download-0.1.0.0@rev:0
```
snoyberg: I've released 0.2.0.0 to Hackage now

src/Stack/New.hs
```diff
@@ -157,24 +157,26 @@ loadTemplate name logIt = do
...
-            dReq = DownloadRequest
-              { drRequest = req
-              , drHashChecks = []
-              , drLengthCheck = Just downloadFileSize
-              , drRetryPolicy = drRetryPolicyDefault
-              }
+        let path' = toFilePath path
+        exist <- liftIO $ fileExist path'
+
+        downloadFileSize <- if exist
```
snoyberg: Maybe I misunderstood our discussions before, but why do we care about the size of the file on the filesystem?

jeffhappily: The file size is also being use here https://github.com/commercialhaskell/stack/blob/cba6996791c2a7253388acab1b056aa5d30fe35e/src/Network/HTTP/StackClient.hs#L177 as the total size for the progress, I was thinking  to not change the behavior

snoyberg: But isn't that known to be wrong?

jeffhappily: Yeah that's true, OK I'll remove it

PR #5218, opened 2020-03-08
Update documentation related to Azure CI

psibi:

@snoyberg I missed reviewing your https://github.com/commercialhaskell/stack/pull/5209 , so here is my proposed changes. We can iterate on this if you disagree on parts of the MR. Thanks!

### DIFF
--- a/doc/azure_ci.md
+++ b/doc/azure_ci.md
@@ -31,11 +31,7 @@ used on your projects:
 * The complex Azure configuration is intended for projects that need
   to support multiple GHC versions and multiple OSes, such as open
   source libraries to be released to Hackage. It tests against Stack
-  on Linux and macOS. The configuration is significantly more involved
-  to allow for all of this branching behavior. In the provided
-  template, we use the Linux build configuration to test against
-  various combinations of stack. The macOS and Windows configuration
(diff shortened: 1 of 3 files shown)


doc/azure_ci.md
```diff
@@ -30,14 +30,13 @@ used on your projects:
...
   to support multiple GHC versions and multiple OSes, such as open
-  source libraries to be released to Hackage. It tests against
-  cabal-install, as well as Stack on Linux and macOS. The
-  configuration is significantly more involved to allow for all of
-  this branching behavior. In the provided template, we use the Linux
-  build configuration to test against various combinations of
-  cabal-install and stack. The macOS and Windows configuration is used
-  only for testing Stack build configuration. These are the files for
-  the complex configuration:
+  source libraries to be released to Hackage. It tests against Stack
```
snoyberg: I think the wording here can be simplified even further, can't it? We now have essentially identical scripts on Mac, Windows, and Linux, and test only against Stack, without any branching behavior.

psibi: I have pushed a fix simplifying it further. Let me know if it needs further refining. Thanks!

PR #5224, opened 2020-03-13
Fix Docker version parsing error

### LINKED ISSUE #5193: docker version isn't being parsed
Current bug happens on `master` and I suspect it is because we switched to Cabal-3

Here is a problem I ran into:
```
$ stack build --fast --verbose
Version 2.2.0, Git revision 681ca331609e555d62e9aacbbc18770f81abd8de PRE-RELEASE x86_64 hpack-0.32.0
2020-02-21 23:40:28.302738: [debug] Checking for project config at: <redacted>stack.yaml
2020-02-21 23:40:28.303210: [debug] Loading project config file stack.yaml
2020-02-21 23:40:28.308326: [debug] SELECT COUNT(*) FROM "last_performed" WHERE ("action"=?) AND ("timestamp">=?); [PersistInt64 1,PersistUTCTime 2020-02-20 20:40:28.308297 UTC]
2020-02-21 23:40:28.308994: [debug] Run process: /usr/bin/docker --version
2020-02-21 23:40:28.624191: [debug] Process finished in 315ms: /usr/bin/docker --version
2020-02-21 23:40:28.624667: [error] Cannot get Docker version (invalid 'docker --version' output).
$ /usr/bin/docker --version
Docker version 19.03.6, build 369ce74a3c
```
As it turns out Cabal has changed their version parser and leading zero is no longer valid:
https://github.com/commercialhaskell/stack/blob/df34c0ad06d884c870ce79f2d672335897cd9976/src/Stack/Docker.hs#L402-L412

This definition is from pantry: `parseVersion = simpleParse`
```
λ> import Distribution.Text
λ> simpleParse "19.03.6" :: Maybe Version
Nothing
λ> simpleParse "19.3.6" :: Maybe Version
Just (mkVersion [19,3,6])
```

One possible solution would be to switch to the more general solution provided by base in `Data.Version`, since `docker` is not a Cabal project anyways:

```haskell
λ> import Text.ParserCombinators.ReadP
λ> import Data.Version
λ> let pv = readP_to_S parseVersion
λ> last (pv "19.03.6," :: [(Version, String)])
(Version {versionBranch = [19,3,6], versionTags = []},",")
```

gdziadkiewicz:
The error was caused by lack of support for leading zeros in Cabal's
Distribution.Version that was used for Docker version parsing. The fix
introduces middle step of parsing the version using Data.Version
and converting it to Distribution.Version. Fix #5193

### Testing
I tested it by running it on my machine which has Docker version 19.03.8, build afacb8b installed
and observed that the error disappeared after introducing my changes.

### DIFF
--- a/src/Stack/Docker.hs
+++ b/src/Stack/Docker.hs
@@ -36,11 +36,12 @@ import           Data.Conduit.Process.Typed hiding (proc)
 import           Data.List (dropWhileEnd,isPrefixOf,isInfixOf)
 import           Data.List.Extra (trim)
 import qualified Data.Map.Strict as Map
+import           Data.Maybe (listToMaybe)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import           Data.Time (UTCTime)
-import           Data.Version (showVersion)
-import           Distribution.Version (mkVersion)
+import qualified Data.Version (showVersion, parseVersion, Version)
+import           Distribution.Version (mkVersion, mkVersion')
 #if MIN_VERSION_path(0,7,0)
 import           Path hiding (replaceExtension)
 #else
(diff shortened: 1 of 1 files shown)


snoyberg:
Thank you!

PR #5250, opened 2020-04-02
Update Fedora repo in the install guide

Iisus:
The Fedora Copr repo link was pointing to an obsolete repository, containing stack 1.9.3, so I've updated the link to point to the newer repo.

PR #5262, opened 2020-04-20
Update Fedora repo in the install guide

Iisus:
The Fedora Copr repo link was pointing to an obsolete repository, containing stack 1.9.3, so I've updated the link to point to the newer repo.

### DIFF
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -141,7 +141,7 @@ Use the [generic Linux option](#linux).
 
 There is also an
 unofficial
-[Fedora Copr repo](https://copr.fedoraproject.org/coprs/petersen/stack/) which
+[Fedora Copr repo](https://copr.fedoraproject.org/coprs/petersen/stack2/) which
 can be enabled with: `sudo dnf copr enable petersen/stack`. Note that this Stack
 version may lag behind, so we recommend running `stack upgrade` after installing
 it.
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -142,7 +142,7 @@ Use the [generic Linux option](#linux).
 There is also an
 unofficial
 [Fedora Copr repo](https://copr.fedoraproject.org/coprs/petersen/stack2/) which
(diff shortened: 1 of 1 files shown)


Iisus:
Original pull request, #5250, was based on master. I've closed that one and opened this one, based on stable.
@snoyberg @juhp

snoyberg:
Looks correct to me, @juhp can correct me if this needs to be reverted

juhp:
Thank you @Iisus for this!

Sorry for taking so long to respond - I wasn't really slacking, but actually writing a small tool to do progressive copr builds... :-) and was also thinking about this issue, so sorry for the long reply. ;o)

Just a few comments first anyway on Stack and Fedora:

First some good news: stack will be available as a package in Fedora 32 (actually subpackaged under haskell-platform - I still want to have it as a proper separate source Fedora package eventually of course - (also, too much detail, but to be accurate I think it will probably be shipped as a 0-day update to F32).

As for the coprs, the reason I made the separate (conservative) `stack2` Copr repo was that when stack-2.1 was released initially there were some incompatibilities etc compared to the previous 1.9 release, so I wanted to have both versions available to users. But by now recent Stackage releases are no longer supported by 1.9, so I am plan to build 2.1 soon in the `stack` copr (and keep a `stack-1.9` copr in case anyone still needs it, eg on RHEL7). It seems a large new stack release is coming, and I may be tempted to put that first into the `stack2` copr initially because it will no longer support ghc7 (yeah I know ghc7 is getting old:)

So in conclusion probably the merged PR is fine for now, thanks.

But in the longer term I may get rid of the `stack2` repo, but we can revisit this before that: because the naming of the current two repos is actually a bit confusing for users - I may well end up with a new `stack-2.1` copr and then use the `stack` copr for the latest release (unfortunately copr doesn't support renaming of projects, so this will be a bit of extra work alas but nevermind). More importantly going forward stack should be available as an official Fedora package for all archs (though I am not sure how many Linux archs are supported upstream), and due to dependencies etc it may at times lag behind the latest upstream release specially for older Fedora and EPEL releases.

PR #5267, opened 2020-04-21
Static linux

snoyberg:

### DIFF
--- a/.azure/azure-linux-template.yml
+++ b/.azure/azure-linux-template.yml
@@ -10,17 +10,27 @@ jobs:
       GHC 8.4:
         BUILD: stack
         STACK_YAML: stack-ghc-84.yaml
+        EXTRA_SUFFIX: ""
       GHC 8.6:
         BUILD: stack
         STACK_YAML: stack-ghc-86.yaml
+        EXTRA_SUFFIX: ""
       GHC 8.8:
         BUILD: stack
         STACK_YAML: stack-ghc-88.yaml
+        EXTRA_SUFFIX: ""
+      Alpine:
+        BUILD: stack
+        STACK_YAML: stack.yaml
+        EXTRA_SUFFIX: "alpine"
+        STACK_ARGS: --docker --system-ghc --no-install-ghc --flag stack:static
       style:
         BUILD: style
+        EXTRA_SUFFIX: ""
       pedantic:
         BUILD: pedantic
         STACK_YAML: stack.yaml
+        EXTRA_SUFFIX: ""
   steps:
   - script: |
       export STACK_ROOT="$(Build.SourcesDirectory)"/.stack-root;
@@ -28,8 +38,8 @@ jobs:
       curl -f -L "https://github.com/fpco/cache-s3/releases/download/${CACHE_S3_VERSION}/cache-s3-${CACHE_S3_VERSION}-${OS_NAME}-x86_64.tar.gz" -o ~/.local/bin/cache-s3.tar.gz
       tar xzf ~/.local/bin/cache-s3.tar.gz -C ~/.local/bin
       export PATH=$HOME/.local/bin:$PATH;
-      cache-s3 --prefix="${CACHE_S3_PREFIX}" --git-branch="$(Build.SourceBranchName)" --suffix="${OS_NAME}" restore stack --base-branch="${BASE_BRANCH}"
-      cache-s3 --prefix="${CACHE_S3_PREFIX}" --git-branch="$(Build.SourceBranchName)" --suffix="${OS_NAME}" restore stack work --base-branch="${BASE_BRANCH}"
+      cache-s3 --prefix="${CACHE_S3_PREFIX}" --git-branch="$(Build.SourceBranchName)" --suffix="${OS_NAME}${EXTRA_SUFFIX}" restore stack --base-branch="${BASE_BRANCH}"
+      cache-s3 --prefix="${CACHE_S3_PREFIX}" --git-branch="$(Build.SourceBranchName)" --suffix="${OS_NAME}${EXTRA_SUFFIX}" restore stack work --base-branch="${BASE_BRANCH}"
       etc/scripts/ci-setup.sh
       case "$BUILD" in
         style)
@@ -38,11 +48,9 @@ jobs:
           ;;
         *)
           export PATH=$HOME/.local/bin:$PATH
-          stack --install-ghc $ARGS test --bench --only-dependencies
+          stack test $STACK_ARGS --bench --only-dependencies
           ;;
       esac
-      GHC_OPTIONS="-Werror"
-      if [ "$GHCVER" = "8.2.1" ]; then GHC_OPTIONS="$GHC_OPTIONS -Wno-missing-home-modules"; fi
       set -ex
       case "$BUILD" in
         style)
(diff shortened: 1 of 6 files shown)


snoyberg:
* [ ] Initial review of CI output makes it look like the nightly Linux job is not doing a static build, need to investigate

snoyberg:
This all works perfectly.... except that the executable in the Linux artifact segfaults!?! Investigating...

snoyberg:
It was the stripping! 4a53af2f41ec8127538566d95a82102a6f1403de

snoyberg:
Not sure why stripping a static executable makes it segfault, but other than that this is ready for review and potential merge.

stack-alpine.yaml
```diff
@@ -0,0 +1,30 @@
+resolver: lts-14.27
```
borsboom, line 1: Will we need to remember to keep this in sync with `stack.yaml`?

I think the same adjustments could be achieved by passing arguments to `stack` rather than needing a separate file, but not sure which approach will end up being the bigger maintenance burden.

.azure/azure-linux-template.yml
```diff
@@ -10,26 +10,40 @@ jobs:
...
       GHC 8.8:
         BUILD: stack
         STACK_YAML: stack-ghc-88.yaml
+        EXTRA_SUFFIX: ""
+        GHC_OPTIONS: "-Werror"
+      Alpine:
+        BUILD: stack
+        STACK_YAML: stack-alpine.yaml
+        EXTRA_SUFFIX: "alpine"
+        ARGS: --system-ghc --no-install-ghc
```
borsboom, line 29: Maybe calling this `STACK_ARGS` would be more clear.

.azure/azure-linux-template.yml
```diff
@@ -51,7 +63,16 @@ jobs:
           hlint test/ --cpp-simple
           ;;
         stack)
-          stack test --haddock --no-haddock-deps --ghc-options="$GHC_OPTIONS"
+          stack $ARGS test --haddock --no-haddock-deps --ghc-options="$GHC_OPTIONS" --copy-bins --local-bin-path bin
+
+          # Get output about whether the exe is dynamically linked
+          # Ignore the exit code though, we don't want to fail just because it is static
+          ldd ./bin/stack || true
```
borsboom, line 70: Since this always succeeds, is it intended to remain here or was it just intended for debugging?  Should we `grep` the output for an expected string (depending on whether we want static or dynamic) and fail if it's wrong?

etc/scripts/release.hs
```diff
@@ -74,7 +74,7 @@ main =
             -- build the default value of type Global, with predefined constants
 
             -- 'stack build --dry-run' just ensures that 'stack.cabal' is generated from hpack
-            _ <- readProcess "stack" ["build", "--dry-run"] ""
+            _ <- readProcess "stack" ["build", "--dry-run", "--system-ghc"] ""
```
borsboom, line 77: Might `--system-ghc` cause problems when running `release.hs` locally rather than from CI?

etc/scripts/release.hs
```diff
@@ -206,7 +212,11 @@ rules global@Global{..} args = do
                 ,[" --haddock" | gTestHaddocks]
                 ,[" stack"]]
             let cmd' c = cmd (AddPath [tmpDir] []) stackProgName (stackArgs global) c
-            () <- cmd' "test" gBuildArgs integrationTestFlagArgs "--pedantic --exec stack-integration-test stack"
+            () <- cmd
```
borsboom, line 215: It looks like this will skip the unit tests, and only run the integration tests.  Is that intended?

etc/scripts/release.hs
```diff
@@ -264,8 +274,10 @@ rules global@Global{..} args = do
                                 ,"/t", "http://timestamp.verisign.com/scripts/timestamp.dll"
                                 ,out])
                             (removeFile out)
+            -- FIXME testing if the stripping was causing the segfault
             Linux ->
-                cmd "strip -p --strip-unneeded --remove-section=.comment -o"
+                -- cmd "strip -p --strip-unneeded --remove-section=.comment -o"
+                cmd "cp" $ reverse
```
borsboom, line 280: Is it possible the reason this causes problems is that you're using the host's `strip` rather than Alpine's `strip`?

etc/scripts/release.hs
```diff
@@ -546,9 +558,13 @@ certificateNameOptName = "certificate-name"
...
 uploadOnlyOptName = "upload-only"
 
+-- | @--stack-yaml@ command-line option name.
+stackYamlOptName :: String
+stackYamlOptName = "stack-yaml"
+
 -- | Arguments to pass to all 'stack' invocations.
 stackArgs :: Global -> [String]
-stackArgs Global{..} = ["--install-ghc", "--arch=" ++ display gArch, "--interleaved-output"]
+stackArgs Global{..} = ["--install-ghc", "--system-ghc", "--stack-yaml=" ++ gStackYaml, "--arch=" ++ display gArch, "--interleaved-output"]
```
borsboom, line 567: Similar to another comment, I'm concerned this will cause problems when running locally.

etc/scripts/release.hs
borsboom: I just did a test, and when I used Alpine's `strip` it didn't lead to the segfault.  However, the `stack` executable was unchanged after `strip` (same size and SHA sum), so it doesn't seem that it actually did anything (Ubuntu's `strip` also didn't actually shrink the file, although it did modify it and cause the segfault).

This executable is quite a bit bigger than that generated by the Nix process (59MiB vs. 27 MiB).  Not sure it really matters.

snoyberg: I'd like something smaller, but I'm not willing to complicate the build process for it. I tried `split-sections`, but it didn't help either. I'll document that this is expected behavior for now.

Question though: do we need the stripping on OS X? `stack copy` should automatically perform stripping, which may be why using Alpine's `strip` causes no improvements.

etc/scripts/release.hs
snoyberg: Semi-intentional. I didn't want to run the tests in this step, but I thought they were run in the previous step. I'll modify the code so it runs the tests above.

stack-alpine.yaml
snoyberg: Originally I thought this would work out really nicely, but after fighting with the CI system all day, I still ended up needing to pass in `--system-ghc` and other such things at unexpected places. I like your suggestion here: I'll remove the extra file.

.azure/azure-linux-template.yml
snoyberg: Originally I just meant it as debug output, but now I think we should add a check. No `grep` is necessary, we can pay attention to the exit code of `ldd` I think.

etc/scripts/release.hs
snoyberg: It shouldn't cause a problem, but with the other changes you recommended it's trivial to get rid of it, so I will.

snoyberg:
New version ready for review. I'd like to rebase before merge; let me know if you'd rather I rebase before your next review.

etc/scripts/release.hs
borsboom: I'm not sure off the top of my head, should I try it out and see?

stack.yaml
```diff
@@ -5,7 +5,9 @@ packages:
 
 docker:
   enable: false
-  repo: fpco/stack-build:lts-14.27
+
+  #image: fpco/alpine-haskell-stack:8.6.5
+  image: fpco/alpine-haskell-stack@sha256:49e7e15f3b1d3f882ba5bb701463b1d508fbf40e5aafce6ea31acd210da570ba
```
borsboom, line 10: Is the intent that `stack --docker` _always_ builds using Alpine?  If so, would that mean we'd always have to remember to pass in extra flags along with `--docker` (e.g. `--flag=stack:static --system-ghc`)?

An alternative would be to leave the image here as it was, and have the release process pass in `--docker-image=fpco/alpine-haskell-stack@sha256:49e7e15f3b1d3f882ba5bb701463b1d508fbf40e5aafce6ea31acd210da570ba`.

etc/scripts/release.hs
borsboom: I just did a test on macOS, and the executable copied by `stack install` is 41 MiB.  After `strip`ing it, it goes down to 32 MiB bytes.  So it does seem to make a significant difference.

snoyberg: Huh... that's surprising. OK, I'll add back the stripping on macOS.

stack.yaml
snoyberg: My concern with that approach is that, then, we have to specify the Alpine image ID in multiple places. I'd like to leave it this way for now. I doubt people are building Stack with the Ubuntu Docker images anyway. I'm investigating a significantly different strategy for building with Github Actions, and may be able to simplify things there and not incur the update-in-two-places problem.

snoyberg:
Thanks for the review, looks like we've hit all comments. I'm going to rebase, and then merge once CI passes.

Issue #5274, opened 2020-04-22
Feature Request: Expose sha256 in ls dependencies json


aherrmann:
**Feature Request**

I would like the output of `stack ls dependencies json` to contain `sha256` and size of packages. AFAIK this information is already available in the snapshots, but currently not exposed through `stack`'s CLI.

**Motivation**

The upcoming release of `stack` contains the new [`stack ls dependencies json`](https://github.com/commercialhaskell/stack/pull/4424) feature which returns the package dependency graph and some additional package metadata for a given project (package names, versions, locations, dependencies) in a JSON format. This is very useful for build tools that would like to call out to `stack` to determine how to fetch sources of Stackage dependencies. The motivating use-case is Bazel with `rules_haskell`, see https://github.com/tweag/rules_haskell/issues/1310. Though this may also be beneficial for other build tools.

sschuberth:
Looks like https://github.com/commercialhaskell/stack/releases/tag/v2.5.1 added this, or?

aherrmann:
@sschuberth Thanks for the ping. The feature is included for `archive` dependencies.

From #5280 
> Note, this does not change the output in case of Hackage or repository dependencies. To my understanding, the [`PackageLocationImmutable`](https://hackage.haskell.org/package/pantry-0.4.0.1/docs/Pantry.html#t:PackageLocationImmutable) data type does not provide this information for Hackage or repository dependencies. Instead, it provides the `BlobKey`/`TreeKey` size and hashes which are specific to pantry and not directly applicable to check the hash of a fetched tarball, as explained [here](https://www.fpcomplete.com/blog/2018/07/pantry-part-2-trees-keys).
> 
> Please let me know if there is a good way to obtain regular `sha256`/`size` values for these dependencies within `stack`, then I'd be happy to extend this change accordingly. Alternatively, users could use [`all-cabal-hashes`](https://github.com/commercialhaskell/all-cabal-hashes/) for Hackage dependencies.

In `rules_haskell` we're currently using the `all-cabal-hashes` workaround. However, it incurs additional complexity and download time. So, if stack could emit this information in `ls dependencies json`, then that would be preferable.

PR #5280, opened 2020-04-27
sha256 and size in ls dependencies json for archive dependencies

aherrmann:
This adds `sha256` and `size` fields to the output of `ls dependencies json` in case of archive dependencies ([`PLIArchive`](https://hackage.haskell.org/package/pantry-0.4.0.1/docs/Pantry.html#v:PLIArchive)). See https://github.com/commercialhaskell/stack/issues/5274 for motivation.

Note, this does not change the output in case of Hackage or repository dependencies. To my understanding, the [`PackageLocationImmutable`](https://hackage.haskell.org/package/pantry-0.4.0.1/docs/Pantry.html#t:PackageLocationImmutable) data type does not provide this information for Hackage or repository dependencies. Instead, it provides the `BlobKey`/`TreeKey` size and hashes which are specific to pantry and not directly applicable to check the hash of a fetched tarball, as explained [here](https://www.fpcomplete.com/blog/2018/07/pantry-part-2-trees-keys).

Please let me know if there is a good way to obtain regular `sha256`/`size` values for these dependencies within `stack`, then I'd be happy to extend this change accordingly. Alternatively, users could use [`all-cabal-hashes`](https://github.com/commercialhaskell/all-cabal-hashes/) for Hackage dependencies.

  *Let me know if this is required, if this change ends up still being included in v2.3.0.1 then it seems unnecessary to add an additional changelog entry about this feature, since `ls dependencies json` is a new feature itself.*
  *Likewise, this is only a small addition to `ls dependencies json`, so additional documentation seems not necessary.*

The integration test suite contains `4101-dependency-tree` which is affected by this change. I have adapted the test to take the additional `sha256` and `size` fields into account.


aherrmann:
> Can you add a ChangeLog entry for this?

Sure, done. I've also rebased on latest master.

Issue #3675, opened 2017-12-19
Add shellcheck matrix on Travis for the etc/scripts/


decentral1se:
Know it? Check out: https://github.com/koalaman/shellcheck. It's pretty much the best helper to write and maintain shell scripts that don't suck. We could add to the Travis CI build to help catch mistakes for all us bash noobs. I can do it if people think we should. I assume we should, at least `get-stack.sh`, which is end-user facing.

mgsloan:
:+1: Sounds good to me!

MaxGabriel:
Fwiw, these are all the issues shellcheck currently finds:

(if you run this yourself shellcheck will do nice colorizing)

```
bash-3.2$ find . -type f -name "*.sh" -exec shellcheck {} \;

In ./.travis-setup.sh line 21:
if [ `uname` = "Darwin" ]; then
     ^-- SC2046: Quote this to prevent word splitting.
     ^-- SC2006: Use $(..) instead of legacy `..`.


In ./.travis-setup.sh line 33:
mkdir -p $HOME/.cabal
         ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./.travis-setup.sh line 34:
cat > $HOME/.cabal/config <<EOF
      ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./etc/scripts/mirror-ghc-bindists-to-github.sh line 62:
    echo "    $alias:" >>stack-setup.yaml
    ^-- SC2129: Consider using { cmd1; cmd2; } >> file instead of individual redirects.


In ./etc/scripts/mirror-ghc-bindists-to-github.sh line 66:
    echo "            sha1: $(sha1sum $destfn |cut -d' ' -f1)" >>stack-setup.yaml
                                      ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./etc/scripts/upload-haddocks.sh line 3:
STACKVER=$(cat stack.cabal|grep '^version:'|head -1|awk '{print $2}')
               ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.


In ./etc/scripts/upload-haddocks.sh line 6:
rm -rf _release/$STACKDOCDIR
                ^-- SC2086: Double quote to prevent globbing and word splitting.


In ./etc/scripts/upload-haddocks.sh line 8:
cp -r $(stack path --local-doc-root)/stack-$STACKVER _release/$STACKDOCDIR
[...]

PR #5285, opened 2020-04-30
Fix stack installation in alpine

coord-e:
Small PR to fix installation using installation script in alpine linux. `do_alpine_install` is commented out in 87ff1220c579ec03cff8874d6ee59e87b1b0aec0 but it still occurred in the installation. That caused an error as shown below.

```
/workdir # ./etc/scripts/get-stack.sh
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
(1/2) Installing libbz2 (1.0.6-r7)
(2/2) Installing perl (5.28.2-r1)
Executing busybox-1.30.1-r2.trigger
OK: 44 MiB in 16 packages
Detected Linux distribution: alpine
./etc/scripts/get-stack.sh: line 465: do_alpine_install: not found
```

This PR fixes this by simply skipping the branch for alpine in `do_distro` and make the script try "generic" installation. I tested this change by running `./etc/scripts/get-stack.sh` in a docker container of `alpine:3.11` image and confirmed that the installation finishes successfully with a warning "there is no guarantee that 'stack' will work at all!..." as expected.

  * I'm not sure whether this should be recorded in changelog
  * I this there is no need for this

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -313,20 +313,20 @@ do_osx_install() {
 #     die "Sorry, there is currently no 32-bit FreeBSD binary available."
 #   fi
 # }
-#
-# # Alpine distro install
-# do_alpine_install() {
-#   install_dependencies() {
-#     apk_install_pkgs gmp libgcc xz make
-#   }
-#   install_dependencies
-#   if is_64_bit ; then
-#     print_bindist_notice
-#     install_64bit_standard_binary
-#   else
-#     die "Sorry, there is currently no 32-bit Alpine Linux binary available."
-#   fi
-# }
+
+# Alpine distro install
+do_alpine_install() {
+  install_dependencies() {
+    apk_install_pkgs gmp libgcc xz make
+  }
+  install_dependencies
+  if is_x86_64 ; then
+    print_bindist_notice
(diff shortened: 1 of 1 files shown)


snoyberg:
I've asked @borsboom to review. One comment though: I think this should target the `stable` branch instead.

borsboom:
Ack, I guess I was a bit overzealous when I was commenting out code for no-longer-supported platforms.  Thanks!

I think a better fix would be to un-comment the `do_alpine_install` function, rather than fallback to `do_sloppy_install`.

And, as @snoyberg suggested, base this on `stable` rather than `master`.

coord-e:
@snoyberg @borsboom Thanks for your reviews! I've fixed the base branch to `stable`, and changed my fix to un-comment `do_alpine_install` with small modifications.

I'm willing to write a pull-req that enables `shellcheck` on the CI workflow to prevent this kind of issue. I found #3675 but it has suddenly closed, is it OK if I try this?

borsboom:
I'd be in favour of running `shellcheck` from CI.  I should have done so myself before I committed my changes.

Issue #5213, opened 2020-03-05
first `stack repl foo:test:bar` needs  `stack build --test --no-run-tests`


jneira:
### General summary/comments (optional)

If you starts a repl session in a test component that has dependencies, you need previously build test dependencies with `stack build --test --no-run-tests`

This causes a bug in `hie-bios`, that uses `stack repl` to get the ghc flags needed to start a ghc session in a ide (like `haskell-ide-engine` or `ghcide`): https://github.com/haskell/haskell-ide-engine/issues/1564

I guess `stack repl project:bench:benchname` will have the same behaviour.

### Steps to reproduce

In a simple stack project with a test component that has dependencies (`hspec`for example) i can reproduce the next cli session:

```
$ rm -rf $(stack path --snapshot-install-root) # to ensure test dependencies are not already cached

$ rm -rf .stack-work/

$ stack repl project:test:test-name
....
Warning: Didn't find expected autogen file:
         D:\dev\ws\haskell\project\.stack-work\dist\e626a42b\build\project\autogen\cabal_macros.h
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
<command line>: cannot satisfy -package hspec-2.7.1
    (use -v for more information)

$ stack build --test --no-run-tests
....
Completed 29 action(s)

$ stack repl project:test:test-name
.....
Configuring GHCi with the following packages: stack-test
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
[1 of 1] Compiling Main             ( D:\dev\ws\haskell\project\test\MyLibTest.hs, interpreted )
Ok, one module loaded.
Loaded GHCi configuration from D:\\dev\\....
*Main>
```

This behaviour is not symmetric with `stack build project:test:testname` cause it builds test dependencies although you dont provide explicitly `--tests` and imho is is a bit illogical that user wants to load a test component explicitly in the target but you have to provide `--test --no-run-tests` to start the repl session.

After a first `stack build --test` or `stack build project:test:testname`, with the dependencies in the snapshot, `stack repl project:test:testname` works until you delete the snapshot dir again.

### Expected

`stack repl project:test:testname` should always build test dependencies if needed (and same for bench components)

### Stack version

```
$ stack --version
Version 2.1.3, Git revision 0fa51b9925decd937e4a993ad90cb686f88fa282 (7739 commits) x86_64 hpack-0.31.2
```

### Method of installation

[...]

PR #5306, opened 2020-05-28
Fix repl for test components

qrilka:

Tested using https://github.com/fduxiao/mal taken from https://github.com/haskell/haskell-ide-engine/issues/1564

It's not quite clear why `stack-head repl mal:exe:mal-exe` doesn't work. I have a suspicion that we could need component-based builds but I didn't investigate it deeper.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,9 @@ Other enhancements:
 Bug fixes:
 
 * When using the `STACK_YAML` env var with Docker, make the path absolute.
+* Fix the problem of `stack repl foo:test:bar` failing without a project
+  build before that. See
+  [#5213](https://github.com/commercialhaskell/stack/issues/5213)
 
 * Fix `stack sdist` introducing unneded sublibrary syntax when using
   pvp-bounds. See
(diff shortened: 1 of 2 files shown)


qrilka:
Fixes the problem with test components from #5213

qrilka:
The main concern that I have is the partiality of this fix,  I'll add a changlog entry and merge but also will take another turn next week and try to understand the reasons this doesn't work for executable targets.
Thank you

qrilka:
I've added a changelog but and as we've discussed fixing the problem for executables doesn't look easy.
@snoyberg I have a question still - I targeted this PR to `stable` - will we have a minor release which it could go into?

snoyberg:
We definitely can make a minor release, but there isn't one currently planned

qrilka:
@psibi maybe you know what could be producing `The AWS Access Key Id you provided does not exist in our records.` in the failed CI jobs for this PR?

psibi:
@qrilka Sorry, no idea. Do you know how long this has been failing ?

qrilka:
It's the first time I see something like this

qrilka:
Restarting the pipelines doesn't change anything, @borsboom do you know if we did any changes to AWS config lately? It looks like cache-s3 can't access the corresponding bucket...

snoyberg:
I've been meaning to move us over to Github Actions instead of Azure CI for a few months now, this looks like a good kick in the pants to actually do it.

qrilka:
@snoyberg it looks like the problem is only about cache-s3 bucket and other steps look to be fine, do you think we could merge it?

snoyberg:
Sounds good

PR #5317, opened 2020-06-15
Make file watching take targets into account

### LINKED ISSUE #5310: File watch watches unrelated libs
I have for example a stack.yaml with two packages in, neither of which reference each other. Let's call them package0 and package1

`stack build package0 --file-watch`

this results in watching all the files of package0 (good) but also all the lib files of package1 (seems bad).

I can't figure out a reason to watch the .hs files and the .md files of the other package. my logic for watching the cabal file also seems a bit weak.

Is this intended behaviour? Am I missing some reasoning where package1 could affect package0 if package0's dependency graph doesn't mention it?

and just in case it's just me i'm on ubuntu 18 under wsl (windows 10 1909) with stack 2.1.3
```
❯ stack --version
Version 2.1.3, Git revision 636e3a759d51127df2b62f90772def126cdf6d1f (7735 commits) x86_64 hpack-0.31.2
```

qrilka:
Resolves #5310

Tested manually

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,10 @@ Major changes:
 
 Behavior changes:
 
+* File watching now takes into account specified targets, old behavior could
+  be restored using the new flag `--watch-all`
+  [#5310](https://github.com/commercialhaskell/stack/issues/5310)
+
 Other enhancements:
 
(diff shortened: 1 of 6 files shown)


dten:
Awesome thanks 😁

PR #5327, opened 2020-06-21
Stable move to actions

snoyberg:
Moves unit tests (aka not the nightly tests) from Azure CI over to Github Actions.


snoyberg:
@psibi would you mind reviewing this? The integration tests (nightly) isn't yet converted, so it's still failing, but I figured it would be good to do the first step of conversion first.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Clone project
+        uses: actions/checkout@v2
+      - name: hlint
+        run: |
+          set -ex
+          ./etc/scripts/get-hlint.sh
+          export PATH="$(pwd)"/hlint:$PATH
+          hlint src/
+          hlint src/ --cpp-define=WINDOWS=1
+          hlint test/ --cpp-simple
```
psibi, line 21: I guess we could use the one available in marketplace: https://github.com/marketplace/actions/haskell-linter

But it's unofficial (not from github) and we have a better control here. From looking at the repo, I wasn't able to find which hlint version it was using.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+          extra-suffix: ""
+          stack-args: ""
+    steps:
+      - name: Clone project
+        uses: actions/checkout@v2
+      - name: Cache dependencies
+        uses: actions/cache@v1
+        with:
+          path: ~/.stack
+          key: ${{ runner.os }}-${{ hashFiles('${{ matrix.stack-yaml }}') }}-${{ matrix.extra-suffix }}
```
psibi, line 75: Do you think `package.yaml` should be part of the cache hash ?

.github/workflows/unit-tests.yml
psibi: I guess adding hlint to the official one would add value: https://github.com/github/super-linter

snoyberg: No objection to doing that in the future, but I don't want to block this PR on it. I also don't want to rely on something unofficial, especially when the solution we have has worked on Azure CI for years.

.github/workflows/unit-tests.yml
snoyberg: I could go both ways, but I lean towards no. Basically: if package.yaml changes, but stack.yaml doesn't, it means that packages were added or removed, but the package set itself is unchanged. That means that the cache shouldn't be invalidated.

Issue #5125, opened 2019-12-19
Cache being busted after restoring directory


aschmois:
### Stack version
Used latest 2.1.3 and latest HEAD

### Method of installation
Download linux stack then run `upgrade --git`

### Problem/Question
We are using semaphore CI to compile our project and have not been able to get our project compilation cached. We run stack inside a docker container (for all intents and purposes, it's a regular stack build --test just inside docker).

We cache the stack-root directory and restore it without issues, meaning ghc does not re-download/compile and packages are not re-compiled.

We also cache .stack-work directory and restore it but our build command always rebuilds as if there was no cache. The relevant bits of information from the log below seem to act as if there is a cache, but it ignores it completely.

It unregisters a single file change (to test) but all 818 modules are recompiled:
```
xxx-22: unregistering (local file changes: library/Executable/xxx.hs)
--
  | xxx> configure (lib + exe + test)
  | Configuring xxx-22...
  | xxx> build (lib + exe + test)
  | Preprocessing library for xxx-22..
  | Building library for xxx-22..
  | [  2 of 818] Compiling Environment.xxx
```

The shas seen here are identical to the shas used in the previously cached build.
```
xxx> copy/register
--
  | Installing library in /work/.stack-work/install/x86_64-linux/01c3488eaf4ca467d4cf5a081d8270e39a6956b896e87ed70692ef5026aa5899/8.8.1/lib/x86_64-linux-ghc-8.8.1/xxx-22-B7rEdpkV2M7EtGsQY02An4
  | Installing executable xxx in /work/.stack-work/install/x86_64-linux/01c3488eaf4ca467d4cf5a081d8270e39a6956b896e87ed70692ef5026aa5899/8.8.1/bin
  | Registering library for xxx-22..
```

My assumption is that the cache restore is messing with some file inside stack-work that's causing the cache to bust. My question is which file(s) is that and is there anyway to force stack to not do that?

simonmichael:
Could be https://stackoverflow.com/a/61178945/84401 ?

aschmois:
The current workaround for this is to run this before doing stack work in your ci steps:

```sh
find ./your ./src ./directories -name "*.hs" -print0 | xargs -0 -I '{}' \
  touch \
  --no-create \
  --reference '{}' \
  --date "1000 days ago" \
  '{}'
```

[...]

Issue #838, opened 2015-08-23
Dirtiness checking triggers unnecessary reinstalls WAS: stack build unnecessarily reinstalls packages from github


ljli:
When the stack.yaml includes a dependency from github and doing `stack build` twice, the second command isn't a no-op despite the first command being successful. If doing `stack build` a third time it is, however. During the second `stack build` run, stack unregisters the github package saying "local file changes" and reinstalls them again. Diffing the package folder in .stack-work/downloaded before and after the second command reveals that only one file changed: `downloaded/<package>/.stack-work/dist/x86_64-osx/Cabal-1.22.4.0/stack-build-cache`

I can reproduce this on OS X with the following .cabal and stack.yaml content:
foobar.cabal:
````
name:                foobar
version:             0.1.0.0
cabal-version:       >=1.10
build-type:          Simple

library
  build-depends:
    base,
    HStringTemplate
  default-language: Haskell2010
````
stack.yaml:
````
packages:
- '.'
- location:
    git: https://github.com/factisresearch/HStringTemplate.git
    commit: 033328ce078426733020a2c0d943332f37ffee05
  extra-dep: true

extra-deps: []

resolver: lts-3.0
````

**Comment by borsboom** [comment]
This isn't related to the package being a dependency from github.  I made a local clone of HStringTemplate and changed the stack.yaml to use
```
packages:
- '.'
- 'HStringTemplate'
```
and the same thing happened.  The reason it's happening is that, because of the way we detect TemplateHaskell dependent files, it sees them as new files in the second build and so runs the build again.  Fixing that would introduce a race condition, and we'd rather have occasional redundant builds than a race condition.

borsboom:
(it's not obvious where the TH dependency is introduced, but `Text/StringTemplate/Instances.hs` adds a dependence on `autogen/cabal_macros.h` -- you can discover that by examining the generated `.dump-hi` files)

snoyberg:
Perhaps that file should be special cased and not trigger dirtiness. The
only way it should ever change is when we're reconfiguring.

loisch:
[...]

PR #5351, opened 2020-07-27
Remove ModTime check during build (#5125)

aschmois:
As the issue (#5125) suggests, modified time makes using a stack cache in CI very difficult (if at all possible). This PR suggests making a change in how the build process marks a file as dirty.

Summary of changes:
- When checking the build cache, always create a digest and use that to check against the cache if it exists.
  - Note: let's discuss here about also checking for size.
- When adding unlisted files to the build cache don't check for pre build time. 
  - This was a strange call to make to begin with and looking into the commit history it looks like it was a patch for an issue about Template Haskell (#838). Removing this check does not break the fix. I ran the two test cases mentioned in the issue with latest stack version and with the modified one from this PR and it did not cause a regression.

Questions and comments:
- I wasn't able to figure out the styling, hindent seems to change too many lines. I tried to keep it as stable as possible.
- After this change mod time becomes obsolete (no uses), should we remove it? I wasn't sure how it would affect the json instance supplied to `FileCacheInfo`.
- If checking against size is not required removing size would also cleanup the process a lot more (no necessary composed tuples).
- In terms of performance I built our team's current library with about 1500 modules and there is negligible time difference between the runs. Creating the digest is pretty fast! I do however realize that I run this on a beasty machine so it would be nice to get some insight here.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -39,6 +39,9 @@ Bug fixes:
 * Fix `stack sdist` introducing unneded sublibrary syntax when using
   pvp-bounds. See
   [#5289](https://github.com/commercialhaskell/stack/issues/5289)
+* Fix modified time busting caches by always calculating sha256 digest
+  during the build process.
+  [#5125](https://github.com/commercialhaskell/stack/issues/5125)
 
 ## v2.3.1
 
@@ -116,7 +119,7 @@ Other enhancements:
   prefixes each build log output line with a timestamp.
 
 * Show warning about `local-programs-path` with spaces on windows
-  when running scripts. See 
+  when running scripts. See
   [#5013](https://github.com/commercialhaskell/stack/pull/5013)
 
 * Add `ls dependencies json` which will print dependencies as JSON.
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -59,7 +59,6 @@ import qualified Distribution.Text as C
 import           Distribution.Types.PackageName (mkPackageName)
 import           Distribution.Types.UnqualComponentName (mkUnqualComponentName)
 import           Distribution.Version (mkVersion)
-import           Foreign.C.Types (CTime)
 import           Path
 import           Path.CheckInstall
 import           Path.Extra (toFilePathNoTrailingSep, rejectMissingFile)
@@ -91,7 +90,6 @@ import           System.FileLock (withTryFileLock, SharedExclusive (Exclusive), 
 import qualified System.FilePath as FP
 import           System.IO.Error (isDoesNotExistError)
 import           System.PosixCompat.Files (createLink, modificationTime, getFileStatus)
-import           System.PosixCompat.Time (epochTime)
 import           RIO.PrettyPrint
 import           RIO.Process
 import           Pantry.Internal.Companion
@@ -1601,11 +1599,10 @@ singleBuild ac@ActionContext {..} ee@ExecuteEnv {..} task@Task {..} installedMap
 
         -- FIXME: only output these if they're in the build plan.
 
-        preBuildTime <- liftIO epochTime
         let postBuildCheck _succeeded = do
                 mlocalWarnings <- case taskType of
                     TTLocalMutable lp -> do
-                        warnings <- checkForUnlistedFiles taskType preBuildTime pkgDir
+                        warnings <- checkForUnlistedFiles taskType pkgDir
                         -- TODO: Perhaps only emit these warnings for non extra-dep?
                         return (Just (lpCabalFile lp, warnings))
                     _ -> return Nothing
@@ -1829,12 +1826,11 @@ checkExeStatus platform distDir name = do
         file = T.unpack name
 
 -- | Check if any unlisted files have been found, and add them to the build cache.
-checkForUnlistedFiles :: HasEnvConfig env => TaskType -> CTime -> Path Abs Dir -> RIO env [PackageWarning]
-checkForUnlistedFiles (TTLocalMutable lp) preBuildTime pkgDir = do
+checkForUnlistedFiles :: HasEnvConfig env => TaskType -> Path Abs Dir -> RIO env [PackageWarning]
+checkForUnlistedFiles (TTLocalMutable lp) pkgDir = do
     caches <- runMemoizedWith $ lpNewBuildCaches lp
     (addBuildCache,warnings) <-
         addUnlistedToBuildCache
-            preBuildTime
             (lpPackage lp)
             (lpCabalFile lp)
             (lpComponents lp)
@@ -1844,7 +1840,7 @@ checkForUnlistedFiles (TTLocalMutable lp) preBuildTime pkgDir = do
         writeBuildCache pkgDir component $
             Map.unions (cache : newToCache)
     return warnings
-checkForUnlistedFiles TTRemotePackage{} _ _ = return []
+checkForUnlistedFiles TTRemotePackage{} _ = return []
 
 -- | Implements running a package's tests. Also handles producing
 -- coverage reports if coverage is enabled.
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -28,7 +28,6 @@ import              Data.List
 import qualified    Data.Map as Map
 import qualified    Data.Map.Strict as M
 import qualified    Data.Set as Set
-import              Foreign.C.Types (CTime)
 import              Stack.Build.Cache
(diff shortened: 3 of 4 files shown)


aschmois:
I don't think the windows integration test failure is correct, looks like it failed in the dependencies section.

snoyberg:
I would prefer to _augment_ rather than remove the modtime check. I would want the heuristics to be something like:

* If the file modification is older, assume that the file is the same
* If the file modification is newer, _then_ check the digest and make a determination

That should bypass most of the slowdown potentially instituted by making the digest checks only occur (1) on creating the cache and (2) when the file mod times are incorrect

tfausak:
I worked with @aschmois on this. I expected there to be a slowdown from replacing the modification time check with a digest comparison. I was surprised to find that if there was a difference, it was lost in the noise (at least for our project). Is there a large project we could test the relative performance with? 

Conceptually I like the idea of only relying on the file's contents rather than its metadata. But also of course a nice concept isn't any good if it's too slow.

snoyberg:
I would test on a monorepo like yesodweb/yesod. But even if testing shows no major performance impact, I'd still be worried that some users, based on type of hard drive, file system, cache settings, etc, would end up having a negative impact.

aschmois:
I can try and test on different configurations and post the findings, I have a few different devices I can try to build on.

aschmois:
With the help of @gera-cameron I ran the tests below, please let me know if something looks off about testing methods.

# AWS EC2 Testing

Ran on ec2 instance types `m5.large`. Specifically avoiding the burstable instances to have stable build times. One test run on spinning disks and another on ssd.

> Up to 3.1 GHz Intel Xeon® Platinum 8175M processors with new Intel Advanced Vector Extension (AVX-512) instruction set.

https://aws.amazon.com/ec2/instance-types/

### Setup:

```sh
$ curl -sSL https://get.haskellstack.org/ | sh # install 2.3.1
$ git clone --recurse-submodules http://github.com/yesodweb/yesod
$ cd yesod && stack test && cd - # download ghc and build dependencies
$ git clone http://github.com/aschmois/stack # clone PR
$ cd stack && stack test --copy-bins --local-bin-path bin --ghc-options '-O2' && cd - # build stack and copy binary into ./bin/
$ cp stack/bin/stack yesod/stackx && chmod u+x yesod/stackx # copy modified stack binary as stackx and make it executable
$ cd stack && git checkout b5d30906ebee25df1f2532255e245d329083b623 && stack test --copy-bins --local-bin-path bin --ghc-options '-O2' && cd - # build unmodified stack
$ cp stack/bin/stack yesod/stackz && chmod u+x yesod/stackz # copy unmodified stack binary as stackx and make it executable
```

## Test 1

Each stack binary test started from cold boot then run sequentially. We don't expect numbers to change wildly here since they all start from a clean install

Yesod Build Stack unmodified:
```sh
$ stack clean --full && TIMEFORMAT='%6R'; time ./stackz build
```

Yesod Build Stack modified:
```sh
$ stack clean --full && TIMEFORMAT='%6R'; time ./stackx build
```

### Results

All results are in seconds

| OS           | EC2      | vCPU | RAM | EBS                 | version    | x1 (no io cache) | x2      | x3      | avg     |
|--------------|----------|------|-----|---------------------|------------|------------------|---------|---------|---------|
| Ubuntu 20.04 | m5.large | 2    | 8   | standard (magnetic) | unmodified | *130.105*        | 107.735 | 108.894 | 115.578 |
| Ubuntu 20.04 | m5.large | 2    | 8   | standard (magnetic) | modified   | *128.374*        | 109.875 | 105.994 | 114.738 |
| Ubuntu 20.04 | m5.large | 2    | 8   | gp2 (ssd) 100 iops  | unmodified | *109.004*        | 103.735 | 104.614 | 105.785 |
| Ubuntu 20.04 | m5.large | 2    | 8   | gp2 (ssd) 100 iops  | modified   | *108.405*        | 104.605 | 104.024 | 105.678 |

### Summary 

All of these numbers are within margin of error of each other and we can assume that the same process is happening on each build. This is because digest is being calculated every time since the project was cleaned.

## Test 2

This is where we expect things to show differences since no stack clean is done after the first one.

Yesod Build Stack unmodified:
```sh
$ stack clean --full && ./stackz build
$ export TIMEFORMAT='%6R'
$ time ./stackz build
$ time ./stackz build
$ time ./stackz build
```

Yesod Build Stack modified:
```sh
$ stack clean --full && ./stackx build
$ export TIMEFORMAT='%6R'
$ time ./stackx build
$ time ./stackx build
$ time ./stackx build
```

### Results

All results are in seconds

| OS           | EC2      | vCPU | RAM | EBS                 | version    | x1      | x2      | x3      | avg     |
|--------------|----------|------|-----|---------------------|------------|---------|---------|---------|---------|
| Ubuntu 20.04 | m5.large | 2    | 8   | standard (magnetic) | unmodified | 0.614   | 0.614   | 0.604   | 0.611   |
| Ubuntu 20.04 | m5.large | 2    | 8   | standard (magnetic) | modified   | 0.626   | 0.635   | 0.634   | 0.632   |
| Ubuntu 20.04 | m5.large | 2    | 8   | gp2 (ssd) 100 iops  | unmodified | 0.594   | 0.594   | 0.594   | 0.594   |
| Ubuntu 20.04 | m5.large | 2    | 8   | gp2 (ssd) 100 iops  | modified   | 0.606   | 0.604   | 0.604   | 0.605   |

### Summary

We notice a ~21ms difference in magnetic drives calculating digests on a very restricted machine and ~11ms difference in ssd drives. 

Based off these results I think we should only use digests since mod times can bring more bugs (such as the CI one I ran into) and performance does not seem to be majorly affected.

snoyberg:
Based on the massive communication misconnect here, I think there's a fundamental misunderstanding here. I think I understand it, but before merging, let's confirm. I said above:

> I would prefer to augment rather than remove the modtime check.

followed by:

> But even if testing shows no major performance impact, I'd still be worried that some users, based on type of hard drive, file system, cache settings, etc, would end up having a negative impact.

Given that despite these comments, you've moved ahead with a whole bunch of performance testing, I think you're trying to imply:

> You're wrong Michael, augmenting is not an option, so let me try to convince you with overwhelming evidence that the performance impact is minimal.

Am I reading this conversation correctly?

tfausak:
We don't think you're wrong, and we're not trying to overwhelm you with benchmarks. 

Like you, we expected the performance of this digest-based approach to be worse than the current approach based on file modification times. After working on this patch we were pleasantly surprised to find that it made effectively no impact whatsoever for our use case. Since your primary concern appeared to be performance, we tried to run a benchmark where we stacked the deck against ourselves: Large project, shared hosting, under powered machine, and spinning disks. Even with those unfavorable conditions we only saw about a 20 ms penalty, which is about 3% of the build time. 

Augmenting is very much an option. However we thought that the approach presented in this PR is both conceptually simpler and results in code that's easier to read, so we figured it was worth a shot. If you're saying that this approach is dead in the water even if the performance impact is minimal, fine.

snoyberg:
Sorry, I didn't mean to imply you're overwhelming me, I just meant that the evidence is overwhelmingly in favor of what you're saying.

I am slightly concerned still, but I'm willing to take this as-is and see if anything complains about their hard drives thrashing later. Thank you!

snoyberg:
Sorry, just one more request: can you update the ChangeLog?

aschmois:
I'd like to try to smooth out the situation, in no way do I want to attack anyone nor do any harm to this code base. I apologise if I came out like that, I can be a little cold sometimes when discussing code. I thought making fancy benchmarks would make a better case for the code written not worse.

With that out of the way if we are moving forward with this I'd like to see if removing the mod time and size (if we don't need it as part of the digest check) is also something you think can be done I'd like to do so to cleanup the tuple composing.

An afterthought, augmenting mod time does seem to be the best option in terms of safety but I think after looking at the benchmarks being more data safe can avoid some future bugs. I've always seen horror stories around mod time checks for caching and have honestly been battling wiht this bug for over a year; not directly but it has been at the back of my mind for that long 😅 . I really want what's best for stack not any random code! Please let me know what you think.

snoyberg:
Really, honestly, nothing to apologize for. I just wanted to hone in on whether there was a correctness issue I was missing, or if there was a different reason. Taylor clarified the situation to my satisfaction in the comment above.

IIUC, we're not using the filesize or timestamp at all, so please do feel free to update the PR by removing them. I'm also not a fan of anonymous tuples in general, so either removing the multiple-data-returns or creating a custom datatype if they are needed would be great.

Thanks!

PR #5358, opened 2020-08-04
Add stack-ghc-810.yaml and Github Actions job

borsboom:
This is the version for `stable`.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -55,6 +55,10 @@ jobs:
           extra-suffix: ""
           stack-args: ""
         - os: ubuntu-latest
+          stack-yaml: stack-ghc-810.yaml
+          extra-suffix: ""
+          stack-args: ""
+        - os: ubuntu-latest
           stack-yaml: stack.yaml
           extra-suffix: "alpine"
           stack-args: "--docker --system-ghc --no-install-ghc --flag stack:static"
--- a/src/Stack/SDist.hs
+++ b/src/Stack/SDist.hs
@@ -232,7 +232,7 @@ getCabalLbs pvpBounds mrev cabalfp sourceMap = do
             [ style Url "https://github.com/commercialhaskell/stack/issues/new"
             , style Url "https://github.com/haskell/cabal/issues/new"
             ]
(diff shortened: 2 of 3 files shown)


borsboom:
@snoyberg The Github Actions checks for this are failing, but it's not obvious how the failures are related to the changes I made.  Have you seen errors like these before?

borsboom:
I also told Github Actions to re-run the jobs from the previous `stable` build, which had succeeded in the past, but it [failed in the re-run with the same error](https://github.com/commercialhaskell/stack/runs/946750601?check_suite_focus=true).  It seems this is more of a general CI problem rather than a problem with this MR.

snoyberg:
I put in a hacky workaround on master, which needs to be improved. I believe what's happening is that Git on Windows had an update which changes how symlinks are handled, and now `git status` _always_ shows changes on Windows for each symlink file. On master, I simply ignored the diffs for Windows, but that's not a great solution, since it means the `stack --version` shows a dirty tree. I think we can merge this PR as-is, and figure out a better fix later.

borsboom:
Since we rely on the integration test job to build bindists, we'll have to fix it properly because we don't want official Windows releases to say they're dirty.

snoyberg:
Agreed, though I'm kind of hoping that Git for Windows just fixes this regression soon and we don't have to do anything.

borsboom:
Do you know if there's an issue filed somewhere for the Git for Windows issue?  I'd like to at least know when it's fixed so I can continue the release process then.

snoyberg:
What I found was a reference in the [release notes](https://github.com/git-for-windows/git/releases/tag/v2.28.0-rc0.windows.1) to [issue 2653](https://github.com/git-for-windows/git/issues/2653). I'm not sure if that's authoritative, since that issue is closed.

borsboom:
Thanks.  A comment in that issue identified an easy workaround: `git config --global core.fscache false`.  I added that in https://github.com/commercialhaskell/stack/pull/5361/commits/7b4ca5c432501ffafb1c5897921f6f8294900a4c and it seems to work.

snoyberg:
Whoa! Good find, thank you!

PR #5359, opened 2020-08-05
Fix typo in Stack/Coverage.hs

DerpyCrabs:
Fixed typo in table's "border" attribute. It didn't affect table style before because its every child draws own border.


snoyberg:
Thanks!

Issue #5379, opened 2020-08-28
Not compatible with optparse-applicative 0.16


Lysxia:
### General summary/comments

optparse-applicative-0.16.0.0 added an argument to the `ShowHelpText` constructor of its `ParseError` type.

I can confirm that 0.15.1.0 fixes the build.

### Steps to reproduce

Build stack with optparse-applicative-0.16.0.0

```
src/Options/Applicative/Complicated.hs:155:17: error:
    • Couldn't match expected type ‘ParseError’
                  with actual type ‘Maybe String -> ParseError’
    • Probable cause: ‘ShowHelpText’ is applied to too few arguments
      In the first argument of ‘abortOption’, namely ‘ShowHelpText’
      In the expression: abortOption ShowHelpText
      In the expression:
        abortOption ShowHelpText
          $ long help <> help Show this help text
    |
155 |     abortOption ShowHelpText $
    |                 ^^^^^^^^^^^^
```

### Expected

(I use cabal) Cabal should pick a compatible version of optparse-applicative so the build succeeds.

### Actual

Build fails.

### Stack version

2.3.3 via Hackage

### Method of installation

cabal-install

chenrui333:
Any update on this issue?

mpilgrem:
The current `package.yaml` for Stack specifies `optparse-applicative >= 0.17.0.0`.

PR #5385, opened 2020-09-14
breaking change with optparse-applicative 0.16.0.0

BebeSparkelSparkel:
cannot compile with cabal since there is a breaking change with optparse-applicative 0.16.0.0 release

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -91,7 +91,7 @@ dependencies:
 - neat-interpolation
 - network-uri
 - open-browser
-- optparse-applicative
+- optparse-applicative >= 0.15.0.0 && < 0.16.0.0
 - pantry
 - casa-client
 - casa-types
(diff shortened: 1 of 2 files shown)


ardamose123:
Apparently, the minimum version constraint for optparse-applicative has to be removed or set to 0.14.0.0 instead of 0.15.0.0.

ardamose123:
Fixes #5379

snoyberg:
We intentionally do not maintain these kinds of bounds in cabal files, they are redundant with the Stack config files.

BebeSparkelSparkel:
@snoyberg  The problem is that I'm trying to build this on OpenBSD and it only has cabal package that I can then build Stack with, but cabal fails to build stack because of this dependency problem

snoyberg:
That sounds like a bootstrapping issue that can be solved with `stack sdist --pvp-bounds` or similar on a Linux machine, or manually adding the bounds once. It doesn't sound like it requires adding a new maintenance burden to the Stack team.

BebeSparkelSparkel:
@snoyberg Ok thanks, didn't know that's how you wanted to do it.

PR #5389, opened 2020-09-17
doc/installation: Fedora now has a stack package

juhp:
update Fedora installation instructions to mention the Fedora package rather than the obsolete copr.


juhp:
I guess I need to do the same to the stable branch?
Or do prefer to fix that first and then forward port to master?

snoyberg:
I cherry-picked onto stable.

PR #5390, opened 2020-09-19
Make on_path() more robust

Martinsos:
I was getting false negatives from the get_stack script in the case where my PATH contained `~/.local/bin`, script saying that `/home/martin/.local/bin` is not on PATH.

With this change, paths are normalized by expanding ~ so false negatives like this one should not be happening any more.

This however does make the function significantly more complicated, and it could be questionable if it is worth it. There is also problem with sed commands - they will fails if paths contain `|`. I couldn't figure out a way to go around this, since from what I read, any character can be in theory used in the filepath. Happy to hear your opinion on it!

----------

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md -> In this case none.
* [x] The documentation has been updated, if necessary. -> In this case, no.

I tested manually in interactive shell with different scenarios, and I also tested by running my install script which has the same function.

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -728,19 +728,11 @@ has_cmd() {
 
 # Check whether the given (query) path is listed in the PATH environment variable.
 on_path() {
-    # Below we normalize PATH and query regarding ~ by ensuring ~ is expanded to $HOME, avoiding
+    # Below we normalize PATH and query by ensuring ~ is expanded to $HOME, avoiding
     # false negatives in case where ~ is expanded in query but not in PATH and vice versa.
-
     # NOTE: If PATH or query have '|' somewhere in it, sed commands bellow will fail due to using | as their delimiter.
-
-    # If ~ is after : or if it is the first character in the path, replace it with expanded $HOME.
-    # For example, if $PATH is ~/martin/bin:~/martin/~tmp/bin,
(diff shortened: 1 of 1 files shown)


etc/scripts/get-stack.sh
```diff
@@ -726,9 +726,14 @@ has_cmd() {
   command -v "$1" > /dev/null 2>&1
 }
 
-# Check whether the given path is listed in the PATH environment variable
+# Check whether the given (query) path is listed in the PATH environment variable.
 on_path() {
-  echo ":$PATH:" | grep -q :"$1":
+    # Below we normalize PATH and query by ensuring ~ is expanded to $HOME, avoiding
```
hasufell, line 731: POSIX doesn't specify tilde-expansion for variables in PATH. That's implementation specific behavior. E.g. in dash shell `~/foo/bin` in PATH is treated as a relative directory.

The only real way to detect whether a directory is listed in PATH is to drop a dummy script in that directory and try to invoke it by filename. But that's gonna be problematic if we don't have write permissions. But I believe the stack script would fail for other reasons if those write permissions aren't in place.

Martinsos: I didn't know that expansion of ~ in PATH was not POSIX, thanks for sharing that!
In that case I will close the PR.

Issue #5392, opened 2020-09-21
Include auto-generated modules in GHCi script


uncle-betty:
Currently, the GHCi script generated by `stack repl` only includes modules from `hs-source-dirs` directories. In particular, it doesn't include any auto-generated modules.

This breaks `stack`'s compatibility with IDE tools such as `haskell-language-server` and `ghcide`, which take the GHCi script to be the single source of truth for an exhaustive list of all modules in a component.

I'm using `ghcide` with the Agda codebase and, as a consequence of this, I had to go back to Cabal builds, which made me sad.

I'll shortly open a pull request with a proposal on how to address this.

uncle-betty:
The pull request has been merged. Thanks!

Closing.

PR #5393, opened 2020-09-21
Also resolve auto-generated component files.

uncle-betty:
This is a proposal to address #5392.

It basically passes the Cabal version down to the place where it's needed to generate the directory paths that hold auto-generated files, `resolveComponentFiles`. These directories are then used in addition to the `hs-source-dirs` directories to find `.hs` files for a component's modules.

I'm only superficially familiar with the `stack` codebase, but it seems that at least this doesn't break any of the tests.

### DIFF
--- a/src/Stack/Package.hs
+++ b/src/Stack/Package.hs
@@ -82,6 +82,7 @@ import qualified RIO.PrettyPrint as PP (Style (Module))
 data Ctx = Ctx { ctxFile :: !(Path Abs File)
                , ctxDistDir :: !(Path Abs Dir)
                , ctxBuildConfig :: !BuildConfig
+               , ctxCabalVer :: !Version
                }
 
 instance HasPlatform Ctx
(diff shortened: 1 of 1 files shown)


uncle-betty:
"... didn't break any of the tests *when I ran them on my machine*" I should have said. :) I'll look into the test failures.

uncle-betty:
Oh, never mind. The failing tests seem to be build issues caused by `casa.fpcomplete.com`.

snoyberg (approved):
LGTM, thanks!

snoyberg:
Also: the server issues should hopefully be fixed. It's a temporary patch right now, but a better fix is coming. Also, we're hoping to get a new Stack release out the door. We already have the commits on `master` to make a server outage non-fatal, they just haven't been released yet.

uncle-betty:
Great! Thanks for the review and for the merge. Current `master` now works like a charm for me.

PR #5409, opened 2020-10-04
Automate signing and uploading stack bindists to Github release

borsboom:
Adds to Github Actions workflow, and removes from `release.hs`.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -7,6 +7,8 @@ on:
       - master
       - stable
       - rc/**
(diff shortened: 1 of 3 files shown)


snoyberg (approved):
This looks great! I think it's ready to merge, I'm just not clicking merge myself in case you need to follow a different workflow.

borsboom:
Cool.  There's not really a lot left in `etc/scripts/release.hs` at this point; we could probably move it all directly into the Github Actions without much difficulty, if we want to go in that direction.

PR #5415, opened 2020-10-15
Support Linux Sparc

tolysz:
Due to monadic nature of the platform detection (`getOSKey`), each platform we want to runs `stack` from needs to be explicitly spelled out.
This change only affects the specified platform by allowing to install additional tools and `ghc`s.


snoyberg (approved):
LGTM, thanks!

tolysz:
Awesome thanks!

PR #5431, opened 2020-11-14
add 'list' command to query package-version in snapshot

juhp:
currently not working for core ghc libs with resolver

```
$ stack list zlib
zlib-0.6.2.2
$ stack --resolver lts-13 list zlib
Selected resolver: lts-13.30
zlib-0.6.2
```

### DIFF
--- a/src/Stack/List.hs
+++ b/src/Stack/List.hs
@@ -1,0 +1,76 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+module Stack.List
+  ( listPackages
+  ) where
+
+import Stack.Prelude
+import qualified RIO.Map as Map
+import RIO.List (intercalate)
+import RIO.Process (HasProcessContext)
+
+newtype ListException
+  = CouldNotParsePackageSelectors [String]
+    deriving Typeable
+instance Exception ListException
+instance Show ListException where
+    show (CouldNotParsePackageSelectors strs) = unlines $ map ("- " ++) strs
+
+-- | Intended to work for the command line command.
+listPackages
+  :: forall env. (HasPantryConfig env, HasLogFunc env, HasProcessContext env)
+  => Maybe RawSnapshot -- ^ when looking up by name, take from this build plan
+  -> [String] -- ^ names or identifiers
+  -> RIO env ()
+listPackages mSnapshot input = do
+    let (errs1, names) = case mSnapshot of
+                   Just snapshot | null input ->
+                                     ([], Map.keys (rsPackages snapshot))
+                   _ -> partitionEithers $ map parse input
+    (errs2, locs) <- partitionEithers <$> traverse toLoc names
+    case errs1 ++ errs2 of
+      [] -> pure ()
+      errs -> throwM $ CouldNotParsePackageSelectors errs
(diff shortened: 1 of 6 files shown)


juhp:
I would appreciate feedback, comments, code review of this.

qrilka (approved):
LGTM but I guess we need to have an opinion from one of Stack's maintainers as well to merge this in. Also a changelog entry looks to be needed in that case and probably mentioning it in the docs.

qrilka:
Also style job error needs to be fixed

qrilka:
@juhp I've mentioned this PR in our discussion with @snoyberg and it looks like there are no particular objections to merge this. Could you address the minor issues about it I pointed above?

juhp:
Sure, thanks, @qrilka, I will try to get to it soon.

(Maybe I will try to check on core library handling too at the same time, though I think it's already quite useful.)

juhp:
I add a Changelog entry and short documentation. Also fixed the hlint warning, and updated package.yaml.
Please have a look, this should be ready now.

juhp:
I slightly tweaked the documentation and changelog.

src/Stack/List.hs
```diff
@@ -0,0 +1,77 @@
...
+    (errs2, locs) <- partitionEithers <$> traverse toLoc names
+    case errs1 ++ errs2 of
+      [] -> pure ()
+      errs -> throwM $ CouldNotParsePackageSelectors errs
+    mapM_ (logInfo . fromString . packageIdentifierString) locs
+  where
+    toLoc | Just snapshot <- mSnapshot = toLocSnapshot snapshot
+          | otherwise = toLocNoSnapshot
+
+    -- FIXME fallback to global or project resolver
```
qrilka, line 40: @juhp what do you mean by this FIXME? Do you want to change the way `stack list` works with no resolver specified? Do you think the way it works in this implementation is bad in some sense?

qrilka (commented):
I'm OK with merging this but would like to understand the point of FIXME in the code.
Thank you.

src/Stack/List.hs
juhp: Good catch - I was looking at this too earlier, and then forgot to remove it.
For now I feel falling back to the Hackage is probably good enough and useful , since `stack list` is not a project command.
So I think I will just drop the comment (I had left it there as a placeholder/reminder about this decision).

juhp:
Okay, I dropped the comment, thanks

qrilka:
Great, thanks

juhp:
Thanks @qrilka for reviewing and helping to get this ready

mpilgrem:
@juhp, may I clarify how this command is intended to work when no packages are specified at the command line? If I command:

~~~text
stack --resolver nightly-2022-09-23 list
~~~

it yields a long list of what looks like the packages in `nightly-2022-09-23`, save that it excludes the packages that come with GHC on Unix-like operating systems (`base`, `Cabal`, `time` etc) - `Win32-2.12.0.1` is listed.

If I command:

~~~text
stack list
~~~

it yields nothing. I was rather hoping it would yield the snapshot specified by Stack's project-level configuration file (`stack.yaml`). Is nothing the intent?

juhp:
I believe that is the current behaviour yes. I mean I would like the boot libraries to be listed too, but I didn't manage to wrangle that - if it's straightforward to do that would be great, thanks!

mpilgrem:
@juhp, many thanks. For now. I've updated the online documentation to describe more fully the existing command.

PR #5443, opened 2020-11-28
Adds Flag to Upload Package Candidate

aaronallen8455:
This adds a `--candidate` flag to the `upload` command to upload
a package candidate rather than publishing the package.
This is useful for checking that everything looks right before
publishing. (see https://hackage.haskell.org/upload#candidates)

I have tested that passing the `--candidate` flag creates a package candidate on hackage.

### DIFF
--- a/src/Stack/Options/UploadParser.hs
+++ b/src/Stack/Options/UploadParser.hs
@@ -1,0 +1,35 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+module Stack.Options.UploadParser
+  ( UploadOpts(..)
+  , UploadVariant(..)
+  , uploadOptsParser
+  ) where
+
+import           Options.Applicative
+import           Stack.Options.SDistParser (sdistOptsParser)
+import           Stack.Prelude
+import           Stack.SDist (SDistOpts(..))
+
+data UploadOpts = UploadOpts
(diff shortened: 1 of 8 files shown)


src/Stack/Options/UploadParser.hs
```diff
@@ -0,0 +1,38 @@
...
+
+-- | Parser for arguments to `stack upload`
+uploadOptsParser :: Parser UploadOpts
+uploadOptsParser =
+  UploadOpts
+    <$> sdistOptsParser
+    <*> uploadVariant
+  where
+    uploadVariant =
+      switch (long "candidate"
```
snoyberg, line 33: How about using [`flag`](https://www.stackage.org/haddock/lts-17.7/optparse-applicative-0.15.1.0/Options-Applicative.html#v:flag) instead of `switch` and `<&>`?

snoyberg (changes requested):
One minor comment, but LGTM, thanks!

src/Stack/Options/UploadParser.hs
aaronallen8455: I've updated it to use `flag`. Thanks

snoyberg (approved):
Thanks!

Issue #3310, opened 2017-08-03
True multi package support in "stack ghci --with-ghc=intero"


mgsloan:
This is just a stub issue to keep track of the goal of having "stack ghci" properly load packages, including switching cwd when loading modules (for TH file loading).  I am removing the support for using intero's `:cd-ghc` command with multiple packages, because it doesn't always work, doesn't work on reload, and was generally a regression in behavior for multi-package projects - see https://github.com/commercialhaskell/stack/issues/3309

A proper solution to this requires an implementation in ghci / intero.  Some relevant issues:

https://ghc.haskell.org/trac/ghc/ticket/10827
https://github.com/commercialhaskell/intero/issues/41
https://github.com/commercialhaskell/intero/issues/39

Ericson2314:
https://github.com/ghc-proposals/ghc-proposals/pull/263 There's a proposal this (and WIP PR for the first step of this).

qrilka:
I've removed `intero` from the title as I suppose there is no need to take it into account as the project has reached its end of life - https://github.com/chrisdone/intero/blob/master/README.md

From the other side there were 2 GSoC projects and some MRs to GHC aiming to resolve the problem of loading multiple packages into GHCi, the main entry point looks to be the wiki page [Multi Session GHC API
](https://gitlab.haskell.org/ghc/ghc/-/wikis/Multi-Session-GHC-API)

flip111:
This might be possible now as https://gitlab.haskell.org/ghc/ghc/-/issues/10827 got resolved through https://gitlab.haskell.org/ghc/ghc/-/merge_requests/6805?commit_id=fd42ab5fa1df847a6b595dfe4b63d9c7eecbf400

Commit fd42ab5fa1df847a6b595dfe4b63d9c7eecbf400 is available in the following git tags (as of this moment):

* ghc-9.4.1-alpha1
* ghc-9.4.1-alpha2
* ghc-9.4.1-alpha3
* ghc-9.4.1-rc1
* ghc-9.4.1-release
* ghc-9.4.2-release
* ghc-9.4.3-release
* ghc-9.4.4-release
* ghc-9.5-start
* ghc-9.6.1-alpha2
* ghc-9.6.1-alpha3
* ghc-9.7-start

PR #5448, opened 2020-12-02
Proper subpackage cwd in stack repl

### LINKED ISSUE #5421: `stack repl subpackage` compile time cwd is different from `stack build subpackage` one

The current working dir at compile time is different between `stack build` and `stack repl`. It likely makes template haskell io computations relying in cwd fail, cause the relative file or directory usually does not exist.
This in turn makes that code not usable in a ide (hls, ghcide) that uses `hie-bios` to manage ghc session, as `hie-bios` uses `stack repl` underneath for stack projects.
(original issue https://github.com/haskell/haskell-language-server/issues/481) 

### Steps to reproduce

1. Create a simple stack project with a main package with an exe and a subpackage with a lib
2. Create in the subpackage lib a module with

```haskell
import System.Directory
import Language.Haskell.TH
import Language.Haskell.TH.Syntax

cwd :: FilePath
cwd = $$(runIO getCurrentDirectory >>= \d-> fmap TExp (lift (d :: FilePath)))
```

and in the main package an executable with

```haskell
main = print cwd
```

3. Run command `stack run` and observe cwd is `/path/to/main/package/subpackage`
4. Run command `stack repl subpackage` or `stack repl` and observe cwd evaluation gives us `/path/to/main/package`

### Expected

Both paths should be equal.

### Actual

`stack build` uses the subpackage path and `stack repl` the main package one

### Stack version

```
$ stack --version
Version 2.5.1  x86_64 hpack-0.33.0
```

### Method of installation

qrilka:

Resolves #5421. Tested manually

### DIFF
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -32,7 +32,7 @@ import           Path.Extra (toFilePathNoTrailingSep)
 import           Path.IO hiding (withSystemTempDir)
 import qualified RIO
 import           RIO.PrettyPrint
-import           RIO.Process (HasProcessContext, exec, proc, readProcess_)
+import           RIO.Process (HasProcessContext, exec, proc, readProcess_, withWorkingDir)
 import           Stack.Build
 import           Stack.Build.Installed
 import           Stack.Build.Source
@@ -412,7 +412,7 @@ runGhci GhciOpts{..} targets mainFile pkgs extraFiles exposePackages = do
     compilerExeName <- view $ compilerPathsL.to cpCompiler.to toFilePath
     let execGhci extras = do
             menv <- liftIO $ configProcessContextSettings config defaultEnvSettings
-            withProcessContext menv $ exec
+            withPackageWorkingDir $ withProcessContext menv $ exec
                  (fromMaybe compilerExeName ghciGhcCommand)
                  (("--interactive" : ) $
                  -- This initial "-i" resets the include directories to
@@ -420,6 +420,10 @@ runGhci GhciOpts{..} targets mainFile pkgs extraFiles exposePackages = do
                  -- is included.
                   (if null pkgs then id else ("-i" : )) $
                   odir <> pkgopts <> extras <> ghciGhcOptions <> ghciArgs)
+        withPackageWorkingDir =
+            case pkgs of
+              [pkg] -> withWorkingDir (toFilePath $ ghciPkgDir pkg)
+              _ -> id
         -- TODO: Consider optimizing this check. Perhaps if no
         -- "with-ghc" is specified, assume that it is not using intero.
         checkIsIntero =
--- a/src/Stack/Ghci.hs
+++ b/src/Stack/Ghci.hs
@@ -751,48 +751,63 @@ wantedPackageComponents _ _ _ = S.empty
 
 checkForIssues :: HasLogFunc env => [GhciPkgInfo] -> RIO env ()
 checkForIssues pkgs = do
-    unless (null issues) $ borderedWarning $ do
-        logWarn "Warning: There are cabal settings for this project which may prevent GHCi from loading your code properly."
-        logWarn "In some cases it can also load some projects which would otherwise fail to build."
-        logWarn ""
-        mapM_ (logWarn . RIO.display) $ intercalate [""] issues
-        logWarn ""
-        logWarn "To resolve, remove the flag(s) from the cabal file(s) and instead put them at the top of the haskell files."
-        logWarn ""
+    when (length pkgs > 1) $ borderedWarning $ do
+        -- Cabal flag issues could arise only when there are at least 2 packages
+        unless (null cabalFlagIssues) $ borderedWarning $ do
+            logWarn "Warning: There are cabal flags for this project which may prevent GHCi from loading your code properly."
+            logWarn "In some cases it can also load some projects which would otherwise fail to build."
+            logWarn ""
+            mapM_ (logWarn . RIO.display) $ intercalate [""] cabalFlagIssues
+            logWarn ""
+            logWarn "To resolve, remove the flag(s) from the cabal file(s) and instead put them at the top of the haskell files."
+            logWarn ""
         logWarn "It isn't yet possible to load multiple packages into GHCi in all cases - see"
         logWarn "https://ghc.haskell.org/trac/ghc/ticket/10827"
   where
-    issues = concat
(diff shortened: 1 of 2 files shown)


src/Stack/Ghci.hs
```diff
@@ -747,14 +751,15 @@ wantedPackageComponents _ _ _ = S.empty
...
 checkForIssues pkgs = do
-    unless (null issues) $ borderedWarning $ do
-        logWarn "Warning: There are cabal settings for this project which may prevent GHCi from loading your code properly."
-        logWarn "In some cases it can also load some projects which would otherwise fail to build."
-        logWarn ""
-        mapM_ (logWarn . RIO.display) $ intercalate [""] issues
-        logWarn ""
-        logWarn "To resolve, remove the flag(s) from the cabal file(s) and instead put them at the top of the haskell files."
-        logWarn ""
+    when (length pkgs > 1) $ borderedWarning $ do
```
snoyberg, line 754: This looks a bit confusing to me. The code reads as "only show the issues if there are multiple packages." I think we know that there are only issues if there are multiple packages, but it's not obvious at this part of the code. I'd recommend one of the following:

* Comment explaining this
* Add an assertion like `assert (length pkgs > 1 || null issues)`
* Only wrapping the 'It isn't yet possible" comment below with `when (length pkgs > 1)`

snoyberg (changes requested):

src/Stack/Ghci.hs
qrilka: You can't have 2 packages with different flags when you have only 1 - does this sound as a good explanation?
But it makes sense to make it more explicit that `issues` in this case are about conflicting flags between different packages.

snoyberg: Yes, that's a good explanation. I'm just thinking that with a generic term like `issues`, it's easy to read it as "all of the potential issues with a package, not just the ones that could arise from conflicting flags or different CWDs."

qrilka: Exactly and even different CWDs don't go into these `issues`.

qrilka: I've pushed an update addressing this

snoyberg (approved):
LGTM, please merge when ready

jneira:
Hi, many thanks for trying fix this.
Unfortunately my tests continue failing, as the subpacakge module continue having cwd at main project root using the repl:

```shell
D:\ws\haskell\stack-test>stack --version
Version 2.6.0, Git revision a636e262579753326cde8d0f38f19e137bb872a2 (dirty) (8261 commits) PRE-RELEASE x86_64 hpack-0.33.0

D:\ws\haskell\stack-test>type subpackage\src\MyLib.hs
{-# LANGUAGE CPP #-}
{-# LANGUAGE TemplateHaskell #-}
module MyLib (cwd, val, val1) where
import System.Directory
import Data.Yaml (Value)
import Data.Yaml.TH (decodeFile)
import Language.Haskell.TH
import Language.Haskell.TH.Syntax

cwd :: FilePath
cwd = $$(runIO getCurrentDirectory >>= \d-> fmap TExp (lift (d :: FilePath)))


val :: Value
val = $$(do
           file <- runIO $ do
                    exist <- doesFileExist "config/value.yaml"
                    return $ if exist then "config/value.yaml"
                                      else "subpackage/config/value.yaml"
           decodeFile file :: Q (TExp Value)
        )

val1 :: Value
val1 = $$(decodeFile "config/value.yaml" :: Q (TExp Value))

D:\ws\haskell\stack-test>dir config /B
No se encuentra el archivo

D:\ws\haskell\stack-test>dir subpackage\config /B
value.yaml

D:\ws\haskell\stack-test>type app\Main.hs
module Main where

import Lib
import MyLib

main :: IO ()
main = do
    putStrLn $ cwd
    putStrLn $ show val
    putStrLn $ show val1

D:\ws\haskell\stack-test>stack run
.................
D:\ws\haskell\stack-test\subpackage
Number 2.0
Number 2.0

D:\ws\haskell\stack-test>stack repl
.............................

D:\\\\ws\haskell\stack-test\subpackage\src\MyLib.hs:24:11: error:
    • Exception when trying toto run compile-time code:
        InvalidYaml (Just (YamlException "Yaml file not found: config/value.yaml"))
      Code: decodeFile "config/value.yaml" :: Q (TExp Value)
    • In the Template Haskell splice:
        $$(decodeFile "config/value.yaml" :: Q (TExp Value))
      In the expression:
        $$(decodeFile "config/value.yaml" :: Q (TExp Value))
      In an equation for ‘val1’:
          val1 = $$(decodeFile "config/value.yaml" :: Q (TExp Value))
   |
24 | val1 = $$(decodeFile "config/value.yaml" :: Q (TExp Value))
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, one module loaded.

<no location info>: error:
    Could not load module ‘MyLib’
    It is a member of the hidden package ‘subpackage-0.1.0.0’.
    You can run ‘:set -package subpackage’ to expose it.
    (Note: this unloads all the modules in the current scope.)
Loaded GHCi configuration from C:\\TEMP\\haskell-stack-ghci\\9dd5157a\\ghci-scri
pt
*Lib>
```

qrilka:
@jneira have you read my comment https://github.com/commercialhaskell/stack/issues/5421#issuecomment-729942615 stating
>  we can't have different working directories for different packages in GHCi.

?
You have even replied that adding a multi-package warning is a suitable way to resolve the situation.
Unfortunately there seem to be no other way around as e.g. https://gitlab.haskell.org/ghc/ghc/-/issues/10827 is 5 years old already and https://gitlab.haskell.org/ghc/ghc/-/wikis/Multi-Session-GHC-API lists a couple of previous GSoC projects on the topic and some still not merged MRs as well.

jneira:
@qrilka uh, my bad, i did read what i want to and no what you said, sorry
I've tested that `stack repl subpackage` works so the pr works as intended.

qrilka:
Yeah, a proper resolution is blocked by upstream but I need to check if there are other issues on this topic so maybe we could at least minimize duplication in the issue tracker :)
Thanks for your feedback @jneira

qrilka:
@jneira I think we could close #5421 in favor of #3310 (probably with intero removed) and add some more info into that issue, does this sound OK to you?

jneira:
@qrilka sure!

PR #5461, opened 2020-12-21
Initialize custom-setup deps for stack dist

### LINKED ISSUE #4006: SDist code deling with custom-setup incorrect
When uploading **http-streams** using `stack upload .` (ie the `sdist` step) I get the following warning:
```
Warning: In getPackageArgs: custom-setup in use, but no dependency map present
Could not find custom-setup dep: Cabal
Could not find custom-setup dep: base
```

which is weird because the [.cabal](https://github.com/afcowie/http-streams/blob/master/http-streams.cabal#L43) file in question has what I was told to put there previously:

```
...
custom-setup
  setup-depends:     base >= 4.5,
                     Cabal >= 1.24

...
```

I asked in the Haskell channel at work and the epic @SamRoberts read through the **stack** code and in short order found the message is emitted from [Execute.hs](https://github.com/commercialhaskell/stack/blob/master/src/Stack/Build/Execute.hs#L1039)'s  function `withSingleContext`. Its fourth argument is a Maybe (Map something something). 

This appears to be called from [SDist.hs](https://github.com/commercialhaskell/stack/blob/master/src/Stack/SDist.hs#L333) where the fourth argument is hard coded to `Nothing`. Hence warning.

Seems either the `Nothing` is wrong and we need to pass in the custom-setup (?) map or the warning code needs to know that it's not going to actually get a map and not warn.

AfC

_feel free to change the issue title to something more helpful_

qrilka:
Resolves #4006

This `explicit-setup-deps` unnecessary and thus it was removed

Tested manually

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,11 @@ Behavior changes:
   one package. See
   [#5421](https://github.com/commercialhaskell/stack/issues/5421)
 
+* `custom-setup` dependencies are now properly initialized for `stack dist`.
+  This makes `explicit-setup-deps` no longer required and that option was
+  removed. See
+  [#4006](https://github.com/commercialhaskell/stack/issues/4006)
+
 Other enhancements:
 
 Bug fixes:
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -781,31 +781,6 @@ is to modify.
 modify-code-page: false
 ```
 
-### explicit-setup-deps
-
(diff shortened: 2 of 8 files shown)


qrilka:
@snoyberg one thing which I'm not quite sure about is removing that option - this makes this PR not backwards compatible. Do we need to do some "deprecation" step before removing it?

snoyberg (approved):
LGTM. I think it's OK to remove the option.

PR #5490, opened 2021-02-22
Add a new option for custom preprocessor extensions

### LINKED ISSUE #3491: Handle change tracking for pre-processor files / don't warn about them
In https://github.com/commercialhaskell/stack/issues/479#issuecomment-322025290 , warnings are emitted for modules that are defined via preprocessor.  Since module name -> file path resolution is used for knowing when to rebuild, I think there is likely a rebuild issue here.

Not sure how to unambiguously resolve module names to files that do not have .hs or .lhs extensions, though.  A hack could be to assume that it is the correct file if the name matches, ignoring the extension, and if it's listed in extra-source-files as suggested by the warning text.  Currently it seems like extra-source-files is not taken into account here.  So, either the warning text should be fixed or the behavior should make it so that its suggestion works.

Repro is at https://github.com/adituv/protobuf/tree/testpreprocess

anka-213:
Instead of just having a hard-coded list of preprocessor extensions, this adds a configuration option for custom preprocessor extensions.

Fixes #3491 as suggested in https://github.com/commercialhaskell/stack/issues/3491#issuecomment-336742261

TODO:
* [ ] Add tests

#### Please also shortly describe how you tested your change. Bonus points for added tests!
I have manually tested the option on an external repo. It updates on changes as expected.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -26,6 +26,10 @@ Other enhancements:
 * `stack list` is a new command to list package versions in a snapshot.
   See [#5431](https://github.com/commercialhaskell/stack/pull/5431)
 
+* `custom-preprocessor-extensions` is a new configuration option for allowing
+  stack to be aware of any custom preprocessors you have added to `Setup.hs`.
+  See [#3491](https://github.com/commercialhaskell/stack/issues/3491)
+
 Bug fixes:
 
 * `stack new` now suppports branches other than `master` as default for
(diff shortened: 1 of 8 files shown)


anka-213:
The build failures seem to be unrelated to my changes. Should I try to fix them anyways?

snoyberg (approved):
LGTM

anka-213:
How often do you usually release new versions of stack? I'm curious about approximately when I can expect this to be in an official release?

snoyberg:
We're typically pretty slow on the releases, maybe every 6-12 months. But since we just added support for GHC 9, we can probably do this one pretty soon. I'll ping internally.

anka-213:
Thanks!

PR #5464, opened 2021-01-04
 Minor refactoring and improvements

theobat:
- Rename lpWanted to lpShouldBeBuilt
- Move printPlan and displayTask in thee Build.hs file (where Plan and Task are defined)
- Isolate the merging of planTasks and planFinals into their own function (and remove mergeWithKeys as it's recommended to not use it in its haddock) defined in Build.hs.
- A reorganization of toAction which transforms a task into an action. This was using very redundantly the Task constructor, and the task -> action is likely to change with component builds.
- Rename W and M in ConstructPlan to more sensible names. Minor other changes there (changed the misnamed packageDepsWithTools).
- Refactor Execute.hs and split the Precompiled/installed stuff into their own module.
- Various haddock additions.
- Remove PackageDepsOracle (unused)
- Split ConstructPlan in 2 modules (the current one and its types)

@qrilka  Note that I added small things since the pull request on master. Notably there was a non pure packageDepsWithTools accessor to package deps, which was only used once and had nothing to do with the actual accessing of package deps.
I have another pull request almost ready after this one which does some actual rework on Setup.hs, the Package datatype and the sublibrary dependencies.

PR #5504, opened 2021-03-10
Comp/package component refactoring

theobat:
So this a work in progress, I'm stuck because I don't know how to get the proper set of components a local package should build (current behavior is, I try to ubild every component, which probably cause some of the errors in the tests).

@qrilka this is the PR I mentioned earlier.

Maybe you can review this (once the other doc PR is merged) and give me some feedback (it's still very messy)

### DIFF
--- a/src/Control/Concurrent/Execute.hs
+++ b/src/Control/Concurrent/Execute.hs
@@ -16,6 +16,8 @@ import           Control.Concurrent.STM   (retry)
 import           Stack.Prelude
 import           Data.List (sortBy)
 import qualified Data.Set                 as Set
+import Stack.Types.NamedComponent
+-- import Prelude (print)
 
 data ActionType
     = ATBuild
@@ -31,7 +33,7 @@ data ActionType
     | ATRunBenchmarks
       -- ^ Task for running the package's benchmarks.
     deriving (Show, Eq, Ord)
-data ActionId = ActionId !PackageIdentifier !ActionType
+data ActionId = ActionId !(PackageIdentifier, Set CompName) !ActionType
     deriving (Show, Eq, Ord)
 -- | This is mainly the result of one or more 'Task' in "Stack.Types.Build".
 -- Its actual content is completely contained in actionDo.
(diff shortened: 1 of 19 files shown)


qrilka:
@theobat the diff seems to contain #5464 - would you mind rebasing?

theobat:
yeah sorry I already did it

qrilka:
Not sure about internal libraries and getting installed components info is one of questions that need to be resolved if we keep the way Stack currently works.

As for components - by default all of them should be included but filtered out by build flags which set whether tests/benches are to be included.

Hope this helps at least a bit.

I'll start reviewing this tomorrow

qrilka:
@theobat I had some (not very thorough) overview of this PR and I'm not sure what kind of a feedback do you want. In the code I don't see any obvious problems (but I had rather high-level overview of the diff) and tests are surely failing as GitHub shows. Maybe you have some concrete questions I could help finding an answer for? Do you have a plan how to continue here? If I was attacking this I would start with some MVP, e.g. doing a very minimal example with only `base` as a dependency but also involving components and then increase number of features used as you progress.

qrilka:
@theobat are you still working on this?

theobat:
@qrilka I'll admit I don't have a lot of time, I wish I did, so I guess my answer is "not right now, and probably not enough in the coming weeks/months". The other significant issue is stack does not build on my mac anymore which does hinder the whole process (I also have ubuntu but it's not my day to day work computer so it's less convenient).
Finally I don't know what's wrong with my changes, I only added a component name to the build proceses, so I'll have to look at the tests one by one to see what's wrong. I suspect it's related to inconsistent flags between component based builds and tests but it's hand wavy...

qrilka:
It looks like I have also not that much time, I thought about trying out your branch but didn't find time yet. As about mac - that's surely not something I could help with. If you have any ideas about immediate tasks here I could look into - please share, probably that could make it more easy to find some time if it were something more concrete.

theobat:
So, I managed to fix the issue with cocoa on mac thanks to this : https://github.com/purescript/purescript/issues/3960#issuecomment-743069720

I'll try to look at the failing tests in the coming weeks

PR #5511, opened 2021-03-31
Bust the cache

snoyberg:
CC @borsboom


borsboom:
It looks like the original problem solved itself (see https://github.com/commercialhaskell/stack/runs/2238653076?check_suite_focus=true), so I'll close this for now.  Sorry about the noise!

PR #5512, opened 2021-03-31
Cache stack.yaml  from a variable properly

qrilka:
Found this when taking Stack's github actions as an example for my project :)
E.g. on https://github.com/commercialhaskell/stack/runs/2238843841?check_suite_focus=true checking the "Cache dependencies" step logs shows:

```
Cache not found for input keys: Linux---bust1.
```

Which clearly has no hashes in cache name

Could this be related to #5511 @snoyberg ?

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -81,7 +81,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ~/.stack
-          key: ${{ runner.os }}-${{ hashFiles('${{ matrix.stack-yaml }}') }}-${{ matrix.extra-suffix }}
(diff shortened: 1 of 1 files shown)


snoyberg (approved):

qrilka:
@snoyberg I've done a force-push to trigger a new build using caches and the result look good https://github.com/commercialhaskell/stack/actions/runs/706000445 show Linux runs from 29 to 34 minutes and in https://github.com/commercialhaskell/stack/actions/runs/706293548 I see them getting down to 8 minutes maximum. macOS is a bit slow - almost 12 minutes but that's way better than 33 minutes. The only problem is with Windows, it looks like we're hitting this bug - https://github.com/actions/cache/issues/198
But this will be still a very good speedup so merging

PR #5529, opened 2021-04-26
Fix path to Hadrian build script for GHC source builds

### LINKED ISSUE #5528: Stack invokes the wrong Hadrian build script when building GHC from source
### General summary

Building a recent version of GHC from source by specifying `compiler: ghc-git-COMMIT-FLAVOUR` in a `stack.yaml` ([as described here](https://github.com/commercialhaskell/stack/blob/master/doc/yaml_configuration.md#building-ghc-from-source-experimental)) fails with the following error:

```
Building GHC from source with `quick` flavour. It can take a long time (more than one hour)...
Did not find executable at specified path: /tmp/with-repo468829/cloned/hadrian/build.stack.sh
```

### Steps to reproduce

For example:

1. Create `stack.yaml` with

```yaml
compiler: ghc-git-7bc7eea3897dcb8a87fdb0921f451b9bc77309f6-quickest
```

(This is the newest GHC commit at the time of writing)

2. Run `stack build`

### Expected

The script `hadrian/build-stack` should be invoked.

### Actual

The script `hadrian/build.stack.sh` was invoked (apparently the script has been renamed).

### Stack version

```
$ stack --version
Version 2.5.1, Git revision d6ab861544918185236cf826cb2028abb266d6d5 x86_64 hpack-0.33.0
```

### Method of installation

Official binary

fwcd:
The build script has been renamed from `build.stack.sh` to `build-stack` (and `build-stack.bat` on Windows).

Fixes #5528.

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md (no user-facing changes, so not necessary)

The fix has been tested manually with recent GHC versions.

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -841,7 +841,7 @@ buildGhcFromSource getSetupInfo' installed (CompilerRepository url) commitId fla
              | otherwise   = hadrianScriptsPosix
 
          foundHadrianPaths <- filterM doesFileExist $ (cwd </>) <$> hadrianScripts
-         let hadrianPath = fromMaybe (error "No Hadrian build script found") $ listToMaybe foundHadrianPaths
+         hadrianPath <- maybe (throwString "No Hadrian build script found") pure $ listToMaybe foundHadrianPaths
 
          logSticky $ "Building GHC from source with `"
             <> RIO.display flavour
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -836,9 +836,12 @@ buildGhcFromSource getSetupInfo' installed (CompilerRepository url) commitId fla
                , "--flavour=" <> flavour -- selected flavour
                , "binary-dist"
                ]
-           hadrianCmd
-             | osIsWindows = hadrianCmdWindows
-             | otherwise   = hadrianCmdPosix
+           hadrianScripts
+             | osIsWindows = hadrianScriptsWindows
+             | otherwise   = hadrianScriptsPosix
+
+         foundHadrianPaths <- filterM doesFileExist $ (cwd </>) <$> hadrianScripts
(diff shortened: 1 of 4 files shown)


snoyberg:
My only concern is that this seems like it will break old builds with old versions of GHC. I'm not sure if we really care about that. But it may be worthwhile to include logic to check for both versions of the file and then use the one that exists. Thoughts?

fwcd:
Sure, I've updated the logic to include both the new and the old path and choose the first one that exists.

snoyberg:
> Any changes that could be relevant to users have been recorded in the ChangeLog.md (no user-facing changes, so not necessary)

It would actually still be nice to include this in the changelog so that, if someone is hitting this issue on an older version, looking at the changelog will let them know it's fixed.

src/Stack/Setup.hs
```diff
@@ -836,9 +836,12 @@ buildGhcFromSource getSetupInfo' installed (CompilerRepository url) commitId fla
...
                ]
-           hadrianCmd
-             | osIsWindows = hadrianCmdWindows
-             | otherwise   = hadrianCmdPosix
+           hadrianScripts
+             | osIsWindows = hadrianScriptsWindows
+             | otherwise   = hadrianScriptsPosix
+
+         foundHadrianPaths <- filterM doesFileExist $ (cwd </>) <$> hadrianScripts
+         let hadrianPath = fromMaybe (error "No Hadrian build script found") $ listToMaybe foundHadrianPaths
```
snoyberg, line 844:
```suggestion
         hadrianPath <- maybe (throwString "No Hadrian build script found") pure $ listToMaybe foundHadrianPaths
```

fwcd:
The branch is now rebased on master and includes the updated changelog and the suggestion.

snoyberg (approved):

snoyberg:
Thanks!

Issue #5452, opened 2020-12-07
Official support for KaOS


cosmos64:
Hi there :hugs: 

I implemented KaOS support, which is an independent distro, that packages itself. 

The current situation shows some generic text: 

```
No setup information was found for ghc-8.8.4 on your platform.
This probably means a GHC bindist has not yet been added for OS key 'linux64-ncurses6'.
Supported versions: ghc-7.10.3, ghc-8.0.1, ghc-8.0.2, ghc-8.2.1, ghc-8.2.2
```

Proposed integration: 

```sh
while true; do
     printf "For stack to work in KaOS is it important to install libtinfo.\nDo you like to do that?" 
     read -p " y or n: " yn
    case $yn in
        [Yy]*)  printf "\e[1;31m Do not edit the PKGBUILD\e[0m " && kcp -u && kcp -i libtinfo; break;;
        [Nn]*)  break;;
        *) printf "Please answer with y or n. ";;
    esac
done

while true; do
    echo
    printf  "To be detectable by IDEs and other applications is it recommended to symlink \e[34mstack\e[0m to the PATH.\nDo you like to do that?" 
    read -p " y or n: " yn
    case $yn in
        [Yy]*)  sudo ln -s /home/$USER/.stack/programs/x86_64-linux/ghc-tinfo6-8.8.4/bin/ghc /usr/bin; break;;
        [Nn]*)  exit;;
        *) printf "Please answer with y or n. ";;
    esac
done
```
Please respect, that I am very new at this. 
I tested the script and like to know if the symlink of ghc is sufficient.

chreekat:
@cosmos64 , it's unusual for a project to support individual Linux distros directly in its own installer. There's just too many of them. Plus, it's usually the job of the distro itself to package up software so it's compatible with the rest of the system. In fact, that's what a distro is all about.

I would suggest bringing this up in the KaOS community!

ShalokShalom:
Stack is in fact 'unusual'  and does precisely that

https://github.com/commercialhaskell/stack/blob/master/etc/scripts/get-stack.sh#L192

cosmos64:
Can I have an update on this?

mpilgrem:
I am closing this issue as attempts to find a user of Stack on KaOS failed: see https://github.com/commercialhaskell/stack/pull/5535#issuecomment-1236781892.

PR #5535, opened 2021-05-08
Adds KaOS support

ShalokShalom:
See this: https://github.com/commercialhaskell/stack/issues/5452

Is this something I am supposed to mention in the change.log?

Then I do it

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -330,6 +330,31 @@ do_alpine_install() {
   fi
 }
 
+# KaOS distro install
+
+do_kaos_install() {
+    while true; do
+     printf "For stack to work in KaOS is it fundamental to install libtinfo.\nDo you like to do that?" 
+     read -p " y or n: " yn
+    case $yn in
+        [Yy]*)  printf "\e[1;31m Do not edit the PKGBUILD\e[0m " && kcp -u && kcp -i libtinfo; break;;
+        [Nn]*)  break;;
+        *) printf "Please answer with y or n. ";;
+      esac
+    done
+
+    while true; do
+      echo
+      printf  "To be detectable by IDEs and other applications is it recommended to symlink \e[34mGHC\e[0m to the PATH.\nDo you like to do that?" 
+      read -p " y or n: " yn
+    case $yn in
+        [Yy]*)  sudo ln -s /home/$USER/.stack/programs/x86_64-linux/ghc-tinfo6-8.8.4/bin/ghc /usr/bin; break;;
+        [Nn]*)  exit;;
+        *) printf "Please answer with y or n. ";;
+      esac
+    done
+}
+
 # Attempts to install on unsupported Linux distribution by downloading
 # the bindist.
 do_sloppy_install() {
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -333,6 +333,7 @@ do_alpine_install() {
 # KaOS distro install
 
 do_kaos_install() {
+    install_x86_64_linux_binary
     while true; do
      printf "For stack to work in KaOS is it fundamental to install libtinfo.\nDo you like to do that?" 
      read -p " y or n: " yn
@@ -433,6 +434,9 @@ distro_info() {
       "Arch Linux"*)
         echo "arch;" # n.b. Version is not available in /etc/issue on Arch
         ;;
+      "KaOS"*)
+        echo "kaos;" # no version used, rolling release. 
+        ;;  
       "Ubuntu"*)
         echo "ubuntu;$(perl -ne 'if(/Ubuntu (\d+\.\d+)/) { print $1; }' < /etc/issue)"
         ;;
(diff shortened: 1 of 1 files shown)


etc/scripts/get-stack.sh
```diff
@@ -328,6 +328,31 @@ do_alpine_install() {
...
+        *) printf "Please answer with y or n. ";;
+      esac
+    done
+
+    while true; do
+      echo
+      printf  "To be detectable by IDEs and other applications is it recommended to symlink \e[34mGHC\e[0m to the PATH.\nDo you like to do that?" 
+      read -p " y or n: " yn
+    case $yn in
+        [Yy]*)  sudo ln -s /home/$USER/.stack/programs/x86_64-linux/ghc-tinfo6-8.8.4/bin/ghc /usr/bin; break;;
```
ShalokShalom, line 350: This line can only work when we soft code the version

ShalokShalom (commented):
I did a mistake and hardcoded that one, how can we detect the version of ghc-tinfo6 easily?

etc/scripts/get-stack.sh
hasufell: I don't think install scripts should ever do this. `/usr/bin` is controlled by your distros package manager.

Also note that there is a `sudocmd` function:

https://github.com/commercialhaskell/stack/blob/708b516a8d34cc42b6c533d161b18b31f913c3d6/etc/scripts/get-stack.sh#L138

ShalokShalom: So you suggest a different dir for the PATH?
And using sudocmd instead sudo?

hasufell: My opinion is to not do anything here. Global GHC installations is probably not in the scope the install script. IDEs like HLS work with stack just fine without exposing a global GHC.

ShalokShalom: its exactly what this script does for Debian and Co, no?

ShalokShalom: And vscode does not

hasufell: > its exactly what this script does for Debian and Co, no?

I don't think so.

mpilgrem:
I would like to resolve this pull request, one way or another. I am a Windows user, so forgive me if I take things step-by-step and ask questions that may seem obvious to a Linux user.

I understand that KaOS is a Linux distribution that is x86-64-specific, and `kcp` is a KaOS package manager. I also understand from the above that KaOS can use the generic Linux GHC binary, but needs `kcp` package `libtinfo`(https://kaos-community-packages.github.io/items/libtinfo.html) as a pre-requisite. Is that correct? Does KaOS not come with something that other Linux distributions provide as standard?

Other parts of the script (for other Linux distribtions) have things like this:

~~~sh
do_fedora_install() {
  install_dependencies() {
    dnf_install_pkgs perl make automake gcc gmp-devel libffi zlib-devel xz tar git gnupg
  }

  if is_x86_64 ; then
    install_dependencies "$1"
    print_bindist_notice
    install_x86_64_linux_binary
  else
    die "Sorry, currently only 64-bit (x86_64) Linux binary is available."
  fi
}

# Install packages using dnf
dnf_install_pkgs() {
  if ! sudocmd "install required system dependencies" dnf install -y ${QUIET:+-q} "$@"; then
    die "\nInstalling dnf packages failed.  Please run 'dnf check-update' and try again."
  fi
}
~~~

Should KaOS/kcp follow the same 'pattern' for its `libtinfo` need? What is it about `libtinfo` that a warning is required before it is installed?

Is `libtinfo` the only dependency that is required if the distribution is KaOS? Distributions such as Ubuntu and Fedora seem to have a list of things that are needed?

I wondered why KaOS is introduced via `do_sloppy_install()` and not as its own `case "$DISTRO"` in `do_distro()`, if it is now being catered for expressly?

The addition of a link in ` /usr/bin` to a particular version of GHC in Stack's 'sandbox' looks wrong to me. If a KaOS user also wanted a specific version of GHC on the PATH (as well as the Stack-supplied ones), would they not just install it?

ShalokShalom:
Wonderful 😄 

Yes, you are basically correct about everything - the only thing I like to mention, is that libtinfo is build by the community and kcp is also a package manager for the community repo. 

I only asked, if said dependency is fine to be added, since some people are quite sensitive about that. 

I personally think its perfectly fine. 👍🏼 

I love to introduce KaOS on the same level and with the same syntax as other distros, I think I was just to shy and hesitant to actually change something of the existing code. 

You are also right about hard coding a specific version. 

The only thing that is important to me, is to have an option to let the user choose to automatically add the compiler to the PATH. 

Preferably, by scanning the sub directories, or simply by choosing the most recent version. 

This way, people can just start coding. 🥳

ShalokShalom:
Sorry, I am not on KaOS anymore.

mpilgrem:
I asked for testers on the [Haskell Community](https://discourse.haskell.org/t/kaos-linux-distro-and-stacks-installation-script/4953) but had no response. I am going to close this pull request. It can be reopened if another KaOS user and Haskeller emerges.

ShalokShalom:
I just have no time at the moment. I tried to launch it in Virtualbox, but failed. 

Will try if I find the time

PR #5544, opened 2021-05-17
Fixes #732: Combine multiple stack.yaml files

### LINKED ISSUE #732: Method to combine multiple project stack.yaml files
A I understand it, `stack.yaml` is a passive/declarative format.  It doesn't have any executable bits, or `#include` syntax, or whatever.  That's all well and good, but it does seem to leave us doing ad-hock `cat` commands on fragments of `stack.yaml` to build the right one.

In particular this comes up when *varying things*.  Like a really simple example would be to build in docker and non-docker environments (and multiple resolvers of course).  Currently, we would put the normal base config in `stack.yaml`, and then have a separate file, say `enable_docker.yaml`, with just this entry:

```yaml
docker:
  enable: true
  repo: ...
```

Then we could build by running:

```bash
$ cat stack.yaml enable_docker.yaml > stack2.yaml
$ stack --stack-yaml=stack2.yaml build
```

This is ok, but it is kind of ugly/messy, and it doesn't scale super well.  If we had command line arguments for all the docker config bits, we could vary this by varying the command line arguments, but we currently don't.

Initially, by the way, I kind of thought the following might work:

```bash
$ stack --stack-yaml=stack.yaml --stack-yaml=enable_docker.yaml build
```

But currently this is an error.

Internally, there's already the ability to combine configs (Monoid), right?  And we already mix configs from global / user / project.  So the restriction to only mix exactly those three configs seems a bit arbitrary.

This is somewhat related to #392 and #378.

mmhat:
This PR implements support for combining mulitple stack.yaml files. Consider a command line like
`stack --stack-yaml stack.yaml --stack-yaml stack.docker.yaml --stack-yaml stack.local.yaml ...`.
The first file (stack.yaml) in the list is the "main" file holding our project config like it used to be while the following files supplement that project configuration. In particular they may
* add new or adjust existing extra dependencies.
* set the resolver.
* set the compiler.
* set flags.

Fixes #732

The following has been done according to the contribution guidelines:

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -861,7 +861,7 @@ loadYaml parser path = do
 getProjectConfig :: HasLogFunc env
                  => StackYamlLoc
                  -- ^ Override stack.yaml
-                 -> RIO env (ProjectConfig (Path Abs File))
+                 -> RIO env (ProjectConfig (NonEmpty (Path Abs File)))
 getProjectConfig (SYLOverride stackYaml) = return $ PCProject stackYaml
 getProjectConfig SYLGlobalProject = return PCGlobalProject
 getProjectConfig SYLDefault = do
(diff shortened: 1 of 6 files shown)


mmhat:
Hello everbody, I'd like to have some feedback on that draft; In particular regarding the merge semantics of the different project configurations. The code in question is https://github.com/mmhat/stack/blob/5c23717e9ed7d9d8b9e92d1f54a80fc7f35be358/src/Stack/Types/Config.hs#L770

snoyberg:
I'm concerned about this kind of ordering decision being confusing to work with. It also doesn't seem to extend naturally to things like the `STACK_YAML` environment variable. Overall, I think I'd be more in favor of something more explicit, like having an `include` or `extend` pragma within a `stack.yaml` file. Even that would be something I'd be on the fence about though to be honest.

mmhat:
@snoyberg Thank you for the feedback! I agree with you that those semantics are kind of confusing, but they were the ones that made most sense to me when I implemented this.
Regarding an  `extend` or `include` pragma: I am not a fan of those either.

I'll have a look if I can come up with another solution but TBH right now I am inclined to close this PR.

mmhat:
Closing since I do not intend to continue working on this.

PR #5552, opened 2021-05-28
WIP: Bump resolver used in release.hs

snoyberg:

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -55,7 +55,7 @@ jobs:
             #   https://github.com/NixOS/nix/issues/2733
             (for i in {1..5}; do bash <(curl -sSL https://nixos.org/nix/install) && exit 0; done; exit 1)
             . ~/.nix-profile/etc/profile.d/nix.sh
-            nix-channel --add https://nixos.org/channels/nixos-19.09 nixpkgs
+            nix-channel --add https://nixos.org/channels/nixos-21.05 nixpkgs
             nix-channel --update # Get GHC 8.2.2
           elif [[ "${{ matrix.os }}" == "windows-latest" ]]
           then
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -22,6 +22,8 @@ Bug fixes:
   [#5545](https://github.com/commercialhaskell/stack/issues/5545)
 * Bump `pantry` version for better OS support. See
   [pantry#33](https://github.com/commercialhaskell/pantry/issues/33)
+* When building the sanity check for a new GHC install, make sure to clear
+  `GHC_PACKAGE_PATH`.
 
 
 ## v2.7.1
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ Bug fixes:
 * `stack setup` always looks for the unpacked directory name to support
   different tar file naming conventions. See
   [#5545](https://github.com/commercialhaskell/stack/issues/5545)
+* Bump `pantry` version for better OS support. See
+  [pantry#33](https://github.com/commercialhaskell/pantry/issues/33)
+
 
 ## v2.7.1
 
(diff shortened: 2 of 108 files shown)


borsboom:
I think the integration tests might be failing due to using an LTS that requires a different GHC version than is already available.  Probably best to bump the integration tests to also use a consistent LTS.

One of these days I might just drop the `release.hs` script entirely and replace it with a simple shell script, since CI is now handling all the builds and there's no longer the need for so much complexity.  Probably not worth doing unless significant changes would be needed though.

snoyberg:
> One of these days I might just drop the release.hs script entirely and replace it with a simple shell script,

+1 from me on that happening ultimately, but agreed, no rush

snoyberg:
I'm in particular unable to fix up the Nix problems here. If someone else is interested in taking a crack at that, I'd really appreciate it.

snoyberg:
I discovered the line in question here:

https://github.com/commercialhaskell/stack/blob/d1eef3febec2957e975f215689c7bb9be536184c/.github/workflows/integration-tests.yml#L58

We locked it down previously to make the test suite reproducible. But I have no idea what we should bump it to instead.

qrilka:
Oh, I completely forgot that it was pinned, sorry, will take a look

snoyberg:
FYI, I'm already including a workaround in this PR:

https://github.com/commercialhaskell/stack/pull/5555/commits/95d214c16674a1b80ccab9b753f73a5b0fb1c8d1

It's not as good, since it still requires two GHCs to be installed, but it does work around using the older GHC for release.hs itself.

qrilka:
@snoyberg it looks like we can't use the last stable version 20.09 as there's no 8.10.4 there:
```
$ NIX_PATH='nixpkgs=https://github.com/NixOS/nixpkgs/archive/20.09.tar.gz'  nix-instantiate --eval -E 'builtins.attrNames((import <nixpkgs> {}).haskell.compiler)'
unpacking 'https://github.com/NixOS/nixpkgs/archive/20.09.tar.gz'...
[ "ghc8101" "ghc8102" "ghc822Binary" "ghc865" "ghc865Binary" "ghc882" "ghc883" "ghc884" "ghcHEAD" "ghcjs" "ghcjs86" "integer-simple" "native-bignum" ]
```
but 21.05 is almost released so I guess this should be good enough as an option
```
$ NIX_PATH='nixpkgs=https://github.com/NixOS/nixpkgs/archive/release-21.05.tar.gz'  nix-instantiate --eval -E 'builtins.attrNames((import <nixpkgs> {}).haskell.compiler)'
unpacking 'https://github.com/NixOS/nixpkgs/archive/release-21.05.tar.gz'...
[ "ghc8102Binary" "ghc8102BinaryMinimal" "ghc8104" "ghc865Binary" "ghc884" "ghc901" "ghcHEAD" "integer-simple" "native-bignum" ]
```
but there's no channel for it yet so some other way of pinning is needed

qrilka:
@snoyberg I guess using env vars in that test could be even more explicit and we'll have the same config both in CI and locally and it looks like `NIX_PATH` is the way to go:
```
NIX_PATH='nixpkgs=https://github.com/NixOS/nixpkgs/archive/20.09.tar.gz' stack exec -- nix-instantiate --eval -E '(import <nixpkgs> {}).lib.version'
"20.09pre-git"
```

mpilgrem:
Rather late in the day, I realised that this pull request was seeking to do part of what I am trying to do in pull request #5697 (move stack to lts-19.7/GHC 9.0.2) - but this request is for lts-8.10.4/lts-17.10. My GHC 9.0.2 pull request passes the unit tests, and the integration tests on Windows, but I am stuck with [1] the integration tests running on `macos-latest` (my current task) and [2] the integration tests that need 'alpine' or 'docker' (see https://github.com/commercialhaskell/stack/pull/5736#issuecomment-1132112610). I have not worked out how I can test CI in my fork, so I am trying to fix the CI in my pull request.

mpilgrem:
I am going to close this pull request, as I think it has been overtaken by either #5697 (which was trying to move Stack to the current LTS, being GHC 9.0.2) or, more likely, #5763 (which is trying to move Stack to GHC 9.2.3 - the subject of the current `nightly` resolver).

PR #5555, opened 2021-05-31
Add a GitHub Actions workflow for ARM64

snoyberg:
This also updates release.hs to use a more recent resolver, since
otherwise we'll have trouble on ARM64 with conflicting LLVM
installations. Unlike #5552, I took the simple workaround here of
installing GHC 8.6.5 in addition to the more recent GHC needed by
release.hs.

### DIFF
--- a/.github/workflows/arm64-release.yml
+++ b/.github/workflows/arm64-release.yml
@@ -1,0 +1,34 @@
+name: ARM64 Release
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - stable
+      - rc/**
+    tags:
+      - '**'
+  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


snoyberg:
The approach I took with modifying the installed GHC didn't work, so I'm going to take another approach instead. I've separated off the sanity check PR to #5556. #5552 can continue to deal with bumping the resolver.

Also, I'm not happy with how much this approach relies on the setup of the GitHub runner instance, so I'm going to rework the ARM installation to use Docker instead.

snoyberg:
@borsboom @qrilka this is ready for review now. I've tested, and the ARM64 bindist seems to work correctly.

Issue #5533, opened 2021-05-06
Stack crashes with "SQLite3 returned ErrorReadOnly ..." on every build


JKTKops:
### General summary/comments

Whenever I try to build any stack project on one of my Windows machines, I get the following error (usually 7 copies of it):
```
SQLite3 returned ErrorReadOnly while attempting to perform step: attempt to write a readonly database
```

I've tried nuking the stack root directory, uninstalling and reinstalling stack, and deleting the .stack-work/ directories. After deleting .stack-work/ directories, I start getting this error:
```
Encountered error while migrating Stack database:
    SQLite3 returned ErrorReadOnly while attempting to perform step: attempt to write a readonly database
Please report this on https://github.com/commercialhaskell/stack/issues
As a workaround you may delete Stack database in C:\<path to project root>\.stack-work\stack.sqlite3 triggering its recreation.
```
The file `.stack-work\stack.sqlite3` didn't exist before executing `stack build` and still does not exist after.

I'm not sure exactly how to reproduce this. On one of my windows machines, it happens consistently and persistently no matter what I do. On the other one, I've never had a problem.

I have the official 2.5.1 binary which was initially install via `stack upgrade` but was reinstalled today from the installer in the user guide.

JKTKops:
On further investigation, it seems possible that this is due to an antivirus trying to protect files in subdirectories of `~/Documents`. I got the error to go away temporarily last week, but didn't know what I did. Today I tried listing `stack` as a trusted program to my antivirus and the problem vanished (I don't use this machine often, hence the slow progress). I'm going to hold off on closing this for a bit in case it comes back.

I don't know if it's possible for such problems to be fixed on Stack's end. Perhaps a note for Windows users that this may be necessary could be good.

ketzacoatl:
@JKTKops, :point_up: there is a doc update based on your feedback.

PR #5560, opened 2021-06-05
README: add note for windows users, to add stack to the list of trusted applications

ketzacoatl:
This contributes to https://github.com/commercialhaskell/stack/issues/5533

IDK if this should target stable, we can change it if that is not the case.


snoyberg (approved):
Thanks!

PR #5568, opened 2021-06-23
Allow passing RTS options to GHC in stack.yaml

### LINKED ISSUE #5180: Passing RTS options to GHC via `ghc-options` in `stack.yaml` fails
When trying to pass in RTS options to GHC via `ghc-options` in `stack.yaml`, the RTS options aren't properly threaded through. 

### Steps to reproduce

1. Create a new project via `stack new`
2. Add the following to `stack.yaml`
```
ghc-options:
  "$locals": +RTS -s -RTS
```
3. Run command `stack build`.

### Expected

The RTS options are passed to GHC, and the statistics are printed out while compiling.

### Actual

```
Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5: unknown RTS option: --ghc-options
```
### Stack version

```
$ stack --version
Version 2.1.3 x86_64
```

### Method of installation

parsonsmatt:

I added an integration test that ensures `stack` builds successfully when passing RTS options to GHC.

Fixes #5180

### DIFF
--- a/src/Stack/Types/Build.hs
+++ b/src/Stack/Types/Build.hs
@@ -629,7 +629,7 @@ configureOptsNoDir econfig bco deps isLocal package = concat
                        flagNameString name)
                     (Map.toList flags)
     , map T.unpack $ packageCabalConfigOpts package
-    , concatMap (\x -> [compilerOptionsCabalFlag wc, T.unpack x]) (packageGhcOptions package)
+    , processGhcOptions (packageGhcOptions package)
     , map ("--extra-include-dirs=" ++) (configExtraIncludeDirs config)
     , map ("--extra-lib-dirs=" ++) (configExtraLibDirs config)
     , maybe [] (\customGcc -> ["--with-gcc=" ++ toFilePath customGcc]) (configOverrideGccPath config)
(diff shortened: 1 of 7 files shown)


parsonsmatt:
it seems like the integration tests on Mac are broken for an unrelated reason - failing to execute GHC for permissions?

src/Stack/Types/Build.hs
```diff
@@ -629,14 +629,36 @@ configureOptsNoDir econfig bco deps isLocal package = concat
...
+        []
+    takeRtsArgs :: [Text] -> (String, [Text])
+    takeRtsArgs ("-RTS" : xs) =
+        ("-RTS", xs)
+    takeRtsArgs (x : xs) =
+        let
+            (other, rest) =
+                takeRtsArgs xs
+        in
+            (T.unpack x ++ " " ++ other, rest)
```
snoyberg, line 659: I'm not sure how important this is in this case, but this has quadratic performance. I'd recommend instead collecting a list of `Text`s and then combining with `T.unwords`. In fact, I think then you'd be able to use `break (== "-RTS")`

stack.yaml
```diff
@@ -8,6 +8,7 @@ docker:
   #repo: fpco/alpine-haskell-stack:8.10.4
   repo: fpco/alpine-haskell-stack@sha256:1024fe4b3b082a8df64d00e8563b3151220ed90af09604a8f7e1d44040500c30
 
+
```
snoyberg, line 11: Looks like an unnecessary whitespace change

snoyberg (changes requested):

src/Stack/Types/Build.hs
parsonsmatt: Thanks for the tip! I've provided a much nicer looking implementation, and it even made me aware of the case where a user *does not* provide a `-RTS` in the argument list.

I decided to splice a `-RTS` on the end of the collected RTS args regardless. This matches the syntax you might use in a CLI invocation - `ghc --version +RTS -N -A128m`. As far as I know, this shouldn't cause any problems - any string without an `-RTS` would completely fail without this, and any string with non-RTS arguments after a `+RTS` would fail in either case.

stack.yaml
parsonsmatt: fixed thanks!

snoyberg (approved):
LGTM, thanks!

PR #5585, opened 2021-07-14
Support custom GHC installation hooks

hasufell:
This adds GHC installation hooks. The feature is described in the documentation. I tested it with the newly added integration test.

### DIFF
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -1200,3 +1200,65 @@ This field is convenient in setups that restrict access to GitHub, for instance 
 
 
 Since 2.5.0
+
+## Hooks
+
+### GHC installation hooks (experimental)
+
+Stack's installation procedure can be fully customized by placing a shell script at
+`~/.stack/hooks/ghc-install.sh` and making it executable.
+
+The script **must** return an exit code of `0` and the standard output **must** be the
+absolute path to the ghc binary that was installed. Otherwise stack will ignore
+the hook and possibly fall back to its own installation procedure.
+
+Hooks are not run when `system-ghc: true`.
+
+When `install-ghc: false`, hooks are still run,
+which allows you to ensure that only your hook will install GHC and stack won't default
+to its own installation logic, even when the hook fails.
+
+An example hook is:
+
+```sh
+#!/bin/sh
+
+set -eu
+
+case $HOOK_GHC_TYPE in
+	bindist)
+		# install GHC here, not printing to stdout, e.g.:
+		#   command install $HOOK_GHC_VERSION >/dev/null
+		;;
+	git)
+		>&2 echo "Hook doesn't support installing from source"
+		exit 1
+		;;
+	*)
+		>&2 echo "Unsupported GHC installation type: $HOOK_GHC_TYPE"
+		exit 2
+		;;
+esac
+		
+echo "location/to/ghc/executable"
+```
+
+The following environment variables are always passed to the hook:
+
+* `HOOK_GHC_TYPE = "bindist" | "git" | "ghcjs"`
+
+For "bindist", additional variables are:
+
+* `HOOK_GHC_VERSION = <ver>`
+
+For "git", additional variables are:
+
+* `HOOK_GHC_COMMIT = <commit>`
+* `HOOK_GHC_FLAVOR = <flavor>`
+
+For "ghcjs", additional variables are:
+
+* `HOOK_GHC_VERSION = <ver>`
+* `HOOK_GHCJS_VERSION = <ver>`
+
+Since 2.8.X
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -34,6 +34,8 @@ Other enhancements:
 * `stack upload` supports authentication with a Hackage API key (via
   `HACKAGE_KEY` environment variable).
 
+* Add GHC installation hooks wrt [#5585](https://github.com/commercialhaskell/stack/pull/5585)
+
 Bug fixes:
 
 * Ensure that `extra-path` works for case-insensitive `PATH`s on Windows.
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -52,7 +52,9 @@ import              Data.List hiding (concat, elem, maximumBy, any)
 import qualified    Data.Map as Map
 import qualified    Data.Set as Set
 import qualified    Data.Text as T
+import qualified    Data.Text.Lazy as TL
 import qualified    Data.Text.Encoding as T
+import qualified    Data.Text.Lazy.Encoding as TL
 import qualified    Data.Text.Encoding.Error as T
 import qualified    Data.Yaml as Yaml
 import              Distribution.System (OS, Arch (..), Platform (..))
@@ -95,7 +97,7 @@ import              Stack.Types.Docker
 import              Stack.Types.SourceMap
 import              Stack.Types.Version
 import qualified    System.Directory as D
-import              System.Environment (getExecutablePath, lookupEnv)
+import              System.Environment (getExecutablePath, lookupEnv, getEnvironment)
 import              System.IO.Error (isPermissionError)
 import              System.FilePath (searchPathSeparator)
 import qualified    System.FilePath as FP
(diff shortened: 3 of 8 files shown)


hasufell:
Any movement here? I'm happy to maintain this feature/codepath.

doc/yaml_configuration.md
```diff
@@ -1200,3 +1200,65 @@ This field is convenient in setups that restrict access to GitHub, for instance
...
+Stack's installation procedure can be fully customized by placing a shell script at
+`~/.stack/hooks/ghc-install.sh` and making it executable.
+
+The script **must** return an exit code of `0` and the standard output **must** be the
+absolute path to the ghc binary that was installed. Otherwise stack will ignore
+the hook and possibly fall back to its own installation procedure.
+
+Hooks are not run when `system-ghc: true`.
+
+When `install-ghc: false`, hooks are still run,
```
chreekat, line 1217: This seems funny: if I don't want GHC to get installed, I don't want a hook up install it, either.

Using implicit config is also a little unnerving (unless stack already has a habit of doing so for hooks?). What if there were two config options, `ghc-install-override: <path/to/script>` and `ghc-install-use-fallback: <bool default True>`?

hasufell: `system-ghc: true` means that stack uses `PATH` to search for ghc, so running the hook makes no sense.

`install-ghc: true` already means that stack falls back to internal installation (e.g. if it can't find a suitable GHC in PATH).

So this perfectly aligns with current config semantics.

chreekat: Ok, we can ignore my suggested changes. :) I'll just try to describe my problems, which I don't think are addressed yet:

(NB: I'm not a regular contributor or maintainer to stack. Just sharing my opinions here.)

##### Problem 1: You're changing the semantics of the "install-ghc" flag.

Currently, `install-ghc: false` means "Do not install ghc". That is very clear and understandable.

It looks like you are modifying the meaning of that flag so that `install-ghc:false` means, "Don't install ghc... unless there is an install hook, in which case *do* install ghc using that hook."

I don't think that's a good change.

##### Problem 2 (Not a blocker): Implicit config

Changing behavior based on the presence and mode of particular files is not one of the good ideas to borrow from git. But I don't hold this position strongly. And I don't know how other users would feel about it.

chreekat: BTW you might be better off just creating a `use-ghcup` option. :D Unless you know of other people that would use the hook?

hasufell: > ##### Problem 1: You're changing the semantics of the "install-ghc" flag.
> 
> Currently, `install-ghc: false` means "Do not install ghc". That is very clear and understandable.
> 

Not really, since the hooks don't necessarily install GHC. The behavior of a hook is only determined by, well, the hook.


> 
> ##### Problem 2 (Not a blocker): Implicit config
> 
> Changing behavior based on the presence and mode of particular files is not one of the good ideas to borrow from git. But I don't hold this position strongly. And I don't know how other users would feel about it.

Pretty much all tools (ghc, stack, cabal, hls) change their behavior depending on presence and mode of files. So I'm not sure I agree.

Thanks for your input though.

mpilgrem:
May I ask what sort of practical problems the availability of this feature is intended to solve? Is it primarily that some users want to use a GHC other than a Stack-supplied one and other than one that is on their PATH? (If so, what prevents those users from adding the path to the desired GHC to the head of their PATH and using `--system-ghc`?) In part, as a Windows user, I am wondering if there are solutions to the main practical problem(s) that do not require a Windows user to have to know about the (MSYS2-supplied) `sh` application.

I appreciate that a hook that is a script is all-flexible/all-powerful, but what troubles me is: could a bad actor not sneak a damaging `ghc-install.hs` file into the directory, and the (possibly naïve) Stack user would not know they were executing it as they used Stack? Many Stack users would never have cause to check the contents of their Stack root.

Would it not be better to specify expressly the script to use on the command line and/or in a Stack configuration file? Stack already has `--setup-info-yaml` to allow a user to change the location where Stack looks for GHC when it needs to install it. Perhaps `--setup-GHC-install=PATH_TO_SCRIPT`?

hasufell:
> May I ask what sort of practical problems the availability of this feature is intended to solve?

Its purpose is for users to be able to decide how GHC is installed. This can be your package manager, can be a nix script or can be ghcup. Stack is not flexible enough currently to cover these use cases. `--system-ghc` is insufficient, because that would require the user first to figure out manually which GHC version stack actually wants (as in: the resolver in use), because otherwise stack will fail the GHC compat check and refuse to run.

Currently, most stack users of haskell-vscode use `--system-ghc` and ghcup, because of two reasons:

1. have only one tool install GHC, not two
2. fix ABI incompatibility issues
    - https://github.com/haskell/vscode-haskell#ghc-abis-dont-match and 
    - https://github.com/haskell/haskell-language-server/issues/2865

This was also the last missing piece during the (failed) unified installer proposal that caused ghcup to not install stack by default, because integration with ghcup was not really possible. This would fix it.

> In part, as a Windows user, I am wondering if there are solutions to the main practical problem(s) that do not require a Windows user to have to know about the (MSYS2-supplied) sh application.

MSYS2 is installed by both stack, ghcup and chocolatey. If you don't have an msys2 toolchain, you won't be able to build a lot of packages, e.g. any package that has a configure script. It can reasonably be assumed that users have it. I won't be implementing powershell solutions or other things. This would complicate the logic and make it non-uniform across platforms.

> could a bad actor not sneak a damaging ghc-install.hs file into the directory, and the (possibly naïve) Stack user would not know they were executing it as they used Stack?

No. If a bad actor can place files on the users system, they can just place a malicious stack binary into PATH.

This approach is used by git itself, many linux package managers and other tools. It's a common design pattern.

> Would it not be better to specify expressly the script to use on the command line and/or in a Stack configuration file?

No, because that would mean any tool (e.g. ghcup) would now have additional complication of trying to figure out if a hook is even installed. This makes the install location somewhat predictable.

It can be debated whether to allow per-project hooks, but that can be solved in a later PR. No one has requested this for now. My suggestion is to not make this unnecessarily complicated for the first round. It can always be adjusted later.

mpilgrem:
Thanks for the explanations. In the spirit of the recent additions to Stack's CONTRIBUTING.md, I'll merge this.

On most users wanting only one tool to install GHC, that is true for me too. I also do not want duplicate copies of GHC on my system; GHC is large and my hard drive is not. So I use only the Stack-supplied GHCs. The same is true for MSYS2 (so I use only the Stack-supplied MSYS2). I have used Stack, VS Code and the Haskell extension for VS Code (or its predecessors) for a long time.

Anecdotally - and briefly, given this Discourse https://discourse.haskell.org/t/get-a-haskell-development-environment-in-three-steps/4822 which looks like it may help me - when the VS Code extension changed recently to prompt for GHCup, I looked up GHCup to see what it was. However, I could not work out how to install it without it also installing a copy of GHC and MSYS2, which seemed to me to duplicate what I knew Stack did and provided. I was also disappointed to read in its User Guide that its basic usage was not available on Windows. So, at that time, I took up the option not to use GHCup with the extension.

hasufell:
Also note that ghcup man easily reuse the stack msys2 installation. The user can supply an alternative location during the install script now. Windows works fine.

hasufell:
Thanks for merging this. As I indicated, I'm happy to maintain this codepath. So ping ce whenever there's an issue with it.

I'll do some follow up patches to ghcup to finalize integration with stack and will ping you for review. Ghcup development happens on GHC GitLab. Do you have an account there?

https://gitlab.haskell.org/haskell/ghcup-hs

hasufell:
Thanks for merging this. As I indicated, I'm happy to maintain this codepath. So ping me whenever there's an issue with it.

I'll do some follow up patches to ghcup to finalize integration with stack and will ping you for review. Ghcup development happens on GHC GitLab. Do you have an account there?

https://gitlab.haskell.org/haskell/ghcup-hs

simonmichael:
Here are the rendered docs for this feature in current stack master:
https://github.com/commercialhaskell/stack/blob/6a7bd5d/doc/yaml_configuration.md#ghc-installation-customisation-experimental

PR #5593, opened 2021-07-26
Fix "Setup configure" not passing --with-ghc{,-pkg}=<..>

hasufell:
This seemed to be a mostly silent bug. It becomes apparent when there's no `ghc-pkg` in PATH or an incompatible one, which is easy to trigger when using `system-ghc: true`. As a consequence `Setup.hs` will be run with a possibly incorrect GHC, causing a straight build/resolution failure that's rather difficult to debug.

The code already tries to pass these flags, so this just fixes already expected behavior.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -9,6 +9,7 @@
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE RankNTypes            #-}
 {-# LANGUAGE TupleSections         #-}
+{-# LANGUAGE TypeApplications      #-}
 -- | Perform a build
 module Stack.Build.Execute
     ( printPlan
@@ -884,8 +885,8 @@ ensureConfig newConfigCache pkgDir ExecuteEnv {..} announce cabal cabalfp task =
         let exes =
               case cpWhich cp of
                 Ghc ->
-                  [ "--with-ghc=" ++ toFilePath (cpCompiler cp)
-                  , "--with-ghc-pkg=" ++ toFilePath pkgPath
+                  [ "--with-ghc=" ++ toFilePath @Abs @File (cpCompiler cp)
+                  , "--with-ghc-pkg=" ++ toFilePath @Abs @File pkgPath
                   ]
         -- Configure cabal with arguments determined by
         -- Stack.Types.Build.configureOpts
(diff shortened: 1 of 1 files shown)


src/Stack/Build/Execute.hs
```diff
@@ -881,21 +881,16 @@ ensureConfig newConfigCache pkgDir ExecuteEnv {..} announce cabal cabalfp task =
...
         let (GhcPkgExe pkgPath) = cpPkg cp
-        let programNames =
+        let exes =
               case cpWhich cp of
                 Ghc ->
                   [ "--with-ghc=" ++ toFilePath (cpCompiler cp)
                   , "--with-ghc-pkg=" ++ toFilePath pkgPath
                   ]
-        exes <- forM programNames $ \name -> do
-            mpath <- findExecutable name
```
hasufell, line 891: Here it'll run `findExecutable "--with-ghc=/path/to/ghc"`, which obviously will not return anything. Both filepaths are already absolute filepaths (on type level), so there's no reason to resolve anything here.

Haskell-mouse: I've made an alternative solution where I leave the `findExecutable` checking. I've just removed adding `--with-ghc/ghcpkg=` to the path, which obviously forces to fail the check: #5597

hasufell: I'm not sure if `findExecutable` on an absolute Path makes much sense (it probably just degrades to`doesFileExist` and checking the exec bit).

Moreover: if those paths don't point to valid ghc/ghci, you'd want proper failure, not hiding it by omitting the `--with-ghc` flag, no?

int-index: > I'm not sure if `findExecutable` on an absolute Path

Does it have to be an absolute path, though?

hasufell: What do you mean? It **is** an absolute path always in this part of the code. That's ensured on type level.

Haskell-mouse: > I'm not sure if `findExecutable` on an absolute Path makes much sense (it probably just degrades to`doesFileExist` and checking the exec bit).

I think that checking the exec bit makes the error message more convenient for a user. 
What message will see the user if he runs a file without exec bit?

> Moreover: if those paths don't point to valid ghc/ghc-pkg, you'd want proper failure, not hiding it by omitting the `--with-ghc` flag, no?
> 
> One reason this bug was so hard to find is exactly the silent failure that causes other toolchain issues down the line.

I agree and in case of proper error messaging your patch is better.
In another case, I would throw an exception here.

snoyberg:
Closing in favor of #5597

PR #5597, opened 2021-08-02
Fixed a bug with ignoring `cpPkg` and `cpCompiler` fields.

Haskell-mouse:
It looks like a user can't override location `ghc` and `ghc-pkg`, because these options are ignored in  `ensureConfig`. 
I prefer to leave `findExecutable` checking and I've just removed obviously mistaken adding "--with-ghc/ghc-pkg=" to the path.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,6 +17,8 @@ Bug fixes:
 
 * Ensure that `extra-path` works for case-insensitive `PATH`s on Windows.
   See [rio#237](https://github.com/commercialhaskell/rio/pull/237)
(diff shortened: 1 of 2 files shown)


snoyberg (changes requested):
Looks great, thanks! Would you mind adding a ChangeLog entry for this bug fix?

Haskell-mouse:
> Looks great, thanks! Would you mind adding a ChangeLog entry for this bug fix?

Thanks. I've added a short note to the `Bug fixes` section.

snoyberg (approved):

snoyberg:
Thanks!

Issue #5578, opened 2021-07-04
Unnecessary configuring on (almost?) any file change


snoyberg:
I've noticed that almost any time I have a file change in a project, `stack build` will run the Cabal configure step, slowing down compilations. I've traced this down to some issue with `stack-setup-config-mod`, and it appears that the file we're checking for dirtiness does not exist, triggering regular rebuilds.

I did a blame, and I think this comes down to this commit:

https://github.com/commercialhaskell/stack/commit/681ca331609e555d62e9aacbbc18770f81abd8de

It looks to me like `stack-setup-config-mod` is never being created in this commit. If I'm not mistaken, we need to add an equivalent to the `writeCabalMod` call that appears in this module.

CC @qrilka, I can take a stab at a PR, but input on this would be appreciated.

qrilka:
I do remember this and I thought that fixed this in some follow-up PR but it looks that the fix didn't land on github somehow, that's odd...

PR #5613, opened 2021-08-27
Fix for local dependency build issue

drone-rites:
Fix for bug where a local dependency that has been built from a local consumer project would be installed in the package db for the consumer project but not in the local dependency's own package db. This results in an error where ghc-pkg is unable to find the cached local package in its own package db and the build fails. This can be resolved by running `stack clean` and then `stack build` for the local dependency, but then a build for the consumer of the package will fail as it tries to find the local dependency in its own package db. The only fix is then to again run `stack clean` for the local dependency and then run `stack build` for the consumer of the dependency.

This fix is extremely simple: remove one line that modifies the setup-config file and causes the trouble. A more satisfying fix for this issue may be to figure out how to adjust the package db search so that cached builds can be used both by the local package and any consumers of the local package.

I'm happy to provide a simple repro case and a short message for the change log.

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -907,11 +907,6 @@ ensureConfig newConfigCache pkgDir ExecuteEnv {..} announce cabal cabalfp task =
             TTLocalMutable{} -> writeConfigCache pkgDir newConfigCache
             TTRemotePackage{} -> return ()
         writeCabalMod pkgDir newCabalMod
-        -- This file gets updated one more time by the configure step, so get
-        -- the most recent value. We could instead change our logic above to
-        -- check if our config mod file is newer than the file above, but this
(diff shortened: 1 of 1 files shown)


borsboom:
@drone-rites Thank you for this PR!  By coincidence I was working on the the same bug when you filed #5680 and this PR was a great hint about where the problem was.  Upon closer inspection it looks like this PR basically reverts the meat of f64188016aea9fb34554f29858423aec0860ae4e (which I confirmed introduced this regression), so I think it would reintroduce the problem that commit was trying to solve (#5578).  Therefore I'm going to close this PR.

However, I've opened a new PR #5682 that I believe has a more complete fix; it basically keeps track of the last project from which a package was built and reconfigures the package if the project changes.  Can you test #5682 to confirm that it also fixes your use case?

PR #5622, opened 2021-09-29
Add FreeBSD aarch64 support.

MikaelUrankar:


hasufell:
We don't even have GHC bindists for this... where do you intend to get them from?

MikaelUrankar:
Here:
https://github.com/commercialhaskell/stackage-content/pull/94

PR #5631, opened 2021-11-05
stack upload: Support authentication with Hackage API key

### LINKED ISSUE #5513: Support API Keys in stack-upload
Hackage supports API Key based authentication (see `https://hackage.haskell.org/user/{username}/manage`), specifically for uploading packages from CI. These keys can be created, named, and revoked specifically, so that you don't have to store your (only) Username/Password in CI configuration.

I'd like to see `stack-upload` support this, perhaps by another env var and/or extension to the `credentials.json` file?

Here is a simplified example request that I'm using for now:

```console
        stack sdist ...
        curl \
            --silent \
            --header "Accept: text/plain" \
            --header "Authorization: X-ApiKey $HACKAGE_API_KEY" \
            --form "package=@$archive" \
            https://hackage.haskell.org/packages/
```

In reality, I want to output the response body but still catch unsuccessful exit codes, which `curl` doesn't make easy. I also have to go look for the `sdist`-produced archive. These two points actually make it quite a bit more complicated to do from the outside vs `stack upload` doing it internally.

kocielnik:
Closes #5513

Usage example:

```bash
$ HACKAGE_KEY=foobar stack upload .
```

> Please also shortly describe how you tested your change.

`stack test` was all green.

Tested scenarios:

1. Key set but wrong: Tried to upload the package with wrong API key set. `stack upload` reported finding the key, and then bounced off Hackage authentication as expected.
2. Key set and correct: Successfully uploaded a package candidate to Hackage. Removed the candidate afterwards.
3. Key not set: `stack upload` asked for username and password. I gave these (correct ones), and uploaded a package candidate to Hackage. Removed the candidate afterwards.

> Bonus points for added tests!

Included a test for the function detecting a set `HACKAGE_KEY` env var.

I want to thank @pbrisbin here for helping me get back on track with this after I got lost trying to understand what GHC was expecting me to write:)

### DIFF
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -81,7 +81,7 @@ maybeGetHackageKey = lookupEnv (T.unpack "HACKAGE_KEY")
 getCredsWithApiKey :: String -> FilePath -> HackageCreds
 getCredsWithApiKey key fp = HackageCreds {
     hcUsername = ""
-  , hcPassword = fromString key
+  , hcPassword = T.pack key
   , hcCredsFile = fp
 }
 
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -49,7 +49,6 @@ import           System.Directory                      (createDirectoryIfMissing
                                                         removeFile, renameFile)
 import           System.Environment                    (lookupEnv)
 import           System.FilePath                       ((</>), takeFileName, takeDirectory)
-import           System.IO                             (putStrLn, putStr, print) -- TODO remove putStrLn, use logInfo
 import           System.PosixCompat.Files              (setFileMode)
 
 
@@ -81,16 +80,16 @@ instance FromJSON (FilePath -> HackageCreds) where
 withEnvVariable :: Text -> IO Text -> IO Text
 withEnvVariable varName fromPrompt = lookupEnv (T.unpack varName) >>= maybe fromPrompt (pure . T.pack)
 
-maybeGetHackageKey :: IO (Maybe HackageKey)
-maybeGetHackageKey = fmap (HackageKey . T.pack) <$> lookupEnv (T.unpack "HACKAGE_KEY")
+maybeGetHackageKey :: RIO m (Maybe HackageKey)
+maybeGetHackageKey = fmap (HackageKey . T.pack) <$> (liftIO $ lookupEnv (T.unpack "HACKAGE_KEY"))
 
 
-loadCreds :: Config -> IO HackageAuth
+loadCreds :: HasLogFunc m => Config -> RIO m HackageAuth
 loadCreds config = do
   maybeHackageKey <- maybeGetHackageKey
   case maybeHackageKey of
     Just key -> do
-      putStrLn "HACKAGE_KEY found in env, using that for credentials."
+      logInfo "HACKAGE_KEY found in env, using that for credentials."
       return $ HAKey key
     Nothing -> HACreds <$> loadUserAndPassword config
 
@@ -98,10 +97,10 @@ loadCreds config = do
 -- line.
 --
 -- Since 0.1.0.0
-loadUserAndPassword :: Config -> IO HackageCreds
+loadUserAndPassword :: HasLogFunc m => Config -> RIO m HackageCreds
 loadUserAndPassword config = do
-  fp <- credsFile config
-  elbs <- tryIO $ L.readFile fp
+  fp <- liftIO $ credsFile config
+  elbs <- liftIO $ tryIO $ L.readFile fp
   case either (const Nothing) Just elbs >>= \lbs -> (lbs, ) <$> decode' lbs of
     Nothing -> fromPrompt fp
     Just (lbs, mkCreds) -> do
@@ -110,14 +109,15 @@ loadUserAndPassword config = do
       writeFilePrivate fp $ lazyByteString lbs
 
       unless (configSaveHackageCreds config) $ do
-        putStrLn "WARNING: You've set save-hackage-creds to false"
-        putStrLn "However, credentials were found at:"
-        putStrLn $ "  " ++ fp
+        logWarn "WARNING: You've set save-hackage-creds to false"
+        logWarn "However, credentials were found at:"
+        logWarn $ "  " <> fromString fp
       return $ mkCreds fp
   where
+    fromPrompt :: HasLogFunc m => FilePath -> RIO m HackageCreds
     fromPrompt fp = do
-      username <- withEnvVariable "HACKAGE_USERNAME" (prompt "Hackage username: ")
-      password <- withEnvVariable "HACKAGE_PASSWORD" (promptPassword "Hackage password: ")
+      username <- liftIO $ withEnvVariable "HACKAGE_USERNAME" (prompt "Hackage username: ")
+      password <- liftIO $ withEnvVariable "HACKAGE_PASSWORD" (promptPassword "Hackage password: ")
       let hc = HackageCreds
             { hcUsername = username
             , hcPassword = password
@@ -127,10 +127,10 @@ loadUserAndPassword config = do
       when (configSaveHackageCreds config) $ do
         shouldSave <- promptBool $ T.pack $
           "Save hackage credentials to file at " ++ fp ++ " [y/n]? "
-        putStrLn "NOTE: Avoid this prompt in the future by using: save-hackage-creds: false"
+        logInfo "NOTE: Avoid this prompt in the future by using: save-hackage-creds: false"
         when shouldSave $ do
           writeFilePrivate fp $ fromEncoding $ toEncoding hc
-          putStrLn "Saved!"
+          logInfo "Saved!"
           hFlush stdout
 
       return hc
@@ -167,26 +167,26 @@ addAPIKey :: HackageKey -> Request -> Request
 addAPIKey (HackageKey key) req =
   setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ T.unpack key] req
 
-applyKeyOrCreds :: HackageAuth -> Request -> IO Request
+applyKeyOrCreds :: HasLogFunc m => HackageAuth -> Request -> RIO m Request
 applyKeyOrCreds haAuth req0 = do
     case haAuth of
         HAKey key -> return (addAPIKey key req0)
         HACreds creds -> applyCreds creds req0
 
-applyCreds :: HackageCreds -> Request -> IO Request
+applyCreds :: HasLogFunc m => HackageCreds -> Request -> RIO m Request
 applyCreds creds req0 = do
-  manager <- getGlobalManager
-  ereq <- applyDigestAuth
+  manager <- liftIO $ getGlobalManager
+  ereq <- liftIO $ applyDigestAuth
     (encodeUtf8 $ hcUsername creds)
     (encodeUtf8 $ hcPassword creds)
     req0
     manager
   case ereq of
       Left e -> do
-          putStrLn "WARNING: No HTTP digest prompt found, this will probably fail"
+          logWarn "WARNING: No HTTP digest prompt found, this will probably fail"
           case fromException e of
-              Just e' -> putStrLn $ displayDigestAuthException e'
-              Nothing -> print e
+              Just e' -> logWarn $ fromString (displayDigestAuthException e')
+              Nothing -> logWarn $ fromString $ displayException e
           return req0
       Right req -> return req
 
@@ -194,12 +194,13 @@ applyCreds creds req0 = do
 -- sending a file like 'upload', this sends a lazy bytestring.
 --
 -- Since 0.1.2.1
-uploadBytes :: String -- ^ Hackage base URL
+uploadBytes :: HasLogFunc m
+            => String -- ^ Hackage base URL
             -> HackageAuth
             -> String -- ^ tar file name
             -> UploadVariant
             -> L.ByteString -- ^ tar file contents
-            -> IO ()
+            -> RIO m ()
 uploadBytes baseUrl auth tarName uploadVariant bytes = do
     let req1 = setRequestHeader "Accept" ["text/plain"]
                (fromString $ baseUrl
@@ -209,33 +210,36 @@ uploadBytes baseUrl auth tarName uploadVariant bytes = do
                                Candidate -> "candidates/"
                )
         formData = [partFileRequestBody "package" tarName (RequestBodyLBS bytes)]
-    req2 <- formDataBody formData req1
+    req2 <- liftIO $ formDataBody formData req1
     req3 <- applyKeyOrCreds auth req2
-    putStr $ "Uploading " ++ tarName ++ "... "
+    logInfo $ "Uploading " <> fromString tarName <> "... "
     hFlush stdout
-    withResponse req3 $ \res ->
+    withRunInIO $ \runInIO -> withResponse req3 (runInIO . inner)
+ where
+    inner :: HasLogFunc m => Response (ConduitM () S.ByteString IO ()) -> RIO m ()
+    inner res =
         case getResponseStatusCode res of
-            200 -> putStrLn "done!"
+            200 -> logInfo "done!"
             401 -> do
-                putStrLn "authentication failure"
+                logError "authentication failure"
                 case auth of
-                  HACreds creds -> handleIO (const $ return ()) (removeFile (hcCredsFile creds))
+                  HACreds creds -> handleIO (const $ return ()) (liftIO $ removeFile (hcCredsFile creds))
                   _ -> pure ()
                 throwString "Authentication failure uploading to server"
             403 -> do
-                putStrLn "forbidden upload"
-                putStrLn "Usually means: you've already uploaded this package/version combination"
-                putStrLn "Ignoring error and continuing, full message from Hackage below:\n"
-                printBody res
+                logError "forbidden upload"
+                logError "Usually means: you've already uploaded this package/version combination"
+                logError "Ignoring error and continuing, full message from Hackage below:\n"
+                liftIO $ printBody res
             503 -> do
-                putStrLn "service unavailable"
-                putStrLn "This error some times gets sent even though the upload succeeded"
-                putStrLn "Check on Hackage to see if your pacakge is present"
-                printBody res
+                logError "service unavailable"
+                logError "This error some times gets sent even though the upload succeeded"
+                logError "Check on Hackage to see if your pacakge is present"
+                liftIO $ printBody res
             code -> do
-                putStrLn $ "unhandled status code: " ++ show code
-                printBody res
-                throwString $ "Upload failed on " ++ tarName
+                logError $ "unhandled status code: " <> fromString (show code)
+                liftIO $ printBody res
+                throwString $ "Upload failed on " <> fromString tarName
 
 printBody :: Response (ConduitM () S.ByteString IO ()) -> IO ()
 printBody res = runConduit $ getResponseBody res .| CB.sinkHandle stdout
@@ -243,19 +247,21 @@ printBody res = runConduit $ getResponseBody res .| CB.sinkHandle stdout
 -- | Upload a single tarball with the given @Uploader@.
 --
 -- Since 0.1.0.0
-upload :: String -- ^ Hackage base URL
+upload :: HasLogFunc m
+       => String -- ^ Hackage base URL
        -> HackageAuth
        -> FilePath
        -> UploadVariant
-       -> IO ()
+       -> RIO m ()
 upload baseUrl auth fp uploadVariant =
-  uploadBytes baseUrl auth (takeFileName fp) uploadVariant =<< L.readFile fp
+  uploadBytes baseUrl auth (takeFileName fp) uploadVariant =<< (liftIO $ L.readFile fp)
 
-uploadRevision :: String -- ^ Hackage base URL
+uploadRevision :: HasLogFunc m
+               => String -- ^ Hackage base URL
                -> HackageAuth
                -> PackageIdentifier
                -> L.ByteString
-               -> IO ()
+               -> RIO m ()
 uploadRevision baseUrl auth ident@(PackageIdentifier name _) cabalFile = do
   req0 <- parseRequest $ concat
     [ baseUrl
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -12,7 +12,7 @@ module Stack.Upload
     , uploadRevision
       -- * Credentials
     , HackageCreds
-    , loadCreds
+    , loadAuth
     , writeFilePrivate
       -- * Internal
     , maybeGetHackageKey
@@ -84,8 +84,8 @@ maybeGetHackageKey :: RIO m (Maybe HackageKey)
 maybeGetHackageKey = fmap (HackageKey . T.pack) <$> (liftIO $ lookupEnv (T.unpack "HACKAGE_KEY"))
 
 
-loadCreds :: HasLogFunc m => Config -> RIO m HackageAuth
-loadCreds config = do
+loadAuth :: HasLogFunc m => Config -> RIO m HackageAuth
+loadAuth config = do
   maybeHackageKey <- maybeGetHackageKey
   case maybeHackageKey of
     Just key -> do
@@ -167,8 +167,8 @@ addAPIKey :: HackageKey -> Request -> Request
 addAPIKey (HackageKey key) req =
   setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ T.unpack key] req
 
-applyKeyOrCreds :: HasLogFunc m => HackageAuth -> Request -> RIO m Request
-applyKeyOrCreds haAuth req0 = do
+applyAuth :: HasLogFunc m => HackageAuth -> Request -> RIO m Request
+applyAuth haAuth req0 = do
     case haAuth of
         HAKey key -> return (addAPIKey key req0)
         HACreds creds -> applyCreds creds req0
@@ -211,7 +211,7 @@ uploadBytes baseUrl auth tarName uploadVariant bytes = do
                )
         formData = [partFileRequestBody "package" tarName (RequestBodyLBS bytes)]
     req2 <- liftIO $ formDataBody formData req1
-    req3 <- applyKeyOrCreds auth req2
+    req3 <- applyAuth auth req2
     logInfo $ "Uploading " <> fromString tarName <> "... "
     hFlush stdout
     withRunInIO $ \runInIO -> withResponse req3 (runInIO . inner)
@@ -276,5 +276,5 @@ uploadRevision baseUrl auth ident@(PackageIdentifier name _) cabalFile = do
     , partBS "publish" "on"
     ]
     req0
-  req2 <- applyKeyOrCreds auth req1
+  req2 <- applyAuth auth req1
   void $ httpNoBody req2
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -52,6 +52,9 @@ import           System.FilePath                       ((</>), takeFileName, tak
 import           System.IO                             (putStrLn, putStr, print) -- TODO remove putStrLn, use logInfo
 import           System.PosixCompat.Files              (setFileMode)
 
+
+newtype HackageKey = HackageKey Text
+
 -- | Username and password to log into Hackage.
 --
 -- Since 0.1.0.0
@@ -61,6 +64,9 @@ data HackageCreds = HackageCreds
     , hcCredsFile :: !FilePath
     }
     deriving Show
+
+data HackageAuth = HAKey HackageKey
+                 | HACreds HackageCreds
 
 instance ToJSON HackageCreds where
     toJSON (HackageCreds u p _) = object
@@ -75,25 +81,18 @@ instance FromJSON (FilePath -> HackageCreds) where
 withEnvVariable :: Text -> IO Text -> IO Text
 withEnvVariable varName fromPrompt = lookupEnv (T.unpack varName) >>= maybe fromPrompt (pure . T.pack)
 
-maybeGetHackageKey :: IO (Maybe String)
-maybeGetHackageKey = lookupEnv (T.unpack "HACKAGE_KEY")
+maybeGetHackageKey :: IO (Maybe HackageKey)
+maybeGetHackageKey = fmap (HackageKey . T.pack) <$> lookupEnv (T.unpack "HACKAGE_KEY")
 
-getCredsWithApiKey :: String -> FilePath -> HackageCreds
-getCredsWithApiKey key fp = HackageCreds {
-    hcUsername = ""
-  , hcPassword = T.pack key
-  , hcCredsFile = fp
-}
 
-loadCreds :: Config -> IO HackageCreds
+loadCreds :: Config -> IO HackageAuth
 loadCreds config = do
   maybeHackageKey <- maybeGetHackageKey
   case maybeHackageKey of
     Just key -> do
       putStrLn "HACKAGE_KEY found in env, using that for credentials."
-      fp <- credsFile config
-      return $ getCredsWithApiKey key fp
-    Nothing -> loadUserAndPassword config
+      return $ HAKey key
+    Nothing -> HACreds <$> loadUserAndPassword config
 
 -- | Load Hackage credentials, either from a save file or the command
 -- line.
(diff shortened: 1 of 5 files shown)


src/Stack/Upload.hs
```diff
@@ -67,12 +75,31 @@ instance FromJSON (FilePath -> HackageCreds) where
...
+  , hcPassword = fromString key
+  , hcCredsFile = fp
+}
+
+loadCreds :: Config -> IO HackageCreds
+loadCreds config = do
+  maybeHackageKey <- maybeGetHackageKey
+  case maybeHackageKey of
+    Just key -> do
+      putStrLn "HACKAGE_KEY found in env, using that for credentials."
```
hasufell, line 93: should probably be `logInfo`?

src/Stack/Upload.hs
```diff
@@ -67,12 +75,31 @@ instance FromJSON (FilePath -> HackageCreds) where
 withEnvVariable :: Text -> IO Text -> IO Text
 withEnvVariable varName fromPrompt = lookupEnv (T.unpack varName) >>= maybe fromPrompt (pure . T.pack)
 
+maybeGetHackageKey :: IO (Maybe String)
+maybeGetHackageKey = lookupEnv (T.unpack "HACKAGE_KEY")
+
+getCredsWithApiKey :: String -> FilePath -> IO HackageCreds
```
hasufell, line 81: Why does this run in `IO`?

src/Stack/Upload.hs
```diff
@@ -67,12 +75,31 @@ instance FromJSON (FilePath -> HackageCreds) where
...
+getCredsWithApiKey key fp = return HackageCreds {
+    hcUsername = ""
+  , hcPassword = fromString key
```
hasufell, line 84:
```suggestion
  , hcPassword = T.pack key
```

src/Stack/Upload.hs
```diff
@@ -136,6 +163,16 @@ credsFile config = do
...
     return $ dir </> "credentials.json"
 
+addAPIKey :: String -> Request -> Request
+addAPIKey key req =
+  setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ key] req
+
+applyKeyOrCreds :: HackageCreds -> Request -> IO Request
+applyKeyOrCreds creds req0 = do
+    case hcUsername creds of
+        "" -> return (addAPIKey (T.unpack $ hcPassword creds) req0)
```
hasufell, line 173: I feel this is a bit wonky... we're re-using the `HackageCreds` type here and stuff the api key into the password field if the username is an empty string.

It seems it's more natural to provide a proper sum type here or use something like `Either HackageCreds HackageAPIKey`, where `newtype HackageAPIKey = HackageAPIKey Text`, or so.

src/Stack/Upload.hs
kocielnik: Right. I tried that but hit an error I couldn't solve. Perhaps you could help me get this straight. The error was:

```haskell
No instance for (MonadReader env0 IO)
        arising from a use of ‘logInfo’
-- (...)
93 |       logInfo "HACKAGE_KEY found in env, using that for credentials."
```

1. Importing `RIO`, as described [here](https://www.fpcomplete.com/haskell/library/rio/) didn't help,
2. Stack.Prelude is already imported in `Upload`, as in some other modules using `logInfo`,
3. `test/integration/lib/StackTest.hs` defines its own internal `logInfo`, but I suppose that's not the approach you have in mind here.

src/Stack/Upload.hs
kocielnik: Good point, eradicated that `IO`!

src/Stack/Upload.hs
hasufell:
```diff
diff --git a/src/Stack/Upload.hs b/src/Stack/Upload.hs
index 66ba4102..4a24daa1 100644
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -52,6 +52,9 @@ import           System.FilePath                       ((</>), takeFileName, tak
 import           System.IO                             (putStrLn, putStr, print) -- TODO remove putStrLn, use logInfo
 import           System.PosixCompat.Files              (setFileMode)
 
+
+newtype HackageKey = HackageKey Text
+
 -- | Username and password to log into Hackage.
 --
 -- Since 0.1.0.0
@@ -62,6 +65,9 @@ data HackageCreds = HackageCreds
     }
     deriving Show
 
+data HackageAuth = HAKey HackageKey
+                 | HACreds HackageCreds
+
 instance ToJSON HackageCreds where
     toJSON (HackageCreds u p _) = object
         [ "username" .= u
@@ -75,25 +81,18 @@ instance FromJSON (FilePath -> HackageCreds) where
 withEnvVariable :: Text -> IO Text -> IO Text
 withEnvVariable varName fromPrompt = lookupEnv (T.unpack varName) >>= maybe fromPrompt (pure . T.pack)
 
-maybeGetHackageKey :: IO (Maybe String)
-maybeGetHackageKey = lookupEnv (T.unpack "HACKAGE_KEY")
+maybeGetHackageKey :: IO (Maybe HackageKey)
+maybeGetHackageKey = fmap (HackageKey . T.pack) <$> lookupEnv (T.unpack "HACKAGE_KEY")
 
-getCredsWithApiKey :: String -> FilePath -> HackageCreds
-getCredsWithApiKey key fp = HackageCreds {
-    hcUsername = ""
-  , hcPassword = T.pack key
-  , hcCredsFile = fp
-}
 
-loadCreds :: Config -> IO HackageCreds
+loadCreds :: Config -> IO HackageAuth
 loadCreds config = do
   maybeHackageKey <- maybeGetHackageKey
   case maybeHackageKey of
     Just key -> do
       putStrLn "HACKAGE_KEY found in env, using that for credentials."
-      fp <- credsFile config
-      return $ getCredsWithApiKey key fp
-    Nothing -> loadUserAndPassword config
+      return $ HAKey key
+    Nothing -> HACreds <$> loadUserAndPassword config
 
 -- | Load Hackage credentials, either from a save file or the command
 -- line.
@@ -164,15 +163,15 @@ credsFile config = do
     createDirectoryIfMissing True dir
     return $ dir </> "credentials.json"
 
-addAPIKey :: String -> Request -> Request
-addAPIKey key req =
-  setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ key] req
+addAPIKey :: HackageKey -> Request -> Request
+addAPIKey (HackageKey key) req =
+  setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ T.unpack key] req
 
-applyKeyOrCreds :: HackageCreds -> Request -> IO Request
-applyKeyOrCreds creds req0 = do
-    case hcUsername creds of
-        "" -> return (addAPIKey (T.unpack $ hcPassword creds) req0)
-        _ -> applyCreds creds req0
+applyKeyOrCreds :: HackageAuth -> Request -> IO Request
+applyKeyOrCreds haAuth req0 = do
+    case haAuth of
+        HAKey key -> return (addAPIKey key req0)
+        HACreds creds -> applyCreds creds req0
 
 applyCreds :: HackageCreds -> Request -> IO Request
 applyCreds creds req0 = do
@@ -196,12 +195,12 @@ applyCreds creds req0 = do
 --
 -- Since 0.1.2.1
 uploadBytes :: String -- ^ Hackage base URL
-            -> HackageCreds
+            -> HackageAuth
             -> String -- ^ tar file name
             -> UploadVariant
             -> L.ByteString -- ^ tar file contents
             -> IO ()
-uploadBytes baseUrl creds tarName uploadVariant bytes = do
+uploadBytes baseUrl auth tarName uploadVariant bytes = do
     let req1 = setRequestHeader "Accept" ["text/plain"]
                (fromString $ baseUrl
                           <> "packages/"
@@ -211,7 +210,7 @@ uploadBytes baseUrl creds tarName uploadVariant bytes = do
                )
         formData = [partFileRequestBody "package" tarName (RequestBodyLBS bytes)]
     req2 <- formDataBody formData req1
-    req3 <- applyKeyOrCreds creds req2
+    req3 <- applyKeyOrCreds auth req2
     putStr $ "Uploading " ++ tarName ++ "... "
     hFlush stdout
     withResponse req3 $ \res ->
@@ -219,7 +218,9 @@ uploadBytes baseUrl creds tarName uploadVariant bytes = do
             200 -> putStrLn "done!"
             401 -> do
                 putStrLn "authentication failure"
-                handleIO (const $ return ()) (removeFile (hcCredsFile creds))
+                case auth of
+                  HACreds creds -> handleIO (const $ return ()) (removeFile (hcCredsFile creds))
+                  _ -> pure ()
                 throwString "Authentication failure uploading to server"
             403 -> do
                 putStrLn "forbidden upload"
@@ -243,19 +244,19 @@ printBody res = runConduit $ getResponseBody res .| CB.sinkHandle stdout
 --
 -- Since 0.1.0.0
 upload :: String -- ^ Hackage base URL
-       -> HackageCreds
+       -> HackageAuth
        -> FilePath
        -> UploadVariant
        -> IO ()
-upload baseUrl creds fp uploadVariant =
-  uploadBytes baseUrl creds (takeFileName fp) uploadVariant =<< L.readFile fp
+upload baseUrl auth fp uploadVariant =
+  uploadBytes baseUrl auth (takeFileName fp) uploadVariant =<< L.readFile fp
 
 uploadRevision :: String -- ^ Hackage base URL
-               -> HackageCreds
+               -> HackageAuth
                -> PackageIdentifier
                -> L.ByteString
                -> IO ()
-uploadRevision baseUrl creds ident@(PackageIdentifier name _) cabalFile = do
+uploadRevision baseUrl auth ident@(PackageIdentifier name _) cabalFile = do
   req0 <- parseRequest $ concat
     [ baseUrl
     , "package/"
@@ -269,5 +270,5 @@ uploadRevision baseUrl creds ident@(PackageIdentifier name _) cabalFile = do
     , partBS "publish" "on"
     ]
     req0
-  req2 <- applyKeyOrCreds creds req1
+  req2 <- applyKeyOrCreds auth req1
   void $ httpNoBody req2
```

src/Stack/Upload.hs
hasufell:
```diff
diff --git a/src/Stack/Upload.hs b/src/Stack/Upload.hs
index 66ba4102..1b8918df 100644
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -49,9 +49,11 @@ import           System.Directory                      (createDirectoryIfMissing
                                                         removeFile, renameFile)
 import           System.Environment                    (lookupEnv)
 import           System.FilePath                       ((</>), takeFileName, takeDirectory)
-import           System.IO                             (putStrLn, putStr, print) -- TODO remove putStrLn, use logInfo
 import           System.PosixCompat.Files              (setFileMode)
 
+
+newtype HackageKey = HackageKey Text
+
 -- | Username and password to log into Hackage.
 --
 -- Since 0.1.0.0
@@ -62,6 +64,9 @@ data HackageCreds = HackageCreds
     }
     deriving Show
 
+data HackageAuth = HAKey HackageKey
+                 | HACreds HackageCreds
+
 instance ToJSON HackageCreds where
     toJSON (HackageCreds u p _) = object
         [ "username" .= u
@@ -75,34 +80,27 @@ instance FromJSON (FilePath -> HackageCreds) where
 withEnvVariable :: Text -> IO Text -> IO Text
 withEnvVariable varName fromPrompt = lookupEnv (T.unpack varName) >>= maybe fromPrompt (pure . T.pack)
 
-maybeGetHackageKey :: IO (Maybe String)
-maybeGetHackageKey = lookupEnv (T.unpack "HACKAGE_KEY")
+maybeGetHackageKey :: RIO m (Maybe HackageKey)
+maybeGetHackageKey = fmap (HackageKey . T.pack) <$> (liftIO $ lookupEnv (T.unpack "HACKAGE_KEY"))
 
-getCredsWithApiKey :: String -> FilePath -> HackageCreds
-getCredsWithApiKey key fp = HackageCreds {
-    hcUsername = ""
-  , hcPassword = T.pack key
-  , hcCredsFile = fp
-}
 
-loadCreds :: Config -> IO HackageCreds
+loadCreds :: HasLogFunc m => Config -> RIO m HackageAuth
 loadCreds config = do
   maybeHackageKey <- maybeGetHackageKey
   case maybeHackageKey of
     Just key -> do
-      putStrLn "HACKAGE_KEY found in env, using that for credentials."
-      fp <- credsFile config
-      return $ getCredsWithApiKey key fp
-    Nothing -> loadUserAndPassword config
+      logInfo "HACKAGE_KEY found in env, using that for credentials."
+      return $ HAKey key
+    Nothing -> HACreds <$> loadUserAndPassword config
 
 -- | Load Hackage credentials, either from a save file or the command
 -- line.
 --
 -- Since 0.1.0.0
-loadUserAndPassword :: Config -> IO HackageCreds
+loadUserAndPassword :: HasLogFunc m => Config -> RIO m HackageCreds
 loadUserAndPassword config = do
-  fp <- credsFile config
-  elbs <- tryIO $ L.readFile fp
+  fp <- liftIO $ credsFile config
+  elbs <- liftIO $ tryIO $ L.readFile fp
   case either (const Nothing) Just elbs >>= \lbs -> (lbs, ) <$> decode' lbs of
     Nothing -> fromPrompt fp
     Just (lbs, mkCreds) -> do
@@ -111,14 +109,15 @@ loadUserAndPassword config = do
       writeFilePrivate fp $ lazyByteString lbs
 
       unless (configSaveHackageCreds config) $ do
-        putStrLn "WARNING: You've set save-hackage-creds to false"
-        putStrLn "However, credentials were found at:"
-        putStrLn $ "  " ++ fp
+        logInfo "WARNING: You've set save-hackage-creds to false"
+        logInfo "However, credentials were found at:"
+        logInfo $ "  " <> fromString fp
       return $ mkCreds fp
   where
+    fromPrompt :: HasLogFunc m => FilePath -> RIO m HackageCreds
     fromPrompt fp = do
-      username <- withEnvVariable "HACKAGE_USERNAME" (prompt "Hackage username: ")
-      password <- withEnvVariable "HACKAGE_PASSWORD" (promptPassword "Hackage password: ")
+      username <- liftIO $ withEnvVariable "HACKAGE_USERNAME" (prompt "Hackage username: ")
+      password <- liftIO $ withEnvVariable "HACKAGE_PASSWORD" (promptPassword "Hackage password: ")
       let hc = HackageCreds
             { hcUsername = username
             , hcPassword = password
@@ -128,10 +127,10 @@ loadUserAndPassword config = do
       when (configSaveHackageCreds config) $ do
         shouldSave <- promptBool $ T.pack $
           "Save hackage credentials to file at " ++ fp ++ " [y/n]? "
-        putStrLn "NOTE: Avoid this prompt in the future by using: save-hackage-creds: false"
+        logInfo "NOTE: Avoid this prompt in the future by using: save-hackage-creds: false"
         when shouldSave $ do
           writeFilePrivate fp $ fromEncoding $ toEncoding hc
-          putStrLn "Saved!"
+          logInfo "Saved!"
           hFlush stdout
 
       return hc
@@ -164,30 +163,30 @@ credsFile config = do
     createDirectoryIfMissing True dir
     return $ dir </> "credentials.json"
 
-addAPIKey :: String -> Request -> Request
-addAPIKey key req =
-  setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ key] req
+addAPIKey :: HackageKey -> Request -> Request
+addAPIKey (HackageKey key) req =
+  setRequestHeader "Authorization" [fromString $ "X-ApiKey" ++ " " ++ T.unpack key] req
 
-applyKeyOrCreds :: HackageCreds -> Request -> IO Request
-applyKeyOrCreds creds req0 = do
-    case hcUsername creds of
-        "" -> return (addAPIKey (T.unpack $ hcPassword creds) req0)
-        _ -> applyCreds creds req0
+applyKeyOrCreds :: HasLogFunc m => HackageAuth -> Request -> RIO m Request
+applyKeyOrCreds haAuth req0 = do
+    case haAuth of
+        HAKey key -> return (addAPIKey key req0)
+        HACreds creds -> applyCreds creds req0
 
-applyCreds :: HackageCreds -> Request -> IO Request
+applyCreds :: HasLogFunc m => HackageCreds -> Request -> RIO m Request
 applyCreds creds req0 = do
-  manager <- getGlobalManager
-  ereq <- applyDigestAuth
+  manager <- liftIO $ getGlobalManager
+  ereq <- liftIO $ applyDigestAuth
     (encodeUtf8 $ hcUsername creds)
     (encodeUtf8 $ hcPassword creds)
     req0
     manager
   case ereq of
       Left e -> do
-          putStrLn "WARNING: No HTTP digest prompt found, this will probably fail"
+          logInfo "WARNING: No HTTP digest prompt found, this will probably fail"
           case fromException e of
-              Just e' -> putStrLn $ displayDigestAuthException e'
-              Nothing -> print e
+              Just e' -> logInfo $ fromString (displayDigestAuthException e')
+              Nothing -> logWarn $ fromString $ displayException e
           return req0
       Right req -> return req
 
@@ -195,13 +194,14 @@ applyCreds creds req0 = do
 -- sending a file like 'upload', this sends a lazy bytestring.
 --
 -- Since 0.1.2.1
-uploadBytes :: String -- ^ Hackage base URL
-            -> HackageCreds
+uploadBytes :: HasLogFunc m
+            => String -- ^ Hackage base URL
+            -> HackageAuth
             -> String -- ^ tar file name
             -> UploadVariant
             -> L.ByteString -- ^ tar file contents
-            -> IO ()
-uploadBytes baseUrl creds tarName uploadVariant bytes = do
+            -> RIO m ()
+uploadBytes baseUrl auth tarName uploadVariant bytes = do
     let req1 = setRequestHeader "Accept" ["text/plain"]
                (fromString $ baseUrl
                           <> "packages/"
@@ -210,31 +210,36 @@ uploadBytes baseUrl creds tarName uploadVariant bytes = do
                                Candidate -> "candidates/"
                )
         formData = [partFileRequestBody "package" tarName (RequestBodyLBS bytes)]
-    req2 <- formDataBody formData req1
-    req3 <- applyKeyOrCreds creds req2
-    putStr $ "Uploading " ++ tarName ++ "... "
+    req2 <- liftIO $ formDataBody formData req1
+    req3 <- applyKeyOrCreds auth req2
+    logInfo $ "Uploading " <> fromString tarName <> "... "
     hFlush stdout
-    withResponse req3 $ \res ->
+    withRunInIO $ \runInIO -> withResponse req3 (runInIO . inner)
+ where
+    inner :: HasLogFunc m => Response (ConduitM () S.ByteString IO ()) -> RIO m ()
+    inner res =
         case getResponseStatusCode res of
-            200 -> putStrLn "done!"
+            200 -> logInfo "done!"
             401 -> do
-                putStrLn "authentication failure"
-                handleIO (const $ return ()) (removeFile (hcCredsFile creds))
+                logError "authentication failure"
+                case auth of
+                  HACreds creds -> handleIO (const $ return ()) (liftIO $ removeFile (hcCredsFile creds))
+                  _ -> pure ()
                 throwString "Authentication failure uploading to server"
             403 -> do
-                putStrLn "forbidden upload"
-                putStrLn "Usually means: you've already uploaded this package/version combination"
-                putStrLn "Ignoring error and continuing, full message from Hackage below:\n"
-                printBody res
+                logError "forbidden upload"
+                logError "Usually means: you've already uploaded this package/version combination"
+                logError "Ignoring error and continuing, full message from Hackage below:\n"
+                liftIO $ printBody res
             503 -> do
-                putStrLn "service unavailable"
-                putStrLn "This error some times gets sent even though the upload succeeded"
-                putStrLn "Check on Hackage to see if your pacakge is present"
-                printBody res
+                logError "service unavailable"
+                logError "This error some times gets sent even though the upload succeeded"
+                logError "Check on Hackage to see if your pacakge is present"
+                liftIO $ printBody res
             code -> do
-                putStrLn $ "unhandled status code: " ++ show code
-                printBody res
-                throwString $ "Upload failed on " ++ tarName
+                logError $ "unhandled status code: " <> fromString (show code)
+                liftIO $ printBody res
+                throwString $ "Upload failed on " <> fromString tarName
 
 printBody :: Response (ConduitM () S.ByteString IO ()) -> IO ()
 printBody res = runConduit $ getResponseBody res .| CB.sinkHandle stdout
@@ -242,20 +247,22 @@ printBody res = runConduit $ getResponseBody res .| CB.sinkHandle stdout
 -- | Upload a single tarball with the given @Uploader@.
 --
 -- Since 0.1.0.0
-upload :: String -- ^ Hackage base URL
-       -> HackageCreds
+upload :: HasLogFunc m
+       => String -- ^ Hackage base URL
+       -> HackageAuth
        -> FilePath
        -> UploadVariant
-       -> IO ()
-upload baseUrl creds fp uploadVariant =
-  uploadBytes baseUrl creds (takeFileName fp) uploadVariant =<< L.readFile fp
+       -> RIO m ()
+upload baseUrl auth fp uploadVariant =
+  uploadBytes baseUrl auth (takeFileName fp) uploadVariant =<< (liftIO $ L.readFile fp)
 
-uploadRevision :: String -- ^ Hackage base URL
-               -> HackageCreds
+uploadRevision :: HasLogFunc m
+               => String -- ^ Hackage base URL
+               -> HackageAuth
                -> PackageIdentifier
                -> L.ByteString
-               -> IO ()
-uploadRevision baseUrl creds ident@(PackageIdentifier name _) cabalFile = do
+               -> RIO m ()
+uploadRevision baseUrl auth ident@(PackageIdentifier name _) cabalFile = do
   req0 <- parseRequest $ concat
     [ baseUrl
     , "package/"
@@ -269,5 +276,5 @@ uploadRevision baseUrl creds ident@(PackageIdentifier name _) cabalFile = do
     , partBS "publish" "on"
     ]
     req0
-  req2 <- applyKeyOrCreds creds req1
+  req2 <- applyKeyOrCreds auth req1
   void $ httpNoBody req2
diff --git a/src/main/Main.hs b/src/main/Main.hs
index bb0f2287..9234f71c 100644
--- a/src/main/Main.hs
+++ b/src/main/Main.hs
@@ -655,21 +655,19 @@ uploadCmd uploadOpts = do
         config <- view configL
         let hackageUrl = T.unpack $ configHackageBaseUrl config
             uploadVariant = uoptsUploadVariant uploadOpts
-        getCreds <- liftIO $ memoizeRef $ Upload.loadCreds config
+        getCreds <- memoizeRef $ Upload.loadCreds config
         mapM_ (resolveFile' >=> checkSDistTarball sdistOpts) files
         forM_ files $ \file -> do
             tarFile <- resolveFile' file
-            liftIO $ do
-              creds <- runMemoized getCreds
-              Upload.upload hackageUrl creds (toFilePath tarFile) uploadVariant
+            creds <- runMemoized getCreds
+            Upload.upload hackageUrl creds (toFilePath tarFile) uploadVariant
         forM_ dirs $ \dir -> do
             pkgDir <- resolveDir' dir
             (tarName, tarBytes, mcabalRevision) <- getSDistTarball (sdoptsPvpBounds sdistOpts) pkgDir
             checkSDistTarball' sdistOpts tarName tarBytes
-            liftIO $ do
-              creds <- runMemoized getCreds
-              Upload.uploadBytes hackageUrl creds tarName uploadVariant tarBytes
-              forM_ mcabalRevision $ uncurry $ Upload.uploadRevision hackageUrl creds
+            creds <- runMemoized getCreds
+            Upload.uploadBytes hackageUrl creds tarName uploadVariant tarBytes
+            forM_ mcabalRevision $ uncurry $ Upload.uploadRevision hackageUrl creds
 
 sdistCmd :: SDistOpts -> RIO Runner ()
 sdistCmd sdistOpts =
```

hasufell:
https://github.com/kocielnik/stack/pull/1

kocielnik:
@hasufell, I tried to get the tests to pass, but so far failed. Maybe you could give me a tip on this?

```haskell
    • Couldn't match expected type ‘IO (Maybe a0)’
                  with actual type ‘RIO m0 (Maybe Stack.Upload.HackageKey)’
```

Played with `liftIO` and `MonadUnliftIO`, without success.

hasufell:
> @hasufell, I tried to get the tests to pass, but so far failed. Maybe you could give me a tip on this?
> 
> ```haskell
>     • Couldn't match expected type ‘IO (Maybe a0)’
>                   with actual type ‘RIO m0 (Maybe Stack.Upload.HackageKey)’
> ```
> 
> Played with `liftIO` and `MonadUnliftIO`, without success.

```diff
diff --git a/src/Stack/Upload.hs b/src/Stack/Upload.hs
index 9ef161a5..0d27f84d 100644
--- a/src/Stack/Upload.hs
+++ b/src/Stack/Upload.hs
@@ -12,6 +12,8 @@ module Stack.Upload
     , uploadRevision
       -- * Credentials
     , HackageCreds
+    , HackageAuth(..)
+    , HackageKey(..)
     , loadAuth
     , writeFilePrivate
       -- * Internal
@@ -53,6 +55,7 @@ import           System.PosixCompat.Files              (setFileMode)
 
 
 newtype HackageKey = HackageKey Text
+    deriving (Eq, Show)
 
 -- | Username and password to log into Hackage.
 --
@@ -62,10 +65,11 @@ data HackageCreds = HackageCreds
     , hcPassword :: !Text
     , hcCredsFile :: !FilePath
     }
-    deriving Show
+    deriving (Eq, Show)
 
 data HackageAuth = HAKey HackageKey
                  | HACreds HackageCreds
+    deriving (Eq, Show)
 
 instance ToJSON HackageCreds where
     toJSON (HackageCreds u p _) = object
diff --git a/src/test/Stack/UploadSpec.hs b/src/test/Stack/UploadSpec.hs
index cccbac51..b1246288 100644
--- a/src/test/Stack/UploadSpec.hs
+++ b/src/test/Stack/UploadSpec.hs
@@ -29,6 +29,6 @@ spec = do
       (fileMode status .&. 0o777) `shouldBe` 0o600
 
   it "finds a HACKAGE_KEY env variable" $ do
-    maybeGetHackageKey `shouldReturn` Nothing
+    runRIO () maybeGetHackageKey `shouldReturn` Nothing
     setEnv "HACKAGE_KEY" "api_key"
-    maybeGetHackageKey `shouldReturn` Just "api_key"
+    runRIO () maybeGetHackageKey `shouldReturn` Just (HackageKey "api_key")
```

kocielnik:
@hasufell, I've retested package uploading with `stack` via the CLI.

What else do you think we'll need to get this change ready for merging?

src/Stack/Upload.hs
```diff
@@ -67,14 +84,26 @@ instance FromJSON (FilePath -> HackageCreds) where
 withEnvVariable :: Text -> IO Text -> IO Text
 withEnvVariable varName fromPrompt = lookupEnv (T.unpack varName) >>= maybe fromPrompt (pure . T.pack)
 
+maybeGetHackageKey :: RIO m (Maybe HackageKey)
+maybeGetHackageKey = fmap (HackageKey . T.pack) <$> (liftIO $ lookupEnv (T.unpack "HACKAGE_KEY"))
```
pbrisbin, line 88: You could do the whole thing in `liftIO` and so move the `<$>` and `$` to drop one parens.

`OverloadedStrings` means the term `"HACKAGE_KEY"` can be either a `String` or `Text`, so treating it as `Text` and `unpack`ing it is functional but unnecessary. You can just treat it like a `String`.

```suggestion
maybeGetHackageKey = liftIO $ fmap (HackageKey . T.pack) <$> lookupEnv "HACKAGE_KEY"
```

I don't think `Stack.Prelude` has `(<$$>)`, but if it did you could do:

```hs
maybeGetHackageKey = liftIO $ HackageKey . T.pack <$$> lookupEnv "HACKAGE_KEY"
```

src/Stack/Upload.hs
```diff
@@ -67,14 +84,26 @@ instance FromJSON (FilePath -> HackageCreds) where
...
+
+loadAuth :: HasLogFunc m => Config -> RIO m HackageAuth
+loadAuth config = do
+  maybeHackageKey <- maybeGetHackageKey
+  case maybeHackageKey of
+    Just key -> do
+      logInfo "HACKAGE_KEY found in env, using that for credentials."
```
pbrisbin, line 95: Personally, I don't think we need to log this. Seems like something added to aid development of this feature more than anything.

src/Stack/Upload.hs
```diff
@@ -136,34 +166,45 @@ credsFile config = do
...
     manager
   case ereq of
       Left e -> do
-          putStrLn "WARNING: No HTTP digest prompt found, this will probably fail"
+          logWarn "WARNING: No HTTP digest prompt found, this will probably fail"
           case fromException e of
-              Just e' -> putStrLn $ displayDigestAuthException e'
-              Nothing -> print e
+              Just e' -> logWarn $ fromString (displayDigestAuthException e')
+              Nothing -> logWarn $ fromString $ displayException e
```
pbrisbin, line 192: These two lines are so close that I wouldn't use different application styles between them. Can you make them the same? I'm partial to `$`, so that is my suggestion.
```suggestion
              Just e' -> logWarn $ fromString $ displayDigestAuthException e'
              Nothing -> logWarn $ fromString $ displayException e
```

src/test/Stack/UploadSpec.hs
```diff
@@ -26,3 +27,8 @@ spec = do
     unless osIsWindows $ do
       status <- getFileStatus fp
       (fileMode status .&. 0o777) `shouldBe` 0o600
+
+  it "finds a HACKAGE_KEY env variable" $ do
+    runRIO () maybeGetHackageKey `shouldReturn` Nothing
+    setEnv "HACKAGE_KEY" "api_key"
```
pbrisbin, line 33: Probably won't bite you, but it could be good to `unsetEnv` after the test, just to avoid any pollution.

pbrisbin (approved):
Thanks for pushing this forward! I left some thoughts, mostly style.

GitHub would let me Approve, apparently, and I do approve, but I'm going to call this a Comment officially because I am very much not an active maintainer of this project and would love a more official eye come along too.

src/Stack/Upload.hs
hasufell: I think it should stay, because if you forget to unset HACKAGE_KEY, you might get unexpected results.

src/test/Stack/UploadSpec.hs
pbrisbin: Something like

```hs

withEnv "HACKAGE_KEY" "api_key"
  $ runRIO () maybeGetHackageKey `shouldReturn` Just (HackageKey "api_key")

withEnv :: String -> String -> IO a -> IO a
withEnv k v f = do
  setEnv k v
  f `finally` unsetEnv k
```

src/Stack/Upload.hs
kocielnik: I agree that at first (and second) sight it looks very much like debug-level info, but I also think it should stay - for the reason Julian gave.

Would be worth reconsidering if this was the only line of output shown for this operation, but for `stack upload` there are at least 5 more log lines shown before it. These include `stack> '/tmp/stack-sdist-3ea4415731d1f1b8/source-files-list'`, which to me seems considerably farther from user interest than the `HACKAGE_KEY` var.

Maybe it would be a good candidate for a separate improvement in `stack upload` to clean up those lines, eg. grouping them into a separate log level.

kocielnik:
> GitHub would let me Approve, apparently, and I do approve, but I'm going to call this a Comment officially because I am very much not an active maintainer of this project and would love a more official eye come along too.

Thanks, @pbrisbin, this helps a lot! :)

src/Stack/Upload.hs
kocielnik: I didn't think of this before, but this kind of change really helps quickly spot the *true* difference between those lines!

kocielnik:
@snoyberg, please let us know what else we can do here in order to make this change ready to merge.

src/Stack/Upload.hs
```diff
@@ -83,14 +112,15 @@ loadCreds config = do
       writeFilePrivate fp $ lazyByteString lbs
 
       unless (configSaveHackageCreds config) $ do
-        putStrLn "WARNING: You've set save-hackage-creds to false"
-        putStrLn "However, credentials were found at:"
-        putStrLn $ "  " ++ fp
+        logWarn "WARNING: You've set save-hackage-creds to false"
```
snoyberg, line 115: Any particular reason to include this change here? It seems unrelated, and also seems to be the cause of quite a large diff here.

kocielnik: We wanted to make sure the user is notified when a `HACKAGE_KEY` is found and used instead of username/password, even if the latter are also already stored in env variables.

It felt lazy (in the negative sense) to do it with another `putStrLn` while we could do it with a `logInfo`. And since we could do *this* with a `logInfo`, we reckoned with just a little more effort we could resolve the TODO from the import section - and leave the module a tiny bit better than we found it.

Julian (@hasufell) was kind enough to contribute this particular set of changes (commit 135826b). If you feel it would help in any way, please let me know and I'll extract that commit into a separate PR.

Alternatively, I can add a mention of this change to the title of this PR to make it less surprising to be found here.

hasufell: > We wanted to make sure the user is notified when a `HACKAGE_KEY` is found and used instead of username/password, even if the latter are also already stored in env variables.
> 
> It felt lazy (in the negative sense) to do it with another `putStrLn` while we could do it with a `logInfo`. And since we could do *this* with a `logInfo`, we reckoned with just a little more effort we could resolve the TODO from the import section - and leave the module a tiny bit better than we found it.
> 
> Julian (@hasufell) was kind enough to contribute this particular set of changes (commit 135826b). If you feel it would help in any way, please let me know and I'll extract that commit into a separate PR.
> 
> Alternatively, I can add a mention of this change to the title of this PR to make it less surprising to be found here.

Basically, there's a TODO in the module already:

```
 
 ​import​           ​System.IO​                             (​putStrLn​, ​putStr​, ​print​) ​--​ TODO remove putStrLn, use logInfo
```

snoyberg: It's not a question of whether it should be changed or not, but of mixing together two unrelated changes into a single PR. It makes it difficult to ensure only relevant code is changing and that no regressions are being introduced. For example, I noticed in the change here that moving to `logInfo` breaks the behavior of having a prompt on the same line without a following newline. It wasn't clear if this is intrinsic to the behavior of this feature, and the change in behavior wasn't called out at all.

snoyberg (approved):
Looks good, thanks!

src/Stack/Upload.hs
kocielnik: I see, thanks for pointing that out.

PR #5644, opened 2021-12-03
Alter command types to be less abstract

danbroooks:
I was hacking around on Stack today, familiarising myself with some of the startup code and playing around adding some logging. I noticed this type, and found it to be adding quite a lot of indirection. Once I figured out what it was doing I thought it seemed logical to make some of the type aliases concrete, as they are only operating on a single type. I think this makes the type signatures less confusing here. I've also moved out a type alias to a place that seems like a suitable home :smile: 

-------

I believe this change should be covered off by the type checker :+1: it is just changing / moving some types.

### DIFF
--- a/src/Options/Applicative/Complicated.hs
+++ b/src/Options/Applicative/Complicated.hs
@@ -7,7 +7,8 @@
 -- pushing upstream to optparse-applicative.
 
 module Options.Applicative.Complicated
-  ( addCommand
+  ( AddCommand
+  , addCommand
   , addSubCommands
   , complicatedOptions
   , complicatedParser
@@ -22,6 +23,9 @@ import           Options.Applicative.Builder.Internal
 import           Stack.Prelude
 import           Stack.Types.Config
 import           System.Environment
+
+type AddCommand env =
+    ExceptT (RIO env ()) (Writer (Mod CommandFields (RIO env (), GlobalOptsMonoid))) ()
 
 -- | Generate and execute a complicated options parser.
 complicatedOptions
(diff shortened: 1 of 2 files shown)


src/Options/Applicative/Complicated.hs
```diff
@@ -37,14 +41,14 @@ complicatedOptions
...
   -- ^ footer
-  -> Parser a
+  -> Parser GlobalOptsMonoid
   -- ^ common settings
-  -> Maybe (ParserFailure ParserHelp -> [String] -> IO (a,(b,a)))
+  -> Maybe (ParserFailure ParserHelp -> [String] -> IO (GlobalOptsMonoid,(RIO env (),GlobalOptsMonoid)))
   -- ^ optional handler for parser failure; 'handleParseResult' is called by
   -- default
-  -> ExceptT b (Writer (Mod CommandFields (b,a))) ()
+  -> AddCommand env
```
mpilgrem, line 49: If moving from polymorphic to concrete, I wondered if `env` was always `Runner`, in practice.

src/main/Main.hs
```diff
@@ -414,24 +412,24 @@ commandLineHandler currentDir progName isInterpreter = complicatedOptions
       where
         -- addCommand hiding global options
         addCommand' :: String -> String -> (a -> RIO Runner ()) -> Parser a
-                    -> AddCommand
+                    -> AddCommand Runner
```
mpilgrem, line 415: If `env` is always `Runner`, this could revert to being just `AddCommand`, with a suitable type synonym.

src/Options/Applicative/Complicated.hs
```diff
@@ -20,12 +21,15 @@ import           Options.Applicative.Types
 import           Options.Applicative.Builder.Extra
 import           Options.Applicative.Builder.Internal
 import           Stack.Prelude
+import           Stack.Types.Config
 import           System.Environment
 
+type AddCommand env =
```
mpilgrem, line 27: I wondered if `Stack.Types.Config` would be a better place for the type synonym. It is imported by `Main` and it already exports some types related to commands.

PR #5651, opened 2021-12-16
Add explicit details for `acme-missiles` example

hs211216:
I tried following these instructions but didn't realize I needed to update both `package.yaml` and `stack.yaml` and 
encountered the same "Could not find module ‘Acme.Missiles’" problem [**godzsa** did three years ago](https://stackoverflow.com/questions/48251027/cannot-add-extra-deps-to-stack-project) 
so I suggest this recap for beginners.

### DIFF
--- a/doc/GUIDE.md
+++ b/doc/GUIDE.md
@@ -444,7 +444,12 @@ extra-deps:
 - acme-missiles-0.3 # not in the LTS
 ```
 
-Now `stack build` will succeed.
+Now `stack build` will succeed.  To recap you needed to 
+* add `acme-missiles` to **dependencies** in `package.yaml` and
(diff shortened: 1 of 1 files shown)


mpilgrem:
I see the problem, but rather than provide a recap, perhaps it would be better to reinforce the introduction? Perhaps something along these lines:
>We've specified the `acme-missiles` package as a dependency in the `package.yaml` file (see above), but `acme-missiles` is not part of that LTS package set, so building failed.
>
>To add `acme-missiles` to the available packages, we'll use the `extra-deps` field in the `stack.yaml` file. That field defines extra packages, not present in the resolver, that will be needed as dependencies. You can add this like so:
~~~yaml
extra-deps:
- acme-missiles-0.3 # not in the LTS
~~~
>Now `stack build` will succeed.

Issue #5680, opened 2022-02-16
Local dependency building is broken


drone-rites:
### General summary/comments (optional)

There is a bug where a local dependency that has been built from a local consumer project would be installed in the package db for the consumer project but not in the local dependency's own package db. This results in an error where ghc-pkg is unable to find the cached local package in its own package db and the build fails. This can be resolved by running stack clean and then stack build for the local dependency, but then a build for the consumer of the package will fail as it tries to find the local dependency in its own package db. The only fix is then to again run stack clean for the local dependency and then run stack build for the consumer of the dependency.

I have provided a PR with a simple fix (#5613). This fix is extremely simple: remove one line that modifies the setup-config file and causes the trouble. A more satisfying fix for this issue may be to figure out how to adjust the package db search so that cached builds can be used both by the local package and any consumers of the local package.

### Steps to reproduce

See description above.

### Expected

See description above.

### Actual

See description above.

### Stack version

2.7.3

### Method of installation

* Official binary, downloaded from stackage.org or fpcomplete's package repository

drone-rites:
Yes, this was resolved with @borsboom's #5613.

PR #5682, opened 2022-02-21
Fix failure to find package when a dependency is shared between projects (#5680)

borsboom:
This fixes a regression that was introduced in f64188016aea9fb34554f29858423aec0860ae4e (the fix for #5578).  Now, a package will be reconfigured if the project it's being built from (and therefore the local package database it gets installed to) changes.

How I tested:
* A new integration test is added in `test/integration/tests/5680-share-package-across-projects`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -26,6 +26,8 @@ Bug fixes:
   See [rio#237](https://github.com/commercialhaskell/rio/pull/237)
 * Fix handling of overwritten `ghc` and `ghc-pkg` locations.
   [#5597](https://github.com/commercialhaskell/stack/pull/5597)
+* Fix failure to find package when a dependency is shared between projects.
+  [#5680](https://github.com/commercialhaskell/stack/issues/5680)
 
 ## v2.7.3
 
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -12,6 +12,7 @@ module Stack.Build.Cache
     , tryGetConfigCache
     , tryGetCabalMod
     , tryGetSetupConfigMod
+    , tryGetPackageProjectRoot
     , getInstalledExes
     , tryGetFlagCache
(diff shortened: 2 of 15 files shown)


borsboom:
@snoyberg Now, in addition to the file `stack-cabal-mod` and `stack-setup-config-mod` files written to the package to track modification times, a new `stack-project-root` file gets written containing the project root when it was last built.  I wonder if it would make sense to consolidate all this information into a single JSON or YAML file rather than writing separate files?

istathar:
Built and tested locally (with `lts-18.25` as it happens) and as far as I can tell the problem we encountered is no longer occurring, at least not using the jumping around that I demonstrated to you. I'll keep using this Stack binary through the next couple days on our main project to see if things stay sane but looking good!

Great work @borsboom and @drone-rites!

drone-rites:
This is great news, thanks for fixing! I'll also switch to this branch for local use until this fix makes it's way to a release.

qrilka:
@borsboom I think single JSON makes sense

qrilka (approved):

snoyberg (approved):
Nice work and nice integration test! Sorry for breaking this.

And yes, I think putting everything into a single JSON is cleaner, but it wouldn't be a blocker.

My recommendation would be cutting a release once this is merged and has at least a bit of testing.

istathar:
@snoyberg I've been testing this for a week and it's been good; @dmvianna was _not_ using it and just hit the bug, tried building Stack from `master` and the problem went away. Thank you so much for getting @borsboom to look on this. We'd be keen for a 2.7.4 release whenever you're ready.

snoyberg:
@borsboom is planning on a release later this week. I believe it will be 2.7.5.

borsboom:
@istathar Now released here: https://github.com/commercialhaskell/stack/releases/tag/v2.7.5

Issue #5670, opened 2022-01-14
current stack fails to build with Cabal-3.4


juhp:
### General summary/comments (optional)
Can't build stack itself with ghc-9.0 (or Stackage Nightly).

### Steps to reproduce
stack --resolver nightly build

(the next nightly after 2022-01-14 should contain pantry, since hackage-security and ed25519 got unlocked for ghc-9.0)

### Actual
```shellsession
$ git diff
diff --git a/stack.yaml b/stack.yaml
index d1f4502f..12e65387 100644
--- a/stack.yaml
+++ b/stack.yaml
@@ -6,6 +6,8 @@ packages:
 extra-deps:
 - rio-0.1.21.0@rev:0
 - pantry-0.5.3@rev:0
+- hackage-security-0.6.0.1
+- ed25519-0.0.5.0
: 
$ stack --resolver nightly build
:
[ 62 of 111] Compiling Stack.Constants.Config [TH]
[ 74 of 111] Compiling Stack.Package

/var/home/petersen/haskell/stack/src/Stack/Package.hs:131:35: error:
    Not in scope: type constructor or class ‘D.Flag’
    Neither ‘Distribution.Package’, ‘Distribution.PackageDescription’,
            ‘Distribution.Text’,
            ‘Distribution.Verbosity’ nor ‘System.Directory’ exports ‘Flag’.
    |
131 |                               -> [D.Flag]
    |                                   ^^^^^^

/var/home/petersen/haskell/stack/src/Stack/Package.hs:938:13: error:
    Not in scope: type constructor or class ‘Flag’
    |
938 | flagMap :: [Flag] -> Map FlagName Bool
    |             ^^^^

/var/home/petersen/haskell/stack/src/Stack/Package.hs:940:17: error:
    Not in scope: type constructor or class ‘Flag’
    |
940 |   where pair :: Flag -> (FlagName, Bool)
    |                 ^^^^

/var/home/petersen/haskell/stack/src/Stack/Package.hs:989:21: error:
    Not in scope: data constructor ‘Flag’
    |
989 |                     Flag flag ->
    |                     ^^^^
Completed 166 action(s).

--  While building package stack-2.7.4 (scroll up to its section to see the error) using:
[...]

PR #5559, opened 2021-06-05
Upgrade to Cabal 3.4.0.0

brandon-leapyear:
Upgrades Stack to use Cabal 3.4.0.0

Resolves https://github.com/commercialhaskell/stack/issues/5558

Doesn't quite fix the haddock issue; where does Stack decide the cabal-the-tool version?

Issue #5558, opened 2021-06-04
Upgrade Cabal 3.4.0.0


brandon-leapyear:
Blocked by https://github.com/commercialhaskell/pantry/pull/38

Support Cabal 3.4.0.0; _might_ resolve https://github.com/commercialhaskell/stack/issues/4428 and https://github.com/commercialhaskell/stack/issues/5284

kk-hainq:
I would like to add that Cabal 3.4 seems to fix https://github.com/haskell/cabal/issues/6470, which makes #5318 a lot easier to handle from `stack`'s side.

brandon-leapyear:
@kk-hainq see the linked PR. Blocked on someone on stack's side reviewing, helping with CI failures, etc

hasufell:
I think this was recently fixed by @mpilgrem

mpilgrem:
From commit a9f00ffe4908b7f55af938f5a85943859f1e7af4 (1 Apr 2022), the `stack` executable can be built with `lts-19.x` (GHC 9.0.2, `Cabal-3.4.1.0`). I have a separate pull request on moving Stack to GHC 9.0.2 (#5736) - a related but different task - which is currently stuck on the CI for Unix-like operating systems. That may be overtaken by the intention of @psibi to move Stack to GHC 9.2.3 (#5763 and https://github.com/commercialhaskell/stack/pull/5736#issuecomment-1152541872).

brandon-leapyear:
Seems like this PR is redundant as of #5763

PR #5696, opened 2022-04-01
Allow stack to build with lts-19.1 (GHC 9.0.2)

mpilgrem:
Uses C pre-processor (CPP) directives to not disturb the existing code that builds with versions of GHC before 9.0.2.

Tested by building `stack` on Windows 11. The built `stack` executable was, in turn, then tested by using it to build `stack` on Windows 11.

After building:
~~~
❯ stack --version
Version 2.7.6, Git revision 79cdfa1b6ddc5b9c06ad8dea1930d4868fe91b56 (8411 commits) RELEASE-CANDIDATE x86_64
Compiled with:
- Cabal-3.4.1.0
...
- aeson-2.0.3.0
...
- base-4.15.1.0
...
- hpack-0.34.6
...
- mintty-0.1.3
...
- pantry-0.5.4
...
~~~

I think documentation of the use of `stack` is unaffected.

One motivation for this was seeing that @snoyberg had updated `pantry-0.5.4` to support `aeson-2.0.0.0`.

### DIFF
--- a/src/Stack/Build.hs
+++ b/src/Stack/Build.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
@@ -19,7 +20,12 @@ module Stack.Build
 
 import           Stack.Prelude hiding (loadPackage)
 import           Data.Aeson (Value (Object, Array), (.=), object)
+#if MIN_VERSION_aeson(2,0,0)
+import qualified Data.Aeson.Key as Key
+import qualified Data.Aeson.KeyMap as KeyMap
+#else
 import qualified Data.HashMap.Strict as HM
+#endif
 import           Data.List ((\\), isPrefixOf)
 import           Data.List.Extra (groupSort)
 import qualified Data.List.NonEmpty as NE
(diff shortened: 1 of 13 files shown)


mpilgrem:
I think this pull request addresses issue #5670.

mpilgrem:
I think this pull request also addresses issue #5558.

mpilgrem:
I've taken a slightly different approach to the `Cabal` dependency than earlier pull request #5559 (which does not currently pass checks). Firstly, I use CPP. Secondly, I change the type of the `packageCabalSpec` field of `Stack.Types.Package.Package` constructor from `!VersionRange` to `!CabalSpecVersion`. In `Stack.Package.applyForceCustomBuild`, I use `cabalSpecToVersionDigits`, while #5559 uses `cabalSpecMinimumLibraryVersion` in the `Package` constructor.

hasufell:
Is there a reason to keep backwardscompat to old Cabal and old aeson? Why not raise lower bound? Aeson is already 2.x in lts-19.1, Cabal is 3.4.1.0.

mpilgrem:
@hasufell, I do not know the approach of the repository to backwards compatability. `CONTRIBUTING.md` is silent on the topic. However, in the case of `pantry`, I saw the comment of @snoyberg on https://github.com/commercialhaskell/pantry/pull/45 'I'd rather avoid a commit that breaks compat with old versions of aeson'. I assumed that desire applied equally to `stack`. Also, I assumed that, without backwards compatability, the CI would fail.

snoyberg:
It's worth documenting, but Stack follows different rules than libraries like pantry. With executables, we don't need to strive for the same broad version compatibility, but instead define a well-known combination that we ship with. This applies doubly to the Cabal library dependency, where we try to support exactly one version of the library with each release. In other words, @hasufell has a good point here.

@mpilgrem I apologize that that wasn't clear and led to you doing extra work on this PR. I'm fine merging as-is, or if you'd rather simplify to remove the CPP that would be great too. And if either of you are interested in clarifying the policy I just specified somewhere that you think contributors would see it easily, that would be great too.

mpilgrem:
In that case, I think I will take things step by step: merge this pull request with the CPP; make sure the CI works with resolver lts-19.1; remove the CPP.

PR #5699, opened 2022-04-05
Use capi calling convention for `ioctl`

minoki:
As described at [GHC blog](https://www.haskell.org/ghc/blog/20210709-capi-usage.html), ccall with variadic function like `ioctl` is problematic on AArch64/Darwin.
`capi` pseudo-calling convention avoids the problem.

I suppose `stack` binary for AArch64 is not officially supported, but this change will help possible future release and unofficial ones.

---


src/unix/System/Terminal.hsc
```diff
@@ -23,7 +25,7 @@ instance Storable WindowWidth where
   poke p (WindowWidth w) = do
     (#poke struct winsize, ws_col) p w
 
-foreign import ccall "sys/ioctl.h ioctl"
+foreign import capi "sys/ioctl.h ioctl"
```
nh2, line 28: Whenever possible, add comments with rationale:

```suggestion
-- `ioctl.h` is variadic, so `capi` is needed, see:
-- https://www.haskell.org/ghc/blog/20210709-capi-usage.html
foreign import capi "sys/ioctl.h ioctl"
```

nh2 (approved):
This change makes sense. Please add your rationale to the comment as suggested.

minoki:
Done. I also changed the order of LANGUAGE pragmas to match 9a839f4eb21c76b1b67db5e844d2c30eba104b2c.

mpilgrem:
This change also made sense to me, so I merged it.

Issue #5707, opened 2022-04-11
`drop-packages: Cabal` is silently ignored


andreasabel:
Entry 
```yaml
drop-packages: 
- Cabal
``` 
in `stack.yaml` seems to be silently ignored.
So working around
- #5706 

according to the recipe in
- #5679 

fails.

mpilgrem:
I think that `stack` may not be ignoring `drop-packages: - Cabal`; I think that recent resolvers do not include `Cabal` to start with - so that it is the version of `Cabal` that ships with the version of GHC that is relevant. See https://github.com/commercialhaskell/stackage/issues/4425#issuecomment-602321624 (from March 2020).

mpilgrem:
With my investigation above and the updating of the documentation, I am going to close this issue.

andreasabel:
Thanks for the explanation, @mpilgrem !

Issue #5706, opened 2022-04-11
`extra-deps: Cabal` is ignored when `.cabal` file does not have `Cabal` dependency


andreasabel:
I am using `cabal-version: 2.4` in my `.cabal` file because I am using `**` globbing patterns in the `data-files` section.

It seems that `Cabal-2.4.0.1` shipped with `lts-14.27` does not correctly implement these yet, at least `stack build` fails:
```
.stack-work/install/x86_64-osx/2d2cad02bc2492d52c57d57f66cc81f62e153722b88c98941d8195159454b5ac/8.6.5/share/x86_64-osx-ghc-8.6.5/MiniAgda-0.2022.4.6/test/should-fail: copyFile: does not exist (No such file or directory)
Completed 10 action(s).
'cabal copy' failed.  Error message:

--  While building package MiniAgda-0.2022.4.6 (scroll up to its section to see the error) using:
      /Users/abel/.stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5 --builddir=.stack-work/dist/x86_64-osx/Cabal-2.4.0.1 copy
    Process exited with code: ExitFailure 1
```
Indeed the `should-fail` directory does not exist, but `**` exists instead (looks like a bug handling globbing pattern `**`):
```
$ ls .stack-work/install/x86_64-osx/2d2cad02bc2492d52c57d57f66cc81f62e153722b88c98941d8195159454b5ac/8.6.5/share/x86_64-osx-ghc-8.6.5/MiniAgda-0.2022.4.6/test
**      fail    succeed
```

However, working around this problem by bumping the `Cabal` version in the `stack.yaml` files fails:
```
extra-deps:
- Cabal-2.4.1.0
```
is seemingly ignored, and also any newer `Cabal` version.

Note that my `.cabal` file does not declare a dependency on `Cabal` directly, the dependency is more indirect via `cabal-version: 2.4`.

Maybe related:
- #5679

mpilgrem:
I understand that the `cabal-version` field in a `.cabal` file just specifies the minimum version of the `Cabal` library needed to interpret the `.cabal` file - it does not affect the version of the `Cabal` library that the `stack` executable was actually compiled with. What version of the `stack` executable are you using? `stack` version 2.7.5 was compiled using `lts-17.15` (GHC 8.10.4) and uses `Cabal-3.2.1.0`.

mpilgrem:
[...]

PR #5724, opened 2022-04-22
Fix #5056 Document how stack uses Cabal-the-library

### LINKED ISSUE #5056: broken link in FAQ

This question in the FAQ https://github.com/commercialhaskell/stack/blob/master/doc/faq.md#what-is-the-relationship-between-stack-and-cabal has a broken link about “Archicture: Plan Execution”, which was pointing at https://github.com/commercialhaskell/stack/blob/0ef138ce0dea5325695efdd358b0804d12ce802f/doc/architecture.md#plan-execution. Given https://github.com/commercialhaskell/stack/commits/master/doc/architecture.md, it seems like the link should now point to somewhere in build-overview.md, given that that replaced architecture.md. But I’m not sure the relevant content is actually in build-overview.

But, I’m not sure fixing the link is the full solution. E.g., the FAQ says “See the Architecture: Plan Execution section for more detail, including how the Cabal version is chosen,” but even in the old architecture.md, I don’t see anything about how the Cabal version is chosen, which is the information I’m looking for. I checked Stack.Build.Execute, as mentioned, but that only seems to _use_ the cabal version, it must be elsewhere that chooses it.

mpilgrem:
The motivation for documenting this comes from investigating #5706 and #5707. It can be stated quite briefly, so I do not think it is out of place under the FAQ.

I also correct the incorrect guidance in the same place about `stack build` initialising a `stack.yaml` file.

Please also shortly describe how you tested your change. Bonus points for added tests! N/A

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -44,18 +44,26 @@ directory. None of this should affect any existing Haskell tools at all.
 
 ## What is the relationship between stack and cabal?
 
-* Cabal-the-library is used by stack to build your Haskell code. See the
-  [Architecture: Plan Execution](architecture.md#plan-execution) section for
-  more detail, including how the Cabal version is chosen.
-* A .cabal file is provided for each package, and defines all package-level
-  metadata just like it does in the cabal-install world: modules, executables,
-  test suites, etc. No change at all on this front.
-* A stack.yaml file references 1 or more packages, and provides information on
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks for clarifying the `Cabal` dependency in the docs!

Issue #1586, opened 2016-01-02
Improve pkg-config/dependencies story on Windows


tejon:
Currently it's up to the Windows user (and undocumented!) to use `stack exec -- pacman` to set up `pkg-config` and any dependencies.

*At worst*, `stack` should be able to handle `pacman -Sy` and the `pkg-config` installation, and then provide a more helpful error message if a package is missing (i.e. how to search for and install it).

If a package's `pkg-config` name is always the suffix of its `pacman` package name -- I'm not sure of this, but it seems likely -- then I don't see why `stack` can't automate the whole process.

borsboom:
Would you be able to provide a more concrete example/proposal?  For example, demonstrate a particular problem and then propose the way Stack should handle it better?

byorgey:
I just ran into this too.  I had no idea I was supposed to do anything involving `pacman` until @mgsloan pointed me at this ticket.  At the very least, a concrete proposal would be to have `stack` print a message at the end of `stack setup` saying something like "You should do `stack exec -- pacman -Sy` and then you can use `stack exec -- pacman -S foobar` to install any `pkg-config` packages needed for building".

tejon:
@byorgey That would be helpful, yeah. But it should be automatable (near as I can tell pacman names do always match pkg-config), and very likely should be performed while building dependencies for reasons relating to #1587: `pacman -S foobar` is generally the wrong library. Ideally we want to run `pacman -S [mingw-[w64|w32]-[x86_64|i686]-]foobar` for each permutation of those found by `pacman -Ss foobar` or, more intelligently, for whichever matches the current build environment. Failing full automation, a `stackman` command that does the right thing would be good.

But none of this really helps until #1587 is solved, because right now stack is building binaries for Windows that think full POSIX support is available. Quite a few will crash if not run within the `msys` shell. I've been getting a hair farther with that every time I tackle it, but I've had limited opportunities to do so.

tejon:
Well, never mind that blockade. :) While working through #1587 I've come to the conclusion that the best solution here is to provide a wrapper for `pacman` which is aware of stack's environment but is able to run outside of it (because installs should optimally happen in the POSIX msys shell, which is _not_ provided by `stack exec` after PR #1909). Doing it any other way runs into the same issues that caused `git` to be dropped (#1046); not to mention that since MSYS2 is a resolver-versioned tool, it seems wrong for `stack`-proper to do anything that might damage build reproduciblity.

[...]

PR #5725, opened 2022-04-23
Add `stack ls tools` subcommand

mpilgrem:
This proposed pull request adds a `tools` subcommand to the `stack ls` command, in order to list stack's installed tools (in order) at the command line. A `--filter TOOL_NAME` option is provided to filter the tools listed by name (the default being no filter).

The existing help for `stack path --programs` is extended to refer to the existence of `stack ls tools`.

The primary motivation for this was to allow Windows users to find easily the full name of the stack-supplied MSYS2 tool(s) (by using `stack ls tools --filter msys2`) - without having to investigate the structure under the stack root. The full name is useful to know in setting the `PKG_CONFIG_PATH` environment variable manually. (See #1586.)

However, it would also allow all users to find easily the installed version(s) of GHC (by using `stack ls tools --filter ghc`), listed with the most recent version first.

`Stack.Setup.getInstalledTool` is modified to eliminate code repetition that would otherwise result (the core of the existing code becoming `Stack.Setup.Installed.filterTools`.)

Please also shortly describe how you tested your change. Bonus points for added tests! Tested by building, and using, `stack`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -21,6 +21,7 @@ Other enhancements:
 * Bump to `hpack-0.35.0`.
 * On Windows, the installer now sets `DisplayVersion` in the registry, enabling
   tools like `winget` to properly read the version number.
+* `tools` subcommand added to `stack ls`, to list stack's installed tools.
 
 Bug fixes:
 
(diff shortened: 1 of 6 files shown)


mpilgrem:
I also tested what happens if, somehow, there are two `msys2-YYYYMMDD.installed` files in the `stack path --programs` directory (which, I think, would require some manual interference with the contents of that directory). `stack ls tools --filter msys2` reports both tools present, with the most recent version listed first (which I consider is the desired behaviour) but `stack` will only use the most recent version of MSYS2 available.

PR #5731, opened 2022-05-02
Fix #5713 Don't assume unified coverage report redundant if only one *.tix file

### LINKED ISSUE #5713: Test coverage for package whose test suite is in another package.

Coverage reports are broken when you use a test suite in one package to test another that doesn't have a test suite.

### Steps to reproduce

1. Make package `foobar` with code in a library `foobar:lib:foobar`.
2. Make package `foobar-gen` with generators and testutils in a library `foobar-gen:lib:foobar-gen`
3. Make test suite `foobar-gen:test:foobar-test` to test `foobar` using the generators in `foobar-gen`.

Run `stack test foobar foobar-gen --coverage`. Find that there is no coverage report for any code in `foobar`.

### Expected

Coverage report for both `foobar` and `foobar-gen`

### Actual

No coverage report for `foobar`.

### Stack version

```
$ stack --version
2.7.3 x86_64 hpack-0.34.6
```

### Method of installation

* Nixpkgs

mpilgrem:
Currently, `Stack.Coverage.generateHpcUnifiedReport` assumes that a unified coverage report will be redundant if there is only one `*.tix` file. However, that is not necessarily the case. For example, one package may test the library of another package that does not test its own library.

As an interim solution, this proposed pull request suggests that a unified report should always be produced if there is one or more `*.tix` files, even if it may be redundant in some circumstances.

A more complex solution would be to have a more complex test that determines whether a unified report would be truely redundant if there is only one `*.tix` file.

Please also shortly describe how you tested your change. Bonus points for added tests! Tested by building, and using, `stack`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,11 @@ Major changes:
 
 Behavior changes:
 
+* `stack build --coverage` will generate a unified coverage report, even if
+  there is only one `*.tix` file, in case a package has tested the library of
+  another package that has not tested its own library. See
+  [#5713](https://github.com/commercialhaskell/stack/issues/5713)
+
 Other enhancements:
 
 Bug fixes:
--- a/src/Stack/Coverage.hs
+++ b/src/Stack/Coverage.hs
@@ -288,10 +288,25 @@ generateHpcUnifiedReport = do
     extraTixFiles <- findExtraTixFiles
     let tixFiles = tixFiles0  ++ extraTixFiles
         reportDir = outputDir </> relDirCombined </> relDirAll
-    if length tixFiles < 2
-        then logInfo $
-            (if null tixFiles then "No tix files" else "Only one tix file") <>
-            " found in " <>
+-- Previously, the test below was:
+--
+--  if length tixFiles < 2
+--      then logInfo $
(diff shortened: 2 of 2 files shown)


NorfairKing:
I just checked and this does indeed solve my problem in #5713 . :tada:

NorfairKing (approved):
This fixes a bug and simplifies code. Double-win.

NorfairKing:
@mpilgrem Do we need anything for this to be merged?
A test would be nice, but I'm not much worried about it.

mpilgrem:
As I have identified in other recent pull requests, I am new to being an outside collaborator, in general and on `stack` in particular, and I don't know the convention/etiquette as regards 'merging your own pull requests'. It seems to me akin to 'marking your own homework', when the change has effects that affect users. However, if nobody objects or comments after a reasonable period of exposure, I will merge the pull request.

NorfairKing:
I have commit rights, but I'm not sure it'd be proper for me to merge this either...
Let's wait to see if anyone objects.

NorfairKing:
I don't see any complaints, this seems very sensible, and I've tried it out, so I'll go ahead and merge.

PR #5760, opened 2022-05-31
Fix #5755 Add --script-no-run-compile flag

### LINKED ISSUE #5755: Add support to only compile script using stack shebang
I have a file
```hs
#!/usr/bin/env stack
{- stack script --resolver ... --package ... -}
```
and running `./Script.hs` or `stack Script.hs` works as expected. But I would like to have a CI job checking that `Script.hs` builds (but not runs). I do see `stack script` has a `--no-run` flag, but it seems like I would have to copy the entire `stack script` command to do so:
```bash
stack script --compile --no-run Script.hs --resolver ... --package ...
```
Doing `stack Script.hs --no-run` or `stack script Script.hs --no-run` would fail; the first would pass `--no-run` as an argument to `Script.hs`, and the second would fail with "`script` requires `--resolver`".

The most straightforward way I see here is to add a `STACK_INTERPRETER_NO_RUN=true` env var that will add the `--no-run` flag to `stack script` and the appropriate flags to `stack runghc`.

mpilgrem:
This proposed pull request adds a `--force-script-no-run-compile` flag (disabled by default) that forces the `--no-run` and `--compile` options with `stack script`.

This enables a command like `stack --force-script-no-run-compile Script.hs` to behave like `stack script ... --no-run --compile -- Script.hs` but without having to list all the other arguments in the stack interpreter options comment (represented by `...`) on the command line.

Also adds an explanation to the online documentation.

Please also shortly describe how you tested your change. Bonus points for added tests! Tested by building and using `stack`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -21,6 +21,14 @@ Other enhancements:
 * Bump to `hpack-0.35.0`.
 * On Windows, the installer now sets `DisplayVersion` in the registry, enabling
   tools like `winget` to properly read the version number.
+* Adds flag `--script-no-run-compile` (disabled by default) that uses the
+  `--no-run` option with `stack script` (and forces the `--compile` option).
+  This enables a command like `stack --script-no-run-compile Script.hs` to
+  behave like `stack script <arguments> --no-run --compile -- Script.hs` but
+  without having to list all the `<arguments>` in the stack interpreter options
+  comment in `Script.hs` on the command line. That may help test that scripts
+  compile in CI (continuous integration). See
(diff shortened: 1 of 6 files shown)


mpilgrem:
Many thanks. The forcing aspect is limited to overriding any `--optimize` flag (the alternative to `--compile`). I'm happy to adopt the shorter flag name.

I did consider a more general form of `--script-no-run=compile` or `--script-no-run=optimize`, but reasoned that this was wanted only to test (mainly in CI) that a script compiled, not to test that it compiled with optimisation.

PR #5736, opened 2022-05-07
Fix #5697 Move stack to lts-19.16 (GHC 9.0.2) from lts-17.15 (GHC 8.10.4)

mpilgrem:
Also corrects a typo in `stack-ghc-902.yaml` comment.

* [x] Any changes that could be relevant to users have been recorded in the ChangeLog.md
* [x] The documentation has been updated, if necessary.

Please also shortly describe how you tested your change. Bonus points for added tests! Tested by building, and using, `stack`.

PR #5763, opened 2022-06-07
Add support for GHC 9.2.4 and its Cabal-3.6.3.0

psibi:

Ran tests locally (excluding the integration tests) and ensured that it passed fine:

```
stack test
....
Finished in 2.6665 seconds
673 examples, 0 failures

stack             > Test suite stack-test passed
Completed 14 action(s).
```


mpilgrem:
On the location of docker images for CI, did you see this: https://github.com/commercialhaskell/stack/pull/5736#issuecomment-1132112610.

psibi:
@mpilgrem Yeah, I did see your comment. I'm planning to create the docker images for GHC 9.2.3. Infact, I have a docker image locally ready for GHC 9.2.2 currently where I have the integration test suite working for Linux.

etc/scripts/build-stack-installer.hs
```diff
@@ -1,5 +1,5 @@
 {- stack script
-    --resolver lts-14.27
+    --resolver lts-19.7
```
mpilgrem, line 2: Should the scripts also be moved on to a GHC 9.2.3 resolver? Otherwise people will need to maintain two versions of GHC.

etc/scripts/release.hs
```diff
@@ -1,5 +1,5 @@
 {- stack script
-    --resolver lts-14.27
+    --resolver lts-19.7
```
mpilgrem, line 2: Should the scripts be moved on to a GHC 9.2.3 resolver?

etc/scripts/build-stack-installer.hs
psibi: Yeah, good point. I will make the changes here and the other script.

.github/workflows/integration-tests.yml
```diff
@@ -56,15 +56,32 @@ jobs:
...
+              extra-lib-dirs:
+              - /lib
+              - /usr/lib
+          EOF
+          fi
+
+          if [[ "${{ matrix.os }}" == "windows-latest" ]]
+          then
+              stack exec pacman -- --sync --refresh --noconfirm autoconf
+          fi
```
mpilgrem, line 83: I was interested in this addition. May I ask, what was the problem with the `windows-latest` CI that required the MSYS2 `autoconf` package to be downloaded?

src/Stack/Package.hs
```diff
@@ -411,7 +414,11 @@ generateBuildInfoOpts BioInput {..} =
               , [ biCabalDir
                 | null (hsSourceDirs biBuildInfo)
                 ]
+#if MIN_VERSION_Cabal(3,6,0)
+              , mapMaybe toIncludeDir (map getSymbolicPath (hsSourceDirs biBuildInfo))
```
mpilgrem, line 418: How about `, mapMaybe (toIncludeDir . getSymbolicPath) (hsSourceDirs biBuildInfo)`?

src/Stack/Package.hs
```diff
@@ -808,7 +815,11 @@ resolveComponentFiles
     -> [DotCabalDescriptor]
     -> RIO Ctx (Map ModuleName (Path Abs File), [DotCabalPath], [PackageWarning])
 resolveComponentFiles component build names = do
+#if MIN_VERSION_Cabal(3,6,0)
+    dirs <- mapMaybeM resolveDirOrWarn (map getSymbolicPath (hsSourceDirs build))
```
mpilgrem, line 819: How about `dirs <- mapMaybe (resolverDirOrWarn . getSymbolicPath) (hsSourceDirs build)`?

src/Stack/PackageDump.hs
```diff
@@ -64,7 +64,8 @@ ghcPkgCmdArgs pkgexe@(GhcPkgExe pkgPath) cmd mpkgDbs sink = do
     case reverse mpkgDbs of
         (pkgDb:_) -> createDatabase pkgexe pkgDb -- TODO maybe use some retry logic instead?
         _ -> return ()
-    sinkProcessStdout (toFilePath pkgPath) args sink'
+    (_,res) <- sinkProcessStderrStdout (toFilePath pkgPath) args CL.sinkNull sink'
+    pure res
```
mpilgrem, line 68: My own version of this was:
~~~
snd <$> sinkProcessStderrStdout (toFilePath pkgPath) args CL.sinkNull sink'
~~~
It has the advantage (or disadvantage, depending on your perspective) of being a single line.

.github/workflows/integration-tests.yml
psibi: IIRC, it was required when I was using `stack-macos.yaml` for Windows. I have removed it now, Thanks!

src/Stack/Package.hs
psibi: Sure!

src/Stack/Package.hs
psibi: Sure!

src/Stack/PackageDump.hs
psibi: I have no strong opinions either wise. I have changed to use your suggestion. Thanks!

src/Stack/PackageDump.hs
```diff
@@ -64,7 +64,7 @@ ghcPkgCmdArgs pkgexe@(GhcPkgExe pkgPath) cmd mpkgDbs sink = do
     case reverse mpkgDbs of
         (pkgDb:_) -> createDatabase pkgexe pkgDb -- TODO maybe use some retry logic instead?
         _ -> return ()
-    sinkProcessStdout (toFilePath pkgPath) args sink'
+    snd <$> sinkProcessStderrStdout (toFilePath pkgPath) args CL.sinkNull sink'
```
snoyberg, line 67: Just to be clear, this changes the behavior of calling the process. Previously, `stderr` from the process would be inherited and therefore error output would show up in the console. Now it will be swallowed. Not terrible, but it would be ideal to avoid that if possible.

stack.yaml
```diff
@@ -1,18 +1,11 @@
-resolver: lts-17.15
+resolver: nightly-2022-06-10 # GHC 9.2.3
```
snoyberg, line 1: It's unusual to make the primary way to build Stack a nightly image, but I don't have a huge issue. @borsboom any thoughts?

snoyberg (approved):

stack.yaml
istathar: At the end of the day it's a snapshot and if it [the dependencies it needs] builds it builds, right? I felt this way too until recently, but the moment we get this patch landed I'm going to take my system to GHC 9.2 and if that means `nightly` for now so be it. I'll return us to `stable` ^W `lts` once we've got one, I imagine. Same would apply here?

src/Stack/PackageDump.hs
mpilgrem: The existing comment at `sinkProcessStdout` warns:
~~~
-- | Consume the stdout of a process feeding strict 'ByteString's to a consumer.
-- If the process fails, spits out stdout and stderr as error log
-- level. Should not be used for long-running processes or ones with
-- lots of output; for that use 'sinkProcessStderrStdout'.
~~~
For some reason, the use of `sinkProcessStdout` on macOS had begun to generate `startProcess: posix_spawnp: illegal operation (Inappropriate ioctl for device)` errors. Substituting `sinkProcessStderrStdout` caused them to disappear.

psibi: I have filed a bug report here: https://github.com/haskell/process/issues/251

@mpilgrem I have been getting various version of the errors, ranging from:

```
startProcess: posix_spawnp: invalid argument (Bad file descriptor)
startProcess: posix_spawnp: does not exist (No such file or directory)
```

I haven't seen the `ioctl` error message that you are seeing, but it is a concern.

src/Stack/Setup.hs
```diff
@@ -474,9 +474,9 @@ warnUnsupportedCompiler ghcVersion = do
         logWarn "For more information, see: https://github.com/commercialhaskell/stack/issues/648"
         logWarn ""
         pure True
-    | ghcVersion >= mkVersion [9, 1] -> do
+    | ghcVersion >= mkVersion [9, 3] -> do
         logWarn $
-          "Stack has not been tested with GHC versions above 9.0, and using " <>
+          "Stack has not been tested with GHC versions above 9.3, and using " <>
```
mpilgrem, line 479: The warning message should, I think, refer to GHC 9.2, not 9.3?

mpilgrem:
@psibi, in the interim of sorting out the CI etc on macOS, I am going to reflect some of this in `master` - the parts that allow `stack` to at least _build_ on some operating systems with GHC 9.2.3, GHC 9.2.4.

psibi:
@mpilgrem I think this PR's macOS CI test already pass. So this can likely be merged.

stack-macos.yaml
```diff
@@ -0,0 +1,30 @@
...
+    developer-mode: true
+
+ghc-options:
+   "$locals": -fhide-source-paths
+
+# Contains macos specific fix for https://github.com/haskell/process/issues/247
+extra-deps:
+- git: https://github.com/haskell/process.git
+  commit: b95e5fbdeb74e0cc36b6878b60f9807bd0001fa8
+- Cabal-3.6.3.0@sha256:ff97c442b0c679c1c9876acd15f73ac4f602b973c45bde42b43ec28265ee48f4,12459
```
mpilgrem, line 26: Is this extra-dep needed? Resolver nightly-2022-07-30 includes Cabal-3.6.3.0. https://www.stackage.org/nightly-2022-07-30/package/Cabal-3.6.3.0.

psibi: I guess it's not required then.

psibi:
@mpilgrem Thanks for taking the PR, let me know if you need any help from me.

mpilgrem:
I rebased against the master branch, hence the forced push. Partly, I was trying to clear the CI fails and I thought it might help to put this on top of the current master.

mpilgrem:
If only I had waited a day ... `nightly-2022-08-01` et seq are now on GHC 9.2.4. I'll bump again, soon.

psibi:
@mpilgrem I would recommend to get this PR for GHC 9.2.3 merged since we would have to create a new Docker image for GHC 9.2.4...

mpilgrem:
@psibi, on your Docker recommendation, is there any possibility that you could do for GHC 9.2.4 what you did for GHC 9.2.3? My thinking is that GHC 9.2.3 will become 'ancient history' and people will not want to download it just to build Stack.

psibi:
@mpilgrem I may have some time during the weekend, but not sure. One thing which I learnt during last time I created docker images for 9.2.3 and 9.2.2 is that it wasn't smooth as the stack build was resulting in some seg fault error using the ghc published by the official alpine builds of the GHC team. So, I had to use the nix path  (which I try to avoid) to create the images. My recommendation would be still to get the 9.2.3 merged.

psibi:
@mpilgrem I have pushed a newer image for ghc 9.2.4 here: https://registry.hub.docker.com/layers/alpine-haskell-stack/psibi/alpine-haskell-stack/9.2.4/images/sha256-6908d1507cdce6bd2f4cf9575a53b002ca1b86262dc4cda3712c0618c4b1b10f?context=explore

You would need to update the stack yaml files to this:

```
docker:
  repo: psibi/alpine-haskell-stack:9.2.4
```

If the test suite passes and everything works fine, I can then publish the same image under fpco's docker registry.

mpilgrem:
I am stumped by the CI failures and have asked the community for some help: https://discourse.haskell.org/t/why-does-ghc-9-2-4-require-process-1-6-13-2/4884

mpilgrem:
When I have a moment, I'll see if I can work out if this (the GHC panic) is a 'GHC 9.2.3 -> GHC 9.2.4' problem or a '-> process-1.6.15.0` problem.

mpilgrem:
@psibi, some good news and some bad news. The good news is that I received an explanation from Bodigrim on the Haskell Community of the `stack-test` build errors that start with:

~~~
Configuring stack-2.7.6...
Warning:
    This package indirectly depends on multiple versions of the same package. This is very likely to cause a compile failure.
      package ghc (ghc-9.2.4) requires process-1.6.13.2
....
~~~

and end with (except, somehow, on macOS, which gets away with it):

~~~
GHC runtime linker: fatal error: I found a duplicate definition for symbol
   mkNamedPipe
whilst processing object file
   C:\Users\runneradmin\AppData\Local\Programs\stack\x86_64-windows\ghc-9.2.4\lib\x86_64-windows-ghc-9.2.4\process-1.6.13.2\HSprocess-1.6.13.2.o
The symbol was previously defined in
   C:\Users\runneradmin\AppData\Roaming\stack\snapshots\48f59125\lib\x86_64-windows-ghc-9.2.4\process-1.6.15.0-7YgH4mi6Un42svYTH51yc2\HSprocess-1.6.15.0-7YgH4mi6Un42svYTH51yc2.o
This could be caused by:
   * Loading two different object files which export the same symbol
   * Specifying the same object file twice on the GHCi command line
   * An incorrect `package.conf' entry, causing some object to be
     loaded twice.
ghc-9.2.4.exe: panic! (the 'impossible' happened)
  (GHC version 9.2.4:
	loadObj "C:\\Users\\runneradmin\\AppData\\Local\\Programs\\stack\\x86_64-windows\\ghc-9.2.4\\lib\\x86_64-windows-ghc-9.2.4\\process-1.6.13.2\\HSprocess-1.6.13.2.o": failed

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
~~~

The bad news is that this is the kotty problem:

GHC 9.2.4 (and GHC 9.2.3) ships with `process-1.6.13.2`, but that lacks the important subsequent macOS bug fixes found in `process-1.6.15.0`.

Builds of Stack are tested extensively using `hspec`, but `hpsec` depends on `hspec-core` which, in turn, depends on `ghc`. `stack-test` built with GHC 9.2.4 and `process-1.6.15.0` can’t build `ghc` on most operating systems.

The release of GHC 9.4 may cut through this knot. `process-1.6.15.0` was released as a precursor for the release of GHC 9.4 and I thinl that release may happen very soon - see https://github.com/haskell/process/issues/256.

psibi:
@mpilgrem I have pushed a commit which might hopefully fix the CI issues, let's see.

I see that the stack.yaml's docker image wasn't properly updated.

mpilgrem:
@psibi, I see your plan - to build `stack` against `process-1.6.15.0` on macOS (where, somehow, the `stack-test` manages to build against two different versions of `process`) and to build `stack` against `process-1.6.13.2` on other operating systems.

As recent changes in MSYS2 can break all versions of GHC before GHC 9.4 on Windows, I am expecting all previous versions of GHC to quickly become 'ancient history' once GHC 9.4 is released. So, in any event, my plan is to move Stack to GHC 9.4 as soon as it is released and available as a resolver on Stackage.

mpilgrem:
I was wrong, previously, when I said `Cabal` was not required as an extra-dep. As a global package (`process`) is being overridden, other global packages that also depend on the overridden package (here, `Cabal`) do need to be expressly added as an extra-dep.

mpilgrem:
My concern about Stack being built against different project-level YAML configuration files for different operating systems is the consequences for any user of `stack upgrade --source-only --git` on macOS. Perhaps the `stack.yaml` for Stack should include the `user-message:` [key](https://docs.haskellstack.org/en/stable/yaml_configuration/#user-message) with a warning that, if on macOS, Stack should be built with `--stack-yaml=stack-macos.yaml`?

psibi:
>  So, in any event, my plan is to move Stack to GHC 9.4 as soon as it is released and available as a resolver on Stackage.

Yeah, agree. Do you think probably we should wait till 9.4 is released ?

With the latest change, I can see that these are the status of integration tests:
- Passing in Windows
- Tests are failing in Linux. I'm guessing they should be easy to fix.
- Some cabal constraint issue in MacOS. 

Do you think it's worth fixing it or waiting till 9.4 is released ? I'm okay with either of the routes, although I do think there is some value in fixing things up and getting it merged. :-)

istathar:
If we can land this now so that Linux users (and production systems) can take advantage of GHC 9.2 that would be lovely. If making it work satisfactorily for Macs has to wait for GHC 9.4 then so be it.

mpilgrem:
We are edging closer to the goal. Only Windows is failing the CI, post build, with:

~~~
stack etc/scripts/release.hs build 
  shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
<command line>: cannot satisfy -package filepattern: 
    filepattern-0.1.2-66wJ8f[5](https://github.com/commercialhaskell/stack/runs/7706097949?check_suite_focus=true#step:5:6)2GB2AJ9AxLRBx22 is unusable due to missing dependencies:
      extra-1.[7](https://github.com/commercialhaskell/stack/runs/7706097949?check_suite_focus=true#step:5:8).10-91WukMj7JUWIbjiASntoX7
    (use -v for more information)
~~~

mpilgrem:
In respect of the Windows CI problem, on my local machine, I tried:

~~~
❯ &"C:\Program Files\Git\bin\bash.exe"

$ stack etc/scripts/release.hs build
<command line>: cannot satisfy -package filepattern:
    filepattern-0.1.2-66wJ8f52GB2AJ9AxLRBx22 is unusable due to missing dependencies:
      extra-1.7.10-91WukMj7JUWIbjiASntoX7
    (use -v for more information)
~~~

and then (extract of output):

~~~
$ stack exec -- ghc-pkg check
There are problems in package shake-0.19.6:
  dependency "extra-1.7.10-91WukMj7JUWIbjiASntoX7" doesn't exist
There are problems in package filepattern-0.1.2:
  dependency "extra-1.7.10-91WukMj7JUWIbjiASntoX7" doesn't exist
~~~

and then (extract of output - notice that `extra-1.7.10` is listed):

~~~
$ stack exec -- ghc-pkg list
C:\sr\snapshots\69802e88\pkgdb
    extra-1.7.10
    {filepattern-0.1.2}
    {shake-0.19.6}
~~~

If I then deleted the snapshot `C:\sr\snapshots\69802e88`, the script would then get all the way through to (extract  of output):
~~~
$ stack etc/scripts/release.hs build

Copied executables to C:\Users\mikep\Documents\Code\GitHub\commercialhaskell\stack\_release\bin\stack-2.7.6-windows-x86_64:
- stack.exe
- stack-integration-test.exe
_release\bin\stack-2.7.6-windows-x86_64\stack.exe --version

cd _release; "c:\Program Files (x86)\NSIS\Unicode\makensis.exe" -V3 installer-stack-2.7.6-windows-x86_64.nsi
Error when running Shake build system:
  at want, called at C:\\\\Users\\mikep\\Documents\\Code\\GitHub\\commercialhaskell\\stack\\etc\\scripts\\release.hs:140:14 in main:Main
* Depends on: build
  at apply1, called at src\Development\Shake\Internal\Rules\Rerun.hs:41:5 in shake-0.19.6-IbcuTwxBUoXFufVfniuvxk:Development.Shake.Internal.Rules.Rerun
* Depends on: _release/stack-2.7.6-windows-x86_64-installer.exe
* Raised the exception:
c:\Program Files (x86)\NSIS\Unicode\makensis.exe: createProcess: does not exist (No such file or directory)
~~~

I think it only fails because I have a modern version of NSIS and the script depends on an out-of-date version.

However, if I now run the script a second time, I'm back to the beginning with a re-built `C:\sr\snapshots\69802e88` causing problems:

~~~
$ stack etc/scripts/release.hs build
<command line>: cannot satisfy -package filepattern:
    filepattern-0.1.2-66wJ8f52GB2AJ9AxLRBx22 is unusable due to missing dependencies:
      extra-1.7.10-91WukMj7JUWIbjiASntoX7
    (use -v for more information)
~~~

mpilgrem:
A little more investigation on my local machine of that problematic snapshot (extracts of output):

~~~
$ stack exec -- ghc-pkg field extra id
id: extra-1.7.10-A2H4uvbJlMq9rs5CtM7Ep4

$ stack exec -- ghc-pkg field shake depends
depends: extra-1.7.10-91WukMj7JUWIbjiASntoX7

$ stack exec -- ghc-pkg field filepattern depends
depends: extra-1.7.10-91WukMj7JUWIbjiASntoX7
~~~

Something odd seems to be happening.

mpilgrem:
@psibi, a cache bust was worth a try (and we will see what happens when the CI completes), but my investigations overnight suggest to me that the problem is that `release.hs` can't be run twice on Windows in a Git Bash shell. Somehow, the running of the script corrupts the snapshot created when the script is compiled. My own investigations are not over but I will not be able to turn back to them until later today.

psibi:
@mpilgrem Yeah, the CI cache bust doesn't seem to help. I'm trying to remove the dependency of the extra package itself as it doesn't seem to be used.

psibi:
Seems I was wrong, it is actually being used.

mpilgrem:
I've been looking at the output to the stderr channel of `stack --color=never --no-terminal etc/scripts/release.hs build --build-args="--verbose --color=never --no-terminal"`. All I have really established is that, when `release.hs` is first built, you have:

~~~
[info] extra                        > Installing library in C:\sr\snapshots\69802e88\lib\x86_64-windows-ghc-9.2.4\extra-1.7.10-91WukMj7JUWIbjiASntoX7
~~~

but when the script builds, you have:

~~~
[info] extra               > using precompiled package

[debug] Run process: C:\Users\mikep\AppData\Local\Programs\stack\x86_64-windows\ghc-9.2.4\bin\ghc-pkg-9.2.4.exe unregister --force extra-1.7.10

[debug] Run process: C:\Users\mikep\AppData\Local\Programs\stack\x86_64-windows\ghc-9.2.4\bin\ghc-pkg-9.2.4.exe register --force C:\sr\snapshots\cd1a0f51\pkgdb\extra-1.7.10-A2H4uvbJlMq9rs5CtM7Ep4.conf
~~~

and when the script is called a second time, it is that `lib\...\extra-1.7.10-91WukMj7JUWIbjiASntoX7` versus `extra-1.7.10-A2H4uvbJlMq9rs5CtM7Ep4.conf` mismatch that causes the reported error.

mpilgrem:
As GHC 9.4.1 is now released, I've started to experiment (elsewhere - I may put my experiments in branch `ghc-941`) to see if I can build Stack on Windows with GHC 9.4.1. At the moment it looks lilke that is blocked by `cereal-0.5.8.2` not supporting GHC 9.4.1 (https://github.com/GaloisInc/cereal/issues/104).

mpilgrem:
@psibi, I have managed to build Stack with GHC 9.4.1 on Windows. See #5820. I am still investigating what is going on with `release.hs` with GHC 9.2.4 on Windows.

borsboom:
As the original author of `release.hs`, let me state: I think it has mostly outlived its usefulness.  If it's causing trouble, my recommendation would be to do away with it.  It's a holdover from before we were doing most of the release work using CI (Github Actions), were building on a lot more plaforms, were building distribution packages (e.g. rpms and debs), etc.  At this point I actually believe it's holding us back from automating more of the release process.  I think it could relatively easily be replaced with a small amount of shell script, probably directly in the Github Actions workflow, to run run the integration test and tar up the bindists.

mpilgrem:
@borsboom, that end state would be great. I can't say if it is `release.hs` itself that is causing the current trouble because I am still trying to work out how `release.hs` does what it does! In this case, what it does is - somehow - to corrupt an existing snapshot database, but on Windows only.

mpilgrem:
I am still investigating the CI problem with the `release.hs` script on Windows. I've returned to basics, trying to understand what `stack <script_module.hs>` actually does. Perhaps it is my misunderstanding, but it seems to me to (a) behave oddly and (b) behave oddly in Stack 2.7.5 (that is, the oddity has not been introduced post-2.7.5). I'll open a separate issue to document my experience.

mpilgrem:
I have rebased on the `master` branch.

psibi:
@mpilgrem  Nice to see that you fixed the Windows CI issue! What was the underlying issue which caused the CI failure initially ?

mpilgrem:
@psibi Thanks! It was a long haul, but it looks good to me and the CI now passes. I've merged it.

mpilgrem:
> Nice to see that you fixed the Windows CI issue! What was the underlying issue which caused the CI failure initially ?

I did not get to the bottom of the underlying issue, but I think (a) something odd is happening with Stack's scripting (at least, on Windows - I've not explored other operating systems) but (b) it exists in Stack 2.7.5, so this pull request was not introducing it. See #5822. The solution here was empirical:

* I tried deleting all the `--package` from `release.hs`. That worked on all the operating systems except Windows, which was somehow calling on `Cabal-3.8.1.0` from Hackage, causing an error (because `Cabal-syntax-3.6.0.0` wanted `Cabal < 3.7`).
* Adding `--extra-dep Cabal-3.6.3.0` to `release.hs` seemed to force Cabal-3.6.30 on all operating systems.

psibi:
@mpilgrem Cool, nice work on getting this merged. The docker images are published under my account, I will publish under fpco's one later and send a PR for the same as it seems to be working well.

Issue #5455, opened 2020-12-08
situation with stack repl --package and multiple packages is unclear


simonmichael:
`stack repl --package`'s help suggests it can support multiple packages, but it gives an obscure error for the obvious syntaxes (comma or space-separated, as supported by [stack script --package](https://docs.haskellstack.org/en/stable/GUIDE/#using-multiple-packages)):

```
$ stack --version
Version 2.5.1, Git revision d6ab861544918185236cf826cb2028abb266d6d5 x86_64 hpack-0.33.0
$ stack repl --help | grep '\--package'
                  [--with-ghc GHC] [--[no-]load] [--package PACKAGE(S)]
  --package PACKAGE(S)     Additional package(s) that must be installed
$ stack repl Shake.hs --package regex,shake
Error parsing targets: Directory not found: regex,shake
$ stack repl Shake.hs --package 'regex shake'
Error parsing targets: Directory not found: regex shake
```

PS I'm interested in this because specifying multiple packages without using a space seems possibly helpful in hie.yaml files.

qrilka:
As far as I understand the way to use multiple package would be by repeating `--package <package` for every one of them, probably we should improve CLI help here.

mpilgrem:
I am closing because the current `stack repl --help` explains:
~~~text
--package PACKAGE        Add a package (can be specified multiple times)
~~~

and that is consistent with `Stack.Config.packagesParser`.

PR #5775, opened 2022-07-11
Help text allows multiple use for --package.

philderbeast:
Changes the help text for `--package`, see #5455.

Tested with:

```
> stack run stack -- repl --help
Usage: stack repl [TARGET/FILE] [--pedantic] [--ghci-options OPTIONS]
                  [--ghc-options OPTIONS] [--flag PACKAGE:[-]FLAG]
                  [--with-ghc GHC] [--[no-]load] [--package PACKAGE(S)]
                  [--main-is TARGET] [--load-local-deps] [--[no-]package-hiding]
                  [--only-main] [--trace] [--profile] [--no-strip] [--[no-]test]
                  [--[no-]bench] [--setup-info-yaml URL]
                  [--snapshot-location-base URL] [--help]
  Run ghci in the context of package(s) (experimental) (alias for 'ghci')

Available options:
  TARGET/FILE              If none specified, use all local packages. See
                           https://docs.haskellstack.org/en/stable/build_command/#target-syntax
                           for details. If a path to a .hs or .lhs file is
                           specified, it will be loaded.
  --pedantic               Turn on -Wall and -Werror
  --ghci-options OPTIONS   Additional options passed to GHCi
  --ghc-options OPTIONS    Additional options passed to both GHC and GHCi
  --flag PACKAGE:[-]FLAG   Override flags set in stack.yaml (applies to local
                           packages and extra-deps)
  --with-ghc GHC           Use this GHC to run GHCi
  --[no-]load              Enable/disable load modules on start-up (default:
                           enabled)
  --package PACKAGE(S)     Each --package adds (installs) one package
  --main-is TARGET         Specify which target should contain the main module
                           to load, such as for an executable for test suite or
                           benchmark.
  --load-local-deps        Load all local dependencies of your targets
  --[no-]package-hiding    Enable/disable package hiding
  --only-main              Only load and import the main module. If no main
                           module, no modules will be loaded.
  --trace                  Enable profiling in libraries, executables, etc. for
                           all expressions and generate a backtrace on exception
  --profile                profiling in libraries, executables, etc. for all
                           expressions and generate a profiling report in tests
                           or benchmarks
  --no-strip               Disable DWARF debugging symbol stripping in
                           libraries, executables, etc. for all expressions,
                           producing larger executables but allowing the use of
                           standard debuggers/profiling tools/other utilities
                           that use debugging symbols.
  --[no-]test              Enable/disable testing the package(s) in this
                           directory/configuration (default: disabled)
  --[no-]bench             Enable/disable benchmarking the package(s) in this
                           directory/configuration (default: disabled)
  --setup-info-yaml URL    Alternate URL or relative / absolute path for stack
                           dependencies
  --snapshot-location-base URL
                           The base location of LTS/Nightly snapshots
  --help                   Show this help text

Run 'stack --help' for global options that apply to all subcommands.
```

### DIFF
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -964,7 +964,7 @@ packagesParser :: Parser [String]
 packagesParser = many (strOption
                    (long "package" <>
                      metavar "PACKAGE(S)" <>
-                     help "Additional package(s) that must be installed"))
+                     help "Each --package adds (installs) one package"))
 
 defaultConfigYaml :: (IsString s, Semigroup s) => s
 defaultConfigYaml =
--- a/src/Stack/Config.hs
+++ b/src/Stack/Config.hs
@@ -54,7 +54,8 @@ import qualified Distribution.Text
 import           Distribution.Version (simplifyVersionRange, mkVersion')
 import           GHC.Conc (getNumProcessors)
 import           Network.HTTP.StackClient (httpJSON, parseUrlThrow, getResponseBody)
-import           Options.Applicative (Parser, help, long, metavar, strOption)
+import           Options.Applicative (Parser, help, long, metavar, strOption, style)
+import           Options.Applicative.Help.Pretty (brackets)
 import           Path
 import           Path.Extra (toFilePathNoTrailingSep)
 import           Path.Find (findInParents)
@@ -963,8 +964,9 @@ getDefaultUserConfigPath stackRoot = do
 packagesParser :: Parser [String]
 packagesParser = many (strOption
                    (long "package" <>
-                     metavar "PACKAGE(S)" <>
-                     help "Each --package adds (installs) one package"))
+                     metavar "PACKAGE" <>
+                     help "Add a package" <>
+                     style brackets))
 
 defaultConfigYaml :: (IsString s, Semigroup s) => s
 defaultConfigYaml =
--- a/src/Stack/Options/ExecParser.hs
+++ b/src/Stack/Options/ExecParser.hs
@@ -5,6 +5,7 @@ module Stack.Options.ExecParser where
 import           Options.Applicative
 import           Options.Applicative.Builder.Extra
 import           Options.Applicative.Args
+import           Options.Applicative.Help.Pretty (brackets)
 import           Stack.Options.Completion
 import           Stack.Prelude
 import           Stack.Types.Config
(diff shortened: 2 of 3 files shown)


mpilgrem:
I wondered if the `metavar` of `PACKAGE(S)` was misleading, and whether `[--package PACKAGE]` would be more accurate.

The goal is to help a user understand that the `--package PACKAGE` option can appear multiple times, but I wondered if (a) there was a subtle change in meaning going from `Additional package(s) that must be installed` to `Each --package adds (installs) one package` and (b) the possibility of multiple use should be stated more directly. How about (if the `metavar` changes):

~~~
--package PACKAGE    Additional package that must be installed. Multiple such options are allowed.
~~~

philderbeast:
Agreed `[--package PACKAGE]` would be a nice way to put it, especially since we're habituated to that list syntax.

One thing I was going for with the change was avoiding line wrapping.

```
--package PACKAGE(S)     Additional package that must be installed. Multiple
                         such options are allowed.
```

I felt that "must" was redundant as we're issuing a command to stack and it should follow through to the letter, options included!

By saying the package is installed aren't we surfacing an implementation detail? The user wants a repl where they can import modules from packages added with `--package` on the command line and from packages in `build-depends` of the `.cabal` file.

philderbeast:
I can get `[--package PACKAGE]` by using `<> style brackets`.

mpilgrem:
I think `Add a package` is good - succinct without being terse - but we were at cross-purposes (my fault) with my `[--package PACKAGE]`. I was actually referring to where it appears in the `Usage: ...` part of the documentation. I think `style brackets` is interesting but think it would be a metaphor unfamilar to most users.

I tried to track down documentation conventions elsewhere, but with little luck. I found something that suggests that [Google](https://developers.google.com/style/code-syntax#arguments-that-can-repeat) and [Microsoft](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/command-line-syntax-key) both use `...` to indicate repetition, but did not find that compelling. Also, I think `[--package PACKAGE ...]` (in `Usage: ...`) would be ambiguous about whether it was `PACKAGE` or all of `--package PACKAGE` that was repeating, and I can't see how you could achieve `[--package PACKAGE]...` in `Usage: ...`.

Closer to home, I looked at other examples of `many` in Stack. Most do not bother to document that the option can be used multiple times. An exception is `stack build [--skip ARG]`, which has:
~~~
--skip ARG               Skip given component, can be specified multiple times
~~~
or the `stack --docker-help`, which has:
~~~
--docker-mount (PATH | HOST-PATH:CONTAINER-PATH)
                         Mount volumes from host in container (may specify
                         multiple times)
--docker-env NAME=VALUE  Set environment variable in container (may specify
                         multiple times)
~~~
If the help text is useful and succinct, I don't think it needs to be constrained to one line. It would be good to find a form of words for `many` that could be used consistently throughout Stack. How about:
~~~
--package PACKAGE        Add a package (option may be repeated)
~~~

PR #5677, opened 2022-02-12
Fix `cabal build` by freezing cabal dependencies

simnalamburt:
To easily package stack to various distributions, one should be able to bootstrap stack without relying on stack binary. So many package maintainers use cabal-install to build stack from source code.

Due to recent update of aeson-2.0.0.0, stack can no longer be built with cabal-install. So I hereby send a PR to freeze cabal dependencies using stackage 17.15 LTS.

###### Reference:
- https://hackage.haskell.org/package/stack2cabal
- https://github.com/Homebrew/homebrew-core/pull/95032

PR #5776, opened 2022-07-11
Ignore dist-newstyle folder.

philderbeast:
I like to be able to build projects with either stack or cabal. I saw #5677 and tried it, noticing that the cabal build products aren't ignored by git. This change ignores those in `dist-newstyle/`.


arrowd:
@mpilgrem Trivial PR to merge.

PR #5777, opened 2022-07-12
WIP: Use standard compiler builds on FreeBSD.

arrowd:
Previosly, Stack had to discern between FreeBSD 11- and FreeBSD 12+ and select
a compiler based on this version check. Compiler builds for 12+ were named "-ino64".
Now that FreeBSD 11 reached EoL, there is no need to have two compiler build
versions, so just use standard name for those builds that were previosly named "ino64".

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -993,13 +993,6 @@ getGhcBuilds = do
                         [] -> CompilerBuildStandard
                         _ -> CompilerBuildSpecialized (intercalate "-" c))
                     libComponents
-            Platform _ Cabal.FreeBSD -> do
-                let getMajorVer = readMaybe <=< headMaybe . (splitOn ".")
-                majorVer <- getMajorVer <$> sysRelease
-                if majorVer >= Just (12 :: Int) then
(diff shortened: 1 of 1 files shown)


hasufell:
What does EOL mean? Users can still use FreeBSD 11, no?

arrowd:
> What does EOL mean?

End-of-Life.

> Users can still use FreeBSD 11, no?

That depends on "can use" definition. There are no official packages nor security updates for FreeBSD 11 anymore. You most likely won't be able to compile any FreeBSD port too. Theoretically, users can compile Stack from the source, but I don't think this scenario should be supported. Stack is a development tool, after all, and you shouldn't do development on an ancient OS versions.

arrowd:
@mpilgrem Trivial PR to merge.

arrowd:
Wait, this require changes to stackage-content first. Do not merge this yet.

mpilgrem:
I think somebody other than me should look at this request. I am a Windows user and do not have knowledge or experience of FreeBSD. I noted that the CI had not passed. I've re-run the failed integration tests, in case that failure was just a glitch.

Issue #5773, opened 2022-07-07
Does Stack support Nix flakes?


malteneuss:
I found a really nice blog post that describes [how to correctly setup Stack with Nix](https://www.tweag.io/blog/2022-06-02-haskell-stack-nix-shell/), which is definitely needed to be done on NixOS. However, they use the "old" nix-shell file style of Nix. Does Stack already support or plan to support the new Nix flake style configuration? Officially Nix flakes are still experimental, but they are so stable that they can be used as a much better replacement.

malteneuss:
Yes, it does in a way. I've added the necessary parts as a NixOS section for the Stack docs in #5774.

PR #5779, opened 2022-07-13
 Expand Nix integration docs with more context and NixOS section

malteneuss:
Rewritten some paragraphs in an active voice, added more background info of Nix in some places to make it easier for Nix beginners to use it with Stack, and added a section on how to use Stack with NixOS and Nix flakes.

It answers my question in https://github.com/commercialhaskell/stack/issues/5773.

Feel free to edit as you wish, maintainers.

### DIFF
--- a/doc/nix_integration.md
+++ b/doc/nix_integration.md
@@ -272,7 +272,7 @@ Nix derivations ("packages"), for instance to change some build options of the
 libraries you use, or to set additional environment variables. See the
 [Nix manual][nix-manual-exprs] for more. The `buildStackProject` utility
 function is documented in the [Nixpkgs manual][nixpkgs-manual-haskell].  In such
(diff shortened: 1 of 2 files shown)


mpilgrem:
Thanks! I am a Windows user and cannot comment on the substance. If nobody spots anything after a short period of exposure here, I'll merge the changes.

I did wonder if the changes to  `stack.cabal` would be problematic when putting this through.

psibi:
I gave it a look and it LGTM (atleast all the part except Nix flakes as I don't use that yet). 

As Mike mentioned, I would avoid changes to the cabal file as it's supposed to be generated from package.yaml (via hpack).

malteneuss:
I reverted those changes to `stack.cabal`. It slipped through somehow.

PR #5789, opened 2022-07-21
Fix #1369 Add build option `--cabal-verbosity=VERBOSITY`

### LINKED ISSUE #1369: --cabal-verbose should allow setting verbosity level
From what I observed, using `--cabal-verbose` passes `--verbose` to Setup program. However, the resulting output is not as detailed is it could be. The `--verbose` flag takes an 0..3 int parameter. 
Passing `--verbose` without value is equivalent to ``--verbose=2`. I'd like to be able to obtain cabal output from `--verbose=3`.

The seemingly natural solution would be to add the same value parameter to `--cabal-verbose` and pass it to Setup.

mpilgrem:
Retains the existing `--[no-]cabal-verbose`, as an alternative flag on the command line.

As with the existing flag, the new option can be set in configuration files. The new option has priority in configuration files.

Error messages are provided if the specified verbosity is not recognised by Cabal (the library).

The `ChangeLog.md` and documentation are updated accordingly.

Tested by building Stack on Windows 11, and using it with various combinations of options at the command line and/or in a `stack.yaml` file.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -32,6 +32,9 @@ Other enhancements:
 * Fuller help is provided at the command line if a subcommand is missing (for
   example, `stack ls` now yields the equivalent of `stack ls --help`). See
   [#809](https://github.com/commercialhaskell/stack/issues/809)
+* Add build option `--cabal-verbosity=VERBOSITY` to specify the Cabal verbosity
+  level (the option accepts Cabal's numerical and extended syntax).
+  See [#1369](https://github.com/commercialhaskell/stack/issues/809)
 
 Bug fixes:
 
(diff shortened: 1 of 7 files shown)


mpilgrem:
The 'style' test is failing because CPP is used to determine which imports are necessary. (Once the block to moving to GHC 9.0.2 and higher is removed, the CPP can be removed.)

mpilgrem:
@ndmitchell, given your interest in resolving the issue (albeit three years ago) I thought perhaps you might review this before it was merged.

ndmitchell:
@mpilgrem I'm definitely not qualified to review anything in the stack codebase anymore, I'm afraid.

mpilgrem:
@ndmitchell, thanks for replying - that's noted.

PR #5796, opened 2022-07-28
Add docs on Stack and VS Code

mpilgrem:
The use of the 'Haskell' extension for Visual Studio Code (VS Code), the extent of its current integration with Stack, and the possible implications of that for Stack's position relative to other build tools for Haskell (namely, Cabal (the tool)) have been the subject of some extensive discusion recently on the [Haskell Community](https://discourse.haskell.org/), under a number of Topics.

This proposed pull request seeks to add documentation on Stack and VS Code and the current position. Its objective is to capture that there are current viable workarounds to the limitations and that future developments may see the extension able to work well with Stack.

### DIFF
--- a/doc/Stack_and_VS_Code.md
+++ b/doc/Stack_and_VS_Code.md
@@ -1,0 +1,128 @@
+<div class="hidden-warning"><a href="https://docs.haskellstack.org/"><img src="https://cdn.jsdelivr.net/gh/commercialhaskell/stack/doc/img/hidden-warning.svg"></a></div>
+
+# Stack and Visual Studio Code
+
+[Visual Studio Code](https://code.visualstudio.com/) (VS Code) is a popular
+source code editor, and
+['Haskell'](https://marketplace.visualstudio.com/items?itemName=haskell.haskell)
+is an extension for VS Code that is popular with Haskell coders.
+
+The 'Haskell' extension can be used with Stack but there are some things to be
+aware of, set out below.
+
+## Haskell Language Server
+
+The VS Code extension makes use of the Haskell Language Server (HLS). To work,
+HLS has to be built with the same version of GHC that it will support. That is,
+a version of HLS is required for each version of GHC in use. It is possible that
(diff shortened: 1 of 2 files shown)


mpilgrem:
@hasufell, I have aimed to 'write up' my understanding of the current position, from the perspective of a Stack user. May I ask, from your perspective, is this accurate? Also, important, from your perspective, is this fair? If not, what would you change or add?

doc/Stack_and_VS_Code.md
```diff
@@ -0,0 +1,115 @@
...
+allow the VS Code extension/GHCup to customise Stack, so that Stack can
+automatically switch between versions of GHC that GHCup has downloaded and
+installed. It is hoped a future release of the extension/GHCup will take
+advantage of that new feature. See
+[Stack #5585](https://github.com/commercialhaskell/stack/pull/5585) and
+[GHCup #392](https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/392).
+
+### Workaround #2
+
+Another workaround is to install GHCup so that is 'empty' except for the current
```
hasufell, line 79: This is not a worakaround. It will cause the ABI mismatch problems, whenever stack supplies a different bindist that HLS was built with.

doc/Stack_and_VS_Code.md
```diff
@@ -0,0 +1,115 @@
...
+version of HLS, allow the VS Code extension to use GHCup to manage HLS
+requirements, and to ignore messages from the extension on start-up that
```
hasufell, line 81: That's not necessary. You can completely disable managing of cabal, stack and GHC: https://github.com/haskell/vscode-haskell#setting-a-specific-toolchain

```json
{
  "haskell.toolchain": {
    "ghc": null,
    "cabal": null,
    "stack": null
  }
}
```

mpilgrem:
@hasufell, many thanks for your advice. I hope the changes that I have made have responded positively to it.

hasufell (approved):

PR #5797, opened 2022-07-29
Lint YAML especially `.hlint.yaml`.

philderbeast:
The vscode editor showed problems with `.hlint.yaml`. I was able to reproduce these before fixing them by installing and configuring yamllint for which I added a GitHub action. In reviewing `.hlint.yaml` I saw that we should have been adding our errors before any ignores. Added a comment introducing why we have the "Use alternative" ignore. The ignore for "Too strict maybe" was repeated so I reduced this duplication collecting the within modules together. For YAML linting I'm ignoring most of `.gitub/workflows/`, `doc` and `test` folders.

I pulled the hlint action out of the unit-tests action and used `haskell/actions/setup-hlint` for the set up.

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -1,8 +1,7 @@
 name: hlint
 
-on:
-  pull_request:
-  push:
+# yamllint disable-line rule:truthy
+on: [pull_request, push]
 
 jobs:
   style:
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -1,0 +1,27 @@
+name: hlint
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  style:
+    name: hlint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - uses: haskell/actions/hlint-setup@v2
+        name: Set up HLint
+        with:
+          version: "3.4"
+
+      - name: hlint
+        run: |
+          set -ex
+          hlint src/
+          hlint src/ --cpp-define=WINDOWS=1
+          hlint test/ --cpp-simple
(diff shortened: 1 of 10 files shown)


mpilgrem:
Looking at the files that have changed, some of them seem to have changed largely because indentation has been added or increased. May I ask, what is the rationale for those changes?

philderbeast:
> what is the rationale for those changes?

Within vscode I used format document. I have the extension [linter](https://github.com/fnando/vscode-linter). This fixed the linting problems.

philderbeast:
This is what I first saw in vscode (or something similar), problems with the HLint configuration:

Screen Shot 2022-07-29 at 3 39 19 PM

mpilgrem:
On the linting of YAML, as you can see, historically, the stylistic preference here has been what I think yamlint [refers to](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.indentation) as `indentation: {spaces: 2, indent-sequences: false}`.

philderbeast:
I use `package.dhall` with the same hpack phraseology. With dhall I get a formatter included out of the box.

```
> dhall-hpack-cabal --force
generated plugins-for-blobs.cabal
> dhall format package.dhall
```

The option of formatting `package.yaml` is new to me. I would guess the historical stylistic preference is to not use an automatic formatter and go with whatever is easiest to type or, if you're like me, copy an existing `package.yaml` from another project before adding and removing items.

mpilgrem:
Thanks! I've merged that, with a little tinkering. It seemed to me sensible to apply it to all of the YAML files in the respository. It also seemed to me sensible to combine all linting (yamllint, HLint) into a single 'linting' workflow. I did change the `.yamllint.yaml` to be tolerant of (in fact, require) an indentation style that followed what was already used in the repository. I dealt with the `truthy` problem by turning it off for the keys (it was only GitHub Actions' `on:` key that was tripping it).

philderbeast:
> I did change the `.yamllint.yaml` to be tolerant of (in fact, require) an indentation style that followed what was already used in the repository.

@mpilgrem the [github workflow sample code](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) uses indented arrays.

PR #5805, opened 2022-07-31
Fix get_isa logic

### LINKED ISSUE #5792: `get_isa()` in `get-stack.sh` does not handle all possible outputs of `uname -m`
`get_isa()` in `get-stack.sh` is as follows:
~~~sh
# determines the the CPU's instruction set
get_isa() {
  if uname -m | grep -Eq 'armv[78]l?' ; then
    echo arm
  elif uname -m | grep -q aarch64 ; then
    echo aarch64
  else
    echo x86
  fi
}
~~~

`get_isa()` is used by `is_x86_64()` which, in turn, is used by `do_distro()` (called in the 'Linux' limb of `do_os()`).

`get_isa()` assumes that if the output of `uname -m` (the machine hardware name) is not `armv6l`, `armv7l` or `aarch64` then it is `x86_64` (or `x86`). However, this [Wikipedia table](https://en.wikipedia.org/wiki/Uname#Examples) indicates that other values are possible when `uname -s` (the kernel name) is `Linux`, including `arm6l`, `sparc64`,`ppc64`,`i686`,`ppc`,`k1om`, and `'mips`.

It has been [suggested](https://github.com/commercialhaskell/stack/pull/5630#discussion_r928196881) that this is a 'bug waiting to happen'.

HallowedDust5:

The fix was relatively simple and did not change intended functionality so I did not conduct any tests. Fixes #5792

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -52,6 +52,7 @@ Bug fixes:
   [#5608](https://github.com/commercialhaskell/stack/pull/5608)
 * On Unix-like operating systems, respect the `with-gcc` option when installing
   GHC. See [#5609](https://github.com/commercialhaskell/stack/pull/5609)
+* Fixed logic in `get_isa()` in `get-stack.sh` to not include systems that don't have x86 in their `uname -m` output. See [5792](https://github.com/commercialhaskell/stack/issues/5792).
 
 ## v2.7.5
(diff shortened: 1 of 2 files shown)


mpilgrem:
Thanks!

Issue #5799, opened 2022-07-29
Commit stack*.yaml.lock files.


philderbeast:
Can we please stop ignoring stack's lock files? The [docs](https://docs.haskellstack.org/en/stable/lock_files/) say these can be put in source control and help make builds more reproducible.

https://github.com/commercialhaskell/stack/blob/8bf6a5afd779c4120c25c6ab5f20d1d5581d8994/.gitignore#L30

mpilgrem:
Without commenting on the merits or demerits, there is an existing issue which discusses this -  https://github.com/commercialhaskell/stack/issues/4795 - and it looks like there were arguments for and against. Can you add your perspective to that discussion?

philderbeast:
Thanks for the link to the previous discussion that seems to be in favour overall with some caution around libraries and CI builds. For my own tastes, reproducibility is top of the list and for my own packages I use something like the following for CI builds:

```
> tree -L 2 stack
stack
├── stack-8.0.2.yaml
├── stack-8.0.2.yaml.lock
├── stack-8.10.7.yaml
├── stack-8.10.7.yaml.lock
├── stack-8.2.2.yaml
├── stack-8.2.2.yaml.lock
├── stack-8.4.4.yaml
├── stack-8.4.4.yaml.lock
├── stack-8.6.5.yaml
├── stack-8.6.5.yaml.lock
├── stack-8.8.4.yaml
├── stack-8.8.4.yaml.lock
├── stack-9.0.2.yaml
├── stack-9.0.2.yaml.lock
├── stack-9.2.3.yaml
└── stack-9.2.3.yaml.lock
```

```yaml
name: stack

on:
  push:
  pull_request:

jobs:
  build:
    name: ghc ${{ matrix.ghc }}
    runs-on: ubuntu-latest
    strategy:
      matrix:
        cabal: ["3.6"]
        ghc:
          - "8.0.2"
          - "8.2.2"
          - "8.4.4"
          - "8.6.5"
          - "8.8.4"
          - "8.10.7"
          - "9.0.2"
          - "9.2.3"

    steps:
      - uses: actions/checkout@v3

      - name: Setup Stack
        uses: haskell/actions/setup@v2
        with:
          ghc-version: ${{ matrix.ghc }}
          cabal-version: ${{ matrix.cabal }}
          enable-stack: true
          stack-version: "latest"

      - name: Cache ~/.stack and .stack-work
        uses: actions/cache@v3
        with:
          path: |
[...]

PR #5808, opened 2022-08-02
Use stack/ghc-*.yaml matrix for locked pedantic build.

philderbeast:
I imagine a lot of people are choosing to run with the `haskell/actions/setup` action for setting up cabal, ghc and stack. We could build some of stack this way too such as the `ubuntu-latest` pedantic builds for a matrix of GHC versions. I've added the lock files, see #5799.

I tested with gh-actions but haven't restricted `on` for this action.


mpilgrem:
I may have misunderstood your intent with this pull request, but I hope the following is apt:

I am not certain, but I think each released version of the Stack executable is tied to one version of GHC and one version of Cabal (the library). See: https://github.com/commercialhaskell/stack/blob/master/CONTRIBUTING.md#backwards-compatability. The presence of `stack-ghc-902.yaml` and `stack-ghc-923.yaml` in the repository (and CPP directives in Stack's code) is, I think, just temporary, to facilitate the moving Stack wholesale to a more recent version of GHC/Cabal than GHC 8.10.4/Cabal-3.2.1.0.

At least, that is why I understand that the `pedantic` job and the `integration-tests` job (which does the bulding of binaries for release - see  https://docs.haskellstack.org/en/stable/maintainers/releases/#release-process) don't have a matrix of GHC versions.

PR #5811, opened 2022-08-02
Add stan gh-action and ignore triggered checks

philderbeast:
Adds a gh-action for [stan](https://kowainik.github.io/projects/stan). Both the project and stan need to be built with the same GHC version. All of the triggered checks are excluded. We can chip away at the hints once this action is set up.

### DIFF
--- a/.github/workflows/stan.yml
+++ b/.github/workflows/stan.yml
@@ -1,0 +1,44 @@
+name: stan
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    name: ghc ${{ matrix.ghc }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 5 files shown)


mpilgrem:
Our of curiosity, and perhaps also a sense of loyalty on Stack's own repository, why did you pick Cabal (the tool) as the build tool in the workflow?

philderbeast:
> why did you pick Cabal (the tool) as the build tool in the workflow?

I'm build tool agnostic (and was reusing something I knew worked).

mpilgrem:
Thanks! I've merged this. Some people use the Stack CI as a sort of tutorial ON how to acheive things with Stack itself. So, if it is possible to acheive the same outcome with Stack as the build tool, it would be good to reflect that in the workflow.

PR #5831, opened 2022-08-20
Documenting additional installation instructions for Apple Silicon chip involving C source compilation

leoslf:
Hi Team,

This is the first PR contributing to the open source community, please do not hesitate to correct me if I am wrong.

This is to document for handling the error `configure: error: C compiler cannot build executables.` during the installation of `stack` itself or some package involving c source compilation for Apple Silicon chip (e.g. M1).

## Issue
The problem was caused by the attempt to link a `x86_64` object to existing `arm64` libraries during the compiler test in `configure` which was invoked by `cabal` running in `x86_64`, and the C compiler just targeted the architecture of the parent process running while compiling the object files.

## Possible fix
It could be fixed by passing `-arch arm64` into the `CFLAGS` environment variable, which would be then picked up by the C compiler to compile an`arm64` object to correctly link with `arm64` libraries.

Thank you very much.

Best Regards,
Leo

-------
Pointing to `stable` as the base branch

Checklist:

### DIFF
--- a/doc/install_and_upgrade.md
+++ b/doc/install_and_upgrade.md
@@ -115,25 +115,32 @@ If you are on macOS 10.12 (Sierra) and encounter GHC panic while building, see
 this [issue](https://github.com/commercialhaskell/stack/issues/2577)
 
 
-On Apple silicon chip (`aarch64`/`arm64` architectures), the installation for stack itself or some packages (e.g. `network`) requiring C source compilation might fail with
-`configure: error: C compiler cannot build executables.` in which case you should pass `-arch arm64` as part of
-the `CFLAGS` environment variable, which will be then picked up by the c compiler of your choice.
+On Apple silicon chip (AArch64/ARM64) architectures, the installation of Stack
+or some packages (e.g. `network`) requiring C source compilation might fail with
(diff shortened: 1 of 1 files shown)


mpilgrem:
Thanks for the contribution. I merged it, but made some minor editorial changes in doing so. I'm trying to keep to lines of no more than 80 characters. For consistency of presentation, I'm also trying to conform the documentation to usage like 'Stack', 'C', 'AArch64', 'ARM64' - rather than lowercase alternatives. I split up some of the longer sentences into shorter ones.

May I ask, under Bash, what is the difference between `${CFLAGS}` and `${CFLAGS:-}`?

leoslf:
Hi @mpilgrem, thanks for merging the changes in the late night!  The editorial changes are cool, it looks a lot better now.

As for Bash, the writing `${VARIABLE_NAME:-DEFAULT_VALUE}` would substitute `DEFAULT_VALUE` when `VARIABLE_NAME` is not set, `${CFLAGS:-}` would have empty string as the default.
That was just in case that `CFLAGS` was not declared and make Bash know not to treat it as an error if `set -u` or `set -e` is on for completeness.

PR #5844, opened 2022-08-27
Fix #5323 New `--no-rsl-in-log` option (by default)

### LINKED ISSUE #5323: --verbosity info is too terse compared to --verbosity debug
### General summary/comments

I would like to have `--verbose`  be an alias for `--verbosity info` with some of the more informative output from `--verbosity debug` promoted to info.

eg I don't think `stack --verbose` should output all of RawSnapshotLayer: for most general stack users it is not useful and it makes for example `stack -v install` output too heavy.

### Steps to reproduce

For example:

1. run `stack -v install`

### Expected

details like resolver snapshot paths and compiler chosen, etc

### Actual

Outputs all of RawSnapshotLayer - too much information :)

### Stack version

2.3.1

### Method of installation

* Via cabal-install
* An unofficial package repository: Fedora

mpilgrem:
This pull request is draft and experimental. It would require pull request https://github.com/commercialhaskell/pantry/pull/55 to be incorporated into a released version of Pantry before it could be incorporated into Stack's code base.

Please also shortly describe how you tested your change. Bonus points for added tests! Tested by building and using Stack, including the `stack --verbose` flag.


mpilgrem:
`pantry-0.5.7` has been released, exposing `loadAndCompleteSnapshotRaw'`. This pull request now depends on that, and I have taken off its 'draft' status.

Issue #5709, opened 2022-04-12
Stack for aarch64 (again)


dboreham:
Apologies, I see from the history that this has come up again and again. I'd like to understand the state of Stack for ARM64. I'm attempting to use Ubuntu rather than macos. 

When I follow the batteries included instructions I get this:

```
$ curl -sSL https://get.haskellstack.org/ | sh
Detected Linux distribution: ubuntu

Sorry, currently only 64-bit (x86_64) Linux binary is available.
```

But I can see from various issues and posts that there is a GHC for ARM64, and there's even a CI job for Stack on ARM64: https://github.com/commercialhaskell/stack/actions/workflows/arm64-release.yml and also binary packages shipped in this release : https://github.com/commercialhaskell/stack/releases/tag/v2.7.1 (although internet lore claims they don't work).

I suppose I'm wondering what's the gap between the things that exist and `get.haskellstack.org` working above? Is there a mailing list, forum or somewhere that I can read to catch up? Also if I can help or contribute to make this happen I'm happy to do that. Thanks, hope this makes sense.

dboreham:
Small update. I tried the stack aarch64 binary in release 2.7.1. It runs, but only as far as:

```
$ stack build
Unable to find installation URLs for OS key: linux-aarch64-tinfo6
```

dboreham:
Well it turns out: we can have nice things after all.

After some poking around in the stack source, I was able to persuade it to build GHC from source using the binary from release 2.7.1.

The resulting newborn stack was able to successfully run `stack upgrade` and then succeeded in creating, building and running the hello world project:

```
$ stack exec helloworld-exe
someFunc
```

dboreham:
I'm not sure if this issue should be resolved with a doc fix (tell users which dependencies need to be present on the system in order to motivate stack to build GHC itself vs bailing out); or if the fix is to arrange for the appropriate pre-built binaries to be available for download.

dboreham:
For posterity these are the dependencies I found were necessary on a stock Ubuntu 20 system:

```
sudo apt install build-essential libnuma1 libnuma-dev libtinfo-dev libtinfo5 libtinfo6 libc6-dev
sudo apt install llvm*
sudo apt install clang ninja-build
sudo apt install zlib1g-dev
sudo apt install libgmp-dev
```

I found some of these from this thread: https://discourse.haskell.org/t/using-stack-on-raspberry-pi/2363/15

hasufell:
[...]

PR #5847, opened 2022-08-29
Update arm64.Dockerfile

mpilgrem:
Experimental, to see if the Dockerfile can be updated without breaking the GitHub workflow.

### DIFF
--- a/etc/dockerfiles/arm64.Dockerfile
+++ b/etc/dockerfiles/arm64.Dockerfile
@@ -16,18 +16,27 @@ RUN curl -sSL https://github.com/commercialhaskell/stack/releases/download/v2.7.
 ARG USERID
 ARG GROUPID
 
-RUN useradd --uid $USERID stack && mkdir -p /home/stack && chown -R stack /home/stack && usermod -aG $GROUPID stack
+RUN useradd --uid $USERID stack
+RUN mkdir -p /home/stack
+RUN chown -R stack /home/stack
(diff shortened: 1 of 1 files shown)


mpilgrem:
@snoyberg, I am having difficulty with the ARM64 workflow on this commit. It is failing with "ApplyLayer exit status 1 stdout:  stderr: write /home/stack/.stack/programs/aarch64-linux/ghc-9.2.4/lib/ghc-9.2.4/ghc-9.2.4/libHSghc-9.2.4_p.a: no space left on device", when it was passing earlier.

Given 'no space left on device', I wondered if it was the same cause as https://github.com/commercialhaskell/stack/issues/5709#issuecomment-1126784503 and https://github.com/commercialhaskell/stack/issues/5709#issuecomment-1126786666, where you cleaned up old Docker images and that resolved matters. Could you do the same again?

snoyberg:
> Given 'no space left on device', I wondered if it was the same cause as https://github.com/commercialhaskell/stack/issues/5709#issuecomment-1126784503 and https://github.com/commercialhaskell/stack/issues/5709#issuecomment-1126786666, where you cleaned up old Docker images and that resolved matters. Could you do the same again?

I logged into the system and saw it had a bunch of free space already. I'd already set up a recurring nightly prune, it seems like there were just enough builds in the past 24 hours to fill up the disk.

If you'd like access to the machine, send me your SSH public key on Slack and I'll give you access.

PR #5856, opened 2022-09-08
Integration tests in forks

jsoref:

Without these changes, the following things can go wrong in forks:
1. when one pulls in an update to `master`, the the workflow will run, and then get stuck waiting for a self-hosted-runner
2. if one _had_ a self-hosted-runner, signing would fail unless one also had a signing secret
3. if one had both of the above and had an organization that restricts permissions, then the workflow could fail because it doesn't ask for write permissions

Beyond that, the [GitHub's `actions/create-release`](https://github.com/actions/create-release) is unsupported, so this PR switches to a maintained action (one that actually documents the permission required required to create a release...).

Note: you should not merge this to this repository without creating a secret named `SELF_HOSTED_RUNNERS` that has a value like `arm64` (well, you can, but it'd result in arm64 builds being skipped until the secret is added).

Here's a run **with** the secret set (that would have passed if I had such a runner):
https://github.com/check-spelling/stack/actions/runs/3012421271

Here's a run **without** the secret set that passes:
https://github.com/check-spelling/stack/actions/runs/3012476622

image

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -151,6 +151,8 @@ jobs:
 
   github-release:
     name: Create GitHub release
+    permissions:
+      contents: write
     needs:
     - integration-tests
     - linux-arm64
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -199,11 +199,11 @@ jobs:
         echo ::set-output name=SOURCE_TAG::${GITHUB_REF#refs/tags/}
     - name: Create GitHub release (final)
       if: "!startsWith(github.ref, 'refs/tags/rc/')"
-      uses: actions/create-release@v1
+      uses: ncipollo/release-action@v1.10.0
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
-        tag_name: ${{ github.ref }}
+        tag: ${{ github.ref }}
         body: |
           See https://haskellstack.org/ for installation and upgrade instructions.
 
@@ -218,11 +218,11 @@ jobs:
         prerelease: false
     - name: Create GitHub release (release candidate)
       if: "startsWith(github.ref, 'refs/tags/rc/')"
-      uses: actions/create-release@v1
+      uses: ncipollo/release-action@v1.10.0
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
-        tag_name: ${{ github.ref }}
+        tag: ${{ github.ref }}
         body: |
           **Changes since v[INSERT PREVIOUS VERSION]:**
 
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -98,8 +98,11 @@ jobs:
   configuration:
     name: Check for self-hosted runners
     runs-on: ubuntu-latest
+    env:
+      CAN_SIGN: ${{ secrets.RELEASE_SIGNING_KEY != '' }}
     outputs:
       arm64: ${{ fromJSON(steps.runners.outputs.arm64) }}
+      can-sign: ${{ env.CAN_SIGN }}
     steps:
     - name: Check for hosted runners
       id: runners
@@ -176,6 +179,7 @@ jobs:
         name: Linux-ARM64
         path: _release
     - name: Hash and sign assets
+      if: needs.configuration.outputs.can-sign
       shell: bash
       env:
         RELEASE_SIGNING_KEY: ${{ secrets.RELEASE_SIGNING_KEY }}
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -112,7 +112,7 @@ jobs:
       run: |
         echo "::set-output name=runners::$SELF_HOSTED_RUNNERS"
         if echo "$SELF_HOSTED_RUNNERS" | grep -q 'arm64'; then
-          echo "::set-output name=arm64::[self-hosted, linux, ARM64]"
+          echo "::set-output name=arm64::['self-hosted', 'linux', 'ARM64']"
         else
           echo '::set-output name=arm64::"ubuntu-latest"'
         fi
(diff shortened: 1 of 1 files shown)


mpilgrem:
@jsoref, we apply linting of YAML files and it is 'opinionated' when it comes to indentation. When the rest of the CI is complete, please can you fix the indentation that it complains about to be consistent with the approach in the rest of the YAML file.

mpilgrem:
@jsoref, on your comments on Slack about testing, once this completes the CI. I will merge it and then test it here by creating a dummy 'version' tag (which I will later delete) and see if it creates a draft release as intended. Can I leave it to you to test that it works as intended in a fork of the repository?

jsoref:
Sure.

mpilgrem:
@jsoref, please can you explain further. The GitHub Actions are following the changed workflow files in the pull request, as I would expect, but it is not finding any of the secrets. If workflows in pull requests can't access secrets, then how does the proposed CI work?

jsoref:
Once the PR is merged, the version _here_ will be the one that would run the next time a commit is added to `master`, or similar it would run.

But while in the context of _this_ PR, the workflow being changed here isn't run.

On secret access, you're right. Oops, this won't work "well enough". It'd work for anyone self-testing, but indeed, it won't help for PRs from forks, and while it's _possible_ to make it work, the risk entailed in that change isn't worth it. Lemme rebase and include a version that special cases this specific repository `commercialhaskell/stack` -- it'll mean that you won't need the secret, but that anyone in a fork who chooses to set up a self-hosted runner will be able to set the secret themselves and use it.

mpilgrem:
On the changed workflow, I don't think that is correct. You can see from this pull request's checks that it is the changed workflow that is being run. I think the only stopping block is the access to the secrets. This https://docs.github.com/en/actions/security-guides/encrypted-secrets#accessing-your-secrets suggests that only people who have permission to edit the workflow file can use secrets in that workflow file.

jsoref:
Strange. I could have sworn that workflows from the PR branch weren't used. I wonder if that changed w/ the "require owner approval" bits...

Anyway, I think I'm done writing the logic change (got interrupted by a work question).

jsoref:
Fwiw, here's a run w/o the secret set: https://github.com/check-spelling/stack/actions/runs/3015938080

---

And here's a run w/ the secret set: https://github.com/check-spelling/stack/runs/8252116879?check_suite_focus=true -- if you expand `Check for self-hosted runners`, you'll see:

```
Run echo "::set-output name=runners::$SELF_HOSTED_RUNNERS"
##[debug]/usr/bin/bash --noprofile --norc -e -o pipefail /home/runner/work/_temp/fa19460b-d927-4ef4-bb3f-539cb4cad371.sh
::set-output name=runners::***
##[debug]steps.runners.outputs.runners='***'
::set-output name=***::"[\"self-hosted\", \"linux\", \"ARM64\"]"
##[debug]steps.runners.outputs.***='"[\"self-hosted\", \"linux\", \"ARM64\"]"'
```
The `***` are because the secret itself is `arm64` and that's being masked (note that the masks are case sensitive, which is why we can still see `ARM64`).

Right now, the: [Linux ARM64](https://github.com/check-spelling/stack/runs/8252120495?check_suite_focus=true#logs) job shows:
```
Started 2m 0s ago
Requested labels: ["self-hosted", "linux", "ARM64"]
Job defined at: check-spelling/stack/.github/workflows/integration-tests.yml@refs/heads/integration-tests-in-forks
Waiting for a runner to pick up this job...
```

Sadly, when I *cancel* the job, that detail is eaten by GitHub.

.github/workflows/integration-tests.yml
```diff
@@ -95,14 +95,45 @@ jobs:
...
+    steps:
+    - name: Check for hosted runners
+      id: runners
+      shell: bash
+      env:
+        SELF_HOSTED_RUNNERS: ${{ secrets.SELF_HOSTED_RUNNERS || (github.repository_owner == 'commercialhaskell' && 'arm64') }}
+      run: |
+        echo "::set-output name=runners::$SELF_HOSTED_RUNNERS"
+        if echo "$SELF_HOSTED_RUNNERS" | grep -q 'arm64'; then
+          echo '::set-output name=arm64::"[\"self-hosted\", \"linux\", \"ARM64\"]"'
```
mpilgrem, line 115: Are the quotation marks causing the problem with the hiatus at:

~~~text
Requested labels: ["self-hosted", "linux", "ARM64"]
Job defined at: commercialhaskell/stack/.github/workflows/integration-tests.yml@refs/pull/5856/merge
Waiting for a runner to pick up this job...
~~~

The original was as follows (no quotation marks):

~~~yaml
linux-arm64:
    name: Linux ARM64
    runs-on: [self-hosted, linux, ARM64]
~~~

mpilgrem (commented):
See my comment about quotaton marks causing problems with requesting a runner for Linux ARM64.

.github/workflows/integration-tests.yml
jsoref: I'm not sure. I don't have a personal self-hosted runner (I have one at work that I could play w/).

In general, I'd expect `fromJSON` to do more or less the right thing. Since I _think_ it's supposed to be a string or an array per:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on

> You can provide `runs-on` as a single string or as an array of strings. If you specify an array of strings, your workflow will run on a self-hosted runner whose labels match all of the specified `runs-on` values, if available. If you would like to run your workflow on multiple machines, use [`jobs.<job_id>.strategy`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategy).

I _guessed_ that this was the right notation to make it happy.

Let's assume that I guessed wrong.

If it turns out I guessed wrong, I'll file at least one bug against this -- I'll probably start w/ the docs repo, although I feel that this is a bug in the impl...

mpilgrem: I've made the change - lets see what happens ...

jsoref:
It seems to do the right thing:
[Linux ARM64](https://github.com/check-spelling/stack/runs/8254097823?check_suite_focus=true#logs)
```
Started 1m 37s ago
Requested labels: [self-hosted, linux, ARM64]
Job defined at: check-spelling/stack/.github/workflows/integration-tests.yml@refs/heads/integration-tests-in-forks
Waiting for a runner to pick up this job...
```

PR #5864, opened 2022-09-14
Update documentation for shell completion scripts (Bash, zsh, fish)

citrusmunch:
* [ ] test rendering of documentation

Mostly looking for feedback on if this is necessary to include.

I'd be happy to extend the scope of this to clean up zsh and bash as well, since I see no reason similar approaches can't be used for all three shells as per the [optparse-applicative documentation](https://github.com/pcapriotti/optparse-applicative#bash-zsh-and-fish-completions).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -42,6 +42,7 @@ Other enhancements:
 * `tools` subcommand added to `stack ls`, to list stack's installed tools.
 * `stack uninstall` shows how to uninstall Stack.
 * `--ghc-variant` accepts `int-native` as a variant.
+* Update documentaion for shell auto-completion with Bash and zsh. Added fish.
 
 Bug fixes:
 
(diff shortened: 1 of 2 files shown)


mpilgrem:
@citrusmunch, thanks! If you are able to extend the scope of this pull request, as you suggest, that would be much appreciated. In particular, I think the Z shell (Zsh) documentation is out of date.

citrusmunch:
Ready to go as far as content goes. I've included what I hope to be best practices for completion script locations, and tested each.

However, currently struggling to test the docs with the requirements listed [here](https://www.mkdocs.org/user-guide/deploying-your-docs/#local-files).

mpilgrem:
Thanks for thinking about testing the online documentation. I will ensure that it renders nicely, when I merge it. (I often have to do that by trial and error, but the errors are becoming fewer as I learn how the extended Markdown works.)

mpilgrem:
I merged at the command line.

PR #5869, opened 2022-09-18
Fix #5868 Try to repair trigger of steps in workflow

### LINKED ISSUE #5868: Integration tests not actioning 'download Linux-ARM64 artifact' or 'hash and sign assets'
@jsoref, unfortunately, when I pushed a tag for the v2.9.1 release, the workflow did not 'download Linux-ARM64 artifact' or 'hash and sign assets'. So, I've deleted that tag and will investigate.

It seems that conditions:

~~~yaml
if: contains(needs.configuration.outputs.runners, 'arm64')
~~~

and

~~~yaml
 if: needs.configuration.outputs.can-sign
~~~

are not being met. I think I see the problem with the first and will open a pull request. I am not sure why the second is not working. Perhaps it needs a more explicit test.

mpilgrem:
Also tries to fix artifacts being uploaded to 'wrong' draft GitHub release.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -178,13 +178,13 @@ jobs:
         name: Windows
         path: _release
     - name: Download Linux-ARM64 artifact
-      if: contains(needs.configuration.outputs.runners, 'arm64')
+      if: needs.configuration.outputs.test-arm64 == 'true'
       uses: actions/download-artifact@v3
       with:
         name: Linux-ARM64
         path: _release
     - name: Hash and sign assets
-      if: needs.configuration.outputs.can-sign
+      if: needs.configuration.outputs.can-sign == 'true'
       shell: bash
       env:
         RELEASE_SIGNING_KEY: ${{ secrets.RELEASE_SIGNING_KEY }}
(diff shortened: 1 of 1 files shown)


mpilgrem:
My first hypothesis for a fix did not work. I'm now trying an express dependency on the `configuration` job, to see if that is what stopping `needs.configuration.outputs.*` from yielding the expected value.

.github/workflows/integration-tests.yml
```diff
@@ -157,6 +157,7 @@ jobs:
     permissions:
       contents: write
     needs:
+    - configuration
```
jsoref, line 160: hmm, interesting. I expect this is an important fix, although, as usual, I'm disappointed in the documentation...

https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context

> The needs context contains outputs from all jobs that are defined as a dependency of the current job. For more information on defining job dependencies, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idneeds)."

Technically the `github-release` job is defined as dependent on the `linux-arm64` job, and that job `needs: configuration`. But it seems likely that this isn't what the developers thought when they implemented it.

jsoref: Indeed, this does appear to be a critical bit, as demonstrated by: https://github.com/jsoref/cascading-needs/actions/runs/3079668494

Reported as https://github.com/github/docs/issues/20701#issue-1377326735

PR #5883, opened 2022-09-25
import `Data.List` qualified as "-Wcompat-unqualified-imports" says

Vlix:
Pretty simple import fixes.
Importing `Data.List` qualified wherever it's used  (and removing it when only `map` is used, because it's from `base`)


mpilgrem:
Thanks!

PR #5884, opened 2022-09-25
Stack upgrade binary linux fix

Vlix:
Check for relative paths in `Location` headers of redirect responses when trying to get a new `stack` binary when using `stack upgrade` and re-add the domain when it begins with a `/`.

Can't really test this, but it's an easy/obvious enough fix, methinks.

### DIFF
--- a/hie.yaml
+++ b/hie.yaml
@@ -1,0 +1,20 @@
+cradle:
+  multi:
+    - path: "./.stack-work"
+      config: { cradle: { none } }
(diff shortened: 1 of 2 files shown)


Vlix:
In  my opinion, this change only _also_ allows paths instead of only absolute URLs to be used as values of the `Location` header, so I see it as just a more robust implementation. But I'm fine with whatever gets decided.

PR #5906, opened 2022-10-19
Import stackage nightly locally minus stack & pantry.

philderbeast:
Maintenance for the cabal build of stack by importing the latest stackage nightly for ghc-9.2.4 but referencing it locally so that we can comment out stack and pantry constraints to avoid the following conflicts.

```
> cabal build all --disable-tests
Resolving dependencies...
Error: cabal: Could not resolve dependencies:
[__0] next goal: stack (user goal)
[__0] rejecting: stack-2.10.0 (constraint from project config
stackage/ghc-9.2.4/cabal.config requires ==2.9.1)
```

```
> cabal build all --disable-tests
Resolving dependencies...
Error: cabal: Could not resolve dependencies:
[__0] trying: stack-2.10.0 (user goal)
[__1] next goal: pantry (dependency of stack)
[__1] rejecting: pantry-0.6.0 (constraint from project config
stackage/ghc-9.2.4/cabal.config requires ==0.5.7)
```


mpilgrem:
Is this the only way to achieve this with Cabal (the tool)? It is unattactive to have to mainain a 3,000+ line configuration file. Can it not, at least, be tailored to Stack's 198 dependencies (which can be listed for the project with Stack's `stack ls dependencies` command)? (If it can, then perhaps a subcommand could be added to `stack ls dependencies` that would also list dependencies in `<package_name> == <version_number>` format, making that sort of thing easier.)

mpilgrem:
@philderbeast, actually Stack already has the feature for listing dependencies in the format of Cabal bounds:

`stack ls dependencies text --separator " == "`

mpilgrem:
@philderbeast, for your information, I've just bumped Stack in the `master` branch to `pantry >= 0.7.0` and `fsnotify >= 0.4.1.0`.

philderbeast:
I added another command to `ls dependencies` in the cabal constraints format. Added a makefile to update those constraints. Kept the location as `stackage/ghc-9.2.4/cabal.config` as most constraints will be coming from stackage. Let me know if you'd prefer another location such as `cabal/ghc-9.2.4/cabal.config`.

```
> stack run -- ls dependencies --help
...
Usage: stack ls dependencies [COMMAND | [--separator SEP] [--[no-]license] 
                               [--[no-]external] [--[no-]include-base] 
                               [--depth DEPTH] [--prune PACKAGES] [TARGET] 
                               [--flag PACKAGE:[-]FLAG] [--test] [--bench] 
                               [--global-hints]]

  View the dependencies
...
Available commands:
  text                     Print dependencies as text (default)
  tree                     Print dependencies as tree
  json                     Print dependencies as JSON
  constraints              Print exact cabal constraints

> stack run -- ls dependencies constraints
...
constraints:
  , Cabal ==3.6.3.0
  , Cabal-syntax ==3.6.0.0
  , Glob ==0.10.2
  , OneTuple ==0.3.1
  , QuickCheck ==2.14.2
  , StateVar ==1.2.2
  , aeson ==2.0.3.0
...
```

philderbeast:
@hasufell could stack2cabal make use of this?

cabal.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -eu
```
mpilgrem, line 2: Stack locates its 'helper' scripts in `etc/scripts`. It may help to give the file a fuller name, more indicative of its function - see the other script file names for examples.

It would also be helpful to document the intended use of the script at its start (even though it is short).

cabal.sh
```diff
@@ -0,0 +1,4 @@
...
+
+stack build && stack run -- ls dependencies constraints > "$1"
```
mpilgrem, line 4: These two commands look wrong to me. Should it not just be `stack ls dependencies constraints > "$1"`? Stack does not have to build Stack to list its dependencies.

See my comment below about the naming of the new subcommand.

(`stack run`, by the way, builds an executable and runs it - see https://docs.haskellstack.org/en/stable/GUIDE_advanced/#the-stack-run-command.)

Makefile
```diff
@@ -0,0 +1,9 @@
+GHC := 9.2.4
```
mpilgrem, line 1: I did not follow the purpose of this file. 'Makefile' files are not used in the Stack repository.

cabal.project
```diff
@@ -9,7 +9,7 @@
 
 -- `import:` is only available to users of Cabal (the tool) >= 3.8.1.0.
 
-import: https://www.stackage.org/nightly-2022-09-05/cabal.config
+import: stackage/ghc-9.2.4/cabal.config
```
mpilgrem, line 12: Now that there is no connection between the `cabal.config` file and a Stackage snapshot, I suggest `cabal.config` is moved the root directory of the repository, to sit beside the `stack.yaml`.

The Haskell comment immediately above is no longer apt - perhaps it should be replaced by a reference to the new `stack ls dependencies` subcommand or the script that makes use of it?

src/Stack/Options/DotParser.hs
```diff
@@ -102,4 +105,5 @@ listDepsOptsParser = subparser
                      (  formatSubCommand "text" "Print dependencies as text (default)" listDepsTextParser
                      <> formatSubCommand "tree" "Print dependencies as tree" listDepsTreeParser
                      <> formatSubCommand "json" "Print dependencies as JSON" listDepsJsonParser
+                     <> formatSubCommand "constraints" "Print exact cabal constraints" listDepsConstraintsParser
```
mpilgrem, line 108: Cabal is now referred to as 'Cabal', rather than 'cabal', consistently in Stack documentation.

I suggest this read 'Print dependencies as Cabal constraints`, to follow the pattern of the other subcommands.

I would place the new subcommand immediately after `text`, given the similarity between 'text' and the new subcommand.

I suggest that the new subcommand is named `cabal`, rather than `constraints` -  it is shorter (always good) and it more directly describes its intended use - that is, with Cabal (the tool).

stackage/ghc-9.2.4/cabal.config
```diff
@@ -0,0 +1,200 @@
+constraints:
```
mpilgrem, line 1: As above, I would move `cabal.config` to sit beside the `stack.yaml` in the repository root.

mpilgrem (commented):
@philderbeast, thanks. I had some thoughts as I read the code, noted above.

cabal.sh
mpilgrem: That said, if I am correct about the commands in the script (see below) is not a single command script overkill?

cabal.sh
philderbeast: `stack ls dependencies` is such a good lightweight command. I've removed the unnecessary `stack build` that preceded it in the script.

Makefile
philderbeast: I was hoping for a lightweight way to maintain the cabal.config file. I'm open to other suggestions to do the same.

cabal.project
philderbeast: I'd like to leave the comment there in the project about the limitations as stack-only users can be quite surprised by this. I've added a comment mentioning that `ls dependencies constraints` supplies the `cabal.config` constraints.

stackage/ghc-9.2.4/cabal.config
philderbeast: Done.

src/Stack/Options/DotParser.hs
philderbeast: All done.

mpilgrem:
@philderbeast, I merged this manually, as I tweaked it a little. I also added some online documentation.

philderbeast:
Thanks for the merge @mpilgrem. The additions to the docs look good.

PR #5910, opened 2022-10-20
enable allow-newer to only apply to specific packages

intolerable:


src/Stack/Types/Config.hs
```diff
@@ -441,6 +443,21 @@ instance FromJSON ApplyGhcOptions where
             "everything" -> pure AGOEverything
             _ -> fail $ "Invalid ApplyGhcOptions: " ++ show t
 
+data AllowNewer
+  = AllowNewerAll
+  | AllowNewerNone
+  | AllowNewerPackages [String]
```
mpilgrem, line 449: I am wondering if it would be better if the type of the final data constructor was `![PackageName]` (ie `Distribution.Types.PackageName.PackageName` from the `Cabal` package), with consequent changes elsewhere. The `PackageName` type is used throughout Stack's code base.

src/Stack/Types/Config.hs
```diff
@@ -441,6 +443,21 @@ instance FromJSON ApplyGhcOptions where
...
+  deriving (Show, Read, Eq, Ord)
+
+instance FromJSON AllowNewer where
```
mpilgrem, line 452: I am wondering if what is needed is `WithJSONWarnings AllowNewer` instance of `FromJSON`.

src/Stack/Types/Config.hs
```diff
@@ -441,6 +443,21 @@ instance FromJSON ApplyGhcOptions where
...
+  parseJSON v =
+    case v of
+      Bool True -> pure AllowNewerAll
+      Bool False -> pure AllowNewerNone
+      String "all" -> pure AllowNewerAll
+      String "none" -> pure AllowNewerNone
+      _ -> AllowNewerPackages <$> parseJSON v
```
mpilgrem, line 459: I am wondering if, rather than the current catch-all `_`, `parseJSON v` should distinguish valid arrays from other `v` and then treat the residual invalid `v` differently.

test/integration/tests/allow-newer-specific-packages/files/stack.yaml
```diff
@@ -0,0 +1,4 @@
+resolver: nightly-2022-10-18
```
mpilgrem, line 1: The integration tests are currently standardised around `nightly-2020-09-05`. It makes them easier to update, if they all remain in line with each other.

mpilgrem (commented):
@intolerable, many thanks. It looks pretty good to me, but I have yet to test it. A few thoughts occurred to me as I read the code changes, which I have indicated.

Thanks also for including an integration test, but is there a missing `Main.hs` associated with the new files in `files`?

test/integration/tests/allow-newer-specific-packages/files/package.yaml
```diff
@@ -0,0 +1,8 @@
+name: foo
+version: 0
+
+dependencies:
+  - base >= 4.14 && < 4.15
```
mpilgrem, line 5: The YAML linter is opinionated when it comes to indentation. Don't indent this, and it will pass.

test/integration/tests/allow-newer-specific-packages/files/stack.yaml
```diff
@@ -0,0 +1,4 @@
...
+
+allow-newer:
+  - foo
```
mpilgrem, line 4: As above, the YAML linter is opinionated when it comes to indentation.

mpilgrem:
The integration CI is failing on x86_64/Linux because of an upstream change in Git. This is fixed in the `master` branch, so should go away if you rebase on `master`.

mpilgrem:
@intolerable, I have one other thought:

By changing the `allow-newer:` key, older versions of Stack will fall over with YAML configuration files that make use of the enhanced functionality. That is an unattractive outcome.

If, however, you were to introduce the array of dependencies in a further top-level key (`allow-newer-deps:`, for example), then older versions of Stack would warn about the unrecognised key but still continue to try to build. So, sophisticated users of Stack would configure:
~~~yaml
allow-newer: true
allow-newer-deps:
- foo
- bar
~~~

The separation would also allow easier toggling off and on of the list of dependencies.

src/Stack/Types/Config.hs
intolerable: Fixed, thank you

test/integration/tests/allow-newer-specific-packages/files/stack.yaml
intolerable: Fixed, thank you

test/integration/tests/allow-newer-specific-packages/files/package.yaml
intolerable: Fixed both of these, thank you

intolerable:
@mpilgrem Thank you for your feedback and suggestions. I have implemented most of your suggestions, but have questions about the other `FromJSON` instance-related ones:

> I am wondering if what is needed is `WithJSONWarnings AllowNewer` instance of `FromJSON`.

I took a look at `WithJSONWarnings`, and it appeared to me to be designed primarily for handling `Object` values. Was that assumption incorrect, and should I be using it anyway, even when the `Value` in this case is an `Array`?

> I am wondering if, rather than the current catch-all `_`, `parseJSON v` should distinguish valid arrays from other `v` and then treat the residual invalid `v` differently.

I've changed this code slightly, since I'm now using a separate `allow-newer-deps` field, so I don't think this is still relevant. Please correct me if I've misunderstood this.

There is however an minor issue with the current implementation, in that it simply parses the array of strings from YAML and calls `Distribution.Types.PackageName.mkPackageName`, a function which does no sanity checking whatsoever. It's very possible that a user could put a string in here which is not a valid package name. As far as I can tell, though, this is unlikely to cause issues, since the list is simply being used as a whitelist to match against, rather than being fed to anything else. In the future, if I add support for the alternative cabal-style syntax, I'll ensure that the parsing validates each `PackageName` correctly.

This brings me to another question I have regarding future changes and your comment suggesting improvements to backwards compatibility:

With this change implemented as a new field (`allow-newer-deps`), it's still very possible (though I don't think very likely) that a project with a `stack.yaml` with `allow-newer: true` and `allow-newer-deps: [...something]` will build on versions of Stack which understand `allow-newer-deps` but not on older versions which don't understand this field. Is this a major issue, or is some small level of breakage acceptable? I don't think there is a way to guarantee that any project using this feature will build on older versions of Stack, but requiring `allow-newer` to be enabled in order to use `allow-newer-deps` gets most of the way there.

In addition, this change allows only package names to be included in the list of `allow-newer-deps`, and if I extend the syntax in future to allow for more specificity, versions of stack with this change implemented will not support the next syntax changes, preventing projects utilizing this feature from building on older versions of Stack. Is there a good pathway forwards that will allow me to incrementally improve this feature without breaking users' workflows? Should this feature be documented as `experimental`, given that it may change in awkward ways in the future?

Final question: `allow-newer` is not a very good name for this field, as it suggests something that is very similar to its actual behaviour ("ignore version _upper_ bounds"), but can sometimes have very unintended consequences (when it ignores version _lower_ bounds). While it makes sense for `allow-newer` to mirror the (confusingly-named) Cabal option, can we give this field a different name in Stack that more accurately describes its effects?

mpilgrem:
@intolerable, thanks for considering my comments. My answers to your questions:

> I took a look at `WithJSONWarnings`, and it appeared to me to be designed primarily for handling `Object` values. Was that assumption incorrect, and should I be using it anyway, even when the `Value` in this case is an `Array`?

I need to look at `WithJSONWarnings` more closely myself. Perhaps I had misunderstood its use.

> With this change implemented as a new field (`allow-newer-deps`), it's still very possible (though I don't think very likely) that a project with a `stack.yaml` with `allow-newer: true` and `allow-newer-deps: [...something]` will build on versions of Stack which understand `allow-newer-deps` but not on older versions which don't understand this field. Is this a major issue, or is some small level of breakage acceptable?

I think that sort of breakage is acceptable. The key will be to make sure that things are well-documented in Stack's online documentation.

> if I extend the syntax in future to allow for more specificity, versions of stack with this change implemented will not support the next syntax changes, preventing projects utilizing this feature from building on older versions of Stack. Is there a good pathway forwards that will allow me to incrementally improve this feature without breaking users' workflows? Should this feature be documented as experimental, given that it may change in awkward ways in the future?

Could you elaborate on the possible future syntax extension? I was, perhaps naively, assuming that if you are telling Stack to ignore Cabal's bounds or not that was specific to a package name. However, you can certainly flag 'experimental' in Stack's in-tool or online documentation if you think that will help users. Past examples are the `stack query` command or https://docs.haskellstack.org/en/stable/yaml_configuration/#building-ghc-from-source-experimental.

> Final question: allow-newer is not a very good name for this field, as it suggests something that is very similar to its actual behaviour ("ignore version upper bounds"), but can sometimes have very unintended consequences (when it ignores version lower bounds). While it makes sense for allow-newer to mirror the (confusingly-named) Cabal option, can we give this field a different name in Stack that more accurately describes its effects?

I think the short answer is yes, and this is how we could do it. (1) introduce synonyms, and give the synomyns precedence in the documentation and deprecate `allow-newer`; (2) after a suitable period - which may be years! - remove the deprecated `allow-newer`. What do you think the synonyns should be? How about `ignore-bounds` and `ignore-bounds-deps`?

intolerable:
> Could you elaborate on the possible future syntax extension? I was, perhaps naively, assuming that if you are telling Stack to ignore Cabal's bounds or not that was specific to a package name.

Cabal's `allow-newer` in `cabal.project` syntax (described here in the [Cabal manual](https://cabal.readthedocs.io/en/3.8/cabal-project.html#cfg-field-allow-newer)) allows the user to specify which version bounds should be ignored in a very granular way, for example:

- `foo` (equivalent to `*:foo`) ignores specified version bounds for package `foo` in all dependencies
- `foo:bar` ignores package `foo`'s specified version bounds for package `bar`
- `foo-1.2.3:bar` ignores package release `foo-1.2.3`'s specified version bounds for package `bar`

as well as some wildcards like `*:*` (equivalent to Stack's `allow-newer: true` option). I'd like to add support for these to Stack, as I've found them very useful recently, although I appreciate that they may be less useful for Stack given its use of snapshots.

> I think the short answer is yes, and this is how we could do it. (1) introduce synonyms, and give the synomyns precedence in the documentation and deprecate `allow-newer`; (2) after a suitable period - which may be years! - remove the deprecated `allow-newer`. What do you think the synonyns should be? How about `ignore-bounds` and `ignore-bounds-deps`?

I think `ignore-bounds` or something very similar is a good name for this option. For implementing this change with minimal disruption, would this work as a plan?:

1. mark `allow-newer-deps` as `experimental` (as part of this change now)
2. add support to `allow-newer-deps` for the other Cabal syntax constructs
3. add a new option `ignore-bounds: $whatever`, synonymous with using both `allow-newer: true` and `allow-newer-deps: $whatever`, which covers the behaviour of both `allow-newer` and `allow-newer-deps` (like how Cabal's `allow-newer` covers both of these use cases`)
4. mark `allow-newer` and `allow-newer-deps` both as deprecated

This would keep `allow-newer-deps` as a useful experimental option until we fully mirror the Cabal syntax, and give a clear migration path for users of this experimental feature to a future supported `ignore-bounds` option which subsumes both `allow-newer` and `allow-newer-deps` (at which point the questionably-named `allow-newer` options can be deprecated and then removed).

Also: I don't think I remembered to rebase from `master`, so the CI builds you've triggered will have failed, I think :disappointed:

src/Stack/Build/ConstructPlan.hs
```diff
@@ -747,6 +744,19 @@ addPackageDeps package = do
                                         warn_ "trusting snapshot over Cabal file dependency information"
                                         pure True
                                     else pure False
+                        if allowNewer of
```
mpilgrem, line 747: The CI is currently failing because of the typo in this line.

mpilgrem:
@intolerable, my thoughts your further comments above:

> I've found them very useful recently, although I appreciate that they may be less useful for Stack given its use of snapshots.

My assumption was this: a Stack user says, 'I want to build with this specific set of package versions (that is, the specified snapshot)'. A Cabal file says, 'Sorry, I will not build with package ABC of version XYZ (that is, the specified Cabal bounds)'. The Stack user then says, 'Ignore what Cabal files say about bounds; I know what I am doing (`allow-newer: true`)'.

Your pull request allows the Stack user to say, 'OK, I want to trust what the Cabal files say generally about bounds, but I know what I am doing with regards to package ABC'.

I see that your further syntax will allow the Stack user to be even more precise in expressing the area in which they know what they are doing when ignoring what Cabal files advise about bounds.

> would this work as a plan?

Looks good to me.

src/Stack/Build/ConstructPlan.hs
intolerable: fixed, bad hlint fix :)

intolerable:
I've updated the documentation to make it clear that `allow-newer-deps` should be considered experimental at the moment, so I think it should be ready to merge now :+1:

Issue #5920, opened 2022-10-30
Refactoring proposition about Package/LocalPackage


theobat:
Hello @mpilgrem,

While trying to renew some effort on #5460, #5504, I envisoned a much needed refactoring first and a more iterative approach. The current merge request is old and too ambitious to land in any reasonable fashion.

So my proposal this time is much more limited and yet I believe very important. My assessment about the Package and LocalPackages type is that they are Cabal 1.0 oriented and they have very ad-hoc code for named components (tests, executables, benchmarks, libraries). This has very bad impacts on the codebase in many places and it should be fixed before trying any component based builds.

Here are some examples of what I mean :
```hs
-- in the definition of LocalPackage in ./src/Stack/Types/Package.hs
data LocalPackage = LocalPackage
    { lpPackage       :: !Package
    -- ^ The @Package@ info itself, after resolution with package flags,
    -- with tests and benchmarks disabled
    , lpComponents    :: !(Set NamedComponent)
    -- ^ Components to build, not including the library component.
    , lpUnbuildable   :: !(Set NamedComponent)
    -- ^ Components explicitly requested for build, that are marked
    -- "buildable: false".
    , lpWanted        :: !Bool -- FIXME Should completely drop this "wanted" terminology, it's unclear
    -- ^ Whether this package is wanted as a target.
    , lpTestDeps      :: !(Map PackageName VersionRange)
    -- ^ Used for determining if we can use --enable-tests in a normal build.
    , lpBenchDeps     :: !(Map PackageName VersionRange)
    -- ^ Used for determining if we can use --enable-benchmarks in a normal
    -- build.
    , lpTestBench     :: !(Maybe Package)
    -- ^ This stores the 'Package' with tests and benchmarks enabled, if
    -- either is asked for by the user.
```
And the main creation of such type is enlightened by an edifying comment : 
```hs
        -- We resolve the package in 4 different configurations:
        --
        -- - pkg doesn't have tests or benchmarks enabled.
        --
        -- - btpkg has them enabled if they are present.
        --
        -- - testpkg has tests enabled, but not benchmarks.
        --
        -- - benchpkg has benchmarks enabled, but not tests.
        --
        -- The latter two configurations are used to compute the deps
        -- when --enable-benchmarks or --enable-tests are configured.
        -- This allows us to do an optimization where these are passed
        -- if the deps are present. This can avoid doing later
        -- unnecessary reconfigures.
        pkg = resolvePackage config gpkg
        btpkg
[...]

PR #5926, opened 2022-11-04
Simple refactoring

theobat:
This is a preparation work for #5920. It really is a simple review of unused fields and I had to move all the Dependency related types to a dedicated module because having them in Package made circular imports when moving dependencies to Components. 

@mpilgrem

### DIFF
--- a/src/Stack/Build/ConstructPlan.hs
+++ b/src/Stack/Build/ConstructPlan.hs
@@ -684,7 +684,8 @@ addEllipsis t
 addPackageDeps :: Package -> M (Either ConstructPlanException (Set PackageIdentifier, Map PackageIdentifier GhcPkgId, IsMutable))
 addPackageDeps package = do
     ctx <- ask
-    deps' <- packageDepsWithTools package
+    checkAndWarnForUnknownTools package
+    let deps' = packageDeps package
     deps <- forM (Map.toList deps') $ \(depname, DepValue range depType) -> do
         eres <- addDep depname
         let getLatestApplicableVersionAndRev :: M (Maybe (Version, BlobKey))
@@ -925,8 +926,8 @@ psLocation PSRemote{} = Snap
 
 -- | Get all of the dependencies for a given package, including build
 -- tool dependencies.
-packageDepsWithTools :: Package -> M (Map PackageName DepValue)
-packageDepsWithTools p = do
+checkAndWarnForUnknownTools :: Package -> M ()
+checkAndWarnForUnknownTools p = do
     -- Check whether the tool is on the PATH before warning about it.
     warnings <- fmap catMaybes $ forM (Set.toList $ packageUnknownTools p) $
       \name@(ExeName toolName) -> do
@@ -938,7 +939,7 @@ packageDepsWithTools p = do
             Left _ -> pure $ Just $ ToolWarning name (packageName p)
             Right _ -> pure Nothing
     tell mempty { wWarnings = (map toolWarningText warnings ++) }
-    pure $ packageDeps p
+    pure ()
 
 -- | Warn about tools in the snapshot definition. States the tool name
 -- expected and the package name using it.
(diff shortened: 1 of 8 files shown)


theobat:
@mpilgrem this one is ready to be merged it's a cleaning worth the merge no matter what. (unused functions etc)

src/Stack/Build/ConstructPlan.hs
```diff
@@ -52,6 +52,7 @@ import           System.Environment (lookupEnv)
 import           System.IO (putStrLn)
 import           RIO.PrettyPrint
 import           RIO.Process (findExecutable, HasProcessContext (..))
+import Stack.Types.Dependency (DepValue(DepValue), DepType (AsLibrary))
```
mpilgrem, line 55: Please could this `import` line adopt the same format as used elsewhere in the module.

src/Stack/Build/Haddock.hs
```diff
@@ -3,7 +3,7 @@
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
-{-# LANGUAGE RecordWildCards       #-}
+{-# LANGUAGE NamedFieldPuns       #-}
```
mpilgrem, line 6: I did not follow what prompted this change.

src/Stack/Build/Source.hs
```diff
@@ -315,25 +315,13 @@ loadLocalPackage pp = do
...
             , packageConfigEnableBenchmarks = not $ Set.null benches
             }
-        testconfig = config
-            { packageConfigEnableTests = True
-            , packageConfigEnableBenchmarks = False
-            }
-        benchconfig = config
-            { packageConfigEnableTests = False
-            , packageConfigEnableBenchmarks = True
-            }
```
mpilgrem, line 325: For my benefit, could you explain further why these can be safely deleted? I am not across the architectural considerations and would appreciate some help in understanding them.

src/Stack/Package.hs
```diff
@@ -77,6 +77,7 @@ import           System.IO.Error
 import           RIO.Process
 import           RIO.PrettyPrint
 import qualified RIO.PrettyPrint as PP (Style (Module))
+import Stack.Types.Dependency (DepValue(..), DepType (..))
```
mpilgrem, line 80: Please could this `import` line adopt the same format as used elsewhere in the module.

src/Stack/Types/Dependency.hs
```diff
@@ -0,0 +1,29 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Stack.Types.Dependency where
```
mpilgrem, line 3: Stack's other modules tend to have express export lists.

src/Stack/Types/Dependency.hs
```diff
@@ -0,0 +1,29 @@
...
+
+
+-- | The value for a map from dependency name. This contains both the
+-- version range and the type of dependency, and provides a semigroup
+-- instance.
+data DepValue = DepValue
+  { dvVersionRange :: !VersionRange
+  , dvType :: !DepType
+  }
+  deriving (Show,Typeable)
```
mpilgrem, line 17: Please could there be a space after the comma? That would be consistent within this module and with other Stack modules. (I appreciate that consistency was missing from the code you moved!)

mpilgrem (commented):
Many thanks. Some (minor, I think) thoughts occured to me as I read your proposed changes.

src/Stack/Build/Haddock.hs
theobat: This change was motivated by a simplified tracking of what is used from the Package type (as I intend to change this type to have a more cabal-like structure as stated in the issue, and thus I have to know what is used where). It's not strictly necessary, but I believe RecordWildCards is a bad and unnecessary extension (especially in this case, when 10+ fields are used it's questionable but obviously unseful).

src/Stack/Build/Source.hs
theobat: Honestly I don't know why it existed in the first place, as it's never been used as far as I know.

theobat: The fields have been there for a very long time  (7+ years) and I don't know at which point they stopped being used.
The reason why its important to remove them is that, we should have dependencies tracked at the component level for mixins/backpack support, not at the LocalPackage/Package level.

mpilgrem:
@theobat, thanks for the explanations.

PR #5942, opened 2022-11-13
refactoring: bring package file gathering logic into dedicated module…

theobat:
Ok so @mpilgrem, this is also a preparation work for #5920.

The idea is that, bringing a proper component design to the Package type will allow us to grab files by component instead of doing the current package level file gathering which consists of :

- When processing the raw Cabal 'PackageDescription' data, create a function which holds a reference to this type.
- When calling the function iterate on many cabal based info from the retained 'PackageDescription', these iterations are mostly at the component level.

This is at a very high level what the 'packageDescModulesAndFiles' is doing.

Now, I didn't want to do all this Module creation and code moving, but I faced very compelling reasons to do so while reviewing the code :

- Splitting file gathering by component (while keeping the overall existing architecture and logic) imposed the use of many utility functions wich were in the Package module. importing them in a (incoming) Component module or in a ComponentFile module created circular imports.
- The File gathering logic is not using any of the overall Package type design and values, so it's weird to keep this in the same place to begin with.
- The file gathering logic take a huge place in the Package module wihtout any hierarchy for functions rendering the overall Package module hard to navigate.

Overall I don't think the current design for file gathering logic is legitimate. Once I bring proper structured components data with BuildInfo and all, I expect that the packageFiles field of the Package type can be simplified to a simple function taking a Package (instead of a field). But this is a side note, and we'll go step by step.

Something like :
```hs
-- Assuming Package has all the cabal level info needed for gathering the files
getPackageFiles :: forall env. HasEnvConfig env
                      => Path Abs File
                      -> Package
                      -> RIO env
                           (Map NamedComponent (Map ModuleName (Path Abs File))
                           ,Map NamedComponent [DotCabalPath]
                           ,Set (Path Abs File)
                           ,[PackageWarning])
```

### DIFF
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -85,6 +85,7 @@ import           Stack.Types.Config
 import           Stack.Types.GhcPkgId
 import           Stack.Types.NamedComponent
 import           Stack.Types.Package
+import           Stack.Types.PackageFile
 import           Stack.Types.Version
 import qualified System.Directory as D
 import           System.Environment (getExecutablePath, lookupEnv)
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -36,6 +36,7 @@ import              Stack.Types.Build
 import              Stack.Types.Config
 import              Stack.Types.NamedComponent
 import              Stack.Types.Package
+import              Stack.Types.PackageFile
 import              Stack.Types.SourceMap
 import              System.FilePath (takeFileName)
 import              System.IO.Error (isDoesNotExistError)
--- a/src/Stack/ComponentFile.hs
+++ b/src/Stack/ComponentFile.hs
@@ -1,0 +1,522 @@
+{-# LANGUAGE NoImplicitPrelude   #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE FlexibleInstances   #-}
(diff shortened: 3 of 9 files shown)


mpilgrem:
@theobat, do you participate in the Haskell Foundation's Slack workspace? I ask because there is a #stack-collaborators channel which might be a better place to discuss your more 'architectural' thinking, as you move towards that.

In respect of this pull request, I may be able to consider it this evening.

theobat:
@mpilgrem I don't have acess to the slack but I must be invited right ? As for the PR it's ok feel free to take the time to digest my comments (although the PR is not doing much, it's a simple module creation/code moving, it's still motivated by by a somewhat "deeper" intention).

mpilgrem:
@theobat, on HF's Slack workspace, I think the invitation is an open one - try the link at the bottom right of this web page: https://haskell.foundation/contact/.

mpilgrem:
@theobat, one other thought: I've made a large commit (f25bd3d3a70ca703545be83a0d3648bbe72efccf) that effects a large number of modules. It may not interfere with what you are doing but, in case it does, you may want to rebase on the current master branch.

mpilgrem:
@theobat, I am pretty sure that I have merged this, manually, into the master branch. Manually, because I wanted to make some purely-formatting changes to the 'new' source code files - their being new provided an opportunity to do that. Those fomatting changes are mostly related to using two spaces, rather than four, for indentation and aiming to avoid longer lines of code. I reordered some of the import lists to be alphabetical.

theobat:
ok great @mpilgrem

PR #5982, opened 2022-12-21
Better heuristic for detecting Debian derivatives

### LINKED ISSUE #5881: Stack can't install GHC 9.4.x on Debian 11 (Bullseye)
# WORKAROUND

* Add `ghc-build: standard` to `~/.stack/config.yaml`
* Make sure libtinfo5 is available, by e.g. `apt install libtinfo5`

### Steps to reproduce

1. `docker run -it debian:bullseye`
2. `apt update && apt install curl`
3. `curl -sSL https://get.haskellstack.org/ | sh`
4. `stack new --resolver nightly`
5. Modify `stack.yaml` to say `compiler: ghc-9.4.2`
6. `stack run`

### Expected

I expected to see some dependency failure because the current nightly snapshot probably doesn't work on GHC-9.4.2.

### Actual

Fails during installation of the Fedora 33 build of GHC:

```
Downloaded ghc-tinfo6-9.4.2.                                      
ar: conftest.a: No such file or directory                                       
/root/.stack/programs/x86_64-linux/ghc-tinfo6-9.4.2/lib/ghc-9.4.2/bin/ghc-pkg: /lib/x86_64-linux-gnu/libc.so.6:
version `GLIBC_2.32' not found (required by /root/.stack/programs/x86_64-linux/ghc-tinfo6-9.4.2/lib/ghc-9.4.2/bin/../lib/x86_64-linux-ghc-9.4.2/libHSrts-1.0.2-ghc9.4.2.so)
```

Seems like there are GHC releases for [various versions of Debian](https://www.haskell.org/ghc/download_ghc_9_4_2.html). [Debian 9 is referenced in stack-setup.yaml](https://github.com/commercialhaskell/stackage-content/blob/4688495868b46d7ce3752df9d026341431df75f8/stack/stack-setup-2.yaml#L687) for the `linux64` variant, maybe it works with that one. (*EDIT: It does, this is what the workaround at the top does*) But I think it should work out of the box. I don't think I can use the `ghc-variant` option to select a Debian build, since it seems I can only append to `linux64-tinfo6`. (*EDIT: Use ghc-build, not variant*) The build I'd like to try is named e.g. `linux64` in the `stack-setup.yaml`. Seems like it could be useful to have a flag similar to `ghc-variant` that would allow me to get rid of `tinfo6`.

### Stack version

~~~text
stack --version
Version 2.9.1, Git revision 409d56031b4240221d656db09b2ba476fe6bb5b1 x86_64 hpack-0.35.0
~~~

### Method of installation

ysangkok:
Tested by removing `~/.stack` on Debian 11 and doing `stack setup` in a fresh project.

Fixes #5881


ysangkok:
@mpilgrem Had forgotten to push the stack.yaml changes, that's why CI failed. They are now pushed.

src/Stack/Setup.hs
```diff
@@ -1226,7 +1227,23 @@ getGhcBuilds = do
...
+            Nothing -> determineGhcBuild
+            Just OsRelease.OsReleaseResult{OsRelease.osRelease=osRelease} -> do
+              let OsRelease.OsRelease{OsRelease.id=osId, OsRelease.id_like=osIdLike} = osRelease
+              -- The heuristic in determineGhcBuild chooses a Fedora build for
+              -- GHC 9.4 on Debian 11 and Ubuntu 20.04, which doesn't work.
+              -- Make sure we use the standard build on Debian-like distributions.
+              -- Up until now, it has always been a Debian build.
+              -- Note that we don't verify the operating system version, so it
+              -- could still be wrong.
+              if elem (fromMaybe osId osIdLike) ["debian", "ubuntu"] then
```
hasufell, line 1242: You probably want:

I'd include `Name` field and then

1. `toLower` and
2. detect the *word* Debian, such that `Debian GNU/Linux` or `GNU/Linux Debian` matches.

Because neither Name, nor ID, nor ID_LIKE are guaranteed to exist and the spec defines defaults.

See here: https://github.com/haskell/ghcup-hs/blob/a5f2067d76fb529d79ae28cca1bb11ed15baa4d0/lib/GHCup/Platform.hs#L124

hasufell: There's also `lsb-release` and the debian version file. GHCup deals with all of them.

ysangkok: Thanks for the quick feedback.

> There's also lsb-release

On my Debian 11 machine, the executable is called `lsb_release`, with an underscore, not a hyphen. I am a bit puzzled as to why GHCup has `lsb_release_cmd = "lsb-release"`. But I don't think the fallback should be necessary.

> toLower

The spec defines ID and ID_LIKE to be lower case. But I have added a comment to call `toLower` anyway as suggested.

> detect the word Debian

I added a call to `words` for ID_LIKE, which the standard indeed suggests could contain spaces.

> neither Name, nor ID, nor ID_LIKE are guaranteed to exist

The Debian derivatives I have tested, Devuan, Mint and Ubuntu, all have either the ID or the ID_LIKE fields (see the gist linked in the issue). The bug has only been observed on Ubuntu 20.04 and Debian 11, so if the heuristic fails on other versions, we'd just be falling back to existing behavior, which I think is fine, since those derivatives/versions are very rare, and it would still be possible to specify the ghc-build manually as is currently needed on Debian 11/Ubuntu 20.04.

If you insist, I can add the name fallback as suggested. It just seems unnecessarily complex, am I missing something?

> the spec defines defaults.

In the case that `ID=linux` as the spec mentions as the default, we'd use existing logic, which is fine! Because the user is on a different distribution then, or a very old Debian derivative.

src/Stack/Setup.hs
```diff
@@ -1226,7 +1228,29 @@ getGhcBuilds = do
     config <- view configL
     case configGHCBuild config of
         Just ghcBuild -> pure [ghcBuild]
-        Nothing -> determineGhcBuild
+        Nothing -> do
+          mOsReleaseResult <- liftIO OsRelease.parseOsRelease
+          case mOsReleaseResult of
+            Nothing -> determineGhcBuild
+            Just OsRelease.OsReleaseResult{OsRelease.osRelease=osRelease} -> do
+              let OsRelease.OsRelease{OsRelease.id=osId, OsRelease.id_like=osIdLike} = osRelease
```
hasufell, line 1236: Still doesn't utilize `name`.

ysangkok: See my previous comment above, I don't understand why it is necessary to use NAME. Do you know of any Debian derivatives that distributions that don't have ID but do have NAME?

hasufell: No. I'm just saying that none of those keys are mandatory as per the spec, so the logical thing to do is consider all of them.

ysangkok: In which case would we fall back from `ID` though? When it has the value `linux`? It is not a `Maybe` in the os-release library.

mpilgrem:
@ysangkok, what has prompted my exploring #5987 as a possible alternative to this pull request, is that - currently - I think Stack's approach is built around the presence/absence of libraries on which flavours of versions of GHC depend, rather than around the identity of Linux distributions.

ysangkok:
Yes, thank you Mike, I'll close this in favour of #5987

PR #5983, opened 2022-12-22
allow persistent-2.14

juhp:
This should allow building stack in current Stackage Nightly.
I tested building 2.9.3 locally also.

It would be nice to have a Hackage revision also for stack-2.9.3.


juhp:
I see, thanks - hmm I don't have any idea about that

PR #5987, opened 2022-12-23
Fix #5881 Add tag if supports only libc6 < 2.32

### LINKED ISSUE #5881: Stack can't install GHC 9.4.x on Debian 11 (Bullseye)

mpilgrem:
My idea is that the default `setup-info` dictionary will continue to use the `linux64-tinfo6` key for Linux distributions that can support `libc6` 2.32 (eg Fedora >= 33 (2.32), Ubuntu 22.04 (2.35)) and will provide a new `linux64-tinfo6-libc6-pre2.32` for Linux distributions that can only support `libc6` < 2.32 (eg Debian 10 (up to 2.28), Debian 11/Ubuntu 20.04 (up to 2.31), Fedora 27 (up to 2.26)). `linux64-tinfo6` and `linux64-tinfo6-libc6-pre2.32` will be the same pre-GHC 9.4.x. From GHC 9.4, `linux64-tinfo6` will use the 'Fedora 33' version of GHC and `linux64-tinfo6-libc6-pre2.32` will use the 'Debian 10' version of GHC.

Please also shortly describe how you tested your change. Bonus points for added tests! I am unable to test this idea. I rely on Linux users to comment.

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -33,8 +33,10 @@ import           Conduit
 import           Control.Applicative ( empty )
 import           Crypto.Hash ( SHA1 (..), SHA256 (..) )
 import qualified Data.Aeson.KeyMap as KeyMap
+import qualified Data.Attoparsec.Text as P
 import qualified Data.ByteString as S
 import qualified Data.ByteString.Lazy as LBS
+import           Data.Char ( isDigit )
 import qualified Data.Conduit.Binary as CB
 import           Data.Conduit.Lazy ( lazyConsume )
 import qualified Data.Conduit.List as CL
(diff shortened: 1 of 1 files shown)


mpilgrem:
As this should not _break_ anything, I will merge it to allow it to be tested more easily. If it is not viable, it can be reversed.

PR #6006, opened 2022-12-30
Increase scope of, and apply, HLint suggestions

mpilgrem:
The hints once considered not useful do appear to be useful when applied to Stack's current code.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -51,49 +51,11 @@
     - Network.HTTP.StackClient
 
 # Not considered useful hints
-- ignore: {name: "Redundant do"}
-- ignore: {name: "Use section"}
-- ignore: {name: "Use camelCase"}
-- ignore: {name: "Use list comprehension"}
-- ignore: {name: "Redundant if"}
-- ignore: {name: "Avoid lambda"}
-- ignore: {name: "Eta reduce"}
-- ignore: {name: "Parse error"}  # we trust the compiler over HLint
-- ignore: {name: "Use =="}  # Creates infinite loops in `EQ` using expressions
-- ignore: {name: "Evaluate"}
-- ignore: {name: "Use &&&"}
-- ignore: {name: "Redundant compare"}
-- ignore: {name: "Use Just"}
-- ignore: {name: "Avoid lambda using `infix`"}
-
-# Added in hlint-2.0.10, ignoring for now
-- ignore: {name: "Unnecessary hiding"}
-- ignore: {name: "Use lambda-case"}
-
 - ignore:
-    name: "Use fewer imports"
+    name: "Use camelCase"
     within:
-    - System.Process.Read  # Related to 'Hide post-AMP warnings' comment
-    - Stack.Exec  # ifdef for System.Process.Read
-
-# Related to 'explicit pattern matching is clearer' comment
+    - System.Terminal
 - ignore:
-    name: "Use fromMaybe"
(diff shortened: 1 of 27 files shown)


test/integration/tests/git-submodules/Main.hs
```diff
@@ -78,4 +78,4 @@ newline :: Char
 newline = '\n'
 
 stripNewline :: String -> String
-stripNewline str = filter (\x -> x /= newline) str
+stripNewline = filter (/= newline)
```
sol, line 81: This is much better than before!  Much easier on the eye!

sol (commented):
@mpilgrem Hey there :wave:

Just saw this while I was actually looking at the GitHub actions setup that `stack` uses.  I did not do a thorough review.

One thing that I think is commonly misunderstood about HLint is that it is not only/primarily a linter in the classical sense but also an education tool that teaches you how you can express thing differently, oftentimes more concise.  Now "more concise" does not always mean "easier to understand" or "expressing intention".  For that reason I think it's important to be mindful when applying HLint suggestions.

Just because HLint suggests something does not mean that applying that suggestion will give you better code.

I left two comments to make that more clear.

.hlint.yaml
```diff
@@ -51,49 +51,7 @@
     - Network.HTTP.StackClient
 
 # Not considered useful hints
-- ignore: {name: "Redundant do"}
```
sol, line 54: Does HLint now disable "Redundant do" by default?  If not, you probably want to keep this here.  There is zero harm in redundant `do`s and in `*Spec.hs` files you literally always want them.  Consider:

```haskell

spec :: Spec
spec = do
  describe "..." $ do
    it ".." $ do
```
Here it would be foolish to add/remove `do`s when you add/remove additional spec items just for the sake of appeasing HLint.  Not only is it a waste of your time but it also gives you larger diffs and potentially more conflicts when rebasing etc.

mpilgrem: Yes, Hlint's "Redundant do" is now off-by-default, since HLint 3.1.2.

src/System/Process/Pager.hs
```diff
@@ -64,7 +64,7 @@ pageWriter writer = do
  where
   cmdspecFromEnvVar = shell <$> MaybeT (lookupEnv "PAGER")
   cmdspecFromExeName =
-    fmap (\path -> proc path []) . MaybeT . findExecutable
+    fmap (`proc` []) . MaybeT . findExecutable
```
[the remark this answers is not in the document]
mpilgrem: I take your general point, although it might be said the original code of which this particular extract is a part is not easy to understand, full stop. I had to look up `System.Directory.findExecutable`, `Control.Monad.Trans.Maybe.MaybeT`, and the `Functor m => Functor (Maybe T m)` instance for `MaybeT` to understand it.

My general approach was (a) to remove each of the existing limitations in `.hlint.yaml` in turn, (b) to see what had been ignored in the existing Stack code base, and (c) if there were only a handful of HLint suggestions and all of them seemed to me to be improvements or, at least, not make things worse, to remove the existing limitation. At this site, I did not think it made things _worse_.

I think this is an application of HLint's "Avoid lambda using `infix`" rule. I'll reintroduce something to ignore that rule.

test/integration/tests/git-submodules/Main.hs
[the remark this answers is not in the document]
mpilgrem: Agreed.

PR #6031, opened 2023-01-18
Bump the cabal project to match lts-20.0 in stack.yaml.

philderbeast:
A small maintenance bump in the cabal project to match the stack project. Added a note about using `with-compiler`.

### DIFF
--- a/cabal.config
+++ b/cabal.config
@@ -1,11 +1,10 @@
 constraints:
-  , Cabal ==3.6.3.0
-  , Cabal-syntax ==3.6.0.0
+  , Cabal ==3.8.1.0
+  , Cabal-syntax ==3.8.1.0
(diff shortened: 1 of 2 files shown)


mpilgrem:
@philderbeast, many thanks - but you have lost Stack's dependency on `Win32`. That has to be taken care of manually, if you update on a non-Windows system. There is a comment in `cabal.project` that warns about that. (A corresponding problem arises if you update automatically on a Windows system, with `unix`.)

mpilgrem:
I am going to merge this before the missing `Win32` is fixed, because I have bumped to `lts-20.8` and I am about to do the same thing for `lts-20.8` in any event.

philderbeast:
Thanks for picking this up and merging it @mpilgrem.

Issue #5603, opened 2021-08-20
Can't add repo to extra-deps if repo contains symlink to git submodule


brandon-leapyear:
### General summary/comments (optional)

Relevant links:
* #4580 
* #5536 
* #5236 
* https://github.com/commercialhaskell/stack/issues/4579
* https://github.com/commercialhaskell/pantry/blob/c7923fc963e36efd179ba301ac426139a9dc182d/src/Pantry/Archive.hs#L406

### Steps to reproduce

```yaml
extra-deps:
- github: google/proto-lens
  commit: 5ee7ac14686cf8d0668b8b3053b055303fc3ce86
  subdirs:
  - proto-lens-protoc
```
Then run `stack build`

### Expected

It should checkout the repo, `git submodule update`, then use it.

### Actual

It downloads the tar archive from GitHub and tries untarring it, with git submodules not included + symlinks not resolving.

```
Unsupported tarball from ...: Symbolic link dest not found from ... to ...
This may indicate that the source is a git archive which uses git-annex
```

### Stack version

2.7.3

qrilka:
Isn't it a duplicate of #5536?

brandon-leapyear:
I'm not sure. This issue is because theres a symlink pointing to a
directory containing the submodule. I'm not sure the other issue has the
symlink aspect to it.

brandon-leapyear:
To be clear, this isn't a duplicate of #5536 (or rather, this _could_ be solved without solving #5536) because the `proto-lens-protoc` package I'm trying to build doesn't use the symlink at all; the symlink just **exists** in the repo and Stack throws an error.

i.e. if [this line](https://github.com/commercialhaskell/pantry/blob/ac0a31956f2c56cecb88e8c197986b6172e1162e/src/Pantry/Archive.hs#L405-L407) were simply changed to
```hs
[] -> Right []
```
My build would work. So the `git-annex` failure is a false positive here

jorpic:
I had a similar issue with submodules and spent some time trying to understand what is going on.

The problem is that with
```yaml
extra-deps:
- github: google/proto-lens
```

`stack` does not clone the repo and can't update submodules, it just downloads an archive from `https://github.com/google/proto-lens/archive/<hash>.tar.gz`. 

On the other hand, this version executes `git clone` and then `git submodule update --init --recursive`:

```yaml
extra-deps:
- git: https://github/google/proto-lens
```

[...]

PR #6037, opened 2023-01-21
Document a difference between `git` and `github` in `extra-deps`

jorpic:
I spent some time trying to understand why `github: user/repo` in `extra-deps` does not work if repo contains git submodules.
Probably it worth to explicitly mention the difference between `git` and `github` in documentation.

ref #5603

### DIFF
--- a/doc/pantry.md
+++ b/doc/pantry.md
@@ -206,6 +206,21 @@ extra-deps:
   commit: a5f4f30f01366738f913968163d856366d7e0342
 ~~~
 
+!!! note
+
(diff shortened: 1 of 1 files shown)


mpilgrem:
@jorpic, I changed the base branch, because pull requests for improvements to documentation that are applicable to the current released version of Stack should target this repository's `stable` branch.

mpilgrem:
@jorpic, post-merging, I made some further changes to this part of the documentation. The help on GitHub archive files was in the wrong location, adding to the confusion.

PR #6081, opened 2023-03-22
stack script: recognize more extra-dep kinds

chrismwendt:
🚧 This is a WIP (it doesn't quite work yet, not sure why) 🚧

When you're working on a Stack script, sometimes you encounter a bug in a dependency and want to use a local patched version. Prior to this PR you couldn't because it only recognized package names. After this PR it will recognize the same kinds of [`extra-deps`](https://docs.haskellstack.org/en/stable/yaml_configuration/#extra-deps) as a proper Stack project, including local paths.

For convenience and backcompat, the argument is first parsed as a literal string, and if that fails then it is parsed as JSON.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,11 @@
 
 Release notes:
 
+Behavior changes:
+
+* `stack script --extra-dep=<...>` now recognizes all the kinds of extra-deps
+  that Stack can use (local packages, git repos, etc.), not just package names.
+
 **Changes since v2.9.3:**
 
 Major changes:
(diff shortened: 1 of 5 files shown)


chrismwendt:
Actually getting this to work is proving to be way harder than I thought it would be 😓

I'm guessing the extra deps need to get translated to a set of import dirs and passed to GHC via `-i`, but it's difficult to figure out how.

Perhaps a better (but even more involved) approach would be to unify the `stack.yaml` config and the Stack script comment with a layer that parses one or the other into a common data type that then gets passed down to the build system so that they have the same capabilities and only differ in representation as YAML vs CLI args.

PR #6123, opened 2023-05-17
don't repeatedly resolve common usage paths

dten:
For some large packages in our project `getPackageFiles` has always quite a long time to complete, but it seems much worse in 9.4.x. (~7.5 seconds on my machine with stack 2.11 and ghc 9.4.5). Profiling showed it spent most of its time checking the the files from `usages` in `.hi` files existed. 

When I checked what paths it was checking and how often I found that every `.hi` file mentioned many `.so` files (~150) and each of these files was being checked it existed for each `.hi` file.

This change attempts to collect up known resolution mappings whilst walking the `.hi` files so the resolution effort doesn't get repeated.

This takes our slowest `getPackageFiles` call down to ~3.6 seconds with the profile now showing the cost is mostly parsing the `.hi` files.

I've included before and after profile results where this was basically `stack test --file-watch --watch-all` on an already build project to get to the point where it's ready to build new changes.

before: note the wide sections calling forgivingResolveFile
stack_before

after: those wide sections are now gone
stack_after

the profiles also show stack does all the `getPackageFiles` work twice which isn't ideal but i haven't looked into yet 👀

### DIFF
--- a/src/Stack/ComponentFile.hs
+++ b/src/Stack/ComponentFile.hs
@@ -20,6 +20,7 @@ module Stack.ComponentFile
 
 import           Control.Exception ( throw )
 import           Data.List ( find, isPrefixOf )
+import           Data.Foldable (foldrM)
 import qualified Data.Map.Strict as M
 import qualified Data.Set as S
 import qualified Data.Text as T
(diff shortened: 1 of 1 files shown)


dten:
sure thing 😃 by the way it looks likely the many `.so` file usings appear when using template haskell

mpilgrem:
Looks good to me.

mbj:
This patch moves my project from 1.5s on `getPackageFiles` to  0.75s. Thanks.

Issue #1345, opened 2015-11-15
Two stack.yamls (for GHC and GHCJS) get their build status mixed up


mtolly:
**Steps to reproduce:**

1. Clone https://github.com/mtolly/stacktest2 -- it is a repo with a single package in the root, and 2 YAML files: `stack.yaml` has `resolver: lts-3.13`, and `stack-ghcjs.yaml` has `resolver: ghcjs-0.2.0_ghc-7.10.2`
2. Build the package with each YAML file
3. Make a change to the code
4. Build again with each YAML file

**Expected:**

Both before and after the change, the module gets compiled anew by both compilers.

**Actual:**

After editing the package, only the first of the two build commands recompiles; the second one assumes everything is already built.

```
╭─mtolly@metal  ~/g/stacktest2  <master>
╰─$ stack build
pkg-a-0.1: configure
Configuring pkg-a-0.1...
pkg-a-0.1: build
Preprocessing library pkg-a-0.1...
[1 of 1] Compiling A                ( A.hs, .stack-work/dist/x86_64-osx/Cabal-1.22.4.0/build/A.o )
In-place registering pkg-a-0.1...
pkg-a-0.1: copy/register
Installing library in
/Users/mtolly/git/stacktest2/.stack-work/install/x86_64-osx/lts-3.13/7.10.2/lib/x86_64-osx-ghc-7.10.2/pkg-a-0.1-IgKVN5mB3raJQ6Q3QCTOYe
Registering pkg-a-0.1...
╭─mtolly@metal  ~/g/stacktest2  <master>
╰─$ stack build --stack-yaml=stack-ghcjs.yaml
pkg-a-0.1: configure
Configuring pkg-a-0.1...
pkg-a-0.1: build
Preprocessing library pkg-a-0.1...
[1 of 1] Compiling A                ( A.hs, .stack-work/dist/x86_64-osx/Cabal-1.22.4.0/build/A.js_o )
In-place registering pkg-a-0.1...
pkg-a-0.1: copy/register
Installing library in
/Users/mtolly/git/stacktest2/.stack-work/install/x86_64-osx/ghcjs-0.2.0_ghc-7.10.2/ghcjs-0.2.0_ghc-7.10.2/lib/x86_64-osx-ghcjs-0.2.0-ghc7_10_2/pkg-a-0.1-1pLnKjAnJO45wYWJywqanC
Registering pkg-a-0.1...

(Make a change to A.hs)

[...]

PR #6162, opened 2023-06-18
distRelativeDir: use ghc version instead of Cabal version 

juhp:
This makes the dist build paths more consistent: clearly differentiated by ghc minor versions.

- [x] Any changes that could be relevant to users have been recorded in ChangeLog.md.

I tested by building stack with a stack including this change.

There is a bit more rationale in the commit message.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,10 @@ Major changes:
 
 Behavior changes:
 
+* The path dist directory path '.stack-work/dist/<platform>/Cabal-*/'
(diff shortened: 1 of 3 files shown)


mpilgrem:
Some notes to self:
* https://github.com/commercialhaskell/stack/issues/1345

mpilgrem (approved):
Looks good to me. I understand that the motivation for this is that it paves the way for a `stack clean` command that can distinguish subdirectories of `dist` by their GHC version.

PR #6166, opened 2023-06-23
Stack.Setup: bump version warnings to ghc-9.8 and Cabal-3.12

juhp:
Current stack seems to work fine with ghc-9.6 and Cabal-3.10 afaict.

TBH I don't find these warnings very helpful and they always lag into the stablized ghc minor versions.
ie we are already at ghc-9.6.2 but stack is still saying it is untested:

```
Warning: Stack has not been tested with GHC versions above 9.6, and using 9.6.2,
         this may fail.

Warning: Stack has not been tested with Cabal versions above 3.10, but version
         3.10.1.0 was found, this may fail.
```

I feel like it would be better even to either bump the warning by another major version
(9.8 is just around the corner) or just drop them entirely?  I don't feel they are
providing that much value: if stack really breaks it should be fairly obvious anyway.

### DIFF
--- a/src/Stack/Setup.hs
+++ b/src/Stack/Setup.hs
@@ -923,7 +923,7 @@ warnUnsupportedCompiler ghcVersion =
         pure True
     | ghcVersion >= mkVersion [9, 7] -> do
         prettyWarnL
-          [ flow "Stack has not been tested with GHC versions above 9.8, and \
+          [ flow "Stack has not been tested with GHC versions 9.8 and above, and \
                  \using"
           , fromString (versionString ghcVersion) <> ","
           , flow "this may fail."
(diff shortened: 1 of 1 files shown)


mpilgrem:
To be fair, there has only been a Stackage snapshot for GHC 9.6.x since 22 June 2023. I think we only accumulate good empirical evidence that Stack 'works' with GHC/Cabal versions once a Stackage snapshot is available. I am not expecting Stack not to work with GHC 9.6.x or Cabal 3.10.x, but I would give it a couple of weeks at least.

mpilgrem:
Regarding the CI failing on Windows/integration - I thought it was a [GHC 9.2.7 -> 9.4.5 effect](https://discourse.haskell.org/t/ghci-9-2-7-to-9-4-5-on-windows-was-there-a-change/6561), but it looks like it must be due to something else. Stack has not changed, so it must be something in the GitHub hosted runner environment.

juhp:
Thanks

Just to add some more context, for example in Fedora we are still shipping 2.9.1 (hopefully 2.9.3.1 in the next version), so users will keep seeing those warnings "forever", even though stack is working fine up to 9.6 currently say. So I may just patch Fedora stack not to print them.

Perhaps the warnings should only be enabled for official builds if you want them - that seems like the best of both worlds to me at least.

juhp:
I applied https://src.fedoraproject.org/rpms/haskell-platform/blob/rawhide/f/stack-disable-ghc-Cabal-version-warnings.patch for Fedora 39.

PR #6177, opened 2023-06-28
Ellipsize build progress output line

develop7:
This PR introduces truncating long build progress lines which are spamming narrow terminal window. Like this:

image
Before this PR _(repro: `stack --stack-root=/home/adziahel/.cache/stacktmp/root --terminal-width=20 build --fast`)_

The issue here is multifold and unclear — it's either the terminal processes `^A^K` before each `logSticky` poorly (because of incorrect terminfo), or it's bug in GNOME VTE, or is it `logSticky` ignores actual terminal width — could be all of these, if we're unlucky enough.

This PR replaces the tail of progress line with ellipsis (see screenshot below), preventing spamming terminal with repeating progress lines.

image
After this PR _(repro: `stack exec stack -- --stack-root=/home/adziahel/.cache/stacktmp/root --work-dir=.stackworktmp build --fast`)_

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,9 @@ Release notes:
 
 * Further to the release notes for Stack 2.3.1, the `-static` suffix has been
   removed from the statically linked Linux/x86_64 binaries.
+* The build progress line gets ellipsized in order to fit one terminal line
+  matching cargo behavior and preventing logSticky spam on narrow 
+  terminals and lots of dependencies building simultaneously
 
 **Changes since v2.11.1:**
 
--- a/src/Stack/Build/Execute.hs
+++ b/src/Stack/Build/Execute.hs
@@ -785,6 +785,7 @@ executePlan' installedMap0 targets plan ee@ExecuteEnv {..} = do
   let keepGoing =
         fromMaybe (not (Map.null (planFinals plan))) (boptsKeepGoing eeBuildOpts)
   terminal <- view terminalL
+  terminalWidth <- view termWidthL
   errs <- liftIO $ runActions threads keepGoing actions $
     \doneVar actionsVar -> do
       let total = length actions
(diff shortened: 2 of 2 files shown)


mpilgrem:
Looks like a good idea, and I would make it the default, but I would also make it configurable on the command line and in YAML configuration files. Something like:
~~~text
stack build --progress-bar=none/capped/full
~~~

where `capped` means capped to the terminal width. That is because I do find the current full information useful, even it is not 'sticky'.

Not 'sticky' for long output occurs because `RIO.Prelude.Logger.stickyImpl` (not exported) uses 'backspace' and, on most terminals (although not all - it is system dependent), that does nothing when the cursor is at the first column.

develop7:
Thanks for the heads up!

> I would also make it configurable on the command line and in YAML configuration files. Something like:
> 
> ```
> stack build --progress-bar=none/capped/full
> ```
> 
> where `capped` means capped to the terminal width. That is because I do find the current full information useful, even it is not 'sticky'.

While I do like your proposal, it looks more of a feature to me and a bit beyond the initial PR; hence it warrants for a separate PR and maybe an issue for the sake of visibility.

> Not 'sticky' for long output occurs because RIO.Prelude.Logger.stickyImpl (not exported) uses 'backspace' and, on most terminals (although not all - it is system dependent), that does nothing when the cursor is at the first column.

Not just that, per the https://gitlab.gnome.org/GNOME/vte/-/issues/62 (via https://superuser.com/a/1474412) it's a known legacy issue (multiple issues) which seem unclear how to approach even, let alone how to fix.

mpilgrem:
@develop7, unless you would like to have a go yourself, I'll accept this PR and raise another for the configuration.

A little off-topic, but I read [ECMA-48](https://www.ecma-international.org/wp-content/uploads/ECMA-48_5th_edition_june_1991.pdf) 8.3.5 to say that 'backspace' (BS) should cause the 'active data position' (i.e. the cursor) to be moved one character position in 'the data component' in 'the direction opposite to that of the implicit movement' (i.e. backwards) - nothing about it stopping at the first column of the current line. I suspect the actual observed BS behavior in software terminals has its origins in the behaviour of physical teleprinters.

develop7:
@mpilgrem I'd raise a new PR anyway, so merge it away

PR #6181, opened 2023-07-02
Fix #6180 Take a different approach to building GHC from source

### LINKED ISSUE #6180: Building GHC 9.4.5 from source fails because of deprecated Hadrian flag

Stack calls Hadrian with `-c` [here](https://github.com/commercialhaskell/stack/blob/1931538a4a3e5058d9432bff68b397c6804a96f2/src/Stack/Setup.hs#L1433), but that option has been deprecated (which in this case means not much less than "removed") for a while: https://gitlab.haskell.org/ghc/ghc/-/issues/20167

Building GHC currently requires running `./boot && ./configure` before calling Hadrian.

### Steps to reproduce

1. Create a Stack project on LTS 21.0 with `stack new`.
2. `stack build --compiler=ghc-git-a213d3676550a0e4d542172de539c0cfa2662431-quickest` (that commit is the one tagged as `ghc-9.4.5-release` in the GHC repo).

### Expected

GHC built from source successfully.

### Actual

Stack builds Hadrian successfully, but then:

~~~text
shake: hadrian --configure has been deprecated (see #20167). Please run ./boot; ./configure manually

Usage: hadrian [options] [target] ...

Standard options:
# ... snip ...
Extra options:
  -c, --configure             Deprecated: Run the boot and configure scripts.
# ... snip ...

Received ExitFailure 1 when running
Raw command: /tmp/with-repo1432778/cloned/hadrian/build-stack -c -j8 --flavour=quickest binary-dist
Run from: /tmp/with-repo1432778/cloned
~~~

### Stack version

In my project's nix-shell:
~~~text
stack --version
2.9.1 x86_64 hpack-0.35.2
~~~

But I've also tried with Stack installed from GHCup:

~~~text
stack --version
Version 2.11.1, Git revision c1167a6abc3f4978ccded5ba0246a57387da0e2f x86_64 hpack-0.35.2
~~~

### Method of installation

* Via GHCup
* Other: Via Nix

### Platform

Linux x86_64, Ubuntu.

mpilgrem:
`Stack.Setup.buildGhcFromSource` no longer uses the `-c` flag of `hadrian` but seeks to run the GHC repository's `boot` python script and `configure` sh script.

`Stack.Setup.buildGhcFromSource` requires MSYS2 tools, so - borrowing the pattern of `WithGHC env` - defined `newtype WithMSYS env` and `runWithMSYS`.

The GHC repository does not provide its `configure` file, so it has to be created by applying `autoreconf`. It also requires `aclocal`. Moved `ensureConfigureScript` out of `Stack.Build.Execute.ensureConfig` to new module `Stack.Config.ConfigureScript` and extended the help on Windows to refer to `aclocal` and the MSYS2 package `autotools`.

GHC's configure script requires ghc, happy and alex on the PATH. So, (1) it is run in Stack's environment, making use of the `stack.yaml` at `hadrian`; and (2) happy and alex are installed using Stack, if not on the PATH.

It appears that GHC's configure script requires a more recent version of GHC than is specified in the `stack.yaml`. So, the `stack.yaml` is overwritten with contents that specify a more recent GHC (LTS Haskell 20.26 (GHC 9.2.8)).

It also appears that `hadrian` has some other pre-requisites that are not included in the Stack-supplied MSYS2 out of the box and have to be installed if absent. They include `python` and the python package `sphinx`. They are not dealt with here, but the online documentation is extended.

Please also shortly describe how you tested your change. Bonus points for added tests! Tested locally on Windows 11.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -19,6 +19,10 @@ Behavior changes:
   be distinguished by GHC version.
 * By default, the `stack build` progress bar is capped to a length equal to the
   terminal width.
+* When building GHC from source, Stack no longer uses Hadrian's deprecated
+  `--configure`\\`-c` flag and, instead, seeks to run GHC's Python `boot` and
+  sh `configure` scripts, and ensure that the `happy` and `alex` executables are
+  on the PATH.
 
 Other enhancements:
 
--- a/doc/maintainers/stack_errors.md
+++ b/doc/maintainers/stack_errors.md
@@ -286,6 +286,7 @@ to take stock of the errors that Stack itself can raise, by reference to the
         [S-5219] | GHCInfoMissingTargetPlatform
         [S-8299] | GHCInfoTargetPlatformInvalid String
         [S-2574] | CabalNotFound (Path Abs File)
+        [S-8488] | GhcBootScriptNotFound
         [S-1128] | HadrianScriptNotFound
         [S-1906] | URLInvalid String
         [S-1648] | UnknownArchiveExtension String
--- a/doc/yaml_configuration.md
+++ b/doc/yaml_configuration.md
@@ -543,12 +543,13 @@ compiler-check: match-exact
 
 [:octicons-tag-24: 2.1.1](https://github.com/commercialhaskell/stack/releases/tag/v2.1.1)
 
-Stack supports building the GHC compiler from source. The version to build and
-to use is defined by a a Git commit ID and an Hadrian "flavour" (Hadrian is the
-build system of GHC) with the following syntax:
+Stack supports building the GHC compiler from source, using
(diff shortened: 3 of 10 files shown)


mpilgrem:
I've subsequently taken a different approach to problematic `Hadrian/stack.yaml` resolvers, as the immediate problem fixed in https://gitlab.haskell.org/ghc/ghc/-/commit/6df15e93171cf6ee9aeee642ede2d816a084efda. Now tested locally on Windows 11 and macOS 13.4.1/Aarch64, with environment variable `STACK=stack --resolver lts-20.26` and:
~~~text
stack --resolver lts-20.26 build --compiler=ghc-git-cec903899234bf9e25ea404477ba846ac1e963bb-quickest
~~~

PR #6184, opened 2023-07-09
Fix digest computation duplication

mbj:
On a project with many local packages I had "suprising" pauses in logged activity when running `stack build --verbose` and these always "nagged" me. Tonight I went in and found thta these pauses contribute to 2-3s of "dead time" before a `--file-watch` triggererd build would start working on my `.hs` file changes.

So I started to add strategic `bracketDebug` till I localized the dead "time" in `checkBuildCache`. And found the work there to be duplicate between the various packages. Lots of `.so` files my packages share (mostly boot libs, that can be quite big) would get hashed many times over.

So I implemented the simplest possible cache that has the lifetime of a build triggered by `--file-watch` to deduplicate these hash attempts. It shaved 2-3 seconds from my build and this is a lot as these 2-3 seconds would get "payed" every time before my code change would even start building. Now they are entirely gone.

On the stack code base as it only has one local source module this change does not make a difference, but on projects with many local source modules if they depend on shared "GHC stuff" its quite noticable.

When testing: Watch out for the `checkCacheResults` debug I added, and try without the last commit in this series to verify.

This is still an RFC, so not fully ready for merge.

TODO:

* [x] Verify with maintainers that this approach is actually correct, and not breaking invariants I did not anticipate.
* [ ] Add changelog entry.
* [x] Remove extra debug bracket added to show the speed difference.

Note the hashing is still not optimal with this change. Other areas of improvement I may do in followups.

* We can likely use pooled concurrency to spread out the work cross multiple cores.
* We can change to not read the entire file into memory to hash it than, but chunk this. This should reduce allocations.

### DIFF
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -24,6 +24,7 @@ import qualified Distribution.PackageDescription as C
 import qualified Pantry.SHA256 as SHA256
 import           Stack.Build.Cache ( tryGetBuildCache )
 import           Stack.Build.Haddock ( shouldHaddockDeps )
+import           Stack.FileDigestCache ( readFileDigest )
 import           Stack.Package ( resolvePackage )
 import           Stack.Prelude
 import           Stack.SourceMap
@@ -435,13 +436,13 @@ loadLocalPackage pp = do
 
 -- | Compare the current filesystem state to the cached information, and
 -- determine (1) if the files are dirty, and (2) the new cache values.
-checkBuildCache :: forall m. (MonadIO m)
+checkBuildCache :: HasEnvConfig env
                 => Map FilePath FileCacheInfo -- ^ old cache
                 -> [Path Abs File] -- ^ files in package
-                -> m (Set FilePath, Map FilePath FileCacheInfo)
+                -> RIO env (Set FilePath, Map FilePath FileCacheInfo)
 checkBuildCache oldCache files = do
   fileTimes <- fmap Map.fromList $ forM files $ \fp -> do
-    mdigest <- liftIO (getFileDigestMaybe (toFilePath fp))
+    mdigest <- getFileDigestMaybe (toFilePath fp)
     pure (toFilePath fp, mdigest)
   fmap (mconcat . Map.elems) $ sequence $
     Map.merge
@@ -454,7 +455,7 @@ checkBuildCache oldCache files = do
   go :: FilePath
      -> Maybe SHA256
      -> Maybe FileCacheInfo
-     -> m (Set FilePath, Map FilePath FileCacheInfo)
+     -> RIO env (Set FilePath, Map FilePath FileCacheInfo)
   -- Filter out the cabal_macros file to avoid spurious recompilations
   go fp _ _ | takeFileName fp == "cabal_macros.h" = pure (Set.empty, Map.empty)
   -- Common case where it's in the cache and on the filesystem.
@@ -518,17 +519,15 @@ getPackageFilesForTargets pkg cabalFP nonLibComponents = do
   pure (componentsFiles, warnings)
 
 -- | Get file digest, if it exists
-getFileDigestMaybe :: MonadIO m => FilePath -> m (Maybe SHA256)
-getFileDigestMaybe fp =
-  liftIO $
-    catch
-      (fmap Just getDigest)
-      (\e ->
-            if isDoesNotExistError e
-                then pure Nothing
-                else throwM e)
- where
-  getDigest = SHA256.hashFile fp
+getFileDigestMaybe :: HasEnvConfig env => FilePath -> RIO env (Maybe SHA256)
+getFileDigestMaybe fp = do
+  cache <- view $ envConfigL.to envConfigFileDigestCache
+  catch
+    (Just <$> readFileDigest cache fp)
+    (\e ->
+          if isDoesNotExistError e
+              then pure Nothing
+              else throwM e)
 
 -- | Get 'PackageConfig' for package given its name.
 getPackageConfig ::
--- a/src/Stack/Build/Source.hs
+++ b/src/Stack/Build/Source.hs
@@ -14,7 +14,6 @@ module Stack.Build.Source
   , hashSourceMapData
   ) where
 
-import           Conduit ( ZipSink (..), withSourceFile )
 import           Data.ByteString.Builder ( toLazyByteString )
 import qualified Data.List as L
 import qualified Data.Map as Map
@@ -523,13 +522,13 @@ getFileDigestMaybe :: MonadIO m => FilePath -> m (Maybe SHA256)
 getFileDigestMaybe fp =
   liftIO $
     catch
-      (fmap Just . withSourceFile fp $ getDigest)
+      (fmap Just getDigest)
       (\e ->
             if isDoesNotExistError e
                 then pure Nothing
                 else throwM e)
  where
-  getDigest src = runConduit $ src .| getZipSink (ZipSink SHA256.sinkHash)
+  getDigest = SHA256.hashFile fp
 
 -- | Get 'PackageConfig' for package given its name.
 getPackageConfig ::
--- a/src/Stack/FileDigestCache.hs
+++ b/src/Stack/FileDigestCache.hs
@@ -1,0 +1,24 @@
+module Stack.FileDigestCache
+  ( FileDigestCache
+  , newFileDigestCache
+  , readFileDigest
+  ) where
+
+import           Stack.Prelude
+import qualified Data.Map.Strict as Map
+import qualified Pantry.SHA256 as SHA256
+
+type FileDigestCache = IORef (Map FilePath SHA256)
+
+newFileDigestCache :: MonadIO m => m FileDigestCache
+newFileDigestCache = newIORef Map.empty
+
+readFileDigest :: MonadIO m => FileDigestCache -> FilePath -> m SHA256
+readFileDigest cache filePath = do
+  digests <- readIORef cache
+  case Map.lookup filePath digests of
+    Just digest -> pure digest
+    Nothing -> do
+      sha256 <- SHA256.hashFile filePath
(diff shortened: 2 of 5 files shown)


src/Stack/Build/Source.hs
```diff
@@ -523,13 +522,13 @@ getFileDigestMaybe :: MonadIO m => FilePath -> m (Maybe SHA256)
...
     catch
-      (fmap Just . withSourceFile fp $ getDigest)
+      (fmap Just getDigest)
       (\e ->
             if isDoesNotExistError e
                 then pure Nothing
                 else throwM e)
  where
-  getDigest src = runConduit $ src .| getZipSink (ZipSink SHA256.sinkHash)
+  getDigest = SHA256.hashFile fp
```
mbj: Pantry does basically the same in `hashFile` so no need for us to redo it here.

src/Stack/Build/Source.hs
```diff
@@ -389,7 +389,7 @@ loadLocalPackage pp = do
   componentFiles <- memoizeRefWith $
     fst <$> getPackageFilesForTargets pkg (ppCabalFP pp) nonLibComponents
 
-  checkCacheResults <- memoizeRefWith $ do
+  checkCacheResults <- memoizeRefWith . debugBracket "checkCacheResults" $ do
```
mbj, line 392: I intend to drop this pre-merge.

stack-ghc-9.4.5.yaml
```diff
@@ -1,7 +1,7 @@
 # This is an experimental project-level configuration, to see if Stack can be
 # built with GHC 9.4.5.
 
-resolver: nightly-2023-06-18 # GHC 9.4.5
+resolver: lts-21.1
```
mbj, line 4: I sneaked this one into the series as I use GHC 9.4 at work and wanted to avoid the download for 9.2 as  I was on a low througput internet connection.

mbj:
I've fixed the linting error. Pls re-run workflow.

mbj:
Nice thanks for the workflow re-run. I've tested this with more repos meanwhile and see improvements across the board.

Now: Is this architecturally the right direction? I do feel like we are "duplicating" and special casing lots of stuff a generic build system like shake would nicer. But I do not see a re-write to a more generic "work deduplicating" backend on the horizon so this approach of deduplicating file hash attempts may be a local optimum.

Curious what others think. Also curious what kind of proof we need to consider this PR as a performance win across the board. All my projects are either: Not affected (one source module so no shared hashes to deduplicate), or significantly affected: Lots of shared hash attempts now deduplicated.

mbj:
@mpilgrem 

>  but please bear with me. 

no worries. Happy you took a look. And yes I'm looking mostly for architectural pointers.

Also a bit more structured:

* During a `stack build` if we have local source modules that depend on the same files. These files would get hashed multiple times. Especially GHC components (which can be quite big).
* This PR intends to share this hash results between these attempts, avoiding duplicate hash attempts.
* Also this PR tries to only ever cache this during a single `stack build` run, so no leaking cross various `stack builds` that may be triggered via `--file-watch`.
* The speed benefit is only observable if you have many local source modules. basically `extra-deps:` pointing to the local FS.
* Without multiple socal source module that depend on these GHC files there is no speed gain/loss.

Also ultimately I think a real rules engine like shake may be a better way to avoid duplicate work. But I think this would be a too big retrofit, so for the moment I'm going to just advance the local optimum.

mpilgrem:
Notes to self:
* I wondered about `IORef`, but then came across the discussion under this [Reddit comment](https://www.reddit.com/r/haskell/comments/7r4wie/comment/dsv7wrd/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) and linking to this [video](https://www.youtube.com/watch?v=KZIN9f9rI34).

mbj:
> I wondered about IORef

Hah, I came to the same conclusion but independently.

chreekat (commented):
Since this is a performance improvement, it would be great if there was some kind of benchmark for it in the test suite. But unless stack already has just such a benchmark test suite and adding another test would be easy, I don't think it's a blocker. Seems cool!

mpilgrem:
Re performance benchmarks, Stack does not have those, but I have been thinking about performance locally. My basic benchmark is 'How long does it take Stack to build Stack?'. That is, this change aims to speed up builds where local/mutable packages in a project happen to have many or large files in common (which I assume is relatively rare). I want to understand what effect it has on 'normal' projects, where there are no, few, or small files in common between local packages.

mbj:
> I want to understand what effect it has on 'normal' projects, where there are no, few, or small files in common between local packages

I saw neither a positive nor a negative change in more normal projects. If you wanted I could make a synthetic project showing the difference.

src/Stack/FileDigestCache.hs
```diff
@@ -0,0 +1,24 @@
...
+newFileDigestCache = newIORef Map.empty
+
+readFileDigest :: MonadIO m => FileDigestCache -> FilePath -> m SHA256
+readFileDigest cache filePath = do
+  digests <- readIORef cache
+  case Map.lookup filePath digests of
+    (Just digest) -> pure digest
+    Nothing -> do
+      sha256 <- SHA256.hashFile filePath
+      writeIORef cache $ Map.insert filePath sha256 digests
```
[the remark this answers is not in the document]
mbj: There is no concurrency, more gains can be made adding concurrency as hashing N files as we do per local source package is a good candidate.

I've made concurrent hash maps before and would happily make one here again, but for the moment I intentionally went with the simplest thing that has a positive benefit to limit the amount of change.

Once we look at concurrency even non "pathologic" more normal projects will benefit. I'll add a comment in the next pass over this code.

mbj:
> Since this is a performance improvement, it would be great if there was some kind of benchmark for it in the test suite. But unless stack already has just such a benchmark test suite and adding another test would be easy, I don't think it's a blocker. Seems cool!

I do not think the performance impact is demonstrable until you hit projects that depend on GHC itself, as than we have lots of these big `.so` files to hash (repeatedly worst case).

I come to the conclusion I'll make a public repo showing the speed gains. We can than think on how to eventually make a persisted test within this repo - if we even consider we want to speed these up. I'm fine to run this patch locally formyself.

mpilgrem:
I am now reasonably confident that this does not result in a measurable performance reduction for building Stack with Stack.

src/Stack/Build/Source.hs
mpilgrem: OK. I think we are close to merging - can you drop this?

src/Stack/FileDigestCache.hs
```diff
@@ -0,0 +1,24 @@
...
+type FileDigestCache = IORef (Map FilePath SHA256)
+
+newFileDigestCache :: MonadIO m => m FileDigestCache
+newFileDigestCache = newIORef Map.empty
+
+readFileDigest :: MonadIO m => FileDigestCache -> FilePath -> m SHA256
+readFileDigest cache filePath = do
+  digests <- readIORef cache
+  case Map.lookup filePath digests of
+    (Just digest) -> pure digest
```
mpilgrem, line 20: HLint does not complain, but I wondered if the brackets around `Just digest` were required?

src/Stack/FileDigestCache.hs
```diff
@@ -0,0 +1,24 @@
+module Stack.FileDigestCache
+  ( FileDigestCache
+  , newFileDigestCache
+  , readFileDigest
+  ) where
+
+import           Stack.Prelude
+import qualified Data.Map.Strict as Map
+import qualified Pantry.SHA256 as SHA256
```
mpilgrem, line 9: Not a substantive point, but other Stack modules list imports alphabetically.

src/Stack/Build/Source.hs
```diff
@@ -436,13 +436,13 @@ loadLocalPackage pp = do
...
 -- | Compare the current filesystem state to the cached information, and
 -- determine (1) if the files are dirty, and (2) the new cache values.
-checkBuildCache :: forall m. (MonadIO m)
+checkBuildCache :: HasEnvConfig env
                 => Map FilePath FileCacheInfo -- ^ old cache
                 -> [Path Abs File] -- ^ files in package
-                -> m (Set FilePath, Map FilePath FileCacheInfo)
+                -> RIO env (Set FilePath, Map FilePath FileCacheInfo)
 checkBuildCache oldCache files = do
   fileTimes <- fmap Map.fromList $ forM files $ \fp -> do
```
mpilgrem, line 444: Not a necessary part of your change, but I assume the name of `fileTimes` reflects some ancient history of this part of the code. Perhaps it should be refactored to be named `fileDigests` and this is an opportunity to do that?

mbj: I intentionalyl wanted to keep the first draft very limited, I'm happy to clean this all up.

mbj:
I should be able to do a pass over this tonight adapting to the discussion results we had meanwhile.

mpilgrem:
@mbj, I am planning to release a new version of Stack soon. Letting you know in case you would like it to reflect this pull request.

mbj:
@mpilgrem I'm on vacation and my time on development is very limited and this PR does not make the cut ATM. Once I'm out of vacation this will change.

src/Stack/FileDigestCache.hs
mbj: adressed.

src/Stack/Build/Source.hs
mbj: done.

src/Stack/Build/Source.hs
mbj: I'll not get to a cleanup in this PR. Sorry.

mbj:
> @mbj, I am planning to release a new version of Stack soon. Letting you know in case you would like it to reflect this pull request.

I've addressed the open feedback, and re-pushed after rebase.

mpilgrem:
Thanks!

mbj:
Ty for merge.

PR #6196, opened 2023-07-24
Don't hard-code program name in help output

sol:


mpilgrem:
I've merged that, but hard-coded `stack` appears throughout Stack's in-app help. A change that was comprehensive in its scope would require quite a bit of work - and then, of course, Stack's online help would still refer to the command as `stack`.

sol:
> I've merged that, but hard-coded `stack` appears throughout Stack's in-app help. A change that was comprehensive in its scope would require quite a bit of work

Appreciated. We already use `progName` at some places, and this was a very low hanging fruit (and the only missing place for the main `stack --help` output).  No need to put more effort into this at this point.

PR #6249, opened 2023-09-21
list: output to stdout

juhp:
That was pretty bad - I had implemented `list` to log its output

I built it locally with stack and tested that output now goes to stdout

Please review the code - I was hoping to find a more rio idiomatic way to output.

Likely worthy of a short release note?
Dunno if this is too late for 2.13?
Once merged I'd be happy to open a backport if it is appropriate: I think it is a good bugfix.


mpilgrem:
@juhp, what was pretty bad? In https://github.com/commercialhaskell/stack/commit/d6b311e7a3818b098276ec02fc255d545c8f91b9 this changed from `logInfo` (your original, on 8 Dec 2020) to `prettyInfo`, but I think that change has no effect when what is being logged is not 'pretty' (as in this case) - all logging in Stack is to the standard error channel.

Do you have in mind a flag like the use of `--stdout` for the `stack ide` commands: https://docs.haskellstack.org/en/stable/ide_command/? See `Stack.IDE.outputFunc`.

juhp:
In comparison `stack ls` gets this right - maybe I should look at that

mpilgrem:
I suppose a distinction could be is whether or not the command output should be pipe-friendly. For example, I often use:
~~~text
> stack path --programs | cd
~~~

mpilgrem:
@juhp, I'll apply this to the release candidate.

mpilgrem:
My thinking is this:

* commands whose central purpose is to provide information to people, should output that information to standard output;
* commands whose central purpose is to provide information to 'machines' and where pipe-friendly is not important, should output that information to standard error but provide a `--stdout` flag. That accommodates `stack ide`, for example; and
* other information produced by commands that is 'logging' in nature can continue to go to standard error.

Issue #6065, opened 2023-03-04
Rationalise `Package` type's `packageDeps`, `packageAllDeps` and/or `packageSubLibDeps` fields


mpilgrem:
This arises out of a code comment in #5659. The data constructor of the `Package` type now includes fields:
~~~haskell
, packageDeps :: !(Map PackageName DepValue)
  -- ^ Packages that the package depends on, both as libraries and build tools.
, packageAllDeps :: !(Set PackageName)
  -- ^ Original dependencies (not sieved).
, packageSubLibDeps :: !(Map MungedPackageName DepValue)
  -- ^ Original sub-library dependencies (not sieved).
~~~

The question is: is some or all of that duplicative and capable of rationalisation? The answer to that question is likely downstream of solving the residual problem of getting sub-library dependencies to work (eg #5839).

PR #6334, opened 2023-11-12
Refactoring package to component centered design

theobat:
This is a first step towards a component based architecture in stack and a package representation closer to the cabal one.
The idea is, following this : https://github.com/commercialhaskell/stack/issues/6065, we had too many scattered representations of component centric data. This is a good stab in regrouping everything in a very cabal-close design.

The main refactoring this MR propose is to have collections of cabal-like components, such as for instance : 
```hs
data StackLibrary = StackLibrary
  { name :: StackUnqualCompName,
    buildInfo :: !StackBuildInfo,
    -- | This is only used for gathering the files related to this component.
    exposedModules :: [ModuleName]
  }
  deriving (Show, Typeable)
```

Which is a mirror of Cabal's Library, with only the things Stack needs.

This MR already brings a lot of change, but keep in mind that :
- I'm still working on implementing all the consequences of this new design, primarily around refactoring the dependencies to simply iterate on each component's dependencies. So I believe this MR is good enough to be merged, but it's paving the way of even better changes (refactorings really).
- It's a pure refactoring nothing is changed functionally


mpilgrem:
@theobat, the integration tests are failing because (I think) you have edited the `stack.cabal` file 'by hand'. The autogenerated `stack.cabal` has the same `other-modules:` modules, but listed in a different order. If you `stack build`, you will get the `stack.cabal` file that is expected, and that is the one that needs to be committed.

Stan is complaining because the ignored `OBS-STAN-0203-fki0nd-1111:21` how has a different location (`OBS-STAN-0203-fki0nd-1118:21`). The `.stan.toml` file needs to be updated.

theobat:
All good, thanks @mpilgrem

mpilgrem:
@theobat, I likely will not work through all of this tonight. However, I have pushed to your branch my progress so far. I am applying consistent formatting as I go (that is what almmost all of my further changes are about); as we discussed elsewhere - I don't ask that you waste your own time thinking about that.

mpilgrem:
Still working through it, also reformatting for consistency as I go (another batch of commits). Some of this has picked up some unrelated formatting.

src/Stack/Package.hs
```diff
@@ -802,3 +810,46 @@ applyForceCustomBuild cabalVersion package
...
+  gatherUnknownTools :: HasBuildInfo x => CompCollection x -> Set Text
+  gatherUnknownTools = foldr' addUnknownTools mempty
+
+packageSubLibrariesNameSet :: Package -> Set Text
+packageSubLibrariesNameSet pkg = getBuildableSetText (packageSubLibraries pkg)
+
+packageExes :: Package -> Set Text
+packageExes pkg = getBuildableSetText (packageExecutables pkg)
+
+packageBenchmarks :: Package -> Set Text
```
mpilgrem, line 854: I wondered if this would be better named `buildableBenchmarks` (with similar names for the corresponding functions for other component types) - partly because of my renaming suggestion elsewhere.

src/Stack/Package.hs
```diff
@@ -802,3 +810,46 @@ applyForceCustomBuild cabalVersion package
...
+  exe = gatherUnknownTools $ packageExecutables pkg
+  addUnknownTools :: HasBuildInfo x => x -> Set Text -> Set Text
+  addUnknownTools = (<>) . Component.sbiUnknownTools . getField @"buildInfo"
+  gatherUnknownTools :: HasBuildInfo x => CompCollection x -> Set Text
+  gatherUnknownTools = foldr' addUnknownTools mempty
+
+packageSubLibrariesNameSet :: Package -> Set Text
+packageSubLibrariesNameSet pkg = getBuildableSetText (packageSubLibraries pkg)
+
+packageExes :: Package -> Set Text
```
mpilgrem, line 851: I wondered if this would be better renamed `buildableExes` (see my similar comment below for explanation).

src/Stack/Package.hs
```diff
@@ -802,3 +810,46 @@ applyForceCustomBuild cabalVersion package
...
+  tests = gatherUnknownTools $ packageTestSuites pkg
+  flib = gatherUnknownTools $ packageForeignLibraries pkg
+  sublib = gatherUnknownTools $ packageSubLibraries pkg
+  exe = gatherUnknownTools $ packageExecutables pkg
+  addUnknownTools :: HasBuildInfo x => x -> Set Text -> Set Text
+  addUnknownTools = (<>) . Component.sbiUnknownTools . getField @"buildInfo"
+  gatherUnknownTools :: HasBuildInfo x => CompCollection x -> Set Text
+  gatherUnknownTools = foldr' addUnknownTools mempty
+
+packageSubLibrariesNameSet :: Package -> Set Text
```
mpilgrem, line 848: I wondered if this would be better renamed `buildableSubLibs` (see my similar comment below for explanation).

Also, good to have `buildableForeignLibs` and `buildableTestSuites`, to complete the set of similar functions.

src/Stack/Ghci.hs
```diff
@@ -957,17 +960,19 @@ makeGhciPkgInfo installMap installedMap locals addPkgs mfileTargets pkgDesc = do
...
      )
+  <> map CExe (S.toList (packageExes pkg))
+  <> map CSubLib buildableSubLibs
+  <> (if boptsTests bopts then map CTest buildableTestSuites else [])
+  <> (if boptsBenchmarks bopts then map CBench buildableBenchmarks else [])
+ where
+  buildableForeignLibs = getBuildableListText $ packageForeignLibraries pkg
+  buildableSubLibs = getBuildableListText $ packageSubLibraries pkg
+  buildableTestSuites = getBuildableListText $ packageTestSuites pkg
+  buildableBenchmarks = getBuildableListText $ packageBenchmarkSuites pkg
```
mpilgrem, line 975: When I refactored to create these four lines (to avoid long lines above), I realised that we now have `Stack.Package.packageSubLibrariesNameSet` (which I would rename `buildableSubLibs`) and `Stack.Package.packageBenchmarks` (which I would rename `buildableBenchmarks`). I was wondering: should we be consistent and have five functions `Stack.Package.buildable<component_type> :: Package -> Set Text` for each of the component types? (So `packageExes` gets renamed `buildableExes`.)

mpilgrem (commented):
@theobat, all looks good to me, but I did ponder about the naming of a small number of things. See my specific comments. What do you think?

src/Stack/Package.hs
mpilgrem: I have given effect to the first part of this suggestion (`buildableSubLibs`).

src/Stack/Ghci.hs
theobat: yes thaat sounds good !

src/Stack/Package.hs
```diff
@@ -802,3 +810,46 @@ applyForceCustomBuild cabalVersion package
   forceCustomBuild =
        packageBuildType package == Simple
     && not (cabalVersion `withinRange` cabalVersionRange)
+
+-- | Check if the package has a main library that is buildable.
+hasMainBuildableLibrary :: Package -> Bool
```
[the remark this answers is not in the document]
theobat: yes way better

src/Stack/Package.hs
mpilgrem: I have also given effect to new `Stack.Package.buildableTestSuites`.

mpilgrem: Finally, I have also given effect to new `Stack.Package.buildableForeignLibs`.

mpilgrem:
@theobat, with the exception of a little reformatting and my renaming a small number of identifiers, I have merged what you proposed. However, I have a question you may be able to help with: the following expressions evaluate to the same value. Can one be rationally be preferred over the other?
~~~haskell
Set.toList $ buildableSubLibs package

getBuildableListText $ packageSubLibraries package
~~~

theobat:
Thanks ! @mpilgrem 
While both are equal **_now_** in terms of computation and meaning, I'd rather use the later, as it allows for direct folds over the  underlying CompCollection structure. In fact, this collection may or may not continue as the combination of a set and a hashmap of buildable components. In particular, it'd be better globally for stack to bend toward the usage of HashSets because benchmarks show that they are more performant, and in this case we'd remove the `asNameSet`in CompCollection as it'd be strictly redundant over a simple HashMap.

mbj:
@theobat, @mpilgrem I just found that on `master` the `--file-watch` feature is broken. And bisected it to commit: a5ff987b0817873666ecf2c8e7dd44ce9747d5c6 from this PR.

I'll try to send a fix PR and open an issue.

mbj:
Sorry for deleting my earlier comment, I thought my bisect went wrong but it did not. So the commit: a5ff987b0817873666ecf2c8e7dd44ce9747d5c6 is (according to my bisect) the first commit showing the symtoms of this issue: https://github.com/commercialhaskell/stack/issues/6365

PR #6357, opened 2023-12-03
refactoring: put installed stuff into a dedicated types package

theobat:
Because it's not going to stay a "Package" thing in the future if we build at the component level

### DIFF
--- a/.stan.toml
+++ b/.stan.toml
@@ -54,25 +54,25 @@
 
 # Anti-pattern: Data.ByteString.Char8.pack
 [[ignore]]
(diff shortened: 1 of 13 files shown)


mpilgrem:
I think the integration tests are failing because you have edited the `stack.cabal` and not the `package.yaml`. I think you need to do the latter, and then `stack build --dry-run` to create the `stack.cabal` from it.

theobat:
@mpilgrem I fixed it thanks, it's only missing the ARM64 now

PR #6384, opened 2023-12-16
Fix #6383 Move Stack on to GHC 9.6.3/LTS 22.0

mpilgrem:
* [x] Any changes that could be relevant to users have been recorded in ChangeLog.md.
* [x] The documentation has been updated, if necessary

Please also shortly describe how you tested your change. Bonus points for added tests! Relying on CI.

PR #6385, opened 2023-12-17
Misc refactorings

theobat:
Minor refactorings and additions for component based builds

### DIFF
--- a/src/Stack/Types/NamedComponent.hs
+++ b/src/Stack/Types/NamedComponent.hs
@@ -5,6 +5,7 @@
 module Stack.Types.NamedComponent
   ( NamedComponent (..)
   , renderComponent
+  , renderComponentTo
   , renderPkgComponents
   , renderPkgComponent
   , exeComponents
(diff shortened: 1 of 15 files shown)


theobat:
This is currently failing due to a bug in ghc.
It's arguably related to 9.4.* and should not be there in 9.6

see 
https://gitlab.haskell.org/ghc/ghc/-/issues/23972

mpilgrem:
@theobat, I'm currently 'almost there' in moving Stack on to GHC 9.6.3 (see #6384).

mpilgrem:
@theobat, I rebased on master (and a GHC 9.6.3-based Stack code base).

mpilgrem:
I am going to close and reopen the pull request, with the intent of forcing the CI to run.

mpilgrem:
I've taken the opportunity to make use of helper function `fromPackageName` consistently (both in these changes and elsewhere).

src/Stack/Types/NamedComponent.hs
```diff
@@ -119,3 +132,4 @@ splitComponents =
   go s e t b (CExe x:xs) = go s (e . (x:)) t b xs
   go s e t b (CTest x:xs) = go s e (t . (x:)) b xs
   go s e t b (CBench x:xs) = go s e t (b . (x:)) xs
+  go s e t b x = go s e t b x
```
mpilgrem, line 135: @theobat, as the righthand side is the same as the lefthand side, how will this ever terminate? Presumably `x` would be `(CFlib x : xs)`, as the other patterns are catered for. I wondered if this was intended to be `go s e t b (CFlib x : xs) = go s e t b xs`?

mpilgrem (changes requested):
I had only one substantive observation.

src/Stack/Types/NamedComponent.hs
theobat: that's right, good catch, I'll fix it

mpilgrem:
In the force-pushing, my reformatting commit has been lost. I'll see if I can reapply it from my local version this evening.

theobat:
@mpilgrem I'm sorry I lost your reformatting I'll try to pull before doing my rebase next time. This last force push should have repaired the broken tests, I only provide the missing info to ghci when there's actually a dependance on a sublibrary.

So, it turns out this MR also contains the fix for the `stack ghci my-project:lib` example that failed before

PR #6398, opened 2024-01-08
Eliminate use of RecordWildCards

mpilgrem:
Motivation: @theobat's [comment](https://discourse.haskell.org/t/ghc2024-community-input/8168/85?u=mpilgrem) at the Haskell Community, and the downstream discussion there.

Only implements 'step 1' (that is, does not eliminate the use of prefixes that indicate the type/data constructor in the naming of fields, to give them unique names). EDIT: My plan would be to implement 'step 2' (elimination of prefixes) separately from this pull request.

Uses `OverloadedRecordDot`, introduced from GHC 9.2.1 (`base-4.16.0.0`) - so bumps lower bound on `base` package.

Replaces:

* `f Constructor{..} = ... prefixFieldName` with (mostly) `f prefix = ... prefix.prefixFieldName`; and
* `pure Constructor{..}` with `pure $ Constructor {prefixFieldName1, ..., prefixFieldNameN}`;

and puts spaces around existing `(.) :: (b -> c) -> (a -> b) -> a -> c` operators.

### DIFF
--- a/.hlint-test.yaml
+++ b/.hlint-test.yaml
@@ -10,9 +10,10 @@
     - NoImplicitPrelude
     - CPP
     - OverloadedLists
+      # Provided from GHC 9.2.1 (base-4.16.0.0):
+    - OverloadedRecordDot
     - OverloadedStrings
     - QuasiQuotes
-    - RecordWildCards
     - Safe
     - ScopedTypeVariables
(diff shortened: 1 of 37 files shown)


theobat (approved):
That's really great @mpilgrem thanks for putting this together !

PR #6402, opened 2024-01-12
Build with NoFieldSelectors

mpilgrem:
Refactor to allow the `Stack` package to build with `--ghc-options -XNoFieldSelectors`. This is intended as a prelude to removing prefixes from the names of fields.

This current step is done by making use of `OverloadedRecordDot`.

Something like `prefixFieldName prefix` becomes `prefix.prefixFieldName`. Something like `f prefixFieldName` becomes `f (.prefixFieldName)` - and 'chains of dots' are simplified.

In a couple of places, the compiler needed help with types:

* `Stack.Types.ConfigMonoid` (`(mempty :: Map GhcOptionKey GhcOptions)`)
* `Stack.Build.Cache` (`let decode :: MonadIO m => m BuildCache`)
* the use of `undefined :: <type>` in `Stack.Types.BuildOpts.defaultBuildOpts` and `defaultTestOpts`

I also did not follow why the `$` was still needed (but it is) in moving from (in `Stack.Build.Execute`):
~~~haskell
(fmap . fmap) (getField @"interface") <$> collectionKeyValueList $ packageTestSuites package
~~~

to

~~~haskell
(fmap . fmap) (getField @"interface") <$> collectionKeyValueList $ package.packageTestSuites
~~~

EDIT: I now follow the need for the `$`. The expression is equivalent to:
~~~haskell
((fmap . fmap) (getField @"interface") <$> collectionKeyValueList) package.packageTestSuites
~~~

### DIFF
--- a/.hlint-test.yaml
+++ b/.hlint-test.yaml
@@ -9,6 +9,7 @@
   - name:
     - NoImplicitPrelude
     - CPP
+    - NoFieldSelectors
     - OverloadedLists
       # Provided from GHC 9.2.1 (base-4.16.0.0):
     - OverloadedRecordDot
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -70,6 +70,7 @@
     - GeneralizedNewtypeDeriving
     - LambdaCase
     - MultiWayIf
+    - NoFieldSelectors
     - OverloadedLists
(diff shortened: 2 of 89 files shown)


theobat:
@mpilgrem yeah thanks a lot, it looks great otherwise.

PR #6411, opened 2024-01-13
Minor refactoring

theobat:
Many things arre happening in the Execute Module, which requires a big split, like prior work, we'll have execute, executePackage, executeComponent


mpilgrem:
@theobat, my plan is [1] to merge a couple of field namings that I have in progress, then [2] rebase this pull request against the master branch and merge it, then [3] continue with other field namings.

theobat:
ok @mpilgrem I just rebased, but feel free to add some more renamings

mpilgrem:
@theobat, after I merged this, I realised that it has code duplicated in `Stack.Build.Execute` and `Stack.Build.ExecuteEnv`. For example, both define and export `withSingleContext`. Was that the intention?

theobat:
@mpilgrem no the datatypes and functions exported by ExecuteEnv should only be imported in Execute and not be redefined (the code is the same, it's only a module breakup)

PR #6451, opened 2024-01-20
Fix #5974 If new Cabal copy is available, use it

### LINKED ISSUE #5974: Rebuilding and running an executable installs every executable configured
### General summary/comments

When only a single executable is required to be (re)built, building and running that executable installs _all_ other binaries configured.

### Steps to reproduce

1. Start a new stack project;
2. Configure more than one (preferably many) executable;
3. Build all executables;
4. Make trivial modification to the `Main.hs` for any executable;
5. Run `stack run that_executable`;
6. Stack installs _all_ exes, even though only a single executable was rebuilt.

### Expected

Since the source changes (step 4 above) affected only a single executable, the reasonable expectation from Stack is that it rebuilds and run that executable alone&mdash;do nothing more.

### Actual

Instead, as mentioned, Stack installs every executable whenever there is _any_ changes made. This is wasteful and unnecessary.

The back story:  was trying my hand at AOC 2022. I configured each day's puzzle as a separate binary, for example, under `app/day12_1` and `app/day12_2`. Due to the incremental nature of development, every time I make changes to `day12_2`, for example, every other executable is installed before `day12_2` is run. That is 12x2, 24 binary installs. Needless to say, this increases the turn around time. And it gets worse with every passing day, when I keep adding more and more executables.

My current work around is to configure one executable at a time, and comment out the rest. I feel shouldn't have to do this. Stack is acting suboptimally. Executables are supposed to be independent of each other; changing one shouldn't cause any action to be taken upon the rest (unless their common dependency changes, which in this case, it didn't).

I believe I've tried all command line options to get Stack to break out of it, but without any success.

### Stack version

Version 2.9.1.

### Method of installation

Official binary, downloaded via haskellstack.org or from Stack's repository

mpilgrem:
Cabal's `copy` used to not allow components to be specified, and Stack had to build all executable components at least once for `copy` not to fail. This pull request continues to do that only if the Cabal version is sufficiently old (before `Cabal-2.0.0.0`). Otherwise, it copies only the library, sub-library, foreign library and executable components that were to be built. As a consquence, it does not need to build all executable components at least once for the Cabal version to be satisfied.

The tutorial part of the online guide to Stack is updated accordingly.

Please also shortly describe how you tested your change. Bonus points for added tests! Tested locally, by reference to output of `internal-libraries` integration test.


mpilgrem:
@theobat, if you could run your eye over this, I would be grateful. if what I have added works, then once Stack drops support for Cabal < 2.2, the historical complexity and the complexity of taking different approaches with different Cabal versions will fall away.

theobat:
@mpilgrem Yes that seems good.

PR #6473, opened 2024-02-04
Minor refactoring

theobat:
This includes two types of refactorings : 
- Some simplifications and function isolation of basic tasks (plus docs) in ExecutePackage
- A change in the CompCollection to use a simple Map rather than the original HashMap. The rationale for using a HasMap was that it scaled better for text/bytestring based keys (see [this benchmark](https://github.com/haskell-perf/dictionaries)), but after a careful attempt at reproducing this in smaller scenarios, It seemed the opposite was true (slightly better times for simple Map below 1000 keys). Given the expected number of components in packages, it seems it made sense to go back to a simple Map.

### DIFF
--- a/src/Stack/Build/Cache.hs
+++ b/src/Stack/Build/Cache.hs
@@ -71,9 +71,9 @@ import           Stack.Types.EnvConfig
                    , installationRootDeps, installationRootLocal
                    , platformGhcRelDir
                    )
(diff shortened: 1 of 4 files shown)


mpilgrem:
@theobat, I merged manually, after a little reformatting, for consistency.

PR #6483, opened 2024-02-12
Fix #6377 Drop support of Cabal < 2.2

### LINKED ISSUE #6377: Deprecate, then remove, Stack's support of Cabal < 2.2
Stack makes use of Cabal (the library) by building/using a small executable which depends on the version of Cabal that is provided with the specified version of GHC.

The motivation for Stack ceasing to support Cabal < 2.2 in that way is @theobat's explanation at https://github.com/commercialhaskell/stack/issues/6356#issuecomment-1839589226. In short, the barriers to introducing Stack support for component-based builds (which might help unlock things like better HLS support for Stack) would be lowered.

`Cabal-2.2.0.0` was uploaded to Hackage on 8 March 2018. It is provided with GHC 8.4.1, released on the same date. Stackage LTS Haskell 12.0, published 9 July 2018, was based on GHC 8.4.3 (replacing LTS 11.22, based on GHC 8.2.2).

On 4 December 2023, at the [Haskell Community](https://discourse.haskell.org/t/stack-support-for-ghc-versions-before-ghc-8-4-1/8275), Reddit, Slack (#stack-users, #stack-collaborators), and Matrix (rooms: Stack and Stackage, Cabal and Hackage), I asked people to flag if important it was important to them _in practice_ that the next version of Stack supports a version of GHC before GHC 8.4.1. My emphasis was on the practical requirement, not matters of principle. So far, although it has been only 10 days, no person has flagged that it is important to them. The Reddit post has had 3,400 views; the HC post 622 views. Some people have expressly said that no weight should be placed on pre-GHC 8.4.1 support, including a past member of the Board of the Haskell Foundation.

@hasufell, helpfully, identified that current versions of certain key GHC boot libraries are dropping support for GHC < 8.6. However, there is a difference between the current version of a library dropping support and a build tool dropping support. In contrast to Stack, Cabal (the tool) does not promise to support GHC versions released more than three years ago.

My proposal is to deprecate Stack's support of Cabal < 2.2 in the next version of Stack after Stack 2.13.1 (by announcing that when it is released) with a plan of removing it (unless there is relevant protest) in the version after that.

mpilgrem:

### DIFF
--- a/.stan.toml
+++ b/.stan.toml
@@ -72,25 +72,25 @@
 
 # Anti-pattern: Data.ByteString.Char8.pack
 [[ignore]]
-  id = "OBS-STAN-0203-erw24B-1028:3"
+  id = "OBS-STAN-0203-erw24B-1021:3"
 # ✦ Description:   Usage of 'pack' function that doesn't handle Unicode characters
 # ✦ Category:      #AntiPattern
 # ✦ File:          src\Stack\Build\ExecuteEnv.hs
 #
-#  1027 ┃
-#  1028 ┃   S8.pack . formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S%6Q"
-#  1029 ┃   ^^^^^^^
+#  1020 ┃
+#  1021 ┃   S8.pack . formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S%6Q"
+#  1022 ┃   ^^^^^^^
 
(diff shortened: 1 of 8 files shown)


mpilgrem:
@theobat, I think this implements what we have discussed. However, before I put it through, may I double-check that you still think that dropping support for Cabal < 2.2 will make it easier for you to make the other changes that you have in mind?

PR #6485, opened 2024-02-14
fix: wrong instance for qualified name of benchmarks

theobat:

@mpilgrem While there is no urge to do so, I think this fix should be backported to stack 2.15, because it'll likely create wrong behaviors in ghci for benchmarks (and probably in hls then).


mpilgrem:
@theobat, noted. I'm OK with releasing a new version of Stack pretty quickly, but lets give it a few weeks to see if any other bugs emerge.

PR #6587, opened 2024-05-18
Hash

Hawthorne001:

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -1,0 +1,43 @@
+name: Haskell CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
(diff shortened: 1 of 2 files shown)


mpilgrem:
@Hawthorne001, please can you elaborate? As set out in `CONTRIBUTING.md`, it is a good idea to raise an issue for discussion before proposing a pull request to resolve it.

I can't immediately see the relevance of `github/super-linter@v4` to this project. It is already has code-quality checks for Haskell (which `super-linter` does not appear to support) and YAML. Also, why is that fork of `super-linter/super-linter` preferable?

I also don't know the relevance of a GitHub workflow that builds Stack with Cabal (the tool). Stack is intended to be built by Stack; building with other tools is not actively maintained. I think an optional workflow 'on request' is fine, but I would not want one that could block Stack development.

PR #6597, opened 2024-06-01
Fix #5925 Add file-watch-hook for post-processing

### LINKED ISSUE #5925: Addition of '--exec-always' flag: a '--exec' that runs even on build failure
**Introduction**

There are times when one would like to use `--exec` even when the build fails. As an example, imagine running 
`stack build ... --exec 'do-stuff-with-error-messages-on-stdout.sh  input.txt' --file-watch"`. In this scenario, we want to process the stdout that was generated every time `--file-watch` causes a rebuild. 

With the current `--exec` flag implementation, that shell script would only run on a successful build. Yet since it is doing something with the error messages, it would be useful to have it run even on failure as that would ultimately help the programmer make the build succeed in this case. 

**Proposal**

The proposal is a new flag `--exec-always` that can execute even if the build fails and also while the `--file-watch` flag has been passed. This would go a long way to helping people build their own tooling workflows around `stack`.

mpilgrem:

This is draft while testing on different operating systems.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -68,6 +68,9 @@ Other enhancements:
 * Add the `ls globals` command to list all global packages for the version of
   GHC specified by the snapshot.
 * Add `stack -h` (equivalent to `stack --help`).
+* In YAML configuration files, the `file-watch-hook` key is introduced to allow
+  `--file-watch` post-processing to be customised with a executable or `sh`
+  shell script.
 
 Bug fixes:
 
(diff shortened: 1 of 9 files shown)


mpilgrem:
`--file-watch` is now misbehaving. I think this because I lifted the `withConfig` up out of the `inner` (because `fileWatchConf` itself needed access to the `file-watch-hook` configuration). I'll reverse that.

PR #6603, opened 2024-06-06
Warn about config limits for scripts

philderbeast:


mpilgrem:
@philderbeast, I am going to close this and put through something similar, but 'lighter'. That is because:
* changes to the documentation for the current version of Stack should target the `stable` branch; and
* this is already documented (although perhaps not as well as it could be) for the `stack script` command. This part of the documentation relates to something different (but related): the Stack interpreter options comment.

philderbeast:
I would have been able to change the merge branch wouldn't I?

I don't see any warnings in either section. There's one note in the script interpreter section. The script interpreter section is the one with more shebang examples. That's what I'm using too. I'm not using `stack script` from the command line.

I see in the stack script section, in one of the windows examples, that `--extra-dep` is used. I can't find that in the docs but [extra-deps](https://docs.haskellstack.org/en/stable/yaml_configuration/#extra-deps) is there.

```
$ grep 'extra-dep\s' *.md
ChangeLog.md:  target is an extra-dep from the commandline.
ChangeLog.md:* Switching a package between extra-dep and local package now forces
ChangeLog.md:  set local extra-dep packages as targets. This was overkill; we
```

mpilgrem:
The problem with changing the 'base' branch to `stable` is your original pull request then modifies hundreds of files (which is why I tried it and then immediately reversed trying it above).

The documentation for both `stack script` and the Stack interpreter options comment spell out that a project's project-level configuration file is ignored. Notes have been added to clarify that non-project specific configuration options in global configuration files are not ignored. The `stack script` documention goes on to explain that `allow-newer` etc can be a useful non-project specific configuration option.

The `--extra-deps` option of the `stack script` command is already documented, see https://docs.haskellstack.org/en/stable/script_command/#the-stack-script-command.

The Stack interpreter options comment can be used with any Stack command that has the correct form (albeit that there are only a small number that do have that form). There is some duplication in the documentation, but the commands' primary documentation is on their own pages.

PR #6607, opened 2024-06-09
Fix #6606 Distinguish global and project-level configuration files

### LINKED ISSUE #6606: Stack does not distinguish global and project-level configuration files in certain error messages
Motivation: #6602

Illustrated by (but not limited to) the error generated by Haskell script file:
~~~haskell
{- stack script
   --snapshot nightly-2024-06-07
   --package dhall
-}

main = pure ()
~~~

Currently, the `stackYaml` field of a value of type `BuildConfig` can refer to either a global or a project-level configuration file, in part because of `Stack.Config.withBuildConfig` (extract):
~~~haskell
 (project', stackYaml) <- case config.project of
    ...
    PCNoProject extraDeps -> do
      p <- ...
      pure (p, config.userConfigPath) -- <<<< refering to the global configuration file!
~~~

It seems to me that types could be used to distinguish values that represent the location of global configuration files from those that represent the location of any project-level configuration file. Also, the field is a potential misnomer, as `stackYaml` could be understood to refer to `stack.yaml`, the default name of a project-level configuration file.

mpilgrem:
Also refactors where:

* 'project' is used to refer to a file or directory that may not be within a project directory but within the Stack root; or
* `stackYaml` is used to refer to a configuration file that may be a global one as well as a project-level one.

### DIFF
--- a/.stan.toml
+++ b/.stan.toml
@@ -83,14 +83,14 @@
 
 # Anti-pattern: Data.ByteString.Char8.pack
 [[ignore]]
-  id = "OBS-STAN-0203-tuE+RG-234:21"
+  id = "OBS-STAN-0203-tuE+RG-234:24"
 # ✦ Description:   Usage of 'pack' function that doesn't handle Unicode characters
 # ✦ Category:      #AntiPattern
 # ✦ File:          src\Stack\Build\ExecutePackage.hs
 #
 #   233 ┃
-#   234 ┃   newProjectRoot <- S8.pack . toFilePath <$> view projectRootL
-#   235 ┃                     ^^^^^^^
+#   234 ┃   newConfigFileRoot <- S8.pack . toFilePath <$> view configFileRootL
+#   235 ┃                        ^^^^^^^
 
 # Anti-pattern: Data.ByteString.Char8.pack
 [[ignore]]
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -78,6 +78,8 @@ Bug fixes:
   specified at the command line.
 * Fix a regression, introduced in Stack 2.11.1, that caused the `script` command
   to parse an (otherwise ignored) project-level configuration file.
+* Stack no longer makes recommendations about a project-level configuration file
+  when only a global configuration file is in use.
 
 ## v2.15.7 - 2024-05-12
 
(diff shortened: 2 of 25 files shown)


philderbeast:
As this includes a renaming, would it make sense to drop `Yaml` altogether in those names and go with more meaningful names, something like `Project*` versus `Global*`, where `*` is something like `Config` or `Settings`?

mpilgrem:
@philderbeast, I am open to suggestions on naming of the field of the `BuildConfig` constructor - I think everything else flows from that choice. At the moment, Stack has:
* at most one project-level YAML configuration file, named `stack.yaml` by default but can be any `*.yaml` file; and
* a user-specific global YAML configuration file, named `config.yaml` in the Stack Root; and
* (according to the documentation; I need to check the code to see if that is correct) perhaps, system-wide global YAML configuration file, named `config.yaml` in `/etc/stack` (but that is a Unix-like file path?).

I picked `yamlFile` for the path to the YAML configuration file because what `*.yaml` and `config.yaml` had in common was the 'YAML' part. The problem with `configYaml` or, even, `configFile` is that people might assume it refers to `config.yaml` only, given the 'config' part. The problem with `settingsFile` or `settingsYaml` is that the term 'settings' is not used anywhere else to refer to these files.

mpilgrem:
Looking at `Stack.Config.loadConfig` and:

~~~haskell
extraConfigs0 <- getExtraConfigs userConfigPath >>=
    mapM (\file -> loadConfigYaml (parseConfigMonoid (parent file)) file)
~~~

the 'extra' global configuration files are under- or un-documented on online. I'll raise that as a separate issue.

mpilgrem:
@philderbeast, on reflection, I decided to prefer `configFile` to `yamlFile` and, where applicable, use `Right projectConfigFile ->`.

philderbeast:
Your choice of `ConfigFile` aligns with Cabal naming a similar thing [`ProjectConfigPath`]( https://github.com/haskell/cabal/blob/2140c0ff7af789bc47e08f9ea5d3929c0082d5b8/cabal-install-solver/src/Distribution/Solver/Types/ProjectConfigPath.hs#L35-L47)[^1].

[^1]: YAML can't do imports but Cabal's projects can so that is why it is a path and not a file over there.

PR #6608, opened 2024-06-09
Warn about config limits for scripts

philderbeast:
Same as #6603 but cherry-picked over the stable branch.

### DIFF
--- a/doc/scripts.md
+++ b/doc/scripts.md
@@ -14,6 +14,16 @@ is most often [`stack script`](script_command.md) but it can be, for example,
 [`stack runghc`](runghc_command.md). If there is no Stack interpreter options
 comment, Stack will warn that one was expected.
 
+!!! warning
+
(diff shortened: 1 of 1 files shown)


mpilgrem:
Closing, as duplicative of:
* the existing online documentation - see https://docs.haskellstack.org/en/stable/scripts/#stack-configuration-for-scripts and https://docs.haskellstack.org/en/stable/script_command/; and
* #6603.

The existing documentation for `stack script` and the Stack interpreter option comment already explain (at some length, contrasting with `stack runghc`) that:
* project-level configuration files are ignored and that this is a deliberate design choice; and
* global configuration is not ignored, and `allow-newer` etc is given as an express example.

If the underlying motivation is that the `stack script` command should have an `--allow-newer` flag, then please raise that as a separate issue.

mpilgrem:
I have, however, attempted to make the existing language in the online documentation clearer and remove any ambiguity.

PR #6609, opened 2024-06-09
get-stack.sh: offer to extend path for user

mathematicalmichael:
I saw a todo while inspecting the shell script that I could handle, having done this for other installers before.

tested on `zsh` and `bash` by setting `PATH` to something short, removing current installation, and running the shell script. Did not test `fish` but used what I believe is the default there.

### DIFF
--- a/etc/scripts/get-stack.sh
+++ b/etc/scripts/get-stack.sh
@@ -745,14 +745,91 @@ on_path() {
   echo ":$PATH:" | grep -q :"$1":
 }
 
+# Check whether the script may be running in well-known CI environments
+has_ci_environment() {
+  if [ -n "$CI" ]; then
+    # GitHub Actions, GitLab CI, Travis, CircleCI and Bitbucket Pipelines all
+    # set CI.
+    return 0
+  elif [ -n "$TR_BUILD" ]; then
+    # Azure Pipelines sets TR_BUILD.
+    return 0
+  elif [ -n "$JENKINS_HOME" ]; then
+    # Jenkins sets JENKINS_HOME.
+    return 0
+  else
+    return 1
+  fi
+}
+
 # Check whether ~/.local/bin is on the PATH, and print a warning if not.
 check_home_local_bin_on_path() {
   if ! on_path "$HOME_LOCAL_BIN" ; then
-    #TODO: offer to add it for the user (pull requests welcome!)
     info "WARNING: '$HOME_LOCAL_BIN' is not on your PATH."
     info "    Stack will place the binaries it builds in '$HOME_LOCAL_BIN' so"
     info "    for best results, please add it to the beginning of PATH in your profile."
     info ""
+    # detect which profile file to use, then print a message about updating it
+    if [ -n "$BASH_VERSION" ]; then
+      if [ -f "$HOME/.bash_profile" ]; then
+        profile_file="$HOME/.bash_profile"
+      else
+        profile_file="$HOME/.bashrc"
+      fi
+    elif [ -n "$ZSH_VERSION" ]; then
+      profile_file="$HOME/.zshrc"
+    elif [ -n "$FISH_VERSION" ]; then
+      profile_file="$HOME/.config/fish/config.fish"
+    elif [ -n "$PROFILE" ]; then
+      profile_file="$PROFILE"
+    elif [ -f "$HOME/.bash_profile" ]; then
+      profile_file="$HOME/.bash_profile"
+    elif [ -f "$HOME/.bashrc" ]; then
+      profile_file="$HOME/.bashrc"
+    elif [ -f "$HOME/.zshrc" ]; then
+      profile_file="$HOME/.zshrc"
+    elif [ -f "$HOME/.config/fish/config.fish" ]; then
+      profile_file="$HOME/.config/fish/config.fish"
+    elif [ -f "$HOME/.profile" ]; then
+      profile_file="$HOME/.profile"
+    else
+      info "    (profile not found; please add it to your PATH manually)"
+    fi
+
+    # print a message about updating profile (if found)
+    if [ -n "$profile_file" ]; then
+      info "    You can do this by running the following command:"
+      info "    echo 'export PATH=\"$HOME_LOCAL_BIN:\$PATH\"' >> \"$profile_file\""
+      info "    (You may need to restart your shell for this to take effect.)"
+      info ""
+
+      # prompt to update it on their behalf, unless QUIET is set.
+      if [ -z "$QUIET" ] && ! has_ci_environment ; then
+        while true; do
+          info "Would you like this installer to add it to your PATH in '$profile_file'?"
+          info "    (This will be done by adding export PATH=\"$HOME_LOCAL_BIN:\$PATH\" to it."
+          info "    You may need to restart your shell for this to take effect.)"
+          info "    [y] Yes, prepend  [n] No (default)"
+          read -p "" yesno
+          # default to no.
+          yesno=${yesno:-n}
+          case $yesno in
+            [Yy]* )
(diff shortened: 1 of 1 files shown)


mpilgrem:
@mathematicalmichael, many thanks! I'll have a look.

etc/scripts/get-stack.sh
```diff
@@ -748,11 +748,70 @@ on_path() {
...
+
+    # if we found a profile file, print a message about updating it
+    if [ -n "$profile_file" ]; then
+      info "    You can do this by running the following command:"
+      info "    echo 'export PATH=\"$HOME_LOCAL_BIN:\$PATH\"' >> \"$profile_file\""
+      info "    (You may need to restart your shell for this to take effect.)"
+      info ""
+    fi
+    # if we found a profile file and we're not in quiet mode, prompt user to update it
+    if [ -z "$QUIET" ]; then
```
mpilgrem, line 790: Despite what line 789 says, this part seems to me to be outside the test for the existence of contents of `profile_file` at line 783, given the `fi` at line 788?

This shell script is not expected to be used on Windows except, possibily, as part of CI (see line 275) but I assume on Windows there is no 'profile' file and so it is important that the prompt is within the scope of the test.

mathematicalmichael: sorry about that. careless oversight. fixed.

etc/scripts/get-stack.sh
```diff
@@ -748,11 +748,71 @@ on_path() {
...
+    
+      # prompt to update it on their behalf, unless QUIET is set.
+      if [ -z "$QUIET" ]; then
+        while true; do
+          # default to no.
+          yesno=n
+          info "Would you like this installer to add it to your PATH in '$profile_file'?"
+          info "    (this will be done by adding export PATH=\"$HOME_LOCAL_BIN:\$PATH\" to it)"
+          info "    (you may need to restart your shell for this to take effect)"
+          info "    (y/N)"
```
mpilgrem, line 797: I think `(y/N)` should be `(y/n)`, given line 810?

mathematicalmichael: afaik convention is to capitalize the default selection. if it were `(Y/n)` then `yes` would be the presumed default.

mpilgrem: On conventions, a search found this discussion: https://ux.stackexchange.com/questions/53640/default-values-for-text-based-confirmation-prompt. I am not a Linux user, which may explain why I had not encountered it.

However, one thing I am seeking to do generally with the Stack project is reduce the level of 'assumed knowledge', including knowledge of conventions. In that regard, I think what the GHCup project does with its prompts is informative. See the `sh` script at https://get-ghcup.haskell.org and `ask_bashrc()`. GHCup 'spells it out'. Its equivalent would be something like:

`[Y] Yes, prepend  [N] No (default is "N").`

mpilgrem:
I tested this on Windows with `sh get-stack.sh` (in `...\etc\scripts`, in the Stack-supplied MSYS2 environment), and got the following output (extract). Pressing ENTER was not recognised as the default. `C:\Users\mike\.bashrc` does not exist:
~~~text
Stack has been installed to: /bin/stack

WARNING: '/bin' is not on your PATH.

WARNING: '/c/Users/mike/.local/bin' is not on your PATH.
    Stack will place the binaries it builds in '/c/Users/mike/.local/bin' so
    for best results, please add it to the beginning of PATH in your profile.

    You can do this by running the following command:
    echo 'export PATH="/c/Users/mike/.local/bin:$PATH"' >> "/c/Users/mike/.bashrc"
    (You may need to restart your shell for this to take effect.)

Would you like this installer to add it to your PATH in '/c/Users/mike/.bashrc'?
    (this will be done by adding export PATH="/c/Users/mike/.local/bin:$PATH" to it)
    (you may need to restart your shell for this to take effect)
    [y] Yes, prepend  [n] No (default)

Please answer 'y' or 'n'
Would you like this installer to add it to your PATH in '/c/Users/mike/.bashrc'?
    (this will be done by adding export PATH="/c/Users/mike/.local/bin:$PATH" to it)
    (you may need to restart your shell for this to take effect)
    [y] Yes, prepend  [n] No (default)
n
Not updating PATH in '/c/Users/mike/.bashrc'
~~~

I am still investigating why.

etc/scripts/get-stack.sh
```diff
@@ -748,11 +748,71 @@ on_path() {
...
+      # prompt to update it on their behalf, unless QUIET is set.
+      if [ -z "$QUIET" ]; then
+        while true; do
+          # default to no.
+          yesno=n
+          info "Would you like this installer to add it to your PATH in '$profile_file'?"
+          info "    (this will be done by adding export PATH=\"$HOME_LOCAL_BIN:\$PATH\" to it)"
+          info "    (you may need to restart your shell for this to take effect)"
+          info "    [y] Yes, prepend  [n] No (default)"
+          read -p "" yesno
```
mpilgrem, line 798: I think this mechanism, together with line 793, does not implement a default value of `n` in variable `yesno` if ENTER is pressed. I think you may need something like:
~~~bash
read -p "" yesno
: ${yesno:=n}
~~~
~~~

mathematicalmichael: thanks for testing on windows! i don't have access to one of those.
fix incoming.

mathematicalmichael: handled. also added a timeout in case someone uses this script in a docker image or something like that and does not set QUIET.

mpilgrem: I think timeout will be important because this script is used in CI, historically it has not needed any human intervention, its `-q` flag is not documented online, and (in any event) `-q` (quiet) does not always imply 'non-interactive'.

Consequently, further, I wonder if it should also skip the prompt if any of the environment variables that indicate popular CI environments are present. ChatGPT offers up the following list (which may need to be checked for hallucinations):
~~~text
Jenkins: JENKINS_HOME
Travis CI: TRAVIS
CircleCI: CIRCLECI
GitHub Actions: GITHUB_ACTIONS
GitLab CI: GITLAB_CI
Azure Pipelines: TF_BUILD
Bitbucket Pipelines: BITBUCKET_BUILD_NUMBER
~~~

mpilgrem: The `-q` flag is now documented online as 'suppresses output and specifies non-intervention (likely a prerequisite for the use of the script in CI environments)'.

mathematicalmichael: agreed that timeout is important. hm. do you think this script needs to cover for all those, or just document that the `QUIET` env should be set in any of those situations?

mpilgrem: I think that something like:
~~~sh
if [ -z "$QUIET" ] || ! has_ci_environment ; then
~~~

would assist because I am concerned that some existing use of the script in CI may simply assume that it is non-interactive.

mpilgrem: I read some more about environment variables in CI and added a commit, accordingly.

mpilgrem:
I now understand the Windows behaviour: when you run a script with the MSYS2-supplied `sh` tool, the `BASH_VERSION` variable is set and available to the script.

mpilgrem:
After testing on Ubuntu, I discoved that POSIX-compliant `read` does not support `-e` or `-t` options. They were giving rise to errors:
~~~text
get-stack.sh: 813: read: Illegal option -e
get-stack.sh: 813: read: Illegal option -t
~~~

So, I have added a commit to remove them and the idea of timeout and will rely on the `has_ci_environment` guard.

mpilgrem:
@mathematicalmichael, I am going to tidy-up the commit history in preparation for merging.

mpilgrem:
@mathematicalmichael, many thanks. That should now be live.

mathematicalmichael:
thanks!
sorry about the radio silence, and thanks for accepting the contribution

Issue #6593, opened 2024-05-26
Refactoring proposal : yet another move towards cabal 


theobat:
So, while working on backpack related processing, I realized that there are many things that stack duplicate in very slightly different ways from cabal, with no obvious benefit and pretty obvious downsides (such as conversion costs). I've come to a point where I either have to duplicate yet another huge chunk of backpack related logic into Stack because the types differ, or I can try to refactor current duplicates from Stack which make no sense. This would benefit Stack by reducing the amount of  conversion and parsing logic done on stack's side (but some conversation will still be needed at places but I'd expect far less than the current state of things). This should also reduce the amount of code in stack at the cost of slightly hampered flexibility and maintenance burden for the imported logic.
This issue is a request for comment for the following refactorings (That I would do if I'm allowed):
- The entire [DumpPackage](https://github.com/commercialhaskell/stack/blob/49fe9c8608f49be83417834f5465c6295cb25972/src/Stack/Types/DumpPackage.hs#L5) type and parsing logic is a duplicate of the [InstalledPackageInfo](https://hackage.haskell.org/package/Cabal-syntax-3.10.1.0/docs/Distribution-InstalledPackageInfo.html) in cabal. The parsing function in cabal is expecting strict ByteString whereas the stack one operates on a stream of Text, but I'd expect the transition to be relatively smooth.
- The [GhcPkgId](https://github.com/commercialhaskell/stack/blob/49fe9c8608f49be83417834f5465c6295cb25972/src/Stack/Types/GhcPkgId.hs#L37) is the mix of many different types in Cabal (but mostly, as of now in stack, it's the mirror of the [ComponentId](https://hackage.haskell.org/package/Cabal-syntax-3.10.1.0/docs/Distribution-Types-ComponentId.html#t:ComponentId)). This is a problem for Backpack because Backpack has a notion of OpenModule (that is component which lacks implementation of its signatures) and this notion is hard to represent correctly with current GhcPkgId type.
- StackUnqualComponentName should be a newtype over [UnqualComponentName](https://hackage.haskell.org/package/Cabal-syntax-3.10.1.0/docs/Distribution-Types-UnqualComponentName.html#t:UnqualComponentName) from cabal, there are little reasons for not having this and the current Text based version is less efficient than the cabal one.
- The NamedComponent from Stack should also use the StackUnqualComponentName as the underlying representation for better interop. A Text translating function can be implemented to limit the amount of refactoring needed.

[...]

PR #6614, opened 2024-06-21
Refactoring toward cabal

theobat:
This is a first step for #6593, the changes in DumpPackage are not done here, because a first move was to bring some types closer to cabal, step by step.
This includes 2 major changes :
- GhcPkgId is now a wrapper over UnitId
- StackUnqualCompName is now a nextype over its cabal counterpart, and generalized wherever we can (instead of Text)

The second change is also a refactoring towards better type safety and documentation. It should also increase overall stack's speed by two aspects : better memory layout with ShortText, and less translation work from cabal.

Indicative integration tests speed:
1185 ubuntu
951 macos
1491 windows
So this corroborates a small speedup (compared to baseline https://github.com/commercialhaskell/stack/pull/6615)

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -83,6 +83,7 @@ dependencies:
 - fsnotify >= 0.4.1
 - generic-deriving
 - ghc-boot
+- hashable
 - hi-file-parser >= 0.1.6.0
 - hpack >= 0.36.0
 - hpc
--- a/.stan.toml
+++ b/.stan.toml
@@ -83,14 +83,14 @@
 
 # Anti-pattern: Data.ByteString.Char8.pack
 [[ignore]]
-  id = "OBS-STAN-0203-tuE+RG-234:24"
+  id = "OBS-STAN-0203-tuE+RG-236:24"
 # ✦ Description:   Usage of 'pack' function that doesn't handle Unicode characters
 # ✦ Category:      #AntiPattern
 # ✦ File:          src\Stack\Build\ExecutePackage.hs
 #
-#   233 ┃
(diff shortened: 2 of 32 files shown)


package.yaml
```diff
@@ -83,6 +83,7 @@ dependencies:
 - fsnotify >= 0.4.1
 - generic-deriving
 - ghc-boot
+- hashable
```
mpilgrem, line 86: You may not need a direct dependency on `hashable` because package `rio` depends on it.

theobat: I need to import
```
import           Data.Hashable ( Hashable(..) )
```
So I don't think so

mpilgrem: `Stack.Prelude` re-exports `RIO` and `RIO` exports `Hashable`?

theobat: My bad, I thought it didn't reexport the methods

mpilgrem: You may well be correct; I'll have a look.

mpilgrem: I think you are correct.

theobat: It seems I was :

> Module ‘RIO’ does not export ‘Hashable(hashWithSalt)’typecheck

While simply importing Hashable works (I find it odd to only reexport the class by the way, but...)

mpilgrem: The RIO project's reasoning is here: https://github.com/commercialhaskell/rio/pull/219.

mpilgrem:
#6615 is now merged, so this could be re-based on `master`.

PR #6645, opened 2024-08-11
Fix #6643 Respect `--no-run-tests`, `--no-run-benchmarks` when listing actions

### LINKED ISSUE #6643: Respect `--no-run-tests`, `--no-run-benchmarks` when listing actions
The motivation of:
* #6637
* #6639

taken together, was to allow a user to specify configuration:
~~~yaml
build:
  test: true
  test-arguments:
    no-run-tests: true
  bench: true
  benchmark-opts:
    no-run-benchmarks: true
notify-if-no-run-tests: false
notify-if-no-run-benchmarks: false
~~~

such that `stack build` (after `stack build`) could appear be an 'no op'. Such a user could command `stack build --run-tests`, if they wished for test suites of project packages to be run.

However, consider a simple two-project package (`packageA`, `packageB`) each with two test suites (1 and 2) and two benchmarks (1 and 2). The second `stack build` yields:
~~~text
Completed 4 action(s).
~~~

The four actions in question are the running of test suites and benchmarks for each of two project packages; these actions do nothing because `--no-run-tests` and `--no-run-benchmarks`.

With notification restored, the output is:
~~~text
packageA> Test running disabled by --no-run-tests flag.
packageB> Test running disabled by --no-run-tests flag.
packageA> Benchmark running disabled by --no-run-benchmarks flag.
packageB> Benchmark running disabled by --no-run-benchmarks flag.
Completed 4 action(s).
~~~

I propose that `--no-run-tests` and `--no-run-benchmarks` should be respected when the list of actions is created, as opposed to when the list of actions that has been created is subsequently executed.

After the change, with notification restored, the output is (but in colour):
~~~text
All test running disabled by --no-run-tests flag. To mute this message in future, set
notify-if-no-run-tests: false in Stack's configuration.
All benchmark running disabled by --no-run-benchmarks flag. To mute this message in future, set
notify-if-no-run-benchmarks: false in Stack's configuration.
~~~

After the change, with notification disabled, the second `stack build` is the desired 'no op'.

mpilgrem:
See:
* #6643 

Puts notification (if enabled) in `Stack.Build.Execute.executePlan'`.

In `Stack.Build.Execute.toActions`:

* does not add action `singleTest` if `not runTests`; and
* does not add action `singleBench` if `not runBenchmarks`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -21,6 +21,10 @@ Behavior changes:
   character in Stack's 'programs' path, as GHC 9.4.1 and later do not work if
   there is a space in the path to the `ghc` executable. S-8432 now presents as a
   warning and not an error.
+* Stack respects the `--no-run-tests` and `--no-run-benchmarks` flags when
+  determining build actions. Previously Stack respected the flags when executing
+  the run test suites or run benchmarks actions for each targeted project
+  package.
 
 Other enhancements:
 
--- a/doc/maintainers/stack_errors.md
+++ b/doc/maintainers/stack_errors.md
@@ -5,7 +5,7 @@
 In connection with considering Stack's support of the
 [Haskell Error Index](https://errors.haskell.org/) initiative, this page seeks
 to take stock of the errors that Stack itself can raise, by reference to the
-`master` branch of the Stack repository. Last updated: 2024-08-02.
+`master` branch of the Stack repository. Last updated: 2024-09-07.
 
 *   `Stack.main`: catches exceptions from action `commandLineHandler`.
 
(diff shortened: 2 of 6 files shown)


mpilgrem:
@theobat, as you have been looking at the architecture of Stack, I would be very grateful if you could spare any time to look at this pull request, as I consider its change to be architectural (even if it is relatively minor).

theobat:
@mpilgrem It looks good to me, though I'd place a user visible warning in case the unreachable code is reached, with an invitation to open an issue.
Also I'm wondering if stack should not prevent tests from being built in case they are disabled, I don't know, the good build behavior is not obvious to me...

PR #6658, opened 2024-11-11
fix(doc/commands/haddock_command.md): mismatching header for command name

andwu137:
Found that the documentation has a mismatching header for the command name `stack bench` instead of `stack haddock`.

Is not affected by the version of stack, so I've put this on the master branch.

No tests are necessary.


mpilgrem:
@andwu137, thanks! But please can you target the `stable` branch with this - that will then get merged into `master` branch also.

PR #6659, opened 2024-11-11
fix(doc.haddock_command): mismatching header for command

andwu137:
Found that the documentation has a mismatching header for the command name stack bench instead of stack haddock.

No tests are necessary.

(Not sure if creating a new fork was the right way to change the base branch)


mpilgrem:
Many thanks!

PR #6703, opened 2025-04-02
Depend on `Cabal-3.14.*`

sol:
Using `Cabal-3.14.*` is one of the required changes for GHC 9.12.  I need this for a personal project right now.  The purpose of this PR is not to necessarily have it merged right away, but rather to avoid duplicated work when Stack upgrades to GHC 9.12.

@mpilgrem from what I understand, we don't want CPP to support multiple versions of Cabal, right?

Do we ever use a more recent version of Cabal than what ships with GHC?  If "no", then this PR has to wait until we upgrade to GHC 9.12.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/stack.cabal linguist-generated=true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -349,7 +349,8 @@ Stack aims to depend on well-known packages. The specific versions on which it
 depends at any time are specified by `package.yaml` and `stack.yaml`. It does
 not aim to be compatible with more than one version of the `Cabal` package at
 any time. At the time of writing (August 2025) the package versions are
-primarily ones in Stackage snapshot LTS Haskell 24.4 (for GHC 9.10.2) and
+primarily ones in Stackage snapshot LTS Haskell 24.4 (for GHC 9.10.2), the
+latest version of `Cabal` released on Hackage (`Cabal-3.14.2.0`) and
 `pantry-0.11.0`.
 
 A Stack executable makes use of Cabal (the library) through a small 'Setup'
--- a/Setup.hs
+++ b/Setup.hs
@@ -25,6 +25,7 @@ import           Distribution.Simple.Utils
 import           Distribution.Types.PackageName ( unPackageName )
 import           Distribution.Types.UnqualComponentName
                    ( unUnqualComponentName )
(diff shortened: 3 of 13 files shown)


mpilgrem:
@sol, there is no 'fixed' rule, but for a while I have based Stack's project-level configuration on Stackage's most recent LTS Haskell snapshot (currently, GHC 9.8.4) with, perhaps, a few `extra-deps` (to avoid bugs or to keep pace with `hpack` releases). In part, that is because it made updating all the tests when the snapshot changed a less painful exercise. (Having everything on the same snapshot used to be important because the old self-hosted runner on Linux/AArch64 only had sufficient space for one or two versions of GHC. Now, it is just a search-replace convenience to move everything forward together.)

Formally, the `stack` library exists only to serve the Stack executable, and so does not aim to build with anything other than the packages in the project-level configuration. As each version of Stack is built with only one specified version of GHC, CPP has not been needed.

Stack does not use its `Cabal` dependency to build. So, I have not experienced a pressing need to 'jump ahead' to build Stack itself against more recent versions of `Cabal`.

While the Stackage project has no direct support for GHC 9.12, I am a little reluctant to accelerate Stack's dependency on `Cabal-3.14` but I would be happy to do that once Nightly moves to GHC 9.12. I assume that will happen once GHC 9.10.2 is released.

sol:
> While the Stackage project has no direct support for GHC 9.12, I am a little reluctant to accelerate Stack's dependency on `Cabal-3.14` but I would be happy to do that once Nightly moves to GHC 9.12.

@mpilgrem great, that certainly works for me.

BTW, can you make sure that this settings is enabled under https://github.com/commercialhaskell/stack/settings?

IMG_1668

sol:
> BTW, can you make sure that this settings is enabled under https://github.com/commercialhaskell/stack/settings?

Wait, maybe it's already enable, not sure. I think in this case the issue was that there were conflicts.  I rebased manually.

juhp:
Stackage LTS 24 has been out for a week now with ghc-9.10.2 :-)

mpilgrem:
@juhp, https://github.com/commercialhaskell/stack/commit/6a672dd12f25151707cf45e9823447334728d245 sets out the changes made to Stack's code to move from building against GHC 9.8.4 to GHC 9.10.2. There were only a few, `import Data.Type.Equality ( type (~) )` was one of them. It caused my to raise:
* https://github.com/commercialhaskell/rio/issues/255

PR #6764, opened 2025-07-18
Dev Containers: Install GitLens extension from VSIX

benz0li:
- [x] Any changes that could be relevant to users have been recorded in ChangeLog.md.
- [x] The documentation has been updated, if necessary

No updates to ChangeLog.md and documentation required.


benz0li:
I pinned GitLens extension to v11.7.0 due to unsolicited AI content in recent versions.

Installing directly from the VS Code Extension Marketplace does not work anymore.
ℹ️ Newer versions of VS Code require extensions to be signed.

GitLens extension v11.7.0 is not signed and can therefore only be installed via VSIX.

PR #6834, opened 2026-01-02
allow persistent-2.18

juhp:
This should allow stack to go into nightly and lts25 :-)

Tested locally with `stack --resolver nightly --allow-newer`.


juhp:
I don't really understand the failures?

Issue #6871, opened 2026-04-11
`stack ls snapshots remote` should use new /api/v1/snapshots endpoint


chreekat:
https://github.com/commercialhaskell/stack/blob/d67585934f769b92cd924b6e0d7d67917c8784f0/src/Stack/Ls.hs#L226

This slipped my notice when I decided to split the http and json responses into separate endpoints.

mpilgrem:
@chreekat, noted. Where is the proposal to change Stackage's API documented?

chreekat:
Unfortunately the only public discussion for the change happened at https://github.com/commercialhaskell/stackage-server/issues/355.

I've opened https://github.com/commercialhaskell/stackage-server/issues/365 to track reverting the API to make Stack work again. As I mention there, I would like to return to the current state of affairs in the long run, so this present issue is still worth doing. I'm happy to do it myself.

chreekat:
NB: The failure no longer reproduces because I fixed stackage.org. I definitely reproduced it before that, though. :)

PR #6875, opened 2026-04-14
Re #6871 Add `recent-snapshots` configuration option

mpilgrem:
See:
* #6871

Please also shortly describe how you tested your change. Bonus points for added tests! Tested locally on Windows 11.


mpilgrem:
@chreekat, to future-proof it, I decided to make the URL configurable rather than only hard-coded.

PR #6916, opened 2026-05-25
Fix #6915 `config set` recreates global-project directory, when required

### LINKED ISSUE #6915: `stack config set` fails if the `global-project` directory is missing/incomplete
As identified by @hasufell:

If the `global-project` directory in the Stack root is missing, or exists but is missing its `stack.yaml` file, `stack config set` fails when run outside of a Stack project and that file is to be consulted. For example:
~~~text
> stack config set system-ghc false
D:\sr\global-project\stack.yaml: withFile: does not exist (No such file or directory)
~~~

Stack's online documentation currently states:

> If the project-level configuration file is deleted, and Stack needs to consult it, then Stack will recreate the contents of the directory.

and I think that is the desired behaviour for `stack config set` too.

mpilgrem:
See:
* #6915

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -55,6 +55,9 @@ Bug fixes:
   depends on project package B and package B's executables (only) depend on
   package A and the name of A is before that of B, alphabetically. That bug is
   fixed.
+* Stack's `config set` commands will recreate the `global-project` directory
+  contents, if Stack needs to consult its project-level configuration file and
+  there is no file.
 
 ## v3.9.3 - 2026-02-19
 
--- a/package.yaml
+++ b/package.yaml
@@ -304,6 +304,7 @@ library:
   - Stack.Types.Component
   - Stack.Types.ComponentUtils
   - Stack.Types.Config
+  - Stack.Types.ConfigExtra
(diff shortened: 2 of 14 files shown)


mpilgrem:
Currently, the recreation of the `global-project` directory happens in the first part of `Stack.Config.withBuildConfig`. The final part does other things, which are not of interest to Stack's `config set` command.

The first part also outputs a user message, if one is specified in the project-level configuration file. I think Stack's `config set` command would want to skip that.

This pull request:
*   separates the first part of `withBuildConfig` as:
    ~~~haskell
    withConfigExtra ::
         forall a env. (HasConfig env, HasTerm env)
      => Bool
         -- ^ Report user message in the project-level configuration file, if
         -- a user message is present?
      -> (ConfigExtra -> RIO env a)
      -> RIO env a
    ~~~

*   introduces the new type `ConfigExtra` for the information provided by the first part of     `withBuildConfig` used by the remaining part;

*   makes use of `withConfigExtra` in `Stack.ConfigCmd.cfgCmdSet`;

*   adds an integration test for the bug fix `6915-config-set-outside-project`.

PR #6945, opened 2026-07-07
Fix #6861 Use nix-instantiate command, with failure exception

### LINKED ISSUE #6861: If `nix-shell <args>` fails, Stack just outputs the Nix error
With Nix integration, Stack executes `nix-shell` with arguments. If the command fails, Stack does not attempt to put that error in any context.

mpilgrem:
See:
* #6861 

If any users of Stack on NixOS or other users of Stack's Nix integration see this, I would appreciate any feedback on it.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -24,6 +24,9 @@ Behavior changes:
 * When Stack's Nix integration is enabled and a `shell-file` is not specified,
   Stack sets `LANG=C.UTF-8` (rather than `en_US.UTF-8`, which may not be
   available if Stack has been built from source).
+* With Nix integration, Stack uses `nix-instantiate` before `nix-shell`,
+  enabling Stack to catch and report when the Nix expression cannot be
+  evaluated by Nix. The `nix-instantiate-options` option is added, accordingly.
 
 Other enhancements:
 
--- a/doc/maintainers/stack_errors.md
+++ b/doc/maintainers/stack_errors.md
@@ -5,7 +5,7 @@
 In connection with considering Stack's support of the
 [Haskell Error Index](https://errors.haskell.org/) initiative, this page seeks
 to take stock of the errors that Stack itself can raise, by reference to the
-`master` branch of the Stack repository. Last updated: 2026-04-24.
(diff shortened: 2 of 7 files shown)


mpilgrem:
I have converted to draft, as there is unexplained adverse changed behaviour in respect of locale when `nix-shell` is used with a store derivation of an expression rather than with the expression directly. 

On my 'out of the box' Ubuntu (via WSL2), `locale --all` yields:
~~~text
C
C.utf8
POSIX
~~~

With a one-package Stack project, with a library module `Lib.hs`
~~~haskell
module Lib
  ( unicode
  ) where

{-# WARNING unicode "Warning emits unicode 😊" #-}
unicode :: IO ()
unicode = putStrLn "😊"
~~~

and executable `myApp` with `Main.hs`:
~~~haskell
module Main
  ( main
  ) where

import           Lib ( unicode )

main :: IO ()
main = unicode
~~~

With 'official' Stack 3.11.1 binary (statically-linked):

`stack build --nix --ghc-options=-Wno-extended-warnings` and `stack exec --nix -- myApp` both work and `stack build --nix --ghc-options=-Wextended-warnings` also works.

With Stack-3.11.1 built from source (dynamically-linked):

`stack build --nix --ghc-options=-Wextended-warnings` (only) fails when GHC emits the unicode character (extract):

~~~text
...
Preprocessing executable 'myApp' for myPackage-0.0.0...
Building executable 'myApp' for myPackage-0.0.0...
[1 of 1] Compiling Main
/home/mpilgrem/Code/Haskell/testNix/app/Main.hs:8:8: warning: [GHC-63394] [-Wdeprecations]
    In the use of `unicode' (imported from Lib):
    "Warning emits unicode
Error: [S-7282]
       Stack failed to execute the build plan.

       While executing the build plan, Stack encountered the error:

       <stderr>: commitAndReleaseBuffer: invalid argument (cannot encode character '\128522')
~~~

The problem is that locale `en_US.UTF-8` is not available and the GHC runtime silently defaults to `ASCII` locale and encoding of `stdout` and `stderr`.

I understand the solution is to refer to a UTF-8 locale that is more likely to be available on the host, namely `C.UTF-8`.

Issue #6944, opened 2026-07-07
'Immutable' Haskell library depends on a C library that has changed - How to fix?


ulidtko:
Related materials:
* #1476

----
#### Background

Once a dependency gets installed into Stack's compiled package database `~/.stack/snapshots`, it gets aggressively cached for as long as the snapshot hash keeps matching. This is as designed and as desired:

```
$ stack install lzma
[...]
$ stack exec -- ghc-pkg unregister lzma
$ stack install lzma
lzma> using precompiled package
$
```

This `using precompiled package` log (and the fast build) — is exactly as expected.

#### The issue

It's hard to surgically yank one package/subtree from this precompiled_cache. I met reasons to want to do this many times; will give one below.

Despite `stack install` offering the `--[no-]force-dirty` option, it's not effective for this purpose, and behaves the same:
```
$ stack exec -- ghc-pkg unregister lzma
$
$ stack install --force-dirty lzma
lzma> using precompiled package
$
```
Expected: fresh rebuild of the lzma package. Actual: `using precompiled package` again.

Thus, the most precise option to force a rebuild of one dependency is still "nuclear":
* identify the active snapshot hash, e.g. `stack exec -- ghc-pkg list | grep /pkgdb`,
* drop the entire snapshot, `rm -rf ~/.stack/snapshots/{{arch}}/{{hash}}`.

More precision is doable with pointwise file removals and `DELETE FROM precompiled_cache WHERE library LIKE '%/lzma-%';` — but this feels like violating API boundaries & messing with internal guts. I think anyone would prefer if `--force-dirty` would instead do that.

#### Sample rationale

Unlike other linkers, MacOS `ld` warns about nonexistent `-L` paths. A real-world scenario I sorted out today proceeded like this:
1. `xz` library 5.8.1 installed via Homebrew at /opt/homebrew/Cellar/xz/5.8.1;
2. A Haskell project having [lzma](https://hackage.haskell.org/package/lzma) (a libxz FFI wrapper) in transitive dependency closure was built;
3. `xz` library upgraded in Homebrew to 5.8.3, /opt/homebrew/Cellar/xz/5.8.3.

Ever since, rebuilds of the Haskell project had been repeatedly emitting:
> **ld: warning: search path '/opt/homebrew/Cellar/xz/5.8.1/lib' not found**

[...]

Issue #1476, opened 2015-12-07
How to reinstall a package?


FranklinChen:
Here's a problem I seem to have run into with Stack: I don't know how to reinstall a package! I can't find documentation on how to do that. Before Stack, `cabal --reinstall` was simple.

The reason I want to reinstall is packages that depend on a C library that I've updated, such that I should rebuild.

mgsloan:
I actually never used the `cabal --reinstall` flag (only `--force-reinstall` a lot, heh..).  It only reinstalls the target package, right?  So, for local packages you should be able to do this either by touching some files or `stack clean`.

It's a bit trickier for snapshot dependencies.  `stack exec -- ghc-pkg unregister --force PKG` before building should do the trick.

I'm undecided on whether to add support for `--reinstall`, as this is hopefully a rather specialized case for people to handle.  On the other hand, `stack exec -- ghc-pkg unregister --force PKG` is a bit of a magical incantation and having a `--reinstall` flag would be consistent with cabal (reuse people's knowledge of it).

filib:
I ran into this issue today. I have a package that depends on a dynamic library that has since changed and would like to reinstall the package.

mgsloan:
I suppose if we've got 2 people in 10 days facing the same issue, that's common enough to warrant a `P2: Should`. So, `stack build --reinstall mtl` seems like a reasonable thing to have.  Note that it would only reinstall target packages.  Dependencies would only be built as necessary, and so usually won't be reinstalled.

mgsloan:
Could also deprecate `--force-dirty` in favor of `--reinstall`.  There would likely be a difference in logic, though - `--force-dirty` causes all local packages to be considered dirty, so if

sebeaumont:
Also if flags change for a package even in ```stack.yaml``` stack doesn't detect that necessitating a manual rebuild by first removing the package...

qrilka:
@sebeaumont it would be great if you could show some minimal reproduction of this

dpwiz:
Git commits are indeed immutable. The problems with connection and configuration however will break this.
Like in my case, where `git` did its part fetching the data, but the filtering failed. This resulted in a botched package data poisoning the cache.

ulidtko:
Another example of when reinstalling is reasonably needed:

 * relinking FFI packages after its native library updates.

[...]

PR #6948, opened 2026-07-10
Emit a hint when caching immutable precompiled build with --force-dirty

ulidtko:
This is a tiny UX-level addition regarding #6944 & #1476. No behavioral change, just an added log:

Screenshot 2026-07-10 at 18 23 35

The Note is guarded to appear only when:
* `using precompiled package`,
* that package is in the explicitly passed targets list,
* and `--force-dirty` is supplied.

### DIFF
--- a/src/Stack/Build/ExecutePackage.hs
+++ b/src/Stack/Build/ExecutePackage.hs
@@ -128,7 +128,7 @@ import           Stack.Types.ComponentUtils
                    ( StackUnqualCompName, toCabalName, unqualCompToString
                    , unqualCompToText
                    )
-import           Stack.Types.Config ( Config (..), HasConfig (..) )
+import           Stack.Types.Config ( Config (..), HasConfig (..), buildOptsL )
 import           Stack.Types.ConfigureOpts
                    ( BaseConfigOpts (..), ConfigureOpts (..) )
 import           Stack.Types.Curator ( Curator (..) )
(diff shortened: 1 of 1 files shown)


ulidtko:
@mpilgrem please review, I hope it's an easy merge. I'll promptly address feedback, if any.

I gave it a round of tests locally, seems to be conservative and uncontroversial.

mpilgrem:
@ulidtko, you want Stack to remind a user that flag `--force-dirty` has no effect if the targets of `build` do not include any that can, in principle, be dirty. I think that is fine, but my preference would be a message like:

> --force-dirty is specified but no build targets or dependencies are capable of being dirty. 
> If you wish Stack to rebuild an installed package, first unregister it from the package database
> by commanding:
> 
> stack exec -- ghc-pkg unregister --force <package>

I'm not sure that warning is required _during_ the build: can't Stack work it out, upfront, once it has parsed the build targets? (I need to think about that...)

EDIT1: With the currently implementation, a user only gets a warning if a build target happens to be a _reusable_ installed package, installed in a _different_ package database - and so only sees it once (after the resuable package is installed in the current package database, it no longer meets the criteria). If we aim to warn on the redundant use of `--force-dirty`, I think we should aim to warn whenever is has been mis-specified.

~~EDIT2: I think Stack knows whether any build targets or dependencies are capable of being dirty once it has made the context (`ctx`) in `Stack.Build.ConstructPlan.constructPlan`.~~

EDIT3: When Stack is reusing a pre-compiled package, I think to get Stack to rebuid it you have to unregister it from the package database it is _coming from_. If you unregister it from the package datbase it has been _copied to_, I think Stack will just bring over another copy on the next build.

EDIT4: @ulidtko, I have an alternative to this pull request at #6949, which includes your hint but at the start of the build and is triggered more frequently.

mpilgrem:
Closed, as overtaken by:
* #6949

Issue #6941, opened 2026-07-04
Second `stack build` fails, if `*.hsig` file is  changed


mpilgrem:
See:
* https://github.com/commercialhaskell/stack/pull/6940#issuecomment-4882136182

By reference to the integration test `backpack-x-pkg-transitive`:

The first `stack build` works, as expected, but if the `Logger.hsig` file is then changed to be:
~~~haskell
signature Logger where

logMessage :: String -> String
logName :: String
~~~

the second `stack build` fails, with:
~~~text
❯ stack build

Warning: Found Backpack signature file for Cabal entry Str, but that module is not listed in the
         component's signatures field:
         * Str.hsig

         If this file is meant to be a Backpack signature, add Str to the signatures field in
         the package description.

Warning: Backpack: no dependency exposes module Str needed to instantiate str-sig. Ensure the
         implementing package is listed in build-depends.
consumer-pkg-0.0.0: unregistering (missing dependencies: logger-sig)
logger-sig-0.0.0: unregistering (local file changes: src\Logger.hsig)
logger-sig-0.0.0: unregistering (local file changes: src\Logger.hsig)
str-sig-0.0.0: unregistering
str-sig-0.0.0: unregistering
logger-sig  > build (lib) with ghc-9.10.3
logger-sig  > Preprocessing library for logger-sig-0.0.0...
logger-sig  > Building library instantiated with
logger-sig  >   Logger = <Logger>
logger-sig  >   Str = <Str>
logger-sig  > for logger-sig-0.0.0...
logger-sig  > <command line>: cannot satisfy -package-id str-sig-0.0.0-HPF0111dNK1KIRtIq2ktMM[Str=<Str>]
logger-sig  >     (use -v for more information)
str-sig     > configure (lib)
str-sig     > Configuring str-sig-0.0.0...
str-sig     > Error: [Cabal-5000]
str-sig     > The following package dependencies were requested
str-sig     >     --dependency='str-sig=str-sig-0.0.0-HPF0111dNK1KIRtIq2ktMM'
str-sig     > however the given installed package instance does not exist.
str-sig     >
~~~

A third `stack build` then fails with (where the GHC error message makes sense):
~~~text
> stack build

Warning: Backpack: no dependency exposes module Str needed to instantiate str-sig. Ensure the
         implementing package is listed in build-depends.
str-sig     > configure (lib)
str-sig     > Configuring str-sig-0.0.0...
str-sig     > build (lib) with ghc-9.10.3
[...]

PR #6952, opened 2026-07-22
Fix stale Backpack CInst planning for installed signatures

philippedev101:
Follow-up to #6940. This is intended to answer the question there about the failing `backpack-x-pkg-transitive` case, which was split out as #6941.

My read is that the test is exposing a problem in the existing Backpack planning, rather than a problem caused by the `.hsig` discovery change itself.

The failure happens around signature packages that are already installed but still need to be used as templates for Backpack instantiations. In the build plan, those packages are `ADRFound`: already installed and usable as dependencies. `ADRToInstall` means something different: Stack has a `Task` for the package and intends to build it before dependents use it.

The previous approach crossed those meanings by upgrading some installed indefinite packages from `ADRFound` to `ADRToInstall` so CInst creation could get at their `Package` metadata. That gave Stack enough information to create the CInst task, but it also made a clean installed signature library look like a normal build-plan task. In the failing case, that can make Stack unregister or rebuild the indefinite unit at the wrong time, leaving later Backpack configure/build steps with stale or missing package ids.

This PR keeps those two ideas separate:

- installed signature packages stay `ADRFound`
- Stack loads source-backed task templates separately when Backpack needs metadata for CInst construction
- those templates are only blueprints for CInst tasks, not ordinary build-plan tasks
- CInst-only tasks are ignored when choosing representative packages for local unregistering, because the CInst configure step still needs the existing indefinite CLib registration to be present

There are two dependency-resolution paths after this change, but they are doing different jobs. The normal `ConstructPlan` resolver still decides the real build plan: which packages are found, which packages must be installed, and which dependencies are missing or present. The Backpack template resolver only reconstructs enough dependency information for templates corresponding to clean installed indefinite packages, because those packages intentionally remain `ADRFound` and therefore do not have normal `Task` values in the plan.

To keep the scope small, the template loader follows Backpack mixin dependencies only. That is enough to discover inherited holes, including inherited signatures from source-backed snapshot packages, without walking ordinary dependencies as if they were Backpack targets.

The PR also shares the `AddDepRes -> MissingPresentDeps` conversion used by the normal resolver and the Backpack template resolver, so `ADRFound` and `ADRToInstall` dependencies are represented the same way in both paths.

Tests add focused coverage for stale CInst planning when implementation packages, signature packages, and consumers change, plus inherited-hole cases for both local and snapshot-backed signature packages. The existing transitive Backpack test remains the baseline for the clean rebuild path.

### DIFF
--- a/src/Stack/Build/Backpack.hs
+++ b/src/Stack/Build/Backpack.hs
@@ -16,7 +16,7 @@ tasks that instantiate those packages with concrete implementations.
 
 module Stack.Build.Backpack
   ( addInstantiationTasks
-  , upgradeFoundIndefinites
+  , loadFoundIndefiniteTasks
   ) where
 
 import           Crypto.Hash ( hashWith, SHA256 (..) )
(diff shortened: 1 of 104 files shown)


philippedev101:
@mpilgrem could you let me know if this is going in the right direction or if you would like some changes? one thing that comes to my mind is that the amount of tests could be reduced

mpilgrem:
@philippedev101, you are right to chase. Despite my talk of not losing momentum, other things (work and Haskell) have had my attention. However, it is still my intention to get this merged for a Stack 4.1.1 for end Q3 2026.

philippedev101:
@mpilgrem no rush, it's not urgent on my side. Q3 sounds fine. I was just wondering if there was something wrong with the PR

